PR #61, opened 2017-05-24
Better change detection

kflynn:
Do a better job of handling corner cases in detecting changes in mappings.

### DIFF
--- a/ambassador/ambassador.py
+++ b/ambassador/ambassador.py
@@ -295,18 +295,37 @@ def ambassador_stats():
 
 def new_config(envoy_base_config=None, envoy_tls_config=None,
                envoy_config_path=None, envoy_restarter_pid=None):
+    # logging.debug("new_config entry...")
+    
     rc = fetch_all_mappings()
     if not rc:
         # Failed to fetch mappings from DB.
+        logging.debug("new_config could not fetch mappings: %s" % rc)
         return rc
 
     # Suppose the fetch "succeeded" but we got no mappings?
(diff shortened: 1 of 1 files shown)


ambassador/ambassador.py
```diff
@@ -295,18 +295,37 @@ def ambassador_stats():
...
+        app.current_mappings = None
+
+        logging.debug("new_config got no mappings at all? %s" % str(rc))
+        return RichStatus.fromError("no mappings found at all, original rc %s" % str(rc))
+
+    num_mappings = 0
+
+    try:
+        num_mappings = len(rc.mappings)
+    except:
```
ark3: if `except TypeError:` doesn't do the job here, I would be very surprised.

kflynn: I thought about that, and decided to deliberately go broad. 
Open to persuasion about why that would be a bad idea.

ambassador/ambassador.py
```diff
@@ -295,18 +295,37 @@ def ambassador_stats():
...
+        # Huh. This can really only happen if something isn't quite initialized
+        # in the database yet.
+        logging.debug("new_config mappings is not an array, assuming empty")
```
ark3: Weird case. May as well log the `type(...)` of mappings.

ambassador/ambassador.py
ark3: All the usual reasons why `except:` is usually wrong (as compared to `except Exception:` below), but I doubt this is at all important.

PR #177, opened 2017-12-15
Fixing links in docs

mdkess:
Fixing some bad links that I noticed while reading the docs.

One thing - reference/tls-auth.md is referenced twice, but the file doesn't exist. I updated it to the how-to/tls-termination.md file, but it wasn't clear to me what the here intent was.


richarddli:
Thanks for this. We had a more comprehensive TLS overview, but decided to delete it since it just covered the core essentials of TLS.

PR #191, opened 2017-12-20
Mention `kube-cert-manager` for auto-renewing Let's Encrypt certificates on Kubernetes

kminehart:


kflynn:
Hey, so very sorry for the delay in feedback here -- mea culpa! 

I actually really really like this idea. Can you update these docs with a bit more, about how you deployed `kube-cert-manager` into the cluster and such?

kminehart:
Hey @kflynn, I added a little bit of extra info about `kube-cert-manager`. I'd hesitate to include any more detailed information about deployment.

Originally, `kube-cert-manager` was a project by Kelsey Hightower which used a more primitive, now deprecated, form of `CustomResourceDefinition`s called `ThirdPartyResources`.  I did include a link to the deployment guide, so that if things were to change again, it wouldn't be hardcoded in the docs.

kminehart:
hmm.  Yeah it can be done. I'll see what I can do. :)

I might instead create a separate page specifically for a Let's Encrypt deployment, would that be preferable?

richarddli:
A separate page for Let's Encrypt would be super. Thanks!

kminehart:
I didn't have a whole lot of time to verify and fact check, but I've added the new page and removed let's encrypt information from the old one.

maxfrigge:
Would be super awesome to get automated letsencrypt certificates. It appears though that some projects are deprecating in favour of https://github.com/jetstack/cert-manager. Have you considered using their implementation?

kminehart:
We actually do use cert manager now, and the setup is very similar, yeah.

maxfrigge:
Sweet! Would you mind updating the PR, if you find the time?

kflynn:
Yeah, gotta fix that CI hassle. Thanks!

koliyo:
Yes, please add some documentation on using ambassador together with `cert-manager` and ACME. The learning curve to set this up is pretty steep atm as a new kubernetes user.

kminehart:
Sorry I never got around to updating this. :grimacing: 

Funny though how I suggested that we probably shouldn't include kube-cert-manager specifically as it could be deprecated, and then a month later it gets deprecated in favor of jetstack/cert-manager.

ofcourseican:
@kminehart what do you mean with "We actually do use cert manager now, and the setup is very similar"? Is this somehow integrated with ambassador already?
What do I need to do to run a Letsencrypt protected service with Ambassador?

koliyo:
Yes, what @ofcourseican mentions is what I think is missing to make ambassador more approachable for this use case.

I would like to see a explicit step-by-step guide to
1. Install cert manager, configured for ACME certificates (referencing existing cert-manager documentation is fine)
2. Install and configure ambassador with TLS termination, using cert manager for acquiring certificate 
3. Install and configure an example service, eg httpbin, on a subdomain with https. Eg `httpbin.mydomain.com`

bhack:
@richarddli What about https://github.com/jetstack/cert-manager/issues/231?

maxfrigge:
I just went through the process of having ambassador terminate a certificate from `jetstack/cert-manager`. There is basically one major thing you need to know and it's [mentioned in the docs](https://github.com/datawire/ambassador/blob/master/docs/user-guide/tls-termination.md#3-youll-need-to-store-the-certificate-in-a-kubernetes-secret). Ambassador will look for secret `ambassador-certs` on startup and use this secret for tls termination. So all you need to do is tell cert manager to store the certificate in that secret. 

Once cert manager is running and you have successfully created the issuer, you can request a certificate like so. What's important here is `secretName: ambassador-certs`.
```yaml
apiVersion: certmanager.k8s.io/v1alpha1
kind: Certificate
metadata:
  name: cloud-foo-com
  namespace: default
spec:
  secretName: ambassador-certs
  issuerRef:
    name: letsencrypt-prod
    kind: ClusterIssuer
  commonName: cloud.foo.com
  dnsNames:
  - cloud.foo.com
  acme:
    config:
    - dns01:
        provider: clouddns
      domains:
      - cloud.foo.com
```
When the certificate has been stored in the secret you need to restart ambassador to pick up on it.

After realising that you don't actually have to set anything up in ambassador to get this working, I think a link to the `jetstack/cert-manager` docs and a short note would be a good solution.

kminehart:
> After realising that you don't actually have to set anything up in ambassador to get this working, I think a link to the jetstack/cert-manager docs and a short note would be a good solution.

Agreed. And this is what I've been saying since I created the issue originally.

Setting up kube-cert-manager is easy and domain specific. Some people might want to use DNS validation, http validation, or whatever.

Kubernetes is a fast moving ecosystem. Doing anything other than saying, "You can use `kube-cert-manager` to automatically renew a TLS certificate into the k8s secret `ambassador-certs`" is futile.

What proves my point even more is that since I opened this in February, `kube-cert-manager` has been deprecated in favor of `cert-manager`.

richarddli:
Thanks. I've added this in #505.

PR #197, opened 2018-01-03
Rdl/prom monitoring

richarddli:


richarddli:
@kflynn Not sure if there's a way to avoid hard coding the Ambassador version into the sample file. Take a look and let me know your thoughts on how I structured these changes.

kflynn:
OK. Upon reflection, your `ambassador-rbac-statsd.yaml` should move into `templates/ambassador/ambassador-rbac-prometheus.yaml` (I think) and become a template, so that the version can be managed by the build process. Use 

```
        image: {AMREG}ambassador:{VERSION}
```

for the Ambassador image, then `make` should fill in the template and put the result in `docs/yaml/ambassador` for you. That way people can refer to it via HTTPS, like the others.

The rest of this looks good. :)

richarddli:
Alright, that's done now, and some documentation on that file has been added.

PR #219, opened 2018-01-19
Update ambassador docker registry -- add quay.io

mpatelpa:


richarddli:
Thank you for the fix!

PR #221, opened 2018-01-19
Mapping objects support for Envoy request_headers_to_add

alexgervais:
Added support for `request_headers_to_add`, https://github.com/datawire/ambassador/issues/217

``` 
apiVersion: ambassador/v0
kind:  Mapping
name:  default
prefix: /default
service: echo-svc:8080
request_headers_to_add:
- key: x-test-proto
  value: "%PROTOCOL%"
- key: x-test-ip
  value: "%CLIENT_IP%"
- key: x-test-static
  value: This is a test header
```

### DIFF
--- a/ambassador/ambassador/mapping.py
+++ b/ambassador/ambassador/mapping.py
@@ -39,6 +39,7 @@ class Mapping (object):
         "use_websocket": True,
         "timeout_ms": True,
         "priority": True,
+        "request_headers_to_add": True,
     }
 
     def __init__(self, _source="--internal--", _from=None, **kwargs):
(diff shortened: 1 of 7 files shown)


kflynn:
This looks fine, with one big caveat: specifying headers that must be present for a `Mapping` to match is done differently. :( From `configuration.md`:

> If present, the `headers` attribute must be a dictionary of `header`: `value` pairs, for example:
> 
> ```yaml
> ---
> apiVersion: ambassador/v0
> kind:  Mapping
> name:  qotm_mapping
> prefix: /qotm/
> headers:
>   x-qotm-mode: canary
>   x-random-header: datawire
> service: qotm
> ```

I'd like to have `request_headers_to_add` work the same way, unless we can come up with a compelling reason not to. (I'd also be OK with calling it something shorter; any suggestions? :) )

PR #230, opened 2018-01-23
Cors support

alexgervais:
Ref: https://github.com/datawire/ambassador/issues/215

- Added `cors` Envoy filter and route settings.
- Added `cors` attribute to `Mapping`.
- Updated unit tests and documentation.

Example:
```
apiVersion: ambassador/v0
kind: Mapping
name: cors_mapping
prefix: /cors/
service: cors:8080
cors:
  methods: POST, GET, OPTIONS
  headers: Content-Type
  credentials: true
  max_age: "86400"
```


kflynn:
Thanks, @alexgervais!

PR #232, opened 2018-01-23
Improved documentation and example around AMBASSADOR_SINGLE_NAMESPACE

alexgervais:
`AMBASSADOR_SINGLE_NAMESPACE` was previously undocumented.

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -62,13 +62,15 @@ Ambassador supports multiple namespaces within Kubernetes. To make this work cor
 
 ```yaml
         env:
+        - name: AMBASSADOR_SINGLE_NAMESPACE
+          value: "true"
         - name: AMBASSADOR_NAMESPACE
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace          
 ```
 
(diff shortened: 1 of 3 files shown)


kflynn:
Hmmmmmm. So `AMBASSADOR_SINGLE_NAMESPACE` is not expected to be all that common -- we expect it to be much more common to have a single Ambassador managing multiple namespaces. So I'd be delighted to have more doc around `AMBASSADOR_SINGLE_NAMESPACE`, but I'd rather _not_ have it included in all the YAML by default.

alexgervais:
Does Ambassador need to know under which namespace it is _running_ if managing multiple namespaces? 

The inclusion of `AMBASSADOR_NAMESPACE` in the sample YAML led me to believe it would manage a single namespace... If it's not required, then I would **remove both** `AMBASSADOR_SINGLE_NAMESPACE` and `AMBASSADOR_NAMESPACE` from the default YAML, but **include both** in the `configuration.md` docs.

alexgervais:
Duplicate of https://github.com/datawire/ambassador/pull/245

PR #264, opened 2018-02-09
Added support for circuit breaker `max_requests`

alexgervais:
* Added support for circuit breaker `max_requests`.
* Using the same default value as Envoy (https://www.envoyproxy.io/docs/envoy/v1.5.0/api-v1/cluster_manager/cluster_circuit_breakers#per-priority-settings)
* Added unit tests

Documentation to follow?


kflynn:
We'll obviously need some end-to-end tests for circuit breakers, but this looks fine as far as it goes.

Issue #275, opened 2018-02-21
Ambassador needs to expose `use_proxy_proto` in Envoy somehow


plombardi89:
The Proxy Protocol is the defacto standard to pass the remote address of a connection to a backend server when there is a proxy in the way. In our situation the "proxy" is an Amazon ELBv1 "classic ELB". In order to support both Secure WebSocket and regular HTTPS communication on an ELB you need to configure it run as TCP+TLS listener and then pass a raw TCP stream back to the instances behind the ELB. In our case Ambassador handles that TCP stream at that point.

Without the proxy protocol enabled the source address information is lost because the ELB rewrites it as the IP address of the load balancer itself. We need this information to be preserved. The mechanism to do this is the proxy protocol being installed on the ELB. Envoy, however, in its default configuration does not enable the proxy protocol but it is available as a listener configuration: https://www.envoyproxy.io/docs/envoy/latest/api-v1/listeners/listeners.html#config-listeners-v1

However, this needs to be installed per-Route because it's not a given that all ELB's that point at Ambassador will have proxy protocol enabled.... Being able to run multiple listeners might be handy and then install particular routes on a particular listener might be a solution...

sbrichardson:
I have notes in Gitter about how I did this successfully.

plombardi89:
@sbrichardson Thanks. Found the notes!
@kflynn https://gitter.im/datawire/ambassador?at=59fca6b72a69af844b44e578

It looks like we need to also enable `use_remote_address` which is documented in here: https://www.envoyproxy.io/docs/envoy/latest/api-v1/network_filters/http_conn_man.html

```
use_remote_address
(optional, boolean) If set to true, the connection manager will use the real remote address of the client connection when determining internal versus external origin and manipulating various headers. If set to false or absent, the connection manager will use the x-forwarded-for HTTP header. See the documentation for x-forwarded-for, x-envoy-internal, and x-envoy-external-address for more information.
```

Hopefully this resolves our Friday frustration with `use_proxy_proto`... we'll see. I have some additional design concerns as well though we need to discuss at some point. Primarily, the setting we've added to the Ambassador Module config is globally scoped in such a way that if there are multiple load balancers pointed at Ambassador (e.g. Argonath) then this could cause a problem if not all load balancers have proxy proto enabled.

plombardi89:
FYI @kflynn `@sbr464 => I didn't have to mess with use_proxy_proto`

[...]

PR #284, opened 2018-02-26
Support use_proxy_proto and use_remote_address

kflynn:
Add support for including `use_proxy_proto` and `use_remote_address` to the `ambassador` Module config.

`use_proxy_proto` defaults to `false`.
`use_remote_address` default to absent.

Both default values are chosen to be in line with a plain Envoy configuration.


kflynn:
This is a (partial?) fix for #275.

PR #290, opened 2018-02-28
Remove magical ConfigMap uuid stuff because it usually doesn't work.

plombardi89:
Defaults to the "Nil" UUID which we will filter out in the future from
scout queries.

I scanned around for some tests but I didn't see any. @kflynn do we test scout initialization at all in Ambassador?

Are we OK with the Nil UUID implementation?

### DIFF
--- a/ambassador/ambassador/config.py
+++ b/ambassador/ambassador/config.py
@@ -68,17 +68,15 @@ class Config (object):
 
     runtime = "kubernetes" if os.environ.get('KUBERNETES_SERVICE_HOST', None) else "docker"
     namespace = os.environ.get('AMBASSADOR_NAMESPACE', 'default')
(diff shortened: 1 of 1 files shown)


ambassador/ambassador/config.py
```diff
@@ -68,17 +68,15 @@ class Config (object):
...
 
-    if scout_install_id:
-        _scout_args['install_id'] = scout_install_id
-    else:
-        _scout_args['id_plugin'] = Scout.configmap_install_id_plugin
-        _scout_args['id_plugin_args'] = { "namespace": namespace }
+    _scout_args['install_id'] = scout_install_id
 
     try:
         scout = Scout(**_scout_args)
```
kflynn: No need for `_scout_args` any more -- just inline the dict into the `Scout` creation:

```
scout = Scout({
    "app": "ambassador",
    ....
})
```

PR #294, opened 2018-03-01
Expose envoy tls alpn_protocol to fix grpc over tls

technicianted:
This PR exposes the ability to set Envoy `alpn_protocol` for the `ssl_context` listeners to enable ALPN protocol requested protocols.

Since gRPC clients use HTTP2, this option must be set to `h2` for it to work over TLS.

[Envoy TLS context reference](https://www.envoyproxy.io/docs/envoy/latest/api-v1/cluster_manager/cluster_ssl.html).

### DIFF
--- a/ambassador/templates/envoy.j2
+++ b/ambassador/templates/envoy.j2
@@ -8,7 +8,10 @@
       "ssl_context": {
         {%- if listener.tls.cert_chain_file -%}
           "cert_chain_file": "{{ listener.tls.cert_chain_file }}",
-          "private_key_file": "{{ listener.tls.private_key_file }}"{{ "," if listener.tls.cacert_chain_file }}
+          "private_key_file": "{{ listener.tls.private_key_file }}"{{ "," if listener.tls.alpn_protocols }}
+          {%- if listener.tls.alpn_protocols -%}
+          "alpn_protocols": "{{ listener.tls.alpn_protocols }}"
+          {%- endif -%}{{ "," if listener.tls.cacert_chain_file }}
         {%- endif -%}
         {%- if listener.tls.cacert_chain_file -%}
           "ca_cert_file": "{{ listener.tls.cacert_chain_file }}"{{ "," if listener.tls.cert_required }}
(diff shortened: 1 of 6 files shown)


kflynn:
Hey, great to see this! Could you add some tests as well? You can probably just add the `alpn_protocol` setting in `ambassador/tests/002-tls-module-1` and verify that the generated files are what you expect. 

Since the docs on creating and modifying the tests haven't been written (sigh), here goes:

----

The CI tests live in `ambassador/tests`; each subdirectory starting with a three-digit number contains a test. Tests run on every build of Ambassador, and test Ambassador's ability to transform a known Ambassador configuration into a known intermediate configuration and a known Envoy configuration.

Call the test directory (e.g. `ambassador/tests/002-tls-module-1`) `$TESTDIR`. Then:

- The test's Ambassador configuration is the set of YAML files in `$TESTDIR/config`.

- The intermediate configuration is generated by `ambassador dump`.
   - The known good intermediate configuration is in `$TESTDIR/gold.intermediate.json`.
   - When the test runs, it generates `$TESTDIR/intermediate.json` and compares it with the known good version.

- The Envoy configuration is generated by `ambassador config`.
   - The known good Envoy configuration is in `$TESTDIR/gold.json`.
   - When the test runs, it generates `$TESTDIR/envoy.json` and compares it with the known good version.

To extend a test, change the configuration in `$TESTDIR/config`, then run a build (`make DOCKER_REGISTRY=-`) which will run all the CI tests. The tests will fail - if not, you didn't really change the test configuration, right? - but they'll output the diffs they found against the known-good versions. If all looks well, you can update the known-good versions:

- `mv $TESTDIR/intermediate.json $TESTDIR/gold.intermediate.json`
- `mv $TESTDIR/envoy.json $TESTDIR/gold.json`

and then commit all the changes you made in `$TESTDIR`.

----

Let me know if you have trouble -- best way is probably to reach me on the [Ambassador Gitter channel](https://gitter.im/datawire/ambassador). And thanks again!!

PR #306, opened 2018-03-07
update to newer alpine to fix vulns

richarddli:
The quay scanner has found some medium security vulnerabilities in our base StatsD image. I've verified that upgrading to the 9 version of the alpine-node image fixes this problem.

I *don't* have e2e tests for statsd, so I haven't actually tested functionality.


kflynn:
Superseded by #322.

PR #321, opened 2018-03-14
regex prefix matching 

### LINKED ISSUE #302: Users want support for regex path (prefix) matching
Envoy supports this; It could be nice that Ambassador expose it.

Working thougth is a `regex_prefix` element in a Mapping, parallel to `prefix`.

radu-c:
option to allow prefix to be a regex.

fixes #302

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@
 
 - As of **0.22.0**, Ambassador is distributed via `quay.io` rather than DockerHub. If you are not using Datawire's published Kubernetes manifests, you will have to update your manifests!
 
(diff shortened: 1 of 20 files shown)


radu-c:
The solution is probably not ideal but it works.

Initially I was going for keeping it envoy compatible adding a regex, but prefix is wired in too many places and making optional threw me to the lino's den.

The build fails due to some missing AWS credentials it seems.

radu-c:
@richarddli @kflynn is there anything else I would need to do here to get this merged ?

kflynn:
Running tests now, hoping to land this today. Thanks!

PR #345, opened 2018-03-26
Disable Istio automatic sidecar injection

majelbstoat:
This adds an annotation to the Ambassador deployments to skip automatic sidecar injection for istio as described [here](https://istio.io/docs/setup/kubernetes/sidecar-injection.html#policy). It also updates the docs to say this is taken care of for you.


majelbstoat:
(I assumed I needed to update the prometheus rbac too, though I haven't used it personally, so I don't know for sure.)

Issue #354, opened 2018-04-02
Prevent "/ambassador/v0/diag/" from shadowing other routes


dougwilson:
Currently I have Ambassador setup to route HTTP into different services based on the incoming hostname, but it seems like the "/ambassador/v0/diag/" route will always respond no matter the incoming hostname. This page includes all the routes and information like Ambassador version number and internal hostname. Is it possible to disable this page and/or add routing rules to the routing rules (like, for example, restrict this page to only show up for a specific hostname)? Also, I may be configuring something wrong that is causing this page to show up for everything, so any help to point in the right direction would be awesome 👍

PR #355, opened 2018-04-02
Mixin the diagnostics section from module definition

dougwilson:
So I'm trying to disable the diagnostics route mappings (from #354) and I think that if the module would allow one to mixin the `diagnostics` section I can use an annotation to disable them. This is my (bad) attempt to try and implement this change. I am still struggling to get `make DOCKER_REGISTRY=-` to run (just keep getting `error: [Errno 2] No such file or directory: 'requirements.txt'`) and I'm not really used to Python so any pointers I would love to get it going and add some tests, even, assuming I'm going down the right path 👍

### DIFF
--- a/ambassador/ambassador/config.py
+++ b/ambassador/ambassador/config.py
@@ -1216,7 +1216,7 @@ class Config (object):
         # as we find them.
         for key in [ 'service_port', 'admin_port', 'diag_port',
                      'liveness_probe', 'readiness_probe', 'auth_enabled',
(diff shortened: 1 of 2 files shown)


kflynn:
So sorry for the delay! Yes, I think it'd be fine to have that flag -- you'll probably need to add an explicit `True` default, too.

As for `requirements.txt` -- what directory are you trying to run `make DOCKER_REGISTRY=-` from?

kflynn:
Closing this one in favor of #382 -- thanks!! :)

PR #370, opened 2018-04-10
Rate limits

alexgervais:
Resolves https://github.com/datawire/ambassador/issues/352

Provides: 
* Envoy configuration
  * virtual_hosts routes config
  * filter
  * cluster_manager
  * grpc_service
  * Unit and end-to-end tests
* Documentation
  * RateLimitService
  * rate_limits mapping attribute
  * Using Envoy's gRPC external rate limit service
  * Tutorial
  * Example

### DIFF
--- a/ambassador/tests/009-rate-limit/config/mapping-qotm.yaml
+++ b/ambassador/tests/009-rate-limit/config/mapping-qotm.yaml
@@ -17,8 +17,8 @@ kind: Mapping
 name: qotm_rate_limit_config_mapping
 prefix: /qotm-rate-custom/
 service: qotm
-rate_limits:
-  - {}
-  - descriptor: custom-label
+rate_limits:                 # Configuring 2 rate_limits rules, so it should generate 2 Envoy rate_limits.actions and 2 external rate limit service calls
+  - {}                         # Rule #1: empty dict with default configuration
+  - descriptor: custom-label   # Rule #2: with configuration overrides
     headers:
       - ":authority"
--- a/ambassador/tests/009-rate-limit/config/mapping-qotm.yaml
+++ b/ambassador/tests/009-rate-limit/config/mapping-qotm.yaml
@@ -10,14 +10,15 @@ kind: Mapping
 name: qotm_rate_limit_mapping
 prefix: /qotm-rate/
 service: qotm
-rate_limit: {}
+rate_limits: [{}]
 ---
 apiVersion: ambassador/v0
 kind: Mapping
 name: qotm_rate_limit_config_mapping
 prefix: /qotm-rate-custom/
 service: qotm
-rate_limit:
-  descriptor: custom-label
-  headers:
-    - ":authority"
+rate_limits:
+  - {}
+  - descriptor: custom-label
+    headers:
+      - ":authority"
--- a/ambassador/tests/009-rate-limit/config/mapping-qotm.yaml
+++ b/ambassador/tests/009-rate-limit/config/mapping-qotm.yaml
@@ -1,0 +1,23 @@
+---
+apiVersion: ambassador/v0
+kind: Mapping
+name: qotm_mapping
+prefix: /qotm/
+service: qotm
+---
+apiVersion: ambassador/v0
+kind: Mapping
+name: qotm_rate_limit_mapping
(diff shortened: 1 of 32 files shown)


kflynn:
Awesome to see this, thanks! Comments inline. :)

docs/reference/rate-limit-external.md
```diff
@@ -0,0 +1,25 @@
...
+  {"<some_request_header>", "<header_value_queried_from_header>"}
+]
+```
+
+This gRPC service must implement the Envoy [ratelimit.proto](https://github.com/envoyproxy/envoy/blob/master/source/common/ratelimit/ratelimit.proto).
+
+If Ambassador cannot contact the rate limit service, it will allow the request to be processed as if there were no rate limit service configuration.
+
+If Ambassador receives an `OK` response from the rate limit service, then Ambassador allows the client request to resume being processed by the normal Ambassador Envoy flow. If Ambassador receives an `OVER_LIMIT` response, then Ambassador will return an HTTP 429 response to the client.
+
```
kflynn: I feel like you need to talk a bit about the actual rate limit returned from the rate-limiting service -- e.g. the sample service you provide responds `OK` with a 1000 rps limit. Talking briefly about what  Envoy actually does with that part of the response will help people understand the overall rate-limiting system.

ambassador/tests/009-rate-limit/config/mapping-qotm.yaml
```diff
@@ -0,0 +1,24 @@
...
+rate_limits: [{}]
+---
+apiVersion: ambassador/v0
+kind: Mapping
+name: qotm_rate_limit_config_mapping
+prefix: /qotm-rate-custom/
+service: qotm
+rate_limits:
+  - {}
+  - descriptor: custom-label
```
kflynn: I'm seeing this empty dict a lot; what's the purpose?

If the idea is to test that the code is rejecting a ratelimit object that doesn't have all the necessary data, let's just make the object fail validation if it doesn't have everything, and only have one bad dict that's clearly commented as "this is testing a failure case" so people using the tests as tutorials don't get confused.

If the purpose is something else, clearly I'm confused and need to be enlightened. :)

end-to-end/ratelimit-service/ratelimit.proto
```diff
@@ -0,0 +1,99 @@
+// FROM https://github.com/envoyproxy/envoy/blob/master/source/common/ratelimit/ratelimit.proto
```
kflynn: Should we just copy this proto in from the envoyproxy github? Is your intention here to pin to specific proto version?

kflynn:
OK, I'll pull this down and run the test suite later today, but the changes look good overall -- thanks again!! :)

ambassador/tests/009-rate-limit/config/mapping-qotm.yaml
kflynn: OK, maybe have the empty dict in just one test, with a comment explaining what's going on? We've been finding that people do indeed read the tests as examples. 😄

alexgervais: Sure, adding descriptive comments to the test case is a good idea :)

end-to-end/ratelimit-service/ratelimit.proto
```diff
@@ -1,4 +1,4 @@
-// FROM https://github.com/envoyproxy/envoy/blob/master/source/common/ratelimit/ratelimit.proto
```
kflynn: Can't the end-to-end test literally compile from ../../common/ratelimit/ratelimit.proto?

kflynn:
I'm going to go ahead and merge this. Room for some edits later, perhaps, but let's start working with it. :) Thanks!!

end-to-end/ratelimit-service/ratelimit.proto
alexgervais: TBH, I kind of like to have it all packaged in the same directory, since the example does not depend on other project files and can easily be moved around or extracted to it's own "samples" repo.

It would work if running node locally, but for the Docker image we would need to add an `ADD` instruction to this file, perhaps from the github URL.

PR #371, opened 2018-04-10
Don't log socat notice messages as it is flooding the logs

alexgervais:
Reducing `socat` log level to prevent flooding notice messages to elasticsearch (or whatever log backend)


kflynn:
Oops. Thanks for the catch! :)

PR #382, opened 2018-04-19
Ambassador module configuration flag to enable diagnostics

alexgervais:
Addressing https://github.com/datawire/ambassador/issues/354 and piggybacking on https://github.com/datawire/ambassador/pull/355 .

Leaving the existing behavior as-is: the diagnostic dashboard routes on the service port is enabled by default.

For security reasons we do need to disable it in production, as not to expose active configurations to attackers and allowing them to change the log level.


kflynn:
Hey @alexgervais, can you modify one of the e2e tests to test this? I'd like to include it -- thanks!

alexgervais:
I did a quick check and given the widespread usage of `check_diag` in end-to-end tests, I think it'll take me a little while to figure out how to tackle this test and refactoring... :/

kflynn:
Make a local `check_diag` for one test that uses `kubectl exec` to run the `curl` on the Ambassador pod? That shouldn't be too bad, and should work just fine.

alexgervais:
@kflynn is this good to go?

kflynn:
@alexgervais I'm going to take it after I get 0.33.0 get (long story, working on the release process for Ambassador). I'll probably add a negative test - to make sure that you cannot actually reach the diag service through Ambassador - after that, unless you want to.

Sorry for the delay!

PR #386, opened 2018-04-23
Support TLS origination for auth-service

dvavili:
Add support to offer certificates to external auth-service. Adds a new field `tls: <Boolean|String>` similar to `Mapping`


kflynn:
Looks good! Can you modify one of the test cases to test it?

dvavili:
Done.

dvavili:
CI seems to be failing because of invalid credentials to push, but the test case that I added passes.

kflynn:
Yeah, gotta fix that CI hassle. Thanks!

PR #414, opened 2018-05-02
Fix broken bootstrap css url

sipplified:


kflynn:
Seems like we might need to vendor that grid.css in later. Merging for now.

Issue #417, opened 2018-05-03
Ambassador provides no option for not rewriting a `regex_prefix` and keep its value.


ereOn:
We are using Ambassador and we tried to setup a passthrough mapping like so:

```
apiVersion: ambassador/v0
kind: Mapping
name: foo
prefix: ^/v1/foo/([^/]*)/bar/$
prefix_regex: true
service: foo
```

As specified in the documentation, the URL prefix gets stripped by default and sent to the `foo` service. Fair enough.

But as far as I understand, the only option to not strip the prefix is to use a `rewrite` directive. Doing this however is not feasible in that case because of the placeholder in the regex prefix.

How can I solve that?

kflynn:
I would delighted to take a PR with documentation and a test! It's definitely not a bug that it works this way, but I confess that the original motivation for the empty-string check was protection against an empty element, rather than this feature. ;) So let's make it supported for real. ;)

ereOn:
@kflynn PR is up ! [#418]

ereOn:
As my PR is merged, I consider my issue fixed and closed! Thanks @kflynn for being so reactive!

irl-segfault:
still an issue, `rewrite: ""` is still the solution.  No docs mentioning this, not a great first day with Ambassador as far as confidence building

PR #418, opened 2018-05-03
Added a test for empty rewrite

ereOn:
This PR adds test for the - until now - unofficially supported empty `rewrite` declaration discussed in #417.


ereOn:
For some reason the build does not pass for me, I assume this has something to do with the fact I can't launch build that access protected Travis variables.

The Python unit-tests are passing though, so all should okay.

kflynn:
Thanks much! and yes, I need to revamp CI to not try to mess with things when it can't mess with them. :)

Issue #494, opened 2018-06-06
CLI error when running config with two files with the same name in different subfolders


lavoiedn:
**Describe the bug**
I'm running into an issue while trying to run `ambassador config <k8s_manifest_path> config.json --k8s` with ambassador cli. If I have two subfolders that contain a yaml with the same name, I get the following error:

```
2018-06-06 13:43:16 ambassador 0.34.1 ERROR: EXCEPTION from parse_config: '<' not supported between instances of 'dict' and 'dict'
Traceback (most recent call last):

  File "/home/user/git/ambassador/venv/lib/python3.6/site-packages/ambassador-0.34.1-py3.6.egg/ambassador/cli.py", line 92, in parse_config
    template_dir_path=template_dir_path, schema_dir_path=schema_dir_path)

  File "/home/user/git/ambassador/venv/lib/python3.6/site-packages/ambassador-0.34.1-py3.6.egg/ambassador/config.py", line 336, in __init__
    self.process_all_objects()

  File "/home/user/git/ambassador/venv/lib/python3.6/site-packages/ambassador-0.34.1-py3.6.egg/ambassador/config.py", line 401, in process_all_objects
    for filename, ocount, obj in sorted(self.objects_to_process):

TypeError: '<' not supported between instances of 'dict' and 'dict'
```

**To Reproduce**
Prepare some configuration with a file structure something like this:
```
ambassador-config
└ default
│   └ ambassador.yaml
│   └ qotm.yaml
└ namespace
    └ qotm.yaml
```

I have attached an archive containing the files I used for testing:
[ambassador-example-config.tar.gz](https://github.com/datawire/ambassador/files/2077616/ambassador-example-config.tar.gz)

Then, run:
`ambassador config </path/to/config> ./config.json --k8s`

**Expected behavior**
I expect configuration to support multiple configuration files with the same name on different levels in the file structure.

**Versions (please complete the following information):**
 - Ambassador: 0.34.1
 - Kubernetes environment: N/A (ambassador cli by itself)
 - Version: N/A

**Additional context**
I think this is due to how the objects to sort are built, they use (filename, ocount, obj) so when trying to sort, if `filename` and `ocount` are the same, it falls back to comparing `object_to_process_a[2] == object_to_process_b[2]` (`obj`) which are both dicts and fails with the above error.

[...]

PR #495, opened 2018-06-06
Change objects_to_process tuple to use full filepath

lavoiedn:
My attempt at a fix for #494

### DIFF
--- a/ambassador/ambassador/config.py
+++ b/ambassador/ambassador/config.py
@@ -27,10 +27,10 @@ from .VERSION import Version
 ##
 ## Ambassador configures itself by creating a new Config object, which calls
 ## Config.__init__().
-## 
+##
 ## __init__() sets up all the defaults for everything, then walks over all the
-## YAML it can find and calls self.load_yaml() to load each YAML file. After 
-## everything is loaded, it calls self.process_all_objects() to build the 
+## YAML it can find and calls self.load_yaml() to load each YAML file. After
+## everything is loaded, it calls self.process_all_objects() to build the
 ## config objects.
 ##
 ## load_yaml() does the heavy lifting around YAML parsing and such, including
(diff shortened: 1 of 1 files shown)


lavoiedn:
I just realised this PR could perhaps use a renaming from `filename` to `filepath` in the `process_all_objects` function? Let me know if this fix seems sound and if you'd prefer the variable be renamed.

kflynn:
Yeah, I think I agree that `filename` needs to become `filepath` in `Config.load_yaml` and `Config.process_all_objects`... which leads us to the ugliness of whether to rename `self.filename`. Sadly, I think it's probably less ugly to do that now rather than waiting. 😦

lavoiedn:
I changed some more instances of `filename` but then I realized it's used in the `source` object which would require changing a lot of test json files.

I'll get a test environment setup tomorrow so I can run python tests and try it out in minikube!

lavoiedn:
@kflynn I wrote another possible fix that involves less drastic changes for the code base and the unit tests seem to work with it: https://github.com/datawire/ambassador/compare/master...lavoiedn:fix-process_all_objects-same-filename-redux?expand=1&w=1

If you'd rather I change the "`filename`" key/value in every instance of the `source` dict, I can do that too! I just don't want to break a portion of the code that uses "`filename`". I didn't seem to find any instance in the code that required `source.filename` to be unique across multiple sources, except maybe in the diagnostic html?

```bash
$ grep -r -e "{{ source.filename }}"          
ambassador/templates/diag.html:              <a href="/ambassador/v0/diag/{{ source.filename }}">
ambassador/templates/diag.html:                <a href="/ambassador/v0/diag/{{ source.filename }}">
```

Let me know which solution you prefer! I think using `filepath` everywhere would be cleaner and more of a long-term fix, but I'm not sure of the implications.

lavoiedn:
Okay just confirmed that the second fix does not work with the diagnostics dashboard. We're out on a team building event this afternoon but I'll be able to invest more time into this tomorrow!

PR #503, opened 2018-06-08
Add probes to example auth service, cleans up docs

containscafeine:
Prior to this commit, the user was expected to do a port forward
of the example auth service per the authentication docs, and then
connect to the auth service to make sure it was up and running.
This is an extra step just to verify the health of the service.

This commit adds probes to the example auth service which make
sure the service is running, and the docs are updated to tell the
user to only check the output of `kubectl get pods` to see if the
service is up and running, instead of doing a port forward and
trying to connect.


richarddli:
Great idea! LGTM.

PR #509, opened 2018-06-11
Add instructions for trying gRPC guide with minikube

containscafeine:
This commit adds the instructions to configure networking if
someone is trying out the tutorial on minikube.


containscafeine:
Well, I spent more time than I should have debugging this, so thought would be worth adding a small note in the docs.

PR #510, opened 2018-06-11
Allow running ambassador as a non-privileged user

### LINKED ISSUE #457: Docker container security
We are running sysdig-falco in our Kubernetes cluster and it is complaining about the following:

`
k8s.node_name=ip-172-16-12-29.ec2.internal 10:53:58.969608520: Error File below /etc opened for writing (user=root command=python3 /application/kubewatch.py sync /etc/ambassador-config /etc/envoy.json parent=entrypoint.sh pcmdline=entrypoint.sh ./entrypoint.sh file=/etc/ambassador-config-1/payment-service-default.yaml program=python3 gparent=<NA> ggparent=<NA> gggparent=<NA>) k8s.pod=<NA> container=21b0bf1f68db
`

I would suggest changing the configuration location from `/etc/ambassador-config` to simply `/ambassador-config`.

I also strongly feel like Ambassador's processes (`entrypoint.sh`, `python3` and `envoy`) should be running as non-root user.

containscafeine:
This commit lets ambassador to be run as a non-root user and
moves all ambassador related configurations to /ambassador
inside the container.

### DIFF
--- a/ambassador/Dockerfile
+++ b/ambassador/Dockerfile
@@ -21,18 +21,18 @@ LABEL PROJECT_REPO_URL         = "git@github.com:datawire/ambassador.git" \
       VENDOR_URL               = "https://datawire.io/"
 
 # This Dockerfile is set up to install all the application-specific stuff into
-# /application.
+# /ambassador.
 #
 # NOTE: If you don't know what you're doing, it's probably a mistake to
 # blindly hack up this file.
 
 RUN apk --no-cache add curl python3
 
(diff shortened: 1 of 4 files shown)


containscafeine:
@kflynn ready for review, PTAL!
One way to test this locally to run the resulting container image using with a random user -
`docker run --user 10001 --rm <image name>`

containscafeine:
Closing in favor of #524

PR #520, opened 2018-06-13
Fix process all objects same filename redux

lavoiedn:
Filepath was not very helpful/clean in logs in some cases so I tried to use `<metadata.name>.<metadata.namespace>` where available.

Let me know if this does not work for you guys, there are still some issues left such as `AMBASSADOR_NAMESPACE` and `AMBASSADOR_SINGLE_NAMESPACE` are not taken into account.

### DIFF
--- a/ambassador/ambassador/config.py
+++ b/ambassador/ambassador/config.py
@@ -27,10 +27,10 @@ from .VERSION import Version
 ##
 ## Ambassador configures itself by creating a new Config object, which calls
 ## Config.__init__().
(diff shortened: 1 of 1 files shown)


kflynn:
The CLI has always been more of a development/testing tool than anything else, so to the extent that I thought about it, I always kind of assumed that you'd only include things for the namespace you wanted in the config you fed to the Ambassador via the CLI. Is there something that'd work better for you there?

Issue #454, opened 2018-05-22
Document and speed up E2E test harness


plombardi89:
The current E2E test harness has some issues we would like to address:

1. It does not execute in parallel.
2. The test harness itself is a hard to debug assembly of shell scripts.
3. Has its own share of flakiness issues (e.g. recently we noticed it failing but still reporting status code 0 and ultimately succeeding).
4. It's slow enough that it doesn't run on anything except RC builds.

With the E2E suite we want to design and implement something that will:

1. Execute quickly enough not to cause agony (understanding that E2E tests are slower than fail-fast system tests)
2. Easy to debug and extend the harness itself.
3. Easy to add new positive and negative test cases as needed.
4. Be mostly or entirely reusable when / if we decide to upgrade Ambassador to use Envoy v2 config or the Envoy ADS.

# Design CoS

- [ ] Discuss and document the current issues with the Ambassador E2E suite. Documentation can be appended to this issue.
- [ ] Write up thoughts around a replacement or alternative to the E2E approach.
- [ ] Design and prototype a new E2E suite. The person designing should at a minimum discuss the design with one other party before further implementation occurs.
- [ ] Ensure the design has clear answers to the above four bullet points.

# Implementation CoS
 
- [ ] The new E2E suite should execute tests in parallel.
- [ ] The new E2E suite should have all the current E2E tests ported into it.
- [ ] The new E2E suite should be able to be run locally or in CI in the same way.
- [ ] Documentation for how to extend or debug the harness.
- [ ] Documentation for how to implement additional test cases.
- [ ] Should be possible to selectively run some E2E tests.

# Additional Nice To Haves

- [ ] It would be nice if the E2E provisioning machinery could use Forge so we continue to eat our own dogfood.

kflynn:
OK, here we go!

## Ambassador Testing

Like most real-world software, Ambassador comprises multiple functional units that work together to get a job done, and we need to test all of them to have confidence that development is improving the product, rather than breaking it. Of course, not all the units can be tested the same way, which makes covering them entertainingly tricky.

### Functional Units 

At a very high level, Ambassador's functional units are:

1. The _translation engine_ takes an Ambassador configuration and spits out an Envoy configuration. It mostly lives in `ambassador/ambassador/config.py`.
[...]

PR #526, opened 2018-06-14
Allow arbitrary secrets for TLS config

### LINKED ISSUE #457: Docker container security

### LINKED ISSUE #463: Users want to be able to load certificates from arbitrary secrets
Presently, Ambassador hardcodes certain Kubernetes secret names for TLS certificates. This is problematic both for TLS origination and for running multiple Ambassadors: a better way would be to allow Ambassador to specify which secret to load a given certificate from, e.g.:

```
apiVersion: ambassador/v0
kind:  Module
name:  tls
config:
  server:
    enabled: True
    secret: my-secret
```

or 

```
---
apiVersion: ambassador/v0
kind:  Mapping
name:  qotm_mapping
prefix: /qotm/
tls: 
  secret: qotm-upstream
service: qotm
```

or the like.

containscafeine:
Allow arbitrary secrets for TLS config
    
    Before this commit, hardcoded secret names "ambassador-certs" and
    "ambassador-cacert" were required to be configured for TLS
    termination and TLS client authentication.
    
    This commit lets users configure both of these secrets in the TLS
    module annotation.
    
    This change is backwards compatible, and if a user has the
    hardcoded secrets set, then it will continue to work as usual and
    hardcoded secrets will take precedence over manually configured
    secrets.
    
    Some refactor that has been done includes taking some Kubernetes
    specific helper functions from kubewatch.py and moving those to
    ambassador/utils.py.

Fix #463


containscafeine:
@kflynn I had some questions regarding the approach -
1. I understand that the entire secrets logic need to be taken out of kubewatch.py, and moved into config.py. Is that correct?
2. Right now, since we support only one secret, we store that one information at /etc/certs/tls.{crt,key} and /etc/cacert/tls.crt, and then update the restarter. How do you suggest we do this with multiple secrets? One way would be to store the secrets at `/etc/certs/server_name/tls.{crt,key} and /etc/cacert/server_name/tls.crt. Another ways could be to store in the cluster itself as another secret or configMap. What do you think?
3. Currently, once the secrets are found, we update in tls_mod like -
```python
                tls_mod = {
                    "apiVersion": "ambassador/v0",
                    "kind": "Module",
                    "name": "tls-from-ambassador-certs",
                    "ambassador_id": ambassador_id,
                    "config": {
                        "server": {
                            "enabled": True,
                            "cert_chain_file": "/etc/certs/tls.crt",
                            "private_key_file": "/etc/certs/tls.key"
                        }
                    }
                }
```
and then trigger the restarter update. How do we trigger the restarter update now? Should we make config.py trigger the restarter update?
4. Right now, we watch for updates at startup using https://github.com/datawire/ambassador/blob/95969cbf4cf085684ecdfbed5556f1e711161d79/ambassador/kubewatch.py#L331, how do we watch for updates now with multiple certs?
5. From what I could find online, we'll need to configure SNI in envoy to enable this - https://www.envoyproxy.io/docs/envoy/latest/faq/sni#faq-how-to-setup-sni, is that correct? I will have to update the envoy.j2 to accommodate this, right?

Not very familiar with the codebase right now, so all these questions. Would appreciate any help :)

kflynn:
These are good questions that may or may not have good answers. ;)

We have conflicting pressures here on `Config` and `kubewatch` (especially in light of the discussion in #454):

- `Config` should really understand the _entire_ Ambassador configuration, but
- `Config` really should _not_ understand the interface with Kubernetes (because `Config` needs to work when we're running standalone for tests, or in Docker for the quickstart). Finally,
- `Config` really should not be responsible for restarting Envoy. That goes double as we think ahead to ADS.

Right now, `kubewatch` understands a bit more of the Ambassador config than it really should, because it's `kubewatch` that understands how to take a `Secret` or a `ConfigMap` and generate Ambassador configuration resources from them. On the other hand, cf #454: we're going to be splitting `kubewatch` in two anyway, most likely. So we can start thinking in terms of `configmgr` and `kubewatch`.

In that light, it feels like this is the sequence we're after:

- Is there stuff present on the filesystem?
   - if so, read each file and send its resources to `configmgr`.
- Is an Ambassador `ConfigMap` present?
   - if so, read its resources and send them to `configmgr`.
- Are any secrets present?
   - if so, read their resources and send them to `configmgr`.
- Are there services present?
   - if so, read their resources and send them to `configmgr`.
- Finally, watch for changes to services or secrets.
   - when they happen, send the new resources or a `DELETE` to `configmgr`.

(The first four steps can possibly be a new program entirely, rather than a mode of `kubewatch`, actually.)

In the short term, we can probably move this direction by separating the initialization and parsing steps in the `Config` class, and giving `Config` an incremental `add` method that accepts a new resource, of whatever type, so that you work with things by:

```python
config = Config()

for resource in random_resource_list:
    config.add(resource)

config.generate_envoy_config()
```

or the like.

@containscafeine, we should talk this over further once @plombardi89 and @rhs and I talk over #454 -- so hopefully tomorrow will be good for that.

ambassador/ambassador/config.py
```diff
@@ -1312,6 +1312,16 @@ def tls_config_helper(self, name, amod, tmod):
...
                     tmp_config.update(tmod['server'])
+
+                     # Check if secrets are supplied for TLS termination and/or TLS auth
+                    secret = context.get('secret')
+                    if secret is not None:
+                        self.logger.info("config.server.secret is {}".format(secret))
+                        (server_cert, server_key, server_data) = read_cert_secret(secret)
+                        if server_cert and server_key:
+                            self.logger.info("saving server cert")
+                            save_cert(server_cert, server_key, TLSPaths.cert_dir.value)
```
containscafeine: add check before overwriting

containscafeine:
Testing:

1.
- `kubectl create secret tls ambassador-certs --cert old-cert.pem --key old-key.pem`
- No other TLS Module mapping present
- `kubectl delete pods -l service=ambassador`
- Cert loaded which points to old.minikube.192.168.39.236.xip.io - :green_heart:, everything works fine

2.
- `kubectl delete secret ambassador-certs`
- `kubectl create secret tls new-secret --cert new-cert.pem --key new-key.pem`
- added mapping to service -
```yaml
      ---
      apiVersion: ambassador/v0
      kind:  Module
      name:  tls
      config:
        server:
          enabled: True
          secret: new-secret
```
- `kubectl delete pods -l service=ambassador`
- New secret loaded, cert points to `new.minikube.192.168.39.236.xip.io` - :green_heart:, everything works fine

3.
- `kubectl create secret tls ambassador-certs --cert old-cert.pem --key old-key.pem`
-
```console
$ kubectl get secrets
NAME                     TYPE                                  DATA      AGE
ambassador-certs         kubernetes.io/tls                     2         16s
new-secret               kubernetes.io/tls                     2         2m
```
- The TLS Module mapping exists
- `kubectl delete pods -l service=ambassador`
- New secret ignored, `ambassador-certs` loaded, points to `old.minikube.192.168.39.236.xip.io` - :green_heart:

containscafeine:
1.
- `kubectl create secret generic ambassador-cacert --from-file=tls.crt=/home/concaf/ambassador/old-cert.pem --from-literal=cert_required=true`
- `kubectl delete pods -l service=ambassador`
- contents of old-cert.pem and /etc/cacert/tls.crt inside the pod are the same :green_heart: 

2.
- `kubectl delete secret ambassador-cacert`
- `kubectl create secret generic new-cacert --from-file=tls.crt=/home/concaf/ambassador/new-cert.pem --from-literal=cert_required=true`
-
```yaml
      ---
      apiVersion: ambassador/v0
      kind:  Module
      name:  tls
      config:
        server:
          enabled: True
          secret: new-secret
        client:
          enabled: True
          secret: new-cacert
```
- `kubectl delete pods -l service=ambassador`
- Contents of new-cert.pem and /etc/cacert/tls.crt inside the pod are the same :green_heart: 

3.
- `kubectl create secret generic ambassador-cacert --from-file=tls.crt=/home/concaf/ambassador/old-cert.pem --from-literal=cert_required=true`
- `new-cacert` and `ambassador-cacert` both secrets exist
- `kubectl delete pods -l service=ambassador`
- Contents of old-cert.pem and /etc/cacert/tls.crt are the same, `new-cacert` has been ignored despite the mapping :green_heart:

kflynn:
Taking into another shared/develop.

PR #529, opened 2018-06-15
Move YAML strings to YAML arrays for CORS configuration

### LINKED ISSUE #466: Support YAML arrays for CORS configuration
At the moment CORS settings use comma separated lists.
It would be nice to supoort YAML arrays as it increases readability.

_Example:_
- Current format:
  ```
  cors:
    methods: POST, GET, OPTIONS
    headers: Content-Type,Accept,Authorization
    credentials: true
    max_age: "86400"
  ```
- Enhanced format:
  ```
  cors:
    methods: 
      - POST
      - GET
      - OPTIONS
    headers: 
      - Content-Type
      - Accept
      - Authorization
    credentials: true
    max_age: "86400"
  ```

containscafeine:
This commit makes defining CORS configuration in ambassador more
readable by moving comma separated strings, which envoy supports,
to YAML arrays.

The previous format looks like -
```yaml
cors:
  methods: POST, GET, OPTIONS
  headers: Content-Type,Accept,Authorization
  credentials: true
  max_age: "86400"
```

The new format looks like -
```yaml
cors:
  origins:
  - http://foo.example
  - http://bar.example
  methods:
  - POST
  - GET
  - OPTIONS
  headers:
  - Content-Type
  - Accept
  - Authorization
  credentials: true
  exposed_headers:
  - X-Custom-Header
  max_age: "86400"
```

Fix #466

### DIFF
--- a/ambassador/ambassador/mapping.py
+++ b/ambassador/ambassador/mapping.py
@@ -122,6 +122,54 @@ class Mapping (object):
         else:
             return self.attrs.get(key)
 
+    def save_cors_element(self, cors_key, route_key, route):
+        """If self.get('cors')[cors_key] exists, and
+        - is a list, e.g. ["1","2","3"], then route[route_key] is set as "1, 2, 3"
+        - is something else, then set route[route_key] as it is
+
+        :param cors_key: key to exist in self.get('cors'), i.e. Ambassador's config
+        :param route_key: key to save to in envoy's cors configuration
+        :param route: envoy's cors configuration
+        """
+        cors = self.get('cors')
+        if cors.get(cors_key) is not None:
+            if type(cors.get(cors_key)) is list:
+                route[route_key] = ", ".join(cors.get(cors_key))
+            else:
+                route[route_key] = cors.get(cors_key)
+
+    def generate_route_cors(self):
+        """Generates envoy's cors configuration from ambassador's cors configuration
+
+        :return generated envoy cors configuration
+        :rtype: dict
+        """
+
+        cors = self.get('cors')
+        if cors is None:
+            return
+
+        route_cors = {'enabled': True}
+        # cors['origins'] cannot be treated like other keys, because if it's a
+        # list, then it remains as is, but if it's a string, then it's
+        # converted to a list
+        origins = cors.get('origins')
+        if origins is not None:
+            if type(origins) is list:
+                route_cors['allow_origin'] = origins
+            elif type(origins) is str:
+                route_cors['allow_origin'] = origins.split(',')
+            else:
+                print("invalid cors configuration supplied - {}".format(origins))
+                return
+
+        self.save_cors_element('max_age', 'max_age', route_cors)
+        self.save_cors_element('credentials', 'allow_credentials', route_cors)
+        self.save_cors_element('methods', 'allow_methods', route_cors)
+        self.save_cors_element('headers', 'allow_headers', route_cors)
+        self.save_cors_element('exposed_headers', 'expose_headers', route_cors)
+        return route_cors
+
     def new_route(self, svc, cluster_name):
         route = SourcedDict(
             _source=self['_source'],
@@ -165,17 +213,9 @@ class Mapping (object):
             for key, value in add_request_headers.items():
                 route['request_headers_to_add'].append({"key": key, "value": value})
 
-        cors = self.get('cors')
-        if cors:
-            route['cors'] = {key: value for key, value in {
-                "allow_origin": cors['origins'].split(',') if cors['origins'] else None,
-                "allow_methods": cors.get('methods'),
-                "allow_headers": cors.get('headers'),
-                "expose_headers": cors.get('exposed_headers'),
-                "allow_credentials": cors.get('credentials'),
-                "max_age": cors.get('max_age'),
-                "enabled": True
-            }.items() if value}
+        envoy_cors = self.generate_route_cors()
+        if envoy_cors:
+            route['cors'] = envoy_cors
 
         rate_limits = self.get('rate_limits')
         if rate_limits != None:
--- a/ambassador/schemas/v0/Mapping.schema
+++ b/ambassador/schemas/v0/Mapping.schema
@@ -25,11 +25,31 @@
         "cors": {
             "type": "object",
             "properties": {
-                "origins":  { "type": "string" },
-                "methods": { "type": "string" },
-                "headers": { "type": "string" },
+                "origins":  {
+                    "anyOf": [
+                        { "type": "string" },
+                        { "type": "array", "items": { "type": "string" } }
+                    ],
+                },
+                "methods": {
+                    "anyOf": [
+                        { "type": "string" },
+                        { "type": "array", "items": { "type": "string" } }
+                    ],
+                },
+                "headers": {
+                    "anyOf": [
+                        { "type": "string" },
+                        { "type": "array", "items": { "type": "string" } }
+                    ],
+                },
                 "credentials": { "type": "boolean" },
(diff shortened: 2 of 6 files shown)


containscafeine:
@kflynn yet to fix docs and add comments, but is the approach fine?

richarddli:
Does this change require a move to YAML arrays? We want to make sure we don't break backwards compatibility.

containscafeine:
Oops, yes this is backward incompatible :man_facepalming:

alexgervais:
My 2 cents, but the reason these were strings is because the values map to HTTP headers, which are themselves CSV values:
```
Access-Control-Allow-Methods: POST, GET
Access-Control-Allow-Headers: X-PINGOTHER, Content-Type
```

I also find it strange for `Access-Control-Allow-Origin` to be an array containing a single `*` value.

kflynn:
In other places (e.g. the `ambassador_id` element) we allow both strings and arrays, so any of these would be fine:

```
cors:
  methods: POST, GET

cors:
  methods:
  - POST
  - GET

cors:
  methods: [ "POST", "GET" ]
```

I don't want to require the array syntax. I'm fine with supporting it, though.

plombardi89:
I don't think I would find `["*"]` too weird in @alexgervais case. It is just configuration after all. Since the real request here was about improving readability I'm wondering if we have one other option: Could folks just use a multi line YAML string? Does Envoy know how to handle newlines?

I think I still prefer the structured format in general though. IMO I would not long term support the non-array format, but that's a different issue.

containscafeine:
Okay, I've made the changes to support both, the older and the newer syntax. Updated and added tests and docs to reflect this. Not particularly happy about the docs update, PTAL.

plombardi89:
Is there a particular concern you have wrt the doc update?

containscafeine:
@plombardi89 it just does not look pretty, that's all!

kflynn:
I think the docs are OK.

ambassador/ambassador/mapping.py
```diff
@@ -122,6 +122,46 @@ def get(self, key, *args):
         else:
             return self.attrs.get(key)
 
+    @staticmethod
+    def add_to_dict_if_key_exists(key1, dict1, key2, dict2):
+        """If dict1[key1] exists, and
```
kflynn: This should probably be something like `cors_save_element`. That's semantically narrow enough that it's OK for it to modify `self.cors` instead of having to have a dict passed in. 

```python
def cors_save_element(self, cors_key, params, param_key):
    ...
    self.cors[cors_key] = params[param_key] # or whatever
    ...
```

and, later, 

```python
self.cors_save_element('credentials', route_cors, 'allow_credentials')
```

ambassador/ambassador/mapping.py
```diff
@@ -122,6 +122,46 @@ def get(self, key, *args):
...
+            if type(dict1.get(key1)) is list:
+                dict2[key2] = ", ".join(dict1.get(key1))
+            else:
+                dict2[key2] = dict1.get(key1)
+
+    def generate_route_cors(self, cors, route_cors):
+        """Generates envoy's cors configuration from ambassador's cors configuration
+
+        :param cors: dict that contains ambassador's representation of cors
+        :param route_cors: envoy's route's cors configuration dict
```
kflynn: Same here -- that's just `self.cors`, no?

ambassador/ambassador/mapping.py
```diff
@@ -166,16 +206,9 @@ def new_route(self, svc, cluster_name):
...
-                "allow_methods": cors.get('methods'),
-                "allow_headers": cors.get('headers'),
-                "expose_headers": cors.get('exposed_headers'),
-                "allow_credentials": cors.get('credentials'),
-                "max_age": cors.get('max_age'),
-                "enabled": True
-            }.items() if value}
+        if cors is not None:
+            route['cors'] = dict()
+            self.generate_route_cors(cors, route['cors'])
```
kflynn: This should probably look like

```python
route['cors`] = self.generate_route_cors(cors)
```

instead of using the `route['cors']` dict as an output parameter.

kflynn:
Some comments inline. Sorry for flipflopping on some review comments!

ambassador/ambassador/mapping.py
containscafeine: makes sense, fixed

ambassador/ambassador/mapping.py
containscafeine: fixed

ambassador/ambassador/mapping.py
```diff
@@ -122,6 +122,46 @@ def get(self, key, *args):
...
+
+        # cors['origins'] cannot be treated like other keys, because if it's a
+        # list, then it remains as is, but if it's a string, then it's
+        # converted to a list
+        if cors.get('origins') is not None:
+            if type(cors.get('origins')) is list:
+                route_cors['allow_origin'] = cors.get('origins')
+            elif type(cors.get('origins')) is str:
+                route_cors['allow_origin'] = cors.get('origins').split(',')
+
```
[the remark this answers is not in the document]
containscafeine: I've added -
```python
            else:
                print("invalid cors configuration supplied - {}".format(cors.get('origins')))
                return 
```
not sure what the idiomatic way to do this is.

containscafeine:
@kflynn np, the more reviews the merrier. Made the changes, PTAL.

ambassador/ambassador/mapping.py
```diff
@@ -122,6 +122,53 @@ def get(self, key, *args):
...
+        """If self.get('cors')[cors_key] exists, and
+        - is a list, e.g. ["1","2","3"], then route[route_key] is set as "1, 2, 3"
+        - is something else, then set route[route_key] as it is
+
+        :param cors_key: key to exist in self.get('cors'), i.e. Ambassador's config
+        :param route_key: key to save to in envoy's cors configuration
+        :param route: envoy's cors configuration
+        """
+        cors = self.get('cors')
+        if cors.get(cors_key) is not None:
```
kflynn: Won't this raise exceptions if `cors_key` isn't present in `cors`? (This applies to many places here... sorry I missed it last time!)

ambassador/ambassador/mapping.py
```diff
@@ -122,6 +122,53 @@ def get(self, key, *args):
...
+
+        cors = self.get('cors')
+        if cors is None:
+            return
+
+        route_cors = {'enabled': True}
+        # cors['origins'] cannot be treated like other keys, because if it's a
+        # list, then it remains as is, but if it's a string, then it's
+        # converted to a list
+        if cors.get('origins') is not None:
```
kflynn: This block will be way simplified by using a temporary variable for `cors.get('origins')` -- it won't make a difference in terms of speed, of course, but it'll make a difference in terms of ease of understanding for the next guy.

ambassador/ambassador/mapping.py
containscafeine: @kflynn will it? Just tried -
```python
>>> a={"a":1,"b":2}
>>> if a.get("c") is not None:
...     print("hi")
... 
```
and it didn't raise an exception even though "c" is not present in dict a.

ambassador/ambassador/mapping.py
containscafeine: fixed

ambassador/ambassador/mapping.py
kflynn: ...just rechecked the docs and I don't know WTF I was thinking. :) :) You're right.

containscafeine:
:tada:

Issue #532, opened 2018-06-16
kubewatch.py busy loop when kube API call fails


mechpen:
**Describe the bug**
We seen lots of failure ``GET /api/v1/services?watch`` from kubewatch.py with a very high frequency.

**To Reproduce**
1. run kubewatch.py against a dummy kube API server which returns error on the API call

**Expected behavior**
Should sleep on failure API call

kflynn:
Fixed by #533

PR #533, opened 2018-06-16
fix kubewatch busy loop issue (#532)

mechpen:


kflynn:
Well that's embarrassing. :) Thanks!!

PR #534, opened 2018-06-18
Add command in docs to check for RBAC

containscafeine:
This commit updates the docs for users to verify is their
Kubernetes installations have RBAC enabled or not. Maybe not the
perfect solution, but a step towards it.

### DIFF
--- a/docs/user-guide/getting-started.md
+++ b/docs/user-guide/getting-started.md
@@ -6,7 +6,15 @@ Ambassador is designed to allow service authors to control how their service is 
 
 ## 1. Deploying Ambassador
 
-To deploy Ambassador in your default namespace, run this command if you're running in a cluster with RBAC enabled:
+To deploy Ambassador in your default namespace, first you need to check if Kubernetes has RBAC enabled, run:
+
+```shell
+kubectl cluster-info dump --namespace kube-system | grep authorization-mode
+```
+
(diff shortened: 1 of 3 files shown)


docs/user-guide/getting-started.md
```diff
@@ -18,6 +18,8 @@ Without RBAC, you can use:
 kubectl apply -f https://getambassador.io/yaml/ambassador/ambassador-no-rbac.yaml
 ```
 
+Note: To check if Kubernetes has RBAC enabled, run `kubectl cluster-info dump --namespace kube-system | grep authorization-mode`. If you see something like `--authorization-mode=Node,RBAC` in the output, then RBAC is enabled.
+
 For production configurations, we recommend you download these YAML files as your starting point, and customize them accordingly (e.g., your namespace).
```
richarddli: I'd put this after the "For production configurations" ... weird to have 2 notes separated by a paragraph of non-note text.

kflynn: I agree with @richarddli -- move the check to right _before_ the choice of which file to use, so they can use the information to get the correct file, rather than using it to find out that they picked the wrong one. ;)

containscafeine: @richarddli @kflynn ack, I've made some changes, PTAL

PR #536, opened 2018-06-19
Added support for service annotations.

psychopenguin:
With this, one can configure the service with custom annotations, for
example:

```
service:
  annotations:
    cloud.google.com/load-balancer-type: "Internal"
```


kflynn:
Merged by hand (with a couple more changes) into `flynn/rel/0.35.1` -- thanks! and sorry for the delay getting to this.

PR #537, opened 2018-06-19
revert to envoy 1.5

plombardi89:


kflynn:
We've decided that we won't need this. Thanks @plombardi89!

PR #558, opened 2018-06-27
document running as non-root

richarddli:


kflynn:
Cherry-picked onto flynn/rel/0.35.0-docs.

PR #567, opened 2018-06-28
Move imagePullPolicy to IfNotPresent from Always

containscafeine:
This commit sets imagePullPolicy for ambassador images to
    IfNotPresent instead of Always, because for running the end to end
    tests locally, where the images are directly pushed to the
    registry inside, for e.g., minikube, the pods fail while pulling
    the images from external registry because they don't exist.


plombardi89:
We actually have it set this way for more accurate metrics. A better solution here is probably to have a way to configure that option in the manifest before deploying.

containscafeine:
Okay, going to close this then, this should be covered in the devex stuff

PR #580, opened 2018-07-02
Enable L7 redirection based on X-FORWARDED-PROTO

### LINKED ISSUE #469: Support global level http -> https redirection
**Use Case**
If TLS terminates externally (Such as AWS ELB), the http request need to be redirected to https.

**Describe the solution you'd like**
The solution that I can think of is twofold. One, to be able to identify and match the protocol. (http/https), and second to be able to add *permenant redirection* for a URL. Both have their own use cases to be able to be used separately.

A config for permanent redirection may look like

```
prefix: /
port: 80
ssl: off
redirect: https://<hostname>
redirection_type: permanent
```

**Describe alternatives you've considered**
Another alternative is to use something like cloudflare to do the redirection.

**Additional context**
Currently we ended up using AWS cloud front for doing redirection. However, cloudfront is not able to handle websocket. Hence we may have to move to nginx-ingress to be able to do this.

containscafeine:
This commit lets users specify a `x_forwarded_proto_redirect`
    field in their ambassador mapping which accepts a boolen value.
    
    An example mapping looks like -
    
    ```yaml
        getambassador.io/config: |
          ---
          apiVersion: ambassador/v0
          kind:  Mapping
          name:  httpbin_mapping
          prefix: /httpbin/
          service: httpbin.org:80
          host_rewrite: httpbin.org
          x_forwarded_proto_redirect: true
    ```
    
    When set to true, ambassador, and in turn envoy, behaves like the
    following -
    
    If an incoming request has `X-FORWARDED-PROTO: https` header
    set, then no redirects take place and a normal response if sent.
    
    If `X-FORWARDED-PROTO` is absent or sent to anything else, like
    `http`, then a 301 redirect to https://host is sent as response.
    
    Keep in mind that ambassador does not have access to any TLS certs
    in the entire workflow. This is typically useful in cases when TLS
    termination happens at an external L7 load balancer, which sets
    the protocol of the incoming request in the `X-FORWARDED-PROTO`
    header of the resulting HTTP request after TLS termination.
    
    Partially fixes #469

### DIFF
--- a/ambassador/ambassador/config.py
+++ b/ambassador/ambassador/config.py
@@ -994,6 +994,7 @@ class Config (object):
             diagnostics = { "enabled": True },
             tls_config = None,
             use_proxy_proto = False,
+            x_forwarded_proto_redirect = False,
         )
 
         # Next up: let's define initial clusters, routes, and filters.
@@ -1099,6 +1100,14 @@ class Config (object):
 
         if 'use_remote_address' in self.ambassador_module:
             primary_listener['use_remote_address'] = self.ambassador_module['use_remote_address']
+
+        # If x_forwarded_proto_redirect is set, then we enable require_tls in primary listener, which in turn sets
+        # require_ssl to true in envoy config. Once set, then all requests that contain X-FORWARDED-PROTO set to
+        # https, are processes normally by envoy. In all the other cases, including X-FORWARDED-PROTO set to http,
+        # a 301 redirect response to https://host is sent
+        if self.ambassador_module.get('x_forwarded_proto_redirect', False):
+            primary_listener['require_tls'] = True
+            self.logger.debug("x_forwarded_proto_redirect is set to true, enabling 'require_tls' in listener")
 
         redirect_cleartext_from = None
         tmod = self.ambassador_module.get('tls_config', None)
@@ -1451,7 +1460,7 @@ class Config (object):
         # as we find them.
         for key in [ 'service_port', 'admin_port', 'diag_port',
                      'liveness_probe', 'readiness_probe', 'auth_enabled',
-                     'use_proxy_proto', 'use_remote_address', 'diagnostics' ]:
+                     'use_proxy_proto', 'use_remote_address', 'diagnostics', 'x_forwarded_proto_redirect' ]:
             if amod and (key in amod):
                 # Yes. It overrides the default.
                 self.set_config_ambassador(amod, key, amod[key])
--- a/ambassador/schemas/v0/Mapping.schema
+++ b/ambassador/schemas/v0/Mapping.schema
@@ -17,7 +17,6 @@
         "prefix": { "type": "string" },
         "prefix_regex": { "type": "boolean" },
         "service": { "type": "string" },
-
         "add_request_headers": { "$ref": "#/definitions/mapStrStr" },
         "auto_host_rewrite": { "type": "boolean" },
         "case_sensitive": { "type": "boolean" },
--- a/ambassador/tests/011-xfp-redirect/config/l7-redirect.yaml
+++ b/ambassador/tests/011-xfp-redirect/config/l7-redirect.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: ambassador/v0
+kind:  Module
+name:  ambassador
+config:
+  x_forwarded_proto_redirect: true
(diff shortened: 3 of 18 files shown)


containscafeine:
@plombardi89 @kflynn WDYT about the field name, and if the behavior is as desired. Once that's good, will add docs.

containscafeine:
Some manual testing with `x_forwarded_proto_redirect: true` -

```console
$ curl -v -s http://gke.35.200.172.124.xip.io/httpbin/ > /dev/null 
*   Trying 35.200.172.124...
* TCP_NODELAY set
* Connected to gke.35.200.172.124.xip.io (35.200.172.124) port 80 (#0)
> GET /httpbin/ HTTP/1.1
> Host: gke.35.200.172.124.xip.io
> User-Agent: curl/7.59.0
> Accept: */*
> 
< HTTP/1.1 301 Moved Permanently
< location: https://gke.35.200.172.124.xip.io/httpbin/
< date: Mon, 02 Jul 2018 18:56:24 GMT
< server: envoy
< content-length: 0
< 
* Connection #0 to host gke.35.200.172.124.xip.io left intact

$ curl -v -s -H "X-FORWARDED-PROTO: http" http://gke.35.200.172.124.xip.io/httpbin/ > /dev/null 
*   Trying 35.200.172.124...
* TCP_NODELAY set
* Connected to gke.35.200.172.124.xip.io (35.200.172.124) port 80 (#0)
> GET /httpbin/ HTTP/1.1
> Host: gke.35.200.172.124.xip.io
> User-Agent: curl/7.59.0
> Accept: */*
> X-FORWARDED-PROTO: http
> 
< HTTP/1.1 301 Moved Permanently
< location: https://gke.35.200.172.124.xip.io/httpbin/
< date: Mon, 02 Jul 2018 18:57:10 GMT
< server: envoy
< content-length: 0
< 
* Connection #0 to host gke.35.200.172.124.xip.io left intact

$ curl -v -s -H "X-FORWARDED-PROTO: https" http://gke.35.200.172.124.xip.io/httpbin/ > /dev/null 
*   Trying 35.200.172.124...
* TCP_NODELAY set
* Connected to gke.35.200.172.124.xip.io (35.200.172.124) port 80 (#0)
> GET /httpbin/ HTTP/1.1
> Host: gke.35.200.172.124.xip.io
> User-Agent: curl/7.59.0
> Accept: */*
> X-FORWARDED-PROTO: https
> 
< HTTP/1.1 200 OK
< server: envoy
< date: Mon, 02 Jul 2018 18:57:23 GMT
< content-type: text/html; charset=utf-8
< content-length: 8344
< access-control-allow-origin: *
< access-control-allow-credentials: true
< via: 1.1 vegur
< x-envoy-upstream-service-time: 552
< 
{ [3967 bytes data]
* Connection #0 to host gke.35.200.172.124.xip.io left intact
```

ambassador/ambassador/config.py
```diff
@@ -1096,6 +1096,14 @@ def generate_intermediate_config(self):
             # ...and route.
             self.add_intermediate_route(mapping['_source'], mapping, svc, cluster_name)
 
+            # If x_forwarded_proto_redirect is set, then we enable require_tls in primary listener, which in turn sets
+            # require_ssl to true in envoy config. Once set, then all requests that contain X-FORWARDED-PROTO set to
+            # https, are processes normally by envoy. In all the other cases, including X-FORWARDED-PROTO set to http,
+            # a 301 redirect response to https://host is sent
+            if mapping.get('x_forwarded_proto_redirect') is not None:
```
kflynn: Let's require a True value here, because it'll be very surprising if someone tries to disable it by setting `x_forwarded_proto_redirect: false` and it stays on. :)

containscafeine: +1, will also add a test to catch that

containscafeine: fixed

containscafeine:
@kflynn fixed the PR, added more tests, and added docs, PTAL!

containscafeine:
@kflynn since this impacts all the ambassador mappings, it has got me thinking if this belongs in a mapping or in the `ambassador` module?

containscafeine:
concaf [11:01 PM]
@flynn quick question about this comment - https://github.com/datawire/ambassador/pull/580#issuecomment-403001479 (edited)
Should this be in a mapping or `ambassador` module? (edited)

flynn [11:12 PM]
If it’s really global, it should be in the ambassador module, yeah.

concaf [11:14 PM]
Yeah, it's sort of global. It changes behavior for all mappings under that ambassador.
I'm going to move that under the `ambassador` module tomorrow and add the tests

containscafeine:
@kflynn added e2e for this, and moved the field to the ambassador module

PR #594, opened 2018-07-06
Populate default TLS config only if files exist

containscafeine:
This commit checks if the certs exist on disk before populating
the default TLS config. This will allow users to set other options
in TLS module without ambassador erroring out because the certs
don't exist.

### DIFF
--- a/ambassador/ambassador/config.py
+++ b/ambassador/ambassador/config.py
@@ -310,14 +310,15 @@ class Config (object):
         # (context.pem).
 
         self.default_tls_config = {
-            "server": {
-                "cert_chain_file": "/etc/certs/tls.crt",
-                "private_key_file": "/etc/certs/tls.key"
-            },
-            "client": {
-                "cacert_chain_file": "/etc/cacert/tls.crt"
-            }
+            "server": {},
+            "client": {},
         }
+        if os.path.isfile("/etc/certs/tls.crt"):
+            self.default_tls_config["server"]["cert_chain_file"] = "/etc/certs/tls.crt"
+        if os.path.isfile("/etc/certs/tls.key"):
+            self.default_tls_config["server"]["private_key_file"] = "/etc/certs/tls.key"
+        if os.path.isfile("/etc/cacert/tls.crt"):
+            self.default_tls_config["client"]["cacert_chain_file"] = "/etc/cacert/tls.crt"
 
         self.tls_config = None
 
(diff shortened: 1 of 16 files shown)


kflynn:
I wasn't trying to close this, _and_ I just learned why GitHub didn't want me to delete its original base branch. Sigh. :)

kflynn:
Code looks good -- how can we add an end-to-end test for it?

richarddli:
I tried to follow the epic thread in #469 and added the docs in https://github.com/datawire/ambassador/pull/616.

ambassador/ambassador/config.py
```diff
@@ -1107,12 +1108,16 @@ def generate_intermediate_config(self):
...
+                primary_listener['tls']['ssl_context'] = True
             redirect_cleartext_from = tmod.get('redirect_cleartext_from')
 
         self.envoy_config['listeners'] = [ primary_listener ]
 
         if redirect_cleartext_from:
-            primary_listener['require_tls'] = True
+            # We only want to set `require_tls` on the primary listener when certs are present on the pod
+            if self.tmod_certs_exist(primary_listener['tls']) > 0:
+                primary_listener['require_tls'] = True
```
containscafeine: @kflynn does this look reasonable?
When we talk about L4 redirection, we want the HTTP requests that arrive at 443 to go right through, for that we don't need `require_tls`.
Does some use-case come to your mind that could break with this change?

end-to-end/0-serial/011-websocket/ambassador/service.yaml
```diff
@@ -0,0 +1,15 @@
+aws:
```
kflynn: Huh. This'll always work in Kubernaut? Hey @plombardi89, is adding an AWS dependency here a good idea right now?

containscafeine: Wait, this file should not even be here. This is auto generated IIRC. Removing.

containscafeine: Removed

ambassador/ambassador/config.py
kflynn: I think this is OK, yeah. Had to think about it for a bit.

kflynn:
My comment for @plombardi89 was addressed by removing the file I was concerned about -- turns out it shouldn't've been included. Taking this one. :smile:

PR #601, opened 2018-07-09
Makefile modifications to run e2e

concaf:
1.     Print virtualenv message if venv was removed

2.     Add generate-rbac-yaml to Makefile, fix e2e
    This commit adds a new target generate-rbac-yaml which is required
    for running solely `make e2e`. Prior to this commit, before
    running `make e2e`, running `make website-yaml` was required
    because that would generate Kubernetes artifacts from templates
    including ambassador RBAC YAML, which is required by e2e tests.
    
    Adding a separate target isolates this behavior of generating
    ambassador's RBAC YAML and allows running `make e2e` independent
    of other make targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -232,7 +232,7 @@ version:
 	@echo "Generating and templating version information -> $(VERSION)"
 	sed -e "s/{{VERSION}}/$(VERSION)/g" < VERSION-template.py > ambassador/ambassador/VERSION.py
 
-e2e-versioned-manifests:
+e2e-versioned-manifests: venv website-yaml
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -245,6 +245,14 @@ website-yaml:
 			-e "s|{{AMBASSADOR_DOCKER_IMAGE}}|$(AMBASSADOR_DOCKER_REPO):$(VERSION)|g;s|{{STATSD_DOCKER_IMAGE}}|$(STATSD_DOCKER_REPO):$(VERSION)|g" \
 			{} \;
 
+generate-rbac-yaml:
```
kflynn: This is a problem. `create-manifests.sh` should already be doing this, and `make e2e-versioned-manifests` calls `create-manifests.sh`.

PR #616, opened 2018-07-13
document TLS redirection

richarddli:
This PR documents how to configure ELBs for TLS redirection as discussed in #469.


richarddli:
Hold on merging -- I'd like feedback on this from @plombardi89, dvavili, and @containscafeine.

kflynn:
I'm taking this into `shared/develop`. I would welcome some example YAML files.

PR #617, opened 2018-07-14
Provide default CORS handling

KowalczykBartek:
Hi !
recently I had occasion to play with Ambassador CORS support, and it turned out that support for default CORS options would be helpful :) 
Idea is that Ambassador has for example such piece of config
```yaml
apiVersion: ambassador/v0
kind:  Module
name:  cors_default_envoy
config:
  cors:
    allow_origin:
      - "*"
    allow_methods: "POST, GET, OPTIONS"
    max_age: "86400"
    enabled: true
    allow_headers: Content-Type, Authorization, host, user-agent, content-length, custom-header1, custom-header2
```
and this will be set as CORS option for all other services inside cluster (if such service has no explicit cors config).

as @kflynn advices I added test (it was an adventure :D ).

Waiting for feedback ! 
thanks :)


ambassador/ambassador/config.py
```diff
@@ -995,7 +995,8 @@ def generate_intermediate_config(self):
             if ((module_name == 'ambassador') or
                 (module_name == 'tls') or
                 (module_name == 'authentication') or
-                (module_name == 'tls-from-ambassador-certs')):
+                (module_name == 'tls-from-ambassador-certs') or
+                (module_name == 'cors_default_envoy')):
```
kflynn: Probably this should just be "cors-default" -- but I'm curious why you went with a separate module instead of just adding into the `ambassador` `Module`'s configuration?

ambassador/ambassador/config.py
```diff
@@ -1166,6 +1167,12 @@ def generate_intermediate_config(self):
             self.envoy_clusters[cluster_key] for cluster_key in sorted(self.envoy_clusters.keys())
         ]
 
+        # Set default cors settings - default for all services without customized CORS section.
+        # Expecation is that this is directly mapped to envoy configuration, no renaming.
```
kflynn: Is this consistent with the per-`Mapping` CORS config? (I think so.)
Should we define a schema that both can validate against? (I think so, again?)

KowalczykBartek:
@kflynn I fixed your comments :) now it is in 100% consistent with "normal" CORS settings,
```yaml
---
apiVersion: ambassador/v0
kind:  Module
name:  ambassador
config:
  cors:
    methods: POST, GET, OPTIONS
    headers: Content-Type
    credentials: true
    max_age: "86400"
```

containscafeine:
@KowalczykBartek one last rebase, please? :)

PR #618, opened 2018-07-16
Add ability to override targetPort for http and https

kostyrev:
I'm using ELB for SSL termination and without those changes I'm not able to render following spec:
```
spec:
  ports:
  - name: http
    port: 443
    protocol: TCP
    targetPort: 80
  - name: https
    port: 80
    protocol: TCP
    targetPort: 80
 ```


kflynn:
@kostyrev Hey, thanks! I like this -- can you update the docs in `helm/README.md` to reflect your new values? I think that's all we need to take this. :smile:

kostyrev:
I've updated the docs

kflynn:
Many thanks!

kflynn:
I'm taking this into `shared/develop` for 0.35.3. Thanks again!

Issue #619, opened 2018-07-16
Helm Chart use fixed tag version


jalberto:
**Please describe your use case / problem.**
Helm char uses 'appVersion' for image deployment, righ tnwo it is outdated (0.35.0), this is hard to set on deploy. 
Mostly helm charts uses a `tag` key in `values` with a default value

**Describe the solution you'd like**
2 suggestions:
- keep chart appVersion in sync with releases
- move it to a tag in values

MichielDeMey:
Should be fixed now in https://github.com/datawire/ambassador/pull/622

jalberto:
thanks!

PR #622, opened 2018-07-17
Utilizing imageTag in helm deployment

endrec:
The Readme suggests that the `imageTag` parameter can be used to specify the docker image version, but it was not used in the deployment template.

### DIFF
--- a/helm/ambassador/README.md
+++ b/helm/ambassador/README.md
@@ -45,9 +45,9 @@ The following tables lists the configurable parameters of the Ambassador chart a
 
 | Parameter                       | Description                                | Default                                                    |
 | ------------------------------- | ------------------------------------------ | ---------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


kflynn:
Looks good (especially in light of #619), thanks! 

My only question: should it be `image.tag`? (If it's easier to work with as `imageTag`, I'm fine with that answer.)

endrec:
Very good question, @kflynn :)
It is not about easiness, it's more about consistency.
It says `imageTag` in the readme, so I thought that should be it. Later I realised the other two is not consistent with the actual template, so I modified the readme, but now the three properties do not follow the same naming convention... :)
I will go with `image.tag`. :)

kflynn:
Thanks!

kflynn:
Pulling this into `shared/develop` for 0.35.3.

PR #638, opened 2018-07-19
Exploring testing ideas

rhs:
Do not merge this. This PR is so that interested parties can follow the testing work.


rhs:
This is now ready for some review. You can read the docs here: https://github.com/datawire/ambassador/tree/rhs/testing/ambassador/tests/poc

Keep in mind they are rst written for sphinx, so they are going to render slightly weirdly in a few cases. If you would like to see them as sphinx intended, then check out this branch and type `make html` in the poc directory.

rhs:
This was merged in other forms.

PR #650, opened 2018-07-24
Parallelize all the tests

containscafeine:


containscafeine:
Yeah, that's because of the 001-simple flake

PR #655, opened 2018-07-24
Enabling Ambassador tracing with TracingService, docs and tutorial

alexgervais:
Provides :
- Envoy V1 configuration for enabling tracing on all routes
- Unit and end-to-end tests
- Documentation of `TracingService`, tutorial and Zipkin example

### DIFF
--- a/ambassador/ambassador/config.py
+++ b/ambassador/ambassador/config.py
@@ -681,6 +681,10 @@ class Config (object):
         return self.safe_store(source_key, "ratelimit_configs", obj_name, obj_kind,
                                SourcedDict(_source=source_key, **obj))
 
+    def handle_tracingservice(self, source_key, obj, obj_name, obj_kind, obj_version):
+        return self.safe_store(source_key, "tracing_configs", obj_name, obj_kind,
+                               SourcedDict(_source=source_key, **obj))
+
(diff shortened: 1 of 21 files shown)


alexgervais:
Yes, the `request_headers_for_tags` config is something a user might provide. I did not have any use for it myself so I preferred highlighting the empty array in case someone else wants to contribute and make use of it.

kflynn:
@alexgervais Makes sense to me, thanks!

ambassador/ambassador/config.py
```diff
@@ -1535,6 +1546,44 @@ def module_config_ratelimit(self, ratelimit_config):
...
+        host_rewrite = None
+        sources = []
+
+        if tracing_config:
+            for config in tracing_config.values():
+                sources.append(config['_source'])
+                cluster_hosts = config.get("service", None)
+                driver = config.get("driver", None)
+                driver_config = config.get("config", {})
+                host_rewrite = config.get("host_rewrite", None)
```
containscafeine: looking back at this, should `host_rewrite` be a field in TracingService? And if so, should it be documented?

alexgervais: If I'm not mistaken it's also an undocumented field of AuthService and RateLimitService. I'm not absolutely sure what was the intent in the original AuthService, but I agree, it should either be documented or removed.

Issue #497, opened 2018-06-07
Ambassador crashes when annotation has empty config


richarddli:
**To Reproduce**

```
apiVersion: v1
kind: Service
metadata:
  name: qotm
  annotations:
    getambassador.io/config: |
      ---
spec:
  type: ClusterIP
  selector:
    app: qotm
  ports:
  - port: 80
    name: http-qotm
    targetPort: http-api
```

**Expected behavior**
An error message, or at least not a crash.

You get:

```
2018-06-07 15:03:03 kubewatch 0.34.1 INFO: generating config with gencount 1 (3 changes)
Traceback (most recent call last):
  File "/application/kubewatch.py", line 493, in &lt;module&gt;
    main()
  File "/usr/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/application/kubewatch.py", line 476, in main
    sync(restarter)
  File "/application/kubewatch.py", line 373, in sync
    restarter.restart()
  File "/application/kubewatch.py", line 118, in restart
    config = self.generate_config(output)
  File "/application/kubewatch.py", line 151, in generate_config
    aconf = Config(output)
  File "/usr/lib/python3.6/site-packages/ambassador-0.34.1-py3.6.egg/ambassador/config.py", line 336, in __init__
  File "/usr/lib/python3.6/site-packages/ambassador-0.34.1-py3.6.egg/ambassador/config.py", line 414, in process_all_objects
AttributeError: 'NoneType' object has no attribute 'get'
AMBASSADOR: kubewatch sync exited with status 1
Here's the envoy.json we were trying to run with:
ls: /etc/envoy*.json: No such file or directory
No config generated.
AMBASSADOR: shutting down
```

containscafeine:
Fixed in #691, test in #710

PR #691, opened 2018-08-02
FIX for #497

rohan47:
This PR fixes issue #497 

Empty config in annotations crashed ambassador.
The expected behavior was to give an error message, or at least not a crash.

```yaml
apiVersion: v1
kind: Service
metadata:
  name: qotm
  annotations:
    getambassador.io/config: |
      ---
spec:
  type: ClusterIP
  selector:
    app: qotm
  ports:
  - port: 80
    name: http-qotm
    targetPort: http-api
```

After this fix the above yaml did not crash ambassador and logged an error

Signed-off-by: rohan47 <rohanrgupta1996@gmail.com>

### DIFF
--- a/ambassador/ambassador/config.py
+++ b/ambassador/ambassador/config.py
@@ -446,6 +446,11 @@ class Config (object):
                 # No pragma involved here; just default to the filename.
                 self.source = self.filename
 
+            # Is the object empty?
+            if obj == None :
+                self.logger.debug("Annotation has empty config")
+                return
+
             # Is an ambassador_id present in this object?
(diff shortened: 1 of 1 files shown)


docs/package.json
```diff
@@ -5,6 +5,7 @@
   "main": "index.js",
   "dependencies": {
     "gitbook-cli": "^2.3.0",
+    "gitbook-plugin-anchorjs": "^2.0.1",
```
richarddli: Why is this in this PR?

rohan47: This got added automatically, and I wasn't sure about this so I kept it to get comments

PR #696, opened 2018-08-03
Add release-prep target for 1-Click Releases

### LINKED ISSUE #595: Make 1-click™ releases
It'd be great to go to a point where we run `make release-prep` and then commit all the generated artifacts, open a PR, merge it, and :boom: it's a new release.

To achieve this degree of automation, `make release-prep` will have to -
- prompt for the release's version (reminding user to vet the changes and check with other contributors to determine the appropriate major and minor version)
- prompt (maybe open an editor) for the release notes to go in CHANGELOG.md and GitHub release page
- validate the CHANGELOG.md (assuming any machinery needs to parse it)
- update index.html
- stage all the changes, and commit with an auto-generated commit message
- open/prompt for the PR

Automating this will quicken our release process and make it simple enough that anyone from the team can make a release.

containscafeine:
This commit adds a `release-prep` target to the Makefile which
spins up an interactive script for making a release. The script
- warns and informs the user around picking the right version
- updated the version in docs/index.html
- prompts the user for release notes and generates changelog and
updates CHANGELOG.md
- generates GitHub release notes for the user
- commits the changes to the current branch

All the user will need to do after running `make release-prep` is
push the branch and release from GitHub UI.

Fix #595

### DIFF
--- a/releng/release-prep.sh
+++ b/releng/release-prep.sh
@@ -1,0 +1,118 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 Datawire. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License
+
+set -e
+
+press_enter() {
+    read -s -p "Press enter to continue"
+    echo
+}
+
+RELEASE_PROMPT="
+Before making the release, make sure that -
+- [ ] the new version number depicts the release accurately
+- [ ] if new features have been added, the MINOR version is bumped
+- [ ] if backwards-incompatible changes have been made, the MINOR version is bumped
+- [ ] if only backwards-compatible bug fixes have been made, the PATCH version is bumped
+- [ ] the changes and the version number has been vetted by project maintainers and stakeholders
+"
+
+echo "$RELEASE_PROMPT"
+press_enter
+
+echo "Fetching latest release tag from GitHub"
+CURRENT_VERSION=$(curl --silent "https://api.github.com/repos/datawire/ambassador/releases/latest" | jq -r '.tag_name')
+
+echo
+echo "Current version: ${CURRENT_VERSION}"
+echo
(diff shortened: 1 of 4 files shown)


containscafeine:
demo

@kflynn @rhs this is how it looks like, PTAL

releng/release-prep.sh
```diff
@@ -0,0 +1,108 @@
...
+echo "Fetching latest release tag from GitHub"
+CURRENT_VERSION=$(curl --silent "https://api.github.com/repos/datawire/ambassador/releases/latest" | jq -r '.tag_name')
+
+echo
+echo "Current version: ${CURRENT_VERSION}"
+echo
+git log --pretty=oneline --abbrev-commit ${CURRENT_VERSION}^..
+echo
+echo "^ these changes have been made since the last release, pick the right version number accordingly"
+read -p "Enter new version: " DESIRED_VERSION
```
kflynn: Maybe ask for confirmation?

releng/release-prep.sh
```diff
@@ -0,0 +1,108 @@
...
+
+### Major changes:
+- Feature: <insert feature description here>
+- Bugfix: <insert bugfix description here>
+
+### Minor changes:
+- Feature: <insert feature description here>
+- Bugfix: <insert bugfix description here>"
+
+RELEASE_NOTES=$(echo "${RELEASE_NOTES_TEMPLATE}" | vipe)
```
kflynn: So. Editors. Editors suck, man. 😄 Here's the way that seems to be the best chance of avoiding screaming users filing bugs every day because they use Emacs instead of vi, or vi instead of Emacs, or back-in-my-day-we-used-ed-and-it's-good-enough-for-me! :wink:

```
echo "${RELEASE_NOTES_TEMPLATE}" > tmpfile

if ! ${EDITOR:-vi} tmpfile; then
    exit
fi

RELEASE_NOTES=$(cat tmpfile)
rm -f tmpfile
```

except the `tmpfile` needs to actually be a proper temporary file and you should probably use `trap 0` to clean it up.

Also I wouldn't make the user delete that template line. Use `sed` to strip it out maybe?

kflynn: (Admittedly the idea of `vipe` is really really cool. But it's not installed on my system, so this whole thing will fail...)

kflynn: Oh @containscafeine -- please update `RELEASING.md` as part of landing this, too!

releng/release-prep.sh
containscafeine: done

releng/release-prep.sh
containscafeine: Totally makes sense, fixed.

containscafeine: Updated RELEASING.md

PR #721, opened 2018-08-09
Fix typo in homepage, insuring -> ensuring

boffbowsh:
Ambassador ensures that we get the speed of Envoy, not insures.


richarddli:
Thanks!

PR #729, opened 2018-08-15
Update documentation, and remove zipkin resource limits

danielbryantuk:
I've tweaked a couple of sentences in the Getting Started guide, as I constantly get asked questions around this.

I've also added to the tracing tutorial, and removed the `resources` config in the zipkin deployment, as this causes issues on minikube and Docker for Mac (a friend mention an OOM kill of Zipkin and a CrashLoopBackoff, which I could recreate).


docs/user-guide/tracing-tutorial.md
```diff
@@ -59,25 +63,22 @@ spec:
         ports:
         - name: http
           containerPort: 9411
-        resources:
-          limits:
-            cpu: "1"
-            memory: 256Mi
-          requests:
```
containscafeine: @danielbryantuk are you facing problems with resources and limits set? We had removed this when it was only limits, but then added it back with requests, so it does not fail.

PR #734, opened 2018-08-16
Detailed Ambassador + Istio Integration for Tracing, Statistics and Monitoring

feitnomore:
Added detailed documentation on Ambassador Integration with Istio System.
Documentation covers Tracing Integration, Prometheus Integration as well as Grafana Dashboard usage.
Also added a yaml file with the ConfigMap that needs to be used with Istio's Prometheus.


feitnomore:
I've commited the changes requested...

PR #742, opened 2018-08-23
Generate docs off of master

### LINKED ISSUE #741: Devs want a cleaner doc build model
Docs are built from `stable` instead of `master`, which is sometimes confusing. Make sure that everyone knows how the doc-build process works, and make sure we have the ability to let non-Datawire people see docs for upcoming releases.

containscafeine:
Fix #741


kflynn:
This PR is probably not what we want to do about this -- gotta check a bunch of corner cases, sadly. Closing for now, we'll revisit this shortly.

PR #750, opened 2018-08-29
Using named targetPort to match backend pod

alexgervais:
In helm and templates, using a named Service `targetPort` to match backend Pod. This will make the transition from running as root (port 80, 443) much easier as the Service will be able to target newly created Pods running as non-root (port 8080, 8443) without downtime during the rolling deployment.

No tests as this relate only to documentation.


kflynn:
Thanks!

PR #764, opened 2018-08-31
added `loadBalancerSourceRanges` option to helm chart and tested

mtbdeano:
When deploying to AWS managed Kubernetes, the ability to add `loadBalancerSourceRanges` helps lock down kubes by forcing the AWS ELB created to add these to the security group.

I have added this key to the Chart `values.yaml`, `README.md` and `service.yaml` and it works when I deploy from this modified Chart.

This was recommended by @plombardi89 in his comment on https://github.com/datawire/ambassador/issues/459#issuecomment-417483928


mtbdeano:
this could also be bundled with https://github.com/datawire/ambassador/pull/654

rhs:
Thanks for your contribution!

PR #768, opened 2018-09-04
Basic kat integration.

rhs:


rhs:
This seems to be working now. The *inner loop* quick start should be:

1. Type `make shell`.
2. Run `py.test -k blah`

PR #783, opened 2018-09-07
allows setting route.timeout_ms to zero

### LINKED ISSUE #545: route timeout_ms cannot be set to 0 (for grpc streaming services)
**Describe the bug**

route timeout_ms cannot be set to 0 (for grpc streaming services)
This is because of the logic in https://github.com/datawire/ambassador/blob/master/ambassador/templates/envoy.j2#L49

**To Reproduce**
1. create a backend grpc streaming service like https://github.com/grpc/grpc/tree/master/examples/python/route_guide
2. Use the following ambassador mapping
```
---
apiVersion: ambassador/v0
kind: Mapping
name: route-guide-stream-mapping
grpc: true
prefix: /routeguide.RouteGuide/
rewrite: /routeguide.RouteGuide/
service: svc-route-guide-stream:4443
tls: upstream
timemout_ms: 0
```
3. See envoy config in ambassador pod. It stills shows 3000 for route timeout_ms for that route.

**Expected behavior**
Able to configure 0 for timeout_ms in a mapping for streaming services

 - Ambassador: 0.34.2
 - Kubernetes - cluster created using kubeadm
 - Version [e.g. 1.10]

**Additional context**
reference https://github.com/envoyproxy/envoy/issues/1778

lovers36:
fix #545

### DIFF
--- a/ambassador/templates/envoy.j2
+++ b/ambassador/templates/envoy.j2
@@ -53,7 +53,7 @@
                   "routes": [
                     {% for route in routes %}
                     {
-                      "timeout_ms": {{ route.timeout_ms or 3000 }},
+                      "timeout_ms": {{ route.timeout_ms if (route.timeout_ms == 0 or route.timeout_ms) else 3000 }},
                       {%- if route.prefix -%}"prefix": "{{ route.prefix }}",{% endif %}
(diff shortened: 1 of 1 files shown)


kflynn:
Thanks!

PR #785, opened 2018-09-08
Rhs/local amb

rhs:


rhs:
This is still a bit slower than it could be because the docker build is a bit slow when there are modified files, and the changes to abstract_tests.py are kinda hacky.

All in all though, it's much faster than waiting for docker-push and we should be able to clean this up when isolated config and ADS are in place.

PR #809, opened 2018-09-15
Add support for AMBASSADOR_ID in Helm chart

jstol:
In order to support multiple Ambassador installs, the `AMBASSADOR_ID` env variable needs to be set within the Deployment. This change adds in an optional `deployment.ambassadorId` variable which sets `AMBASSADOR_ID` within `templates/deployment.yaml`.


jstol:
This is also addressed in #657.

jstol:
@rhs should this be closed now that #657 has been merged?

rhs:
Yes it should be closed, thanks for the reminder! (And thanks for submitting the PR in the first place.)

PR #839, opened 2018-09-26
Nkrause/docUpdate/gRPC

nbkrause:
Demonstrates how to configure the gRPC example services for running behind Ambassador. Demos: docker build
Ambassador mapping 
kubernetes deployment 
TLS termination


docs/user-guide/grpc.md
```diff
@@ -4,133 +4,209 @@
...
+Follow the example up through [Run a gRPC application](https://grpc.io/docs/quickstart/python.html#run-a-grpc-application) to get started.
 
-Ambassador `Mapping`s are based on URL prefixes; for gRPC, the URL prefix is the full service name, including the package path. 
+### Dockerize
 
-For `Hello World`, in its [proto definition file](https://github.com/grpc/grpc-go/blob/master/examples/helloworld/helloworld/helloworld.proto), we see
+After building our gRPC application and testing it locally, we need to package it as a Docker container and deploy it to kubernetes.
+
+```shell
+$ touch Dockerfile
```
richarddli: We can skip this; implicit below is you have to create a Dockerfile.

docs/user-guide/grpc.md
```diff
@@ -4,133 +4,209 @@
...
+Create the container and test it:
+
+```shell
+$ docker build -t <docker_hub>/grpc_example .
+$ docker run -p 50051:50051 <docker_hub>/example
+```
+Where `<docker_hub>` is your docker profile.
+
+Switch to another terminal and, from the same directory, run the `greeter_client`.
+The application should run exactly the same.
```
richarddli: I don't think you reference running it outside the container, so maybe just list the expected output?

docs/user-guide/grpc.md
```diff
@@ -4,133 +4,209 @@
...
+
+Once you verify the container works, push it to your docker hub:
+
```
richarddli: Docker registry, not Hub. Note Docker is capitalized.

docs/user-guide/grpc.md
```diff
@@ -4,133 +4,209 @@
...
+```shell
+$ docker push <docker_hub>/grpc_example
+```
+
+### Mapping gRPC Services
+
+Ambassador `Mapping`s are based on URL prefixes; for gRPC, the URL prefix is the full service name, including the package path (`package.service`). These are defined in the `.proto` definition file. In the example [proto definition file](https://github.com/grpc/grpc/blob/master/examples/protos/helloworld.proto) I am using, we see:
```
richarddli: No first person in the docs

docs/user-guide/grpc.md
```diff
@@ -4,133 +4,209 @@
...
+      grpc: True
+      prefix: /helloworld.Greeter/
+      rewrite: /helloworld.Greeter/
+      service: grpc-example
 ```
 
-Note the `grpc: true` line -- this is the necessary magic when mapping a gRPC service. Also note that you'll need `prefix` and `rewrite` the same here, since the gRPC service needs the package and service to be in the request to do the right thing.
-
-## Deploying `Hello World`
+Note the `grpc: true` line —- this is what tells envoy to use HTTP/2 so the request can communicate with the backend service. Also note that you'll need `prefix` and `rewrite` the same here, since the gRPC service needs the package and service to be in the request to do the right thing.
```
richarddli: Envoy should e capitalized

docs/user-guide/grpc.md
```diff
@@ -4,133 +4,209 @@
...
-          - name: PORT
-            value: "9999"
-        command:
-          - greeter_server
+          containerPort: 50051
       restartPolicy: Always
 ```
 
-(We tell the gRPC service to run on port 9999, then map the container's port 80 inbound to simplify the `Mapping`. There's no magic behind these port numbers: anything will work as long as you're consistent in when mapping everything.)
+Once adding the mapping configuration to the service with an annotation, the rest of the kubernetes deployment YAML file is pretty straightforward. We need to identify the container image to use, expose the `containerPort` to listen on the same port the docker container is listening on, and map the service port (80) to the container port (50051).
```
richarddli: Once the mapping configuration has been added to the service ...

(Awkward sentence). Also Kubernetes should be capitalized, as does Docker.

docs/user-guide/grpc.md
```diff
@@ -4,133 +4,209 @@
...
 
-This is available from getambassador.io, so you can simply
+Once you have the YAML file configured, deploy it to your cluster with kubectl:
```
richarddli: `kubectl`

docs/user-guide/grpc.md
```diff
@@ -4,133 +4,209 @@
...
+$ kubectl apply -f grpc_example.yaml
 ```
 
-or, as always, you can use a local file instead.
+### Testing the deployment
 
-## Testing `Hello World`
-
-Now you should be able to access your service. We'll need the hostname for the Ambassador service, which you can get with
+Make sure to test your kubernetes deployment before making more advanced changes (like adding TLS). To test any service with Ambassador, we will need the hostname of the running Ambassador service which you can get with:
```
richarddli: Capitalize Kubernetes

docs/user-guide/grpc.md
```diff
@@ -4,133 +4,209 @@
...
 ```
 
-Note: If you're trying this out using `NodePort` on minikube and running the docker command above on your host machine, make sure to pass the `--network host` parameter to the docker command.
+### gRPC and TLS
+
+Handling TLS with a gRPC service requires a little extra configuration than with other services. RPC calls cannot use certificates stored in the host machines so, in a normal setup, both the server and client applications must load and read certificates to handle TLS encryption. 
+
+![](/docs/images/gRPC-TLS.png)
+
+With Ambassador, we are able to forgo needing the server to load and read certificates since TLS is terminated at Ambassador. This means the client will now request an SSL connection with Ambassador and validate the certs sent by Ambassador, not the gRPC server application.
```
richarddli: s/SSL/TLS for consistency

docs/user-guide/grpc.md
```diff
@@ -4,133 +4,209 @@
...
+
+Handling TLS with a gRPC service requires a little extra configuration than with other services. RPC calls cannot use certificates stored in the host machines so, in a normal setup, both the server and client applications must load and read certificates to handle TLS encryption. 
+
+![](/docs/images/gRPC-TLS.png)
+
+With Ambassador, we are able to forgo needing the server to load and read certificates since TLS is terminated at Ambassador. This means the client will now request an SSL connection with Ambassador and validate the certs sent by Ambassador, not the gRPC server application. 
+
+![](/docs/images/gRPC-TLS-Ambassador.png)
+
+This has the advantage of not needing the server to bother with certs but means that the client will still need to load the root cert. This is can be done rather easily with some gRPC calls.
```
richarddli: "This is can be done"?

docs/user-guide/grpc.md
```diff
@@ -4,133 +4,209 @@
...
+```diff
+- with grpc.insecure_channel(‘$AMBASSADORHOST:$PORT’) as channel:
++ with grpc.secure_channel(‘$AMBASSADORHOST:$PORT’, grpc.ssl_channel_credentials()) as channel:
+        stub = helloworld_pb2_grpc.GreeterStub(channel)
+        response = stub.SayHello(helloworld_pb2.HelloRequest(name='you'))
+    print("Greeter client received: " + response.message)
+```
+
+`grpc.ssl_channel_credentials(root_certificates=None, private_key=None, certificate_chain=None)`returns the root certificate that will be used to validate the certificate and public key sent by Ambassador. 
+The default values of `None` tells gRPC runtime to grab the root cert from the default location packaged with gRPC and ignore the private key and certificate chain fields.
```
richarddli: s/cert/certificate

PR #852, opened 2018-10-04
Add some TLS support

containscafeine:


containscafeine:
Fix #870

ambassador/tests/kat/abstract_tests.py
```diff
@@ -206,9 +216,10 @@ def queries(self):
     def scheme(self) -> str:
         return "http"
 
-    def url(self, prefix) -> str:
+    def url(self, prefix, https=False) -> str:
```
rhs: Couldn't we just infer the https boolean from the value of `self.scheme()`?

kflynn:
I merged shared/dev/tgr606 back into this (twice) and couldn't stop myself from ripping up a bunch of v2listener.py and puuting it back together. *cough* Sorry about that!

**This is still not functional**. We need to modify `kubewatch` to DTRT with cert loading; I'll post more about that on Slack.

PR #886, opened 2018-10-15
Ignore TLS if invalid secret or paths

containscafeine:
This commit does not turn TLS on and change port to 443 if an
invalid secret or an invalid path is supplied by the end user.

### DIFF
--- a/ambassador/ambassador/ir/irtls.py
+++ b/ambassador/ambassador/ir/irtls.py
@@ -18,6 +18,7 @@ from typing import Optional, TYPE_CHECKING
 
 from ..config import Config
 from .irresource import IRResource as IRResource
+from ambassador.utils import RichStatus
 
 if TYPE_CHECKING:
     from .ir import IR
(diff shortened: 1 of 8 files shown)


ambassador/ambassador/ir/irtls.py
```diff
@@ -54,11 +54,24 @@ def setup(self, ir: 'IR', aconf: Config):
...
         secret = self.get('secret')
 
         # Check if secret and certs, both are specified
         cert_specified = False
-        if self.get('cert_chain_file') is not None or self.get('private_key_file') is not None:
+
+        cert_chain_file = self.get('cert_chain_file')
+        if cert_chain_file is not None:
+            if not ir.file_checker(cert_chain_file):
+                self.logger.error('cert_chain_file path {} does not exist'.format(cert_chain_file))
```
kflynn: These should really be calling `self.post_error()` -- that'll log the error but also record it so that the diag UI can show it.

(If the diag UI _doesn't_ show it, that's a bug and we'll fix it.)

plombardi89:
Do we show or log any kind of error in this situation? It feels like a simple misconfiguration could cause someone to end up running non-HTTPS traffic if this PR is merged and that could have really bad consequences.

ambassador/ambassador/ir/irtls.py
containscafeine: fixed

containscafeine:
@plombardi89 agree, I've put in `self.post_error` calls that should report this in the diagnostics.

PR #910, opened 2018-10-21
Add support for AMBASSADOR_VERIFY_SSL_FALSE to helm chart

christianhuening:
Add support for AMBASSADOR_VERIFY_SSL_FALSE to helm chart by adding the variable to deployment and setting the default in values.yaml to false

### DIFF
--- a/helm/ambassador/templates/deployment.yaml
+++ b/helm/ambassador/templates/deployment.yaml
@@ -62,6 +62,8 @@ spec:
           {{- end }}
           - name: AMBASSADOR_ID
             value: {{ .Values.ambassador.id | quote }}
+          - name: AMBASSADOR_VERIFY_SSL_FALSE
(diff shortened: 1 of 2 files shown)


kflynn:
Hey, so sorry for the delay getting back to you! `AMBASSADOR_VERIFY_SSL_FALSE` is actually active if _present_ irrespective of value, so it needs to not be included if you mean it to be inactive. Sorry for the confusion!

christianhuening:
oh i see, thanks!

christianhuening:
hmmm so if I want to set it active, I have to provide the ENV var? In that case I need the env var activateable from the helm chart or am i missing something here?

Flydiverny:
Currently you will have to manually edit the deployment, but #1044 also solves this with some more flexibility :)

PR #913, opened 2018-10-22
Update helm chart 

cliveseldon:
* Run as non privileged user
    * Adds securityContext with a user id (8888 by default)
    * Changes ports to 8080, 8443
    * Adds config to ensure service_port setting is added as its not the default 80 now
 * Add exporter.enabled setting and set to false to not include statsd exporter by default
 * Fix single namespace typo - needs to be string


kflynn:
Sorry for the delay getting back to you on this one! I like the direction here, but I think we need to keep the default prior to 0.50 as running as root, listening on 80/443. Easy for you to tweak this to preserve the default but allow easily deploying as non-root?

Flydiverny:
I think we can close this as the new helm/charts [chart](https://github.com/helm/charts/tree/master/stable/ambassador) covers this :)

kflynn:
I think you're right. @cliveseldon, many thanks for getting this rolling way back when! Please check out the new Ambassador chart in `helm/charts`. :smile:

PR #916, opened 2018-10-23
Fix ambassador-rbac-prometheus.yaml example

flands:
statsd.listen-address does not appear to be valid. Given this example is
for internal consumption, leverage statsd.listen-udp.

Also, update to latest statsd-exporter.

Finally, use dedicated configmap since Ambassador looks for ambassador-config today


richarddli:
@bcatcho do you mind taking a quick look? this looks good to me; I believe you were the last to look at this area.

richarddli:
Going to tag @bcatcho again, and see if @nbkrause can test this change?

richarddli:
I actually took a look at this recently so I now understand this PR. It looks fine to me.

PR #924, opened 2018-10-24
Add flags --v1 and --v2 to config cli command

lavoiedn:
This allows switching the CLI mode to only generate one config or the other.

Also added support in `validate` cli command via `**kwargs`.

### DIFF
--- a/ambassador/ambassador/cli.py
+++ b/ambassador/ambassador/cli.py
@@ -199,19 +199,20 @@ def dump(config_dir_path: Parameter.REQUIRED, *,
         sys.exit(1)
 
 
-def validate(config_dir_path: Parameter.REQUIRED, *, k8s=False):
+def validate(config_dir_path: Parameter.REQUIRED, **kwargs):
     """
-    Validate an Ambassador configuration
+    Validate an Ambassador configuration. This is an extension of "config" that
+    redirects output to devnull and always exits on error.
 
     :param config_dir_path: Configuration directory to scan for Ambassador YAML files
     :param k8s: If set, assume configuration files are annotated K8s manifests
     """
-    config(config_dir_path, os.devnull, k8s=k8s, exit_on_error=True)
+    config(config_dir_path, os.devnull, exit_on_error=True, **kwargs)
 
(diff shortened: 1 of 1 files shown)


ambassador/ambassador/cli.py
```diff
@@ -199,19 +199,20 @@ def dump(config_dir_path: Parameter.REQUIRED, *,
         sys.exit(1)
 
 
-def validate(config_dir_path: Parameter.REQUIRED, *, k8s=False):
+def validate(config_dir_path: Parameter.REQUIRED, **kwargs):
```
kflynn: Should the `:param k8s:` go away?

ambassador/ambassador/cli.py
```diff
@@ -199,19 +199,20 @@ def dump(config_dir_path: Parameter.REQUIRED, *,
...
     """
-    Validate an Ambassador configuration
+    Validate an Ambassador configuration. This is an extension of "config" that
+    redirects output to devnull and always exits on error.
 
     :param config_dir_path: Configuration directory to scan for Ambassador YAML files
     :param k8s: If set, assume configuration files are annotated K8s manifests
     """
-    config(config_dir_path, os.devnull, k8s=k8s, exit_on_error=True)
+    config(config_dir_path, os.devnull, exit_on_error=True, **kwargs)
```
kflynn: If `kwargs` also has `exit_on_error` set, this will be an error. Can that ever happen? (I don't think so.)

lavoiedn: Nah, it can't happen since `clize` doesn't pick up on `exit_on_error` since the default value is `True`, as far as I am aware at least.

Trying to specify it just results in an "Unknown option" error from what I can see.

ambassador/ambassador/cli.py
lavoiedn: Removed from the comment!

ambassador/ambassador/cli.py
kflynn: Good enough, thanks!

PR #925, opened 2018-10-24
Made a small change to fix the --k8s flag for cli config

lavoiedn:
There was an issue with `rkey` collisions when generating `config` via the CLI from kubernetes yaml files.

This is my suggestion of a fix.

### DIFF
--- a/ambassador/ambassador/config/acresource.py
+++ b/ambassador/ambassador/config/acresource.py
@@ -209,9 +209,9 @@ class ResourceFetcher:
             return
 
         self.filename += ":annotation"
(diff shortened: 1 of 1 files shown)


lavoiedn:
I had to overwrite my commit twice for some minor cleanup (missing key for kwarg, extra newline). Oops!

kflynn:
Thanks!

kflynn:
This will turn into things like `service tracingtest-http, namespace default.1` from `kubewatch`, and we may want to change that later -- however, the uniqueness thing is the important bit at the moment.

PR #927, opened 2018-10-25
Make config dirs tunable

n1koo:
**What**

- Make config dirs for both `ambassador` and `envoy` tunable via custom ENV `CUSTOM_CONFIG_BASE_DIR`. Defaults to current dir (`$AMBASSADOR_ROOT`) if not set

- Also fixed a bug how the `PYTHON_EGG_CACHE` is set, it was not overwritable and was writing to root of the app while  the original intent was `$APPDIR}/.cache`

**Why**
This allows folks to eg. use `tmpfs` mounts on k8s when writing to read is not permitted.

**Open**

I didn't quite grasp why both `envoy.json` and `envoy*.json` is used, could be cleaned up if latter could be dropped.

### DIFF
--- a/ambassador/entrypoint.sh
+++ b/ambassador/entrypoint.sh
@@ -18,10 +18,8 @@ export LC_ALL=C.UTF-8
 export LANG=C.UTF-8
 
 AMBASSADOR_ROOT="/ambassador"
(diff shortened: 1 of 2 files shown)


kflynn:
Taking this one for the next 0.40.

PR #936, opened 2018-10-26
Nkrause/docs/grpc upstream tls

nbkrause:
TLS origination to a gRPC service was never covered in the documentation. It turns out, you need to set alpn_protocols: ["h2"] in the tls-context in the the target envoy cluster as well as the listener envoyproxy/envoy#4291. This can be done using the same method used for supplying cert_chain_file and private_key_file for mTLS.

I also reworded the doc since it kind of rambled about useless stuff.


ngalaiko:
Weird that `ambassador` documentation is more helpful than `envoy` documentation 🙂

PR #939, opened 2018-10-27
Use parameterized statsd host value while checking it

maxnasonov:


kflynn:
Wow, what an embarrassing bug. Thanks for the catch!

PR #947, opened 2018-10-30
Adding volumes and volumeMounts to ambassador service helm chart

kyschouv:
We need to use volume mounts to import our SSL certificate into ambassador (due to how it's being stored), and the helm chart doesn't allow us to add volume mounts. This PR adds that capability.

### DIFF
--- a/helm/ambassador/Chart.yaml
+++ b/helm/ambassador/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: "0.40.0"
+appVersion: "0.40.1"
 description: A Helm chart for Datawire Ambassador
 name: ambassador
-version: 0.40.0
+version: 0.40.1
 sources:
 - https://github.com/datawire/ambassador
 maintainers:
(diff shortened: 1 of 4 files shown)


kyschouv:
Is there anything further I need to do to get this merged? I've noticed the build is failing, but that seems to be the case for all open PRs, and the error doesn't seem related to these changes.

kflynn:
My apologies for the delay -- we've been mostly heads down on Ambassador 0.50.0!

Let me verify, though: no `volumeMounts` section generated at all with the default `values.yaml`, correct?

kyschouv:
Yep, using the default values.yaml will result in the same result as the current chart. I wasn't sure about was whether you'd want to bump the patch or minor version number on the chart with the change - let me know if I should change that.

kflynn:
Great, thanks for the confirmation!

We've been generally sync'ing the version number with the version of Ambassador that's current when the Helm chart is updated. Happy to take suggestions, but your switch to 0.40.1 is what I'd do at this point.

kyschouv:
Awesome, sounds great!

kflynn:
I think we have everything we need, yes. I'm planning to land this into 0.40.2 very soon -- thanks!

kflynn:
Taking for the next 0.40

PR #953, opened 2018-11-04
Update doc-links to support new top level linking

sipplified:


richarddli:
@sipplified could you branch off of `stable` and rebase off of that for this commit? doc changes should land there first.

PR #955, opened 2018-11-05
Added envoy 1.8 zipkin driver configurations and allowing trace headers in ExtAuth requests

alexgervais:


kflynn:
@alexgervais Thanks, this is awesome! Do you have a way to test it?

alexgervais:
@kflynn can you take another look? This is needed for tracing in 0.50 without regressions.

kflynn:
Done -- sorry for the delay! Took a little bit to get things in shape on our end.

PR #965, opened 2018-11-09
Update doc-links.yml to support new top-level menu links

sipplified:


sipplified:
I branched from `stable` this time, it should be right. Let me know if it can't be merged.

PR #976, opened 2018-11-15
Add envoy.router filter to redirect listener

containscafeine:
Current redirect listener does not specify the envoy.router field
in envoy filters, which causes redirects to fail. This commit
fixes that.


kflynn:
Thanks for running this down!

PR #982, opened 2018-11-19
Make CONFIG_DIR and PYTHON_EGG_CACHE configurable, fix hardcoded paths

n1koo:
Fixes some harcoded paths and essentially new version of https://github.com/datawire/ambassador/pull/927

This PR allows folks to use read-only roots and Ambassador won't need to explicitely write to its app root


n1koo:
Ah right, should have tested with the defaults. Sorry about that. Setting `CUSTOM_CONFIG_BASE_DIR` explicitly now and defaulting that to `AMBASSADOR_ROOT` so it works as before

kflynn:
Thanks! Looks good and runs clean for me.

PR #985, opened 2018-11-19
Add statsd support to v2

containscafeine:


richarddli:
Per @kflynn "Looked at your PR — I think that we should maybe support setting a host to send stats to? since Envoy in V2 can do that by hostname and not just IP address now."

Yes, I agree -- and then we can remove the `socat` sidecar.  See https://github.com/datawire/ambassador/issues/841

kflynn:
Exactly (re ditching the sidecar). :smile:

ambassador/ambassador/config/config.py
```diff
@@ -74,6 +74,11 @@ def __init__(self, schema_dir_path: Optional[str]=None) -> None:
             # with our ACResource class.
             schema_dir_path = resource_filename(Requirement.parse("ambassador"), "schemas")
 
+        self.statsd = {'enabled': (os.environ.get('STATSD_ENABLED', '').lower() == 'true')}
+        if self.statsd['enabled']:
+            self.statsd['interval'] = os.environ.get('STATSD_FLUSH_INTERVAL', '1')
+            self.statsd['ip'] = socket.gethostbyname(os.environ.get('STATSD_HOST', 'statsd-sink'))
```
kflynn: What happens if `socket.gethostbyname` can't find the host?

containscafeine: Yeah, everything crashes :fire: Thanks!
I've caught this exception, PTAL :)

kflynn:
Not sure about what's up with the git-tag issue in CI here? Tested locally and it looks OK, so I'm taking it.

PR #1023, opened 2018-12-08
Fixed make command when path has whitespace

### LINKED ISSUE #1022: Build fails when there's a whitespace in $PATH
**Describe the bug**
When trying to build on Mac, `make` fails on trying to set the`PATH` when running `pytest` when there is a whitespace character in the path.

**To Reproduce**
1. `mkdir -p "/tmp/this is valid/`
2. `export PATH=$PATH:"/tmp/this is valid"`
3. `make DOCKER_REGISTRY=-`

**Expected behavior**
It should not fail.

 - Ambassador: master-49b069b2
- OS: macOS 10.14.1

**Additional context**
I could easily have a more reasonable `PATH`, but the reason this might be a bigger problem is that Visual Studio Code adds itself as `/Applications/Visual Studio Code.app` on Mac by default.

martincognite:
Fixes #1022. `PATH=$PATH` is already in quotes both in `utils.sh` and `test.sh`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -220,7 +220,7 @@ ambassador/ambassador/VERSION.py:
 version: ambassador/ambassador/VERSION.py
 
 e2e-versioned-manifests: venv website-yaml
-	cd end-to-end && PATH=$(shell pwd)/venv/bin:$(PATH) bash create-manifests.sh $(AMBASSADOR_DOCKER_IMAGE)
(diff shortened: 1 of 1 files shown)


kflynn:
Closing since your later PR, which I already took, superseded this one. Thanks!!

PR #1044, opened 2018-12-19
Use env and annotations from values file

osnagovskyi:
This change will enable users to pass arbitrary env variables to deployment or daemonset during installation instead of editing deployment object afterward.
Same goes for custom pod annotations.


nbkrause:
PR looks good to me. 

@osnagovskyi If you could add something to the [README](https://github.com/datawire/ambassador/blob/master/helm/ambassador/README.md) indicating the added functionality that would be great.

osnagovskyi:
@nbkrause I've updated README, thanks!

PR #1048, opened 2018-12-19
Update helm-chart readme to reflect current image

williamhogman:
Default image tag was incorrectly stated, this commit corrects it.


kflynn:
(I wavered a bit on which branch because of how we're currently handling master, stable, and the release/0.40 branch. I think stable is, in fact, correct. Thanks!!)

Issue #1058, opened 2018-12-20
0.50.0-rc1: v1 authentication documentation does not mention require proto option


flands:
**Describe the bug**

If you use v1 authentication documentation then authentication will not work. This is because the proto option is required currently and does not have a default.

**To Reproduce**

Enable v1-based authentication in Ambassador using the documentation. For example:

```
---
apiVersion: ambassador/v1
kind:  AuthService
name:  authentication
auth_service: "example-auth:3000"
path_prefix: "/extauth"
```

This fails because the `proto` option is required. Two solutions to fix:

1.) Update documentation to state this is required and the valid options
2.) Update documentation to state this is optional with valid options and set a default (e.g. http)

**Expected behavior**

Authentication works when following the documentation.

**Versions (please complete the following information):**
 - Ambassador: 0.50.0-rc1
 - Kubernetes environment kops
 - Version 1.11.5

**Additional context**

PR #1059, opened 2018-12-20
Add required proto option to authentication documentation

flands:
Option 1 to address #1058 -- option 2 will be posted shortly

Note: Unsure process to add to 0.50.0 branch


kflynn:
Taking into `stable` since we normally do doc-only changes there (it's a long story ;) ). Thanks!

PR #1072, opened 2018-12-31
[Fix] Update some additional paths to be properly quoted (fixes paths…

er1c:
… with a space in them)


Makefile
```diff
@@ -360,12 +360,12 @@ clean-test:
 test: setup-develop cluster.yaml
 	cd ambassador && \
 	AMBASSADOR_DOCKER_IMAGE=$(AMBASSADOR_DOCKER_IMAGE) \
```
er1c: I didn't audit this, but likely this should probably be escaped too

Issue #456, opened 2018-05-22
Investigate GRPC-Web/JSON support in Ambassador


plombardi89:
A number of users have asked for GRPC-Web/JSON transcoder support in Ambassador. We need to know how to integrate this into Ambassador before we proceed.

# CoS
- [ ] A design for how the mapping feature would work has been agreed upon by the team.
- [ ] A prototype that supports the agreed design has been produced.
- [ ] A plan for next steps / release has been created or decided.

richarddli:
See issue #411

dvavili:
Is this targeted for a particular release? Would be good to have this.

mprigge:
For what it's worth, I'm very interested to see Ambassador support this as well. I'm sure you're already aware of the support that's in envoy already. Not sure how hard it would be to have Ambassador support a configuration similar to [this one](https://github.com/grpc/grpc-web/blob/master/net/grpc/gateway/examples/echo/envoy.yaml). Is it just an issue of wrapping envoy configuration correctly or does it go deeper than that?

slamdev:
@nigeltiany you can build ambassador from sources and replace **envoy.j2** with custom one, like here https://github.com/slamdev/catalog/tree/master/infra/ambassador/deploy

key parts are: to add [grpc_web filter](https://github.com/slamdev/catalog/blob/master/infra/ambassador/deploy/envoy.j2#L108) and [grpc and cors mappings](https://github.com/slamdev/catalog/blob/master/services/inventory/deploy/service.yaml) for the service

nigeltiany:
I have built ambassador from that ```envoy.j2``` file and it works. I didn't even have to configure cors mappings. Thank you @slamdev 👈 True hero right here.
Now, submitting a PR is easy because it's just a single new object with one line of code but how to write the end-to-end test. Does any of the maintainers @ambassador want to look at this fix?

richarddli:
@nigeltiany Would love to, although be warned that we've changed things up quite a bit in the (imminent) 0.50 ... so it's possible that your change won't just work in the next release. The current code is on the `release/0.50.0` branch, if you want to take a look.

richarddli:
See PR #1201, thank you to @rotemtam.

kflynn:
gRPC-Web is supported in 0.52.0.

shubhamranjan:
I can see that support for grpc-web transcoder has been implemented but what about grpc-json transcoder ? I am unable to find ambassador documentation regarding this. 

[...]

PR #1076, opened 2019-01-02
grpc-web filter addition for envoy template

jtdawso:
Issue #456 
@kflynn 
@richarddli

### DIFF
--- a/ambassador/templates/envoy.j2
+++ b/ambassador/templates/envoy.j2
@@ -104,6 +104,10 @@
               ]
             },
             "filters": [
+              {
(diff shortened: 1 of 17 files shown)


saeidakbari:
No one to solve this conflicts?

kflynn:
I'm going to go ahead and close this, since it won't work in 0.50 -- we'd really love a new PR though! :smile:

gertvdijk:
@jtdawso I think, what was meant is that the Jinja2 template would only work for Envoy v1, not for v2. See now-updated docs in #1175.

PR #1090, opened 2019-01-09
fix: fixes the issue with websocket handshake and authz filter

gsagula:


kflynn:
Let's not merge this into master. Already took #1091 into release/0.50.0. :smile:

PR #1092, opened 2019-01-09
More mypy cleanup

kflynn:
This tackles mypy errors other than in `v2tls.py` (since there's other work outstanding in that file). Also starting the process of dropping the last bits of `envoy_tls`.


kflynn:
The remaining test failure is spurious. Merging.

PR #1117, opened 2019-01-16
Fix redirect_cleartext_from bug

containscafeine:
Do not merge, test and some refactor incoming


kflynn:
Merging this since it looks OK locally, and I'd just as soon tackle test resource issues all at once.

PR #1120, opened 2019-01-17
Refactor to a new cached Dockerfile

containscafeine:


kflynn:
There's a conflict here, but also, `$(AMBASSADOR_DOCKER_IMAGE)-cached` will change for nearly every build, which negates the benefit of the cached base image. Let's separate the names, and assume that we'll manually update the cached base when need be.

kflynn:
Moving back to "In Progress" since this needs more work.

containscafeine:
@kflynn I've made the changes, PTAL! :)

containscafeine:
@kflynn to test this out, you'll first need to run `make ambassador-docker-image-cached` and then build the image.
This now builds 2 cached container images; one for caching install time dependencies and another for runtime dependencies (curl, python3, ambex, kubewatch).

Once both are built and pushed, we can go ahead with this.

Curious though about how do you feel about 2 cached images :confused:

kflynn:
It took a little reading to figure out the point of the two, but I think I have a clue now. Cool.

kflynn:
Oops. I made a conflict for you when I took your validation fix, mea culpa! If you can fix that I think we're good to go here!

PR #1121, opened 2019-01-18
Bypass Authentication via Configuration

### LINKED ISSUE #174: Bypass Authentication via Configuration
Many types of services need to be public or universally accessible without an authentication or authorization check. For example, Datawire's Scout service.

Another use case is dynamically generated hosts or paths, for example, using [Forge](https://forge.sh) it is common to use the `{{ service.name }}-{{ branch.name }}` strategy to provide independent URL's to access development versions of a service. Asking developers to continually update a rules table for that kind of information seems like a pain.

To accomodate these use cases with the external auth module in Ambassador currently the module developer needs to create a way for developers to specify whitelist rules. Some strategies include updating an in-memory data structure such as a set of ignorable Hosts and Paths. Another idea might be an external database or some other piece of shared storage. The service author can then read and use that configuration to configure whitelist rules. However, this is not ideal... it puts a burden on the external auth developer to think about this functionality up front and also exposes a less than ideal UX.'

It would be preferable if Ambassador's configuration mechanism could specify whitelist properties that would tell Envoy to skip routing to the Auth service in certain cases, for example. 

The bypass mechanism ideally should support host header and URL paths.

patricksanders:
This PR adds a `per_filter_config` to the Envoy route when a mapping has `bypass_auth` set to `true` in Ambassador 0.50.0+. To accomplish this, I added:
- a `bypass_auth` field to the v1 Mapping schema
- `bypass_auth` to the ACMapping resource with a default of `False`
- `per_filter_config` to the Envoy v2 route configuration (conditionally)
- `per_filter_config` to the Envoy configuration template
- a test to ensure a 200 response from a mapping with `bypass_auth` set to `true`
- references to `bypass_auth`

Additionally, my editor decided to clean up some whitespace.

Seeking feedback as this is a non-trivial change. The testing is a little light -- I would especially appreciate feedback on where/how to effectively test this.

Fixes #174

### DIFF
--- a/ambassador/ambassador/config/acmapping.py
+++ b/ambassador/ambassador/config/acmapping.py
@@ -23,7 +23,7 @@ from .acresource import ACResource
 ## Each Mapping object has a group_id that reflects the group of Mappings
 ## that it is a part of. By definition, two Mappings with the same group_id
 ## are reflecting a single mapped resource that's going to multiple services.
-## This implies that Mapping.group_id() is a very, very, very important 
+## This implies that Mapping.group_id() is a very, very, very important
 ## thing that can have dramatic customer impact if changed! (At some point,
 ## we should probably allow the human writing the Mapping to override the
 ## grouping, in much the same way we allow overriding precedence.)
@@ -55,6 +55,7 @@ class ACMapping (ACResource):
                  rewrite: Optional[str]="/",
                  case_sensitive: bool=False,
                  grpc: bool=False,
+                 bypass_auth: bool=False,
 
                  # We don't list "method" or "method_regex" above because if they're
                  # not present, we want them to be _not present_. Having them be always
@@ -81,4 +82,5 @@ class ACMapping (ACResource):
                          rewrite=rewrite,
                          case_sensitive=case_sensitive,
                          grpc=grpc,
+                         bypass_auth=bypass_auth,
                          **kwargs)
--- a/ambassador/ambassador/ir/irmapping.py
+++ b/ambassador/ambassador/ir/irmapping.py
@@ -81,6 +81,7 @@ class IRMapping (IRResource):
         "tls": True,
         "use_websocket": True,
         "weight": True,
+        "bypass_auth": True,
 
         # Include the serialization, too.
         "serialization": True,
@@ -275,7 +276,8 @@ class IRMappingGroup (IRResource):
         'prefix': True,
         'prefix_regex': True,
         'rewrite': True,
-        'timeout_ms': True
+        'timeout_ms': True,
+        'bypass_auth': True
     }
 
     DoNotFlattenKeys: ClassVar[Dict[str, bool]] = dict(CoreMappingKeys)
@@ -427,8 +429,8 @@ class IRMappingGroup (IRResource):
         """
         Finalize a MappingGroup based on the attributes of its Mappings. Core elements get lifted into
         the Group so we can more easily build Envoy routes; host-redirect and shadow get handled, etc.
-        
-        :param ir: the IR we're working from 
+
+        :param ir: the IR we're working from
         :param aconf: the Config we're working from
         :return: a list of the IRClusters this Group uses
         """
--- a/ambassador/tests/kat/t_extauth.py
+++ b/ambassador/tests/kat/t_extauth.py
@@ -21,7 +21,7 @@ apiVersion: ambassador/v0
 kind:  Module
 name:  ambassador
 config:
-  buffer:  
+  buffer:
     max_request_bytes: 16384
     max_request_time: 5000
 ---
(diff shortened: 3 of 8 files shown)


gsagula:
Thanks for the contribution. I will take the first pass.

ambassador/ambassador/config/acmapping.py
```diff
@@ -23,7 +23,7 @@
 ## Each Mapping object has a group_id that reflects the group of Mappings
 ## that it is a part of. By definition, two Mappings with the same group_id
 ## are reflecting a single mapped resource that's going to multiple services.
-## This implies that Mapping.group_id() is a very, very, very important 
+## This implies that Mapping.group_id() is a very, very, very important
```
gsagula: nit: I see that this kind of comments is all over Ambassador, but try to avoid subjective language as much as possible.

patricksanders: Agreed, although I didn't add this comment; my IDE just decided to remove the whitespace when I saved. Happy to update this comment for clarity while I'm in here if you'd like though.

kflynn: I wrote it. :smile: Happy to entertain suggestions as a separate PR. :smile:

ambassador/ambassador/ir/irmapping.py
```diff
@@ -427,8 +429,8 @@ def finalize(self, ir: 'IR', aconf: Config) -> List[IRCluster]:
         """
         Finalize a MappingGroup based on the attributes of its Mappings. Core elements get lifted into
         the Group so we can more easily build Envoy routes; host-redirect and shadow get handled, etc.
-        
-        :param ir: the IR we're working from 
+
+        :param ir: the IR we're working from
```
gsagula: Can you elaborate it a bit, please?

docs/reference/mappings.md
```diff
@@ -120,13 +121,14 @@ If absolutely necessary, you can manually set a `precedence` on the mapping (see
...
 
-For example, defining a mapping with only a `/` prefix will catch all requests previously unhandled and forward them to an external service:
+For example, defining a mapping with only a `/` prefix will catch all requests previously unhandled and forward them to an external service, bypassing authentication (if configured):
 
 ```yaml
 ---
 apiVersion: ambassador/v0
 kind: Mapping
 name: catch-all
+bypass_auth: true
```
gsagula: In order to maintain the previous semantic "catch-all", should `bypass_auth` be set to false?

ambassador/tests/kat/t_extauth.py
```diff
@@ -357,6 +369,9 @@ def queries(self):
         # [5]
         yield Query(self.url("target/"), headers={"X-Forwarded-Proto": "https"}, expected=200)
 
+        # [6]
+        yield Query(self.url("target/unauthed/"), headers={"Requested-Status": "200"}, expected=200)
```
gsagula: Thanks for adding this check. I think you want to make sure that Envoy will skip the authorization process. Would make more sense to set Requested-Status to HTTP 401, but expect an HTTP 200 response?

ambassador/ambassador/ir/irmapping.py
gsagula: Can you elaborate it a bit more, please?

ambassador/ambassador/config/acmapping.py
gsagula: @patricksanders Thanks, We don't need to change anything here.

ambassador/ambassador/ir/irmapping.py
patricksanders: This was also just a whitespace change.

ambassador/tests/kat/t_extauth.py
patricksanders: Yep, I'll do that. Is there any documentation that explains what the provided headers do on the backend?

docs/reference/mappings.md
patricksanders: This was a bit of a shot in the dark. The intent was to show a usage of `bypass_auth`, but this is a questionable use case. I'll probably just remove this (since it defaults to `false` anyway).

ambassador/tests/kat/t_extauth.py
patricksanders: Setting `Requested-Status` to 401 makes this test return a 401. I added more assertions to make sure we're not hitting the auth service. Is this doing what I think it is?

docs/reference/mappings.md
patricksanders: Removed.

ambassador/tests/kat/t_extauth.py
gsagula: Right, `requested-status` is actually telling to the backend what to do. The assertions that you added are perfect. 

There is a KAT [tutorial](https://github.com/datawire/kat/blob/master/docs/tutorial.rst), but nothing very specific about what the headers do to the backend. The quickest way to know is by looking at the backend server implementation, e.g. [requested-status](https://github.com/datawire/ambassador/blob/3cd648c1133aac76af3bc54ebe6bf8dfa1e541aa/kat/backend/server.go#L51).

gsagula:
LGTM, thank you!

This should be good to go as soon as the tests passes. @kflynn has a branch that is supposed to fix our CI.

kflynn:
Checks out locally, merging -- many thanks @patricksanders!!

PR #1138, opened 2019-01-22
Allow adding extra env in helm chart

### LINKED ISSUE #1081: Switch to STATSD_ENABLED in Helm Chart
Ambassador 0.39.0 changed the stats export behavior from running socat as a sidecar to using an environment variable `STATSD_ENABLED`. 

Ambassador 0.50.0-ea7 moved away from socat entirely for exporting statistics. 

While Ambassador has moved away from the sidecar approach to exporting stats, the helm chart has not yet. 

#1044 Adds the ability to add environment variables with `values.yaml` so adding `STATSD_ENABLED` is taken care of. Because of #1034 we need a way to turn off the statsd-sink sidecar container that is deployed by default. 

Preferably, we would remove the statsd-sink sidecar from the chart entirely but to avoid regressions we should leave is optionally configurable in the values file.

Flydiverny:
Helps with #1081

#1122 added possibility to disable statsd sidecar

### DIFF
--- a/helm/ambassador/README.md
+++ b/helm/ambassador/README.md
@@ -51,6 +51,7 @@ The following tables lists the configurable parameters of the Ambassador chart a
 | `image.imagePullSecrets` | Image pull secrets | None
 | `daemonSet` | If `true `, Create a daemonSet. By default Deployment controller will be created | `false`
 | `replicaCount` | Number of Ambassador replicas  | `1`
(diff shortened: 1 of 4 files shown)


Flydiverny:
Ups, suppose this duplicates some of #1044

kflynn:
Closing because we took #1044.

kflynn:
Thanks!

PR #1145, opened 2019-01-24
Fix link to statsd-sink.yaml

Flydiverny:


kflynn:
Taking into `stable` (that's where doc PRs need to end up). Thanks!

PR #1146, opened 2019-01-24
Allow specifying optional externalTrafficPolicy

Guyzeroth:
Modified Helm chart to allow specifying LoadBalancer externalTrafficPolicy for correct setting of x-forwarded-for to upstream clients .  

As per https://www.getambassador.io/reference/modules#use_remote_address

' You may also need to set externalTrafficPolicy: Local '


kflynn:
Thanks for the contribution! We've moved the Ambassador Helm chart into [Helm's stable repo](https://github.com/helm/charts/tree/master/stable/ambassador) -- closing this one since the `stable` chart includes this, thanks!

PR #1149, opened 2019-01-25
added ingress resource

bartlettt:
I've added an ingress resource as an alternate method to route traffic to ambassador.

My use-case is such that I want the cluster ingress controller to handle TLS and redirects. I encountered a problem where http->https redirects are applied to all requests, with no way to differentiate internal and external traffic. 

In addition this allows a convenient way to handle Certificate resources, as annotations on the ingress resource will automatically prompt cert-bot to grab the required cert.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,4 @@ ambassador-deployment-*.yaml
 ambassador-secrets-deployment.yaml
 *failures.txt
 kubernaut-claim.txt
+*.swp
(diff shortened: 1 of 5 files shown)


richarddli:
Thanks! Is there a reason why you couldn't use two Ambassadors (internal/external) to handle your redirect case?

Regarding Certificate resources, https://github.com/Mimetis/ambassadorandtls might be a good solution?

PR #1167, opened 2019-01-29
Ignore TLS module host when SNI specified

### LINKED ISSUE #1156: TLS module alongwith SNI crashes Ambassador (crash on invalid configuration)
Envoy goes for a toss when it's fed an invalid configuration, i.e. it goes in an infinite loop with 100% CPU usage.

A configuration like the following where SNI and TLS module are both configured -

```yaml
      ---
      apiVersion: ambassador/v1
      kind: TLSContext
      name: same-context-2
      hosts:
      - tls-context-host-2
      secret: same-secret-2
      ---
      apiVersion: ambassador/v0
      kind: Module
      name: tls
      config:
        server:
          enabled: True
          secret: user-secret
```

When the resulting configuration is fed to envoy, the following message is logged infinitely -
```
[2019-01-28 14:00:34.606][000034][warning][config] [bazel-out/k8-dbg/bin/source/common/config/_virtual_includes/grpc_mux_subscription_lib/common/config/grpc_mux_subscription_impl.h:70] gRPC config for type.googleapis.com/envoy.api.v2.Listener rejected: Error adding/updating listener ambassador-listener-443: error adding listener '0.0.0.0:443': partial wildcards are not supported in "server_names"
```

and no response is received to any request.

containscafeine:
This commit makes Ambassador not honor the TLS module's conflicting
configuration when specified with SNI. Prior to this commit, this
configuration made Envoy go into a spin loop and rendering it
unusable.

Also added a test for this fix.

Fix #1156

### DIFF
--- a/ambassador/ambassador/envoy/v2/v2listener.py
+++ b/ambassador/ambassador/envoy/v2/v2listener.py
@@ -426,7 +426,19 @@ class V2Listener(dict):
             })
 
         for name, hosts, ctx in envoy_contexts:
+            skip_context = False
             if not ctx:
+                skip_context = True
+
+            if global_sni:
+                for host in hosts:
+                    # Wildcard is not accepted in server_names in envoy, so we need to invalidate this
+                    if host == '*':
+                        config.ir.logger.info("V2Listener: invalid host {} found envoy contexts, ignoring context".format(host))
+                        skip_context = True
+                        break
+
+            if skip_context:
                 continue
 
             config.ir.logger.info("V2Listener: SNI route check %s, %s, %s" %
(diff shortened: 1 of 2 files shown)


kflynn:
From #1168:

"An important note here is that this #1156 fix is different from @containscafeine's work in #1167: specifically, #1167 ignores the TLS `Module` if you also have `TLSContext`s that enable SNI, but this PR uses the TLS `Module` to provide a fallback cert instead. It's actually really tough to know which is more correct in the absence of real-world use.

After some agonizing, and because @containscafeine and I aren't going to overlap again for another 10 hours or so (:stuck_out_tongue: ), I'm going to go with this one because I _think_ it's less surprising to a customer migrating from 0.40. I could be wrong, of course..."

No offense @containscafeine! You may turn out to be right... 😄

Issue #1155, opened 2019-01-28
Ambassador crashes when there are duplicate mappings


containscafeine:
Config that looks like -
```
      apiVersion: ambassador/v0
      kind:  Mapping
      name:  httpbin_mapping
      prefix: /httpbin-1/
      service: httpbin.org:80
      host: tls-context-host-1
      host_rewrite: httpbin.org
      ---
      apiVersion: ambassador/v0
      kind:  Mapping
      name:  httpbin_mapping
      prefix: /httpbin-2/
      service: httpbin.org:80
      host: tls-context-host-2
      host_rewrite: httpbin.org
```
crashes Ambassador, and the pod goes in `CrashLoopBackOff` state. This happens because both the mappings are named the same (`httpbin_mapping` in this case)

Ambassador should report the error and continue on, instead.

kflynn:
I already have a partial fix for this.

kflynn:
Fixed in RC7.

PR #1168, opened 2019-01-30
Post-RC6 tweaks

kflynn:
Fixes for #1155 and #1156, in particular.

An important note here is that this #1156 fix is different from @containscafeine's work in #1167: specifically, #1167 ignores the TLS `Module` if you also have `TLSContext`s that enable SNI, but this PR uses the TLS `Module` to provide a fallback cert instead. It's actually really tough to know which is more correct in the absence of real-world use.

After some agonizing, and because @containscafeine and I aren't going to overlap again for another 10 hours or so (😛 ), I'm going to go with this one because I _think_ it's less surprising to a customer migrating from 0.40. I could be wrong, of course...

### DIFF
--- a/ambassador/ambassador/cli.py
+++ b/ambassador/ambassador/cli.py
@@ -187,6 +187,9 @@ def dump(config_dir_path: Parameter.REQUIRED, *,
             od['v2'] = v2config.as_dict()
 
         if dump_diag:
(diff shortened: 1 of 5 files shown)


containscafeine:
This makes sense :) Thank you!

PR #1170, opened 2019-01-30
Fix #1098 for real

### LINKED ISSUE #1098: Ambassador 0.5.0-RC4 attempts to list all secrets in cluster even with AMBASSADOR_SINGLE_NAMESPACE
**Describe the bug**
Ambassador 0.5.0-RC4 attempts to list all secrets in cluster even with AMBASSADOR_SINGLE_NAMESPACE set to true.

**To Reproduce**
1. Run Ambassador with below conditions
AMBASSADOR_SINGLE_NAMESPACE 'true' & AMBASSADOR_NAMESPACE set
Service Account that only has access to secrets within the chosen namespace/project
2. Observe log
[ambassador-test-31-5fm8q.log](https://github.com/datawire/ambassador/files/2746746/ambassador-test-31-5fm8q.log)

**Expected behavior**
Ambassador will only attempt to list secrets within the chosen namespace/project

**Versions:**
 - Ambassador: 0.5.0-RC4 
 - Kubernetes environment : Openshift
 - Version: 3.6

**Additional context**
Does not occur in RC3

kflynn:
...and include a test with RBAC enforcing single-namespace-only access, so that we don't break it again.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,7 +63,7 @@ RUN releng/install-py.sh prd install */requirements.txt
 RUN rm -rf ./multi ./ambassador
 
 # Grab kubewatch
-RUN wget -q https://s3.amazonaws.com/datawire-static-files/kubewatch/0.3.15/$(go env GOOS)/$(go env GOARCH)/kubewatch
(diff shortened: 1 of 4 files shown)


kflynn:
(I merged this because I saw that the CI test passed even though the docs weren't finished building yet.)

PR #1152, opened 2019-01-25
Helm chart bump

Flydiverny:
Would solve #1034, #1081 and #1084 for 0.50

Closes #1138 
Closes #1136 

I took some liberties.
I generated a new chart with helm create and migrated over existing configs.
To avoid confusion with the statsd previous socat solution I renamed the sidecar to prometheus exporter.

I recommend checking the diff without whitespace diffing

PR #1172, opened 2019-01-31
Updated chart with new best practice labels

Flydiverny:
This PR only updates the template with new labels without changing any functionallity


Flydiverny:
See #1152

PR #1174, opened 2019-02-01
docs: correct template filename extension (json -> j2)

gertvdijk:


gertvdijk:
Oh well, don't bother this one. I noticed that since 0.50.0 loading a custom configuration using a ConfigMap [is not supported any longer](https://github.com/datawire/ambassador/blob/0.50.0/CHANGELOG.md#ambassador-0500).

So that means the documentation should reflect that instead.

Issue #384, opened 2018-04-23
Move/copy helm chart to https://github.com/kubernetes/charts


monotek:
I would like to see (at least a copy of) the Helm chart on https://github.com/kubernetes/charts.

Any thoughts?

MichielDeMey:
Agreed, it's the de facto place to look for helm charts. I was actually surprised that it hasn't been added yet.

Flydiverny:
An alternative to that could be adding the datawire repository to http://hub.helm.sh (https://github.com/helm/hub).

But I think it would make sense to move it as well. Think there's enough community interest for the chart to have its own life, version and release cycle. Considering using my #1152 pr as a starting point for a PR in the helm repository.

Flydiverny:
There's now a PR in https://github.com/helm/charts/pull/11294

Flydiverny:
It has been merged!

richarddli:
Thanks so much! @Flydiverny what do you think we should do about https://www.getambassador.io/user-guide/helm?

MichielDeMey:
On a side-note, we should probably get rid of the Helm chart in this repository to prevent confusion and fragmentation of the Helm charts. PR up https://github.com/datawire/ambassador/pull/1215.

scottrigby:
Nice work @Flydiverny  🎉 

re https://github.com/datawire/ambassador/issues/384#issuecomment-457832939, even though we're working to move all Helm charts to distributed repos for [hub.helm.sh](https://hub.helm.sh), the tools to do so aren't as easy yet as managing a chart on github.com/helm/charts, with it's automated testing and Helm repo releases. So was a good idea to move this chart to github.com/helm/charts in the meantime.

However we're working on those tools to make it simple for orgs to to self-host their own charts on GitHub (giving you full control of who is added to the repo, how PRs are merged, etc), and access to the same automated testing and Helm repo release tooling (through one-click install GitHub apps). Once those are in place, I would like to revisit this with you all to move the chart back to the Datawire org – ideally as a separate GitHub repo just for datawire chart(s) – if you're up for that?

PS, you can already see your chart at https://hub.helm.sh/charts/stable/ambassador, since there was a hub deploy since helm/charts#11294 was merged 😄

kflynn:
Oh hey, this is done now! Thanks @Flydiverny!

PR #1213, opened 2019-02-12
Update helm docs to point to helm/charts chart

Flydiverny:
@richarddli something like this maybe (#384)


MichielDeMey:
LGTM, pointing to the stable helm repo should suffice for most installations.

Flydiverny:
Yeah I figured if you use helm you know about the stable repo. :)

PR #1215, opened 2019-02-12
Removed Helm chart to prevent confusion

MichielDeMey:
This should prevent confusion as to which Helm chart to use for the Ambassador installation.
Part of #384.


kflynn:
Thanks!

Issue #1216, opened 2019-02-12
Do not attempt to retrieve stats before Envoy is running


gertvdijk:
**Describe the bug**
Ambassador appears to retrieve Envoy's statistics before it's running. This causes WARNING-level log entries like:

    WARNING: EnvoyStats.update failed: HTTPConnectionPool(host='127.0.0.1', port=8001): Max retries exceeded with url: /stats (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f329c1ddc88>: Failed to establish a new connection: [Errno 111] Connection refused',))

*(Originally discussed in #1203, requested for reporting individual issue as per comments.)*

**To Reproduce**
1. Deploy Ambassador 0.50.1 without any `Service`s for it.
2. Read logs.
3. Find errors about not being able to reach Envoy's `/stats`.

**Expected behavior**
No attempt to read Envoy stats if it hasn't started Envoy.

**Versions (please complete the following information):**
 - Ambassador: 0.50.1
 - Kubernetes environment *not really relevant*
 - Version *not really relevant*

richarddli:
Fixed in 0.50.2

PR #1223, opened 2019-02-14
Clean up startup process

### LINKED ISSUE #1203: 0.50.1 fails to start without any annotated Service deployed
**Describe the bug**
Without any Services deployed yet, using the official Docker image, 0.50.0 works OK, 0.50.1 fails with the following errors in the Kubernetes pod logs:

    diagd 0.50.1 [P42TAmbassadorEventWatcher] WARNING: EnvoyStats.update failed: HTTPConnectionPool(host='127.0.0.1', port=8001): Max retries exceeded with url: /stats (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f329c1ddc88>: Failed to establish a new connection: [Errno 111] Connection refused',))

Full pod logs here: http://paste.openstack.org/show/744836/

**To Reproduce**
1. Deploy Ambassador in single namespace mode with the `quay.io/datawire/ambassador:0.50.0` image. *Ensure that you don't have any Services with Ambassador annotations deployed.*
2. Notice that you can get it ready in Kubernetes, without errors in the pod logs.
3. Change the deployment to use the `quay.io/datawire/ambassador:0.50.1` image.
4. Notice that Ambassador seems not to be able to connect to Envoy's stats endpoint and fails to get ready.
5. Revert the image to `0.50.0` - everything works as expected again.

**Expected behavior**
An upgrade of 0.50.0 -> 0.50.1 should not impact this internal connection Ambassador-Envoy and it should be able to get ready on its own without having to rely on the presence of Services, which may be deployed at a later time.

 - Ambassador: 0.50.1
 - Kubernetes environment: self-deployed, kubeadm
 - Version: 1.12.3

kflynn:
This PR fixes #1203 and #1216 by being more careful about the startup process: 

1. Having a configuration mounted on `/ambassador/ambassador-config` will automatically prevent watching for Kubernetes events (you can override this by setting `AMBASSADOR_FORCE_KUBEWATCH` in the environment).

2. When watching Kubernetes, it is no longer an error if we don't find any annotations. Instead,  Ambassador will come up on port 80 with the diagnostic service mapped (including the readiness and liveless check endpoints). This is what 0.40 used to do.

3. Ambassador won't start trying to fetch Envoy stats until it has started Envoy.

End results:

- Most users will not have anything mounted on `/ambassador/ambassador-config`. For these users, Ambassador will launch, check Kubernetes services for annotations, and start Envoy after successfully fetching everything the first time from Kubernetes.
   - Ambassador won't pass liveness and readiness checks until after that first-time configuration, so if it's totally unable to talk to Kubernetes and you have the checks enabled, Kubernetes will kill the pod after a little while. This is probably the most sane outcome.

- For users who _do_ have a configuration mounted on `/ambassador/ambassador-config`, Ambassador will launch, read that configuration, and ignore Kubernetes entirely. 
   - Kubernetes liveness and readiness checks will start passing once Ambassador has read the disk configuration. If the disk configuration is utterly broken, to the point that Ambassador can't even parse it, Kubernetes will end up killing the pod if the checks are configured. Again, this is probably the most sane outcome.
   - If you want to change Ambassador's configuration, you need to change the disk configuration and then do a rolling restart by hand. There are currently no plans to have Ambassador watch filesystem events, but we'd be happy to take PRs (you could modify `diagd` to publish a `CONFIG_FS` event in response to a `POST` to an external endpoint, or modify the existing `update` endpoint to know the difference between a URL to fetch from and a disk path, or - less good - modify `diagd` itself to watch the FS).


Aracki:
@kflynn  #1668 👀

PR #1225, opened 2019-02-14
update prom yaml

richarddli:


kflynn:
Looks good here, but I'll let @nbkrause give the final OK. 😄

Issue #1224, opened 2019-02-14
Users want to be able to mix and match HTTP and HTTPS upstreams for testing and development


kflynn:
At present you can't do this:

```
---
apiVersion: ambassador/v1
kind: Mapping
prefix: /foo
service: http://fooservice1
---
apiVersion: ambassador/v1
kind: Mapping
prefix: /foo
service: https://fooservice2
```

because the logic that infers `Cluster`s will decide that the HTTP upstream is incompatible with the HTTPS upstream. I don't think that this is an Envoy limitation, so it shouldn't be an Ambassador limitation either.

PR #1229, opened 2019-02-15
Allow TLS context in AuthService

### LINKED ISSUE #1202: AuthService no longer recognized tls options
**Describe the bug**
v1 AuthService throws an error when specifying the `tls` property.

```
auth-web.default.1: not a valid AuthService: Additional properties are not allowed ('tls' was unexpected) Failed validating 'additionalProperties' in schema: {'$schema': 'http://json-schema.org/schema#', 'additionalProperties': False, 'id': 'https://getambassador.io/schemas/mapping.json', 'properties': {'allow_request_body': {'type': 'boolean'}, 'allowed_authorization_headers': {'items': {'type': 'string'}, 'type': 'array'}, 'allowed_request_headers': {'items': {'type': 'string'}, 'type': 'array'}, 'ambassador_id': {'anyOf': [{'type': 'string'}, {'items': {'type': 'string'}, 'type': 'array'}]}, 'apiVersion': {'enum': ['ambassador/v1']}, 'auth_service': {'type': 'string'}, 'kind': {'type': 'string'}, 'name': {'type': 'string'}, 'path_prefix': {'type': 'string'}, 'proto': {'oneOf': [{'enum': ['http']}, {'enum': ['grpc']}]}, 'timeout_ms': {'type': 'integer'}}, 'required': ['apiVersion', 'kind', 'name', 'auth_service'], 'type': 'object'} On instance: {'apiVersion': 'ambassador/v1', 'auth_service': 'auth-web:443', 'kind': 'AuthService', 'name': 'auth-service', 'tls': True}
```

**To Reproduce**
```yaml
"apiVersion": "ambassador/v1"
"auth_service": "auth-web:443"
"kind": "AuthService"
"name": "auth-service"
"tls": true
```
**Expected behavior**
The AuthService annotation continues to support the `tls` flag with the same semantics as before.

 - Ambassador: 0.50.0
 - Kubernetes environment: On-prem (kubeadm)
 - Version: v1.11.5

kflynn:
Fixes #1202. Note that issue #1224 is still a factor -- you cannot mix HTTP and HTTPS upstreams for anything right now, including `AuthService`. This should be fixed soon.


kflynn:
(Also note that the substantive change here is adding `tls` back to `AuthService.schema` -- the code to process it was already present, the only problem was that the schema wouldn't allow setting it.)

PR #1250, opened 2019-02-24
Added missing s to cors headers config

carte7000:
Add missing s to cors headers comma-separated list config example.


kflynn:
Thanks!!

PR #1253, opened 2019-02-25
Added Lua script injection

lolletsoc:


kflynn:
So I really hate to say this, but... the old end-to-end tests are no more, and the brave new world uses KAT, and I'm very sorry for leaving the old stuff there to lead you astray! :frown:

This is probably a decent starting point for a KAT test for this:

```
class LuaTest(AmbassadorTest):

    target: ServiceType

    def init(self):
        self.target = HTTP()

    def config(self):
        # Use self here, not self.target, so that the Ambassador module is
        # be annotated on the Ambassador itself.
        yield self, self.format("""
---
apiVersion: ambassador/v1
kind: Module
name: ambassador
config:
  ...lua stuff here...
""")

        # Use self.target _here_, because we want the mapping to be annotated
        # on the service, not the Ambassador.
        yield self.target, self.format("""
---
apiVersion: ambassador/v1
kind:  Mapping
name:  lua-target-mapping
prefix: /target/
service: {self.target.path.fqdn}
""")

    def queries(self):
        yield Query(self.url("target/"))

    def check(self):
        for r in self.results:
            assert r.headers.get('whatever-lua-added', None) == value
```

Check out `t_extauth.py` and `test_ambassador.py` for how to link that into the world, and I'll be around to help with further questions -- thanks!!

kflynn:
(You can just delete everything in `ambassador/tests` from here to resolve the conflicts, then move forward adding the KAT test.)

kflynn:
Looks great! I've cloned your branch to let CI run over it, we'll see how it goes. :)

kflynn:
Thanks for doing this! and for being patient. :smile:

PR #1266, opened 2019-02-27
Upgrade to teleproxy 0.3.16, and properly handle namespaces so we can.

kflynn:
Ambassador was using Teleproxy 0.1.1, which is ancient. In particular, it didn't distinguish namespaces -- services with the same name in different namespaces would get conflated in Teleproxy's DNS, with unpredictable results. This is probably responsible for a nontrivial percentage of our test flakes.

It does, of course, require fully-qualifying several things, though.


rhs:
Looks good to me! Really glad to be on a more modern version of teleproxy. ;-)

PR #1270, opened 2019-03-01
Change "tcp" to "http" for L7 LB

rclayton-the-terrible:
I think the intent is to have the last example use the layer 7 LB.  Shouldn't that then read "http" and not "tcp"?


richarddli:
Good catch, thanks!

PR #1278, opened 2019-03-05
Added config changed notifications

Andrei-Predoiu:
This is a proposal to generate discussion on the notification issue 
https://github.com/datawire/ambassador/issues/1236

### DIFF
--- a/ambassador/ambassador/config/config.py
+++ b/ambassador/ambassador/config/config.py
@@ -49,6 +49,8 @@ class Config:
     # When using multiple Ambassadors in one cluster, use AMBASSADOR_ID to distinguish them.
     ambassador_id: ClassVar[str] = os.environ.get('AMBASSADOR_ID', 'default')
     ambassador_namespace: ClassVar[str] = os.environ.get('AMBASSADOR_NAMESPACE', 'default')
+    ambassador_pod_name: ClassVar[str] = os.environ.get('AMBASSADOR_POD_NAME', 'default')
+    ambassador_pod_uid: ClassVar[str] = os.environ.get('AMBASSADOR_POD_UID', 'default')
 
     # INSTANCE VARIABLES
     ambassador_nodename: str = "ambassador"     # overridden in Config.reset
--- a/ambassador/ambassador_diag/diagd.py
+++ b/ambassador/ambassador_diag/diagd.py
@@ -14,9 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License
 import copy
-import subprocess
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union, TYPE_CHECKING
-
 import datetime
 import functools
 import json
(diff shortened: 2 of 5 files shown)


Andrei-Predoiu:
I would like some feedback on my changes.
The snapshot counter resets every time the pod is restarted so i had to add a uuid to the event name.
The workflow to using this functionality would be:
1. Deploy ambassador-diag service as node-port or call the pod directly
2. Other services can watch k8s events. 
Right now i have hard-coded the label:
`event: ambassador-cfg`


TODO:
Change event label to include the ambassador-id in order to allow listening on events from the correct specific ambassador instances
Move notification logic outside diagd.py
Create some tests.

richarddli:
Desired workflow:

1. Dev commits something
2. Build agent deploys the code and that triggers (mapping) changes in ambassador
3. Ambassador checks config and updates the underlying envoy mappings
4. Ambassador publishes k8s event that the mappings have changed.
5. Custom app(CA) watches events from ambassador and is notified
6. CA connects to pod diag API and fetches all the mappings
7. CA tries to fetch Swagger specs for all the mappings from each of the services(based on whatever convention we decide on internally)
8. CA takes all the swagger specs and all the mappings and creates a single swagger spec for the whole gateway
9. CA acts as webhost for the combined spec and is mapped to `prefix: /`

richarddli:
@Andrei-Predoiu thanks for opening this, and discussing your workflow was very helpful. We discussed this and here are our thoughts on the best approach.

1. The CA seems well-scoped and an independent service that could be very valuable. Would love for you to OSS it if you're interested.
2. In terms of triggering the CA to do a Swagger rebuild, it seems that there are two approaches we can think of that would be quite a bit simpler than hooking into Ambassador. The first is to just have a CI system trigger a hook on your CA. If you have the CI infrastructure, this is a potentially attractive option as it would not require you to hack on Ambassador ;-). The second approach is to hook into `kubewatch` which Ambassador uses to get event notifications and use those event notifications to trigger your CA to run. 

What do you think?

Andrei-Predoiu:
@richarddli Thanks for engaging with this issue. 

1. I would love to OSS the CA at some point. In the end it is up to my employer, but i think they will agree :)
2.The second option sounds great. By having ambassador publish various state events allows other devs to interact in many useful ways with ambassador and build various other tools around it. 
That being said, kubewatch still needs to publish the config changed event at the end of the ambassador workflow. Just because kubewatch detected a change, It does not mean that those changes will translated 
into new ambassador configs being deployed.
 
The first option is not really possible and i think it might be brittle. Most of my organization is still maturing towards cloud and CI/CD. While in theory it might work, there is a chance that what goes through the CI does not reflect what's in the cluster. 
The other issue is that Ambassador itself does some checks and validations on the mappings and i would have to reproduce those in CI(maybe this is another opportunity for improvement). Same goes for service health information.

Andrei-Predoiu:
@richarddli 
So, should i continue by looking at kubewatch.go? I am thinking along the lines of making it able to run and publish and event written to a file or passed as an argument:
Fx: `kubewatch --publish-event /path/to/file.yml`
Same could be done to obtain the cluster id. If we go this route, it will turn into its own tool and i would recommend separating it from the telepresence project.

richarddli:
Yes, I think looking at `kubewatch` is a reasonable thing. Note that it's not particularly well documented at the moment. I believe it already does output events to a file.

If you hook into `kubewatch` though you would not get any Ambassador checks, though, since kubewatch runs before Ambassador generates the IR.

Andrei-Predoiu:
That's why i said `kubewatch` needs to be changed a bit so that it can be called to publish events. This way we could call it to publish events after ambassador is done.
How does that sound?

richarddli:
I'm not sure I follow -- kubewatch only connects Ambassador to Kubernetes. So it wouldn't be able to publish events after Ambassador is complete?

Andrei-Predoiu:
It sounds like we might be talking past each other a bit. 
The workflow i am suggesting right now looks like:
`kubernetes --- new config (event?)---> kubewatch`
`kubewatch ---new k8s config--->Ambassador(diagd)`
`Ambassador ---new config snap---> Envoy`
`Envoy ---ok--->Ambassador`
`Ambassador ---publish event request---> kubewatch` <-- new stuff
`kubewatch ---new event---> Kubernetes` <-- new stuff

kflynn:
Not stale, just requires changes to `watt` so we can preserve its role as the Only Thing that Talks to Kubernetes™.

Might do webhooks as well.

kflynn:
As mentioned on Slack, I'm going to go ahead and close this -- we like the concept but we need to modify `watt` to support the bidirectional path for events, and for the moment Andrei says that polling the diag service is sufficient.

Thanks! We'll stay in touch about this. :smile:

PR #1287, opened 2019-03-05
bugfix: new Envoy image that fixes the issue with multiple set-cookie headers

gsagula:
This PR fixes https://github.com/datawire/ambassador/issues/1211


gsagula:
I forgot to update the go Kat client. I will do that later.

PR #1290, opened 2019-03-06
TCPMappings! and test infrastructure improvements, round 1

kflynn:
`TCPMapping`s work now (including with SNI) and we're combining HTTP KAT backend services into superpods, so instead of starting 80-something of them, we start... 3.


kflynn:
@richarddli and @nbkrause, if you folks can look at the docs I've added, that'd rock.

docs/reference/tcpmappings.md
```diff
@@ -0,0 +1,111 @@
...
+| Attribute                 | Description               |
+| :------------------------ | :------------------------ |
+| [`weight`](/reference/canary)        | (optional) specifies the (integer) percentage of traffic for this resource that will be routed using this mapping |
+
+The name of the mapping must be unique.
+
+## Example `TCPMapping`s
+
+`TCPMapping` definitions are fairly straightforward. Here's an example that simply relays any connection on port 2222 onward to a Kubernetes service:
+
```
richarddli: Is this config also the right config necessary to proxy raw TLS?

docs/reference/tcpmappings.md
```diff
@@ -0,0 +1,111 @@
...
+---
+apiVersion: ambassador/v1
+kind: TCPMapping
+name: quote_mapping
+port: 2222
+tls: some-context
+service: qotm
+```
+
+Note that in this example, Ambassador will contact the `qotm` service using cleartext HTTP, not HTTPS. You can explicitly ask Ambassador to use HTTPS:
```
richarddli: in this example above (we should specify if it's the example above or below)

docs/reference/tcpmappings.md
kflynn: Yup -- if you expect raw TLS you'd probably want to use `service: qotm:443` though.

docs/reference/tcpmappings.md
kflynn: Will fix -- actually, I need to clean up the docs around ports too.

richarddli:
I added a couple comments on the docs, but the docs LGTM otherwise.

PR #1307, opened 2019-03-12
bug_fix: updated docker-base image and fixed test

gsagula:
Signed-off-by: Gabriel <gsagula@gmail.com>

## Description
Updated the test to a fail state. Ran `make docker-update-base` and the test passed. Also did a quick manual verification on the dirty image.

This PR fixes https://github.com/datawire/ambassador/issues/1211


gsagula:
@kflynn This PR should fix the issue with set-cookie.

PR #1328, opened 2019-03-19
fix typo

macsual:
## Description
This fixes a typo in the tense of a word.


richarddli:
Hi there, thanks for the PR. I believe you branched this pr from `master` and then are opening it on `stable`, which effectively means that we have to merge all the code commits from `master` onto `stable` in this PR, which I'm sure is not what you intended. Do you mind branching off of `stable` so we can get a clean PR? (We're working on some doc infrastructure improvements to make this much cleaner in the future, so please stay tuned.)

macsual:
I fixed it.

richarddli:
Thanks!

PR #1342, opened 2019-03-21
Sticky sessions

containscafeine:
- eliminates `load_balancer.type`
- implements cookie based sticky sessions 

To Do:
- add docs
- add other sticky sessions


kflynn:
Once again, `echo.websocket.org` failures. Time to pull that internal...

PR #1345, opened 2019-03-21
Warn people about the upcoming default-port change

kflynn:
In 0.60 (which will probably be the next release after 0.52) we'll change Ambassador's default ports from 80/443 to 8080/8443 to facilitate running as non-root. The diag service can now warn people about that.

This PR also includes some changes to make it simpler to just run diagd locally for testing.


kflynn:
Failures are due to `echo.websocket.org` having issues. Merging.

Issue #988, opened 2018-11-21
Support for HTTP/1.0 Healthchecks by Loadbalancers


daimoniac:
**Please describe your use case / problem.**
Some loadbalancer vendors are using http/1.0 healthchecks (haproxy default, digitalocean exclusively)
When deploying ambassador behind such a loadbalancer, Ambassador/envoy is replying with 426 UPGRADE REQUIRED answers:
```
ACCESS [2018-11-21T10:12:10.359Z] "GET /ambassador/v0/check_alive HTTP/1.0" 426 - 0 0 0 - "-" "curl/7.58.0" "-" "localhost:8080" "-"
ACCESS [2018-11-21T10:12:18.139Z] "GET /ambassador/v0/check_ready HTTP/1.0" 426 - 0 0 0 - "-" "curl/7.58.0" "-" "localhost:8080" "-"
```
Therefore, the upstream Loadbalancers are marking the service as DOWN (they expect a 200 OK answer).

**Describe the solution you'd like**
I think we should be able to answer simple http/1.0 calls. 
envoy does support it *accept_http_10*
https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/protocol.proto

**Describe alternatives you've considered**
I have now resolved it by manually setting the upstream loadbalancer to TCP only check.
So the Loadbalancer only checks if the port ist open. Thats not optimal, but better than no service

rljohnsn:
Seeing the same thing.

```
"GET /public/ping HTTP/1.0" 426 - 0 0 0 - "-" "NewRelicPinger/1.0 (1144731)"
```

phenoxim:
Same issue with me. Could anyone give a hint to resolve this propblem?

mstrother:
I'm also looking for HTTP 1.0 support.

phenoxim:
> Same issue with me. Could anyone give a hint to resolve this propblem?

FYI

Update http_config here:
https://github.com/datawire/ambassador/blob/7ccc4945f7b50af35c77ffddca2cc7f22979882a/ambassador/ambassador/envoy/v2/v2listener.py#L353

```
http_config['http_protocol_options'] = {
        'accept_http_10': True
}
```

and build a private ambassador image.

Solved this issue.

ryanmaclean:
@phenoxim - thanks for the patch! This helps me quite a bit. I've forked the repo, and am currently testing, but will submit a PR once done.

richarddli:
[...]

PR #1352, opened 2019-03-21
Allow HTTP/1.0

ryanmaclean:
This is required for cloud-based load balancer health checks that use HTTP 1.0.

## Description
Accept HTTP 1.0 traffic

## Related Issues
This relates to #988 - all credit to @phenoxim !

## Testing
Tests passed aside from 4 warnings which were present in the master branch as well. 

## Todos
- [ ] Tests
- [x] Documentation (happy to help here, just point me in the right direction!)

## Other
* If this is a documentation change, please open the pull request against the `stable` branch.
* Code changes should occur against `stable`.


ryanmaclean:
Looks like the build failed with: 

```
openssl aes-256-cbc \
    -K $encrypted_21b00a486c76_key \
    -iv $encrypted_21b00a486c76_iv \
    -in releng/kubernaut_v2_config.enc \
    -out ~/.config/kubernaut/config \
    -d

iv undefined
```

richarddli:
Thanks! That's a known problem with external contribs (we're working on fixing that -- it's because we don't expose secrets outside of the org).

Regarding docs, I think this should go here:

https://github.com/datawire/ambassador/blob/master/docs/reference/core/ambassador.md

(Is it in the Ambassador module?)

We list the different options in alphabetical order.

ryanmaclean:
@richarddli - will have a look at the docs and see if anything's needed!

ryanmaclean:
Ah interesting - I hadn't stumbled across this file yet! @richarddli would it be better to implement this as enabled by default, and handle it using a conf change? I was being optimistic in that I don't see harm in enabling it by default, but now see that may not be the desired approach.

richarddli:
Generally speaking we disable new things like this by default because we have tens of thousands of folks running in prod now, so we try to be fairly conservative on changing defaults (and we announce in advance breaking changes). E.g., the grpc_web support is false by default.

ryanmaclean:
@richarddli that's exactly what I was looking at the, the grpc section. I'll see if I can find where that is called in order to use the same pattern. Would that work?

richarddli:
Yes, that'd be great! Thanks!

cyrus-mc:
@ryanmaclean PR-1484 implements this functionality but does it via a configuration option on the Ambassador module.

ryanmaclean:
Thanks @cyrus-mc !

Issue #1127, opened 2019-01-22
Support Envoy retry policies


manigandham:
[Automatic retries](https://www.envoyproxy.io/learn/automatic-retries) are a core feature of Envoy but the latest v0.50x doesn't support `envoy_override` anymore and there are no appropriate settings in the Mapping config.

This should be part of the Mapping API. It can support a basic subset until `envoy_override` is supported again or match the Envoy API.

caiobegotti:
As a side comment, what is the expectation about supporting something like envoy_override again after 0.50.x? I can't upgrade it myself due to that, that's how I landed on this ticket here :-(

richarddli:
The v2 YAML configuration that Envoy uses is quite a bit more complex than the v1 configuration we used to support.

So, it's on the list, but it will be quite different. In the meantime, though, the Ambassador code is much cleaner, so would welcome direct contributions to the code base to support more features! :)

richarddli:
Merged to master, will be released shortly

PR #1358, opened 2019-03-22
Envoy Retry Policy Support

LiVe:
## Description
Adds support für envoys retry_policy to ambassador module and http mappings.

## Related Issues
#1127 

## Testing
- manual tests
- deployed to development environment
- started adding automated tests, i'd assume this needs more work

## Todos
- [ X ] Tests (check what needs to be done)
- [ X ] Documentation


kflynn:
Hey, thanks for this! I'm afraid I've caused some trouble for you by dropping the old Envoy-V1 stuff and the old end-to-end tests -- can you take care of the merge conflicts (probably be deleting things) and we can talk further about tests? Are you on [our Slack](https://d6e.co/slack)?

richarddli:
@containscafeine does this touch the work you're looking at right now?

containscafeine:
@richarddli not directly, but this will work well in conjunction to circuit breakers :tada:

LiVe:
@kflynn Hey, sorry for the late reply, busy days. i've merged the latest changes and removed the old / unnecessary files. i've also looked at the test descriptions and moved the old end to end test to a mapping test.

and yes, i'm on the datawire slack too if you want to continue the discussion over there :)

ambassador/ambassador/ir/irambassador.py
```diff
@@ -236,6 +237,14 @@ def setup(self, ir: 'IR', aconf: Config) -> bool:
             else:
                 return False
 
+        if amod and ('retry_policy' in amod):
+            self.retry_policy = IRRETRYPOLICY(ir=ir, aconf=aconf, location=self.location, **amod.retry_policy)
+
```
containscafeine: Will be helpful if we can validate the retry policy here. Look at `validate_load_balancer` for reference :+1:

ambassador/ambassador/ir/irretrypolicy.py
```diff
@@ -0,0 +1,28 @@
...
+from ..config import Config
+from ..utils import RichStatus
+
+from .irresource import IRResource
+
+if TYPE_CHECKING:
+    from .ir import IR
+
+
+class IRRETRYPOLICY (IRResource):
```
containscafeine: nit: change this to `IRRetryPolicy`
I believe you must for gotten confused with `IRCORS` being all caps, but that's because of CORS being in caps. The rest of the classes are camel case ;)

ambassador/tests/t_retrypolicy.py
```diff
@@ -0,0 +1,75 @@
...
+  strategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: retry
+    spec:
+      containers:
+      - name: retry
+        image: live/ambassador-retry-service:1.0.0
```
containscafeine: Do you mind committing the Dockerfile for this image with this PR as well? :)

ambassador/tests/t_retrypolicy.py
```diff
@@ -0,0 +1,75 @@
...
+rewrite: /retry/
+service: https://{self.target.path.fqdn}
+retry_policy:
+  retry_on: "5xx"
+  num_retries: 3
+  per_try_timeout: "500ms"
+""")
+
+    def queries(self):
+        yield Query(self.parent.url("%s/retry/" % self.name), expected=200)
```
containscafeine: Can you explain a bit more on how does this test the retry policy behavior? How do we know that it's retrying on `5xx`?

ambassador/ambassador/ir/irambassador.py
LiVe: i've added this to the IRRetryPolicy class and validated retry_on in setup. would that be a correct approach to this?

ambassador/ambassador/ir/irretrypolicy.py
LiVe: ah yes, i guess the ircors got me there you're right. i've changed this.

ambassador/tests/t_retrypolicy.py
LiVe: of course. initially i've added the code and dockerfile to the now unused end-to-end tests (but thats now excluded via gitignore). where should i put those files?

ambassador/tests/t_retrypolicy.py
LiVe: i'm not sure if this test fits any requirements or best practices at all, but i'd need a bit help / guidance on implementing a solid testing logic for this one (i'm lacking a bit of experience here).

when i've added the changes, all i wanted todo was to check if it retries failed requests. so i just wrote a simple service that flips between a 503 and 200 response (first one fails, second one succeeds, and so on). with the retry policy in place, a query should always succeed on this service, therefor expecting the 200. With multiple (parallel) tests the flipping responses, as they are now, would probably cause issues though (maybe running multiple queries would be enough?). but i guess that there is a better solution to this?

containscafeine:
@LiVe I've made some testing and docs changes at #1486, besides your commits. Do you want to take a look there? ;)

flands:
Can this be set once globally instead of per mapping?

containscafeine:
@flands yes!

PR #1367, opened 2019-03-24
Update dependencies for security vulns

richarddli:
This updates some dependencies on Kat services. Technically, this isn't that important as these services are only used in tests, but it's not a hard fix.

### DIFF
--- a/end-to-end/auth-service/requirements.txt
+++ b/end-to-end/auth-service/requirements.txt
@@ -2,6 +2,6 @@ clize==4.0.1
 flask==0.12.3
 jinja2==2.9.6
 jsonschema==2.6.0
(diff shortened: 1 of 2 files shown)


end-to-end/auth-service/requirements.txt
```diff
@@ -2,6 +2,6 @@ clize==4.0.1
 flask==0.12.3
 jinja2==2.9.6
 jsonschema==2.6.0
-pyyaml==3.12
+pyyaml==4.2b1
```
kflynn: May as well go all the way to 5.1.

kflynn:
Dropping this one, I'll replace it shortly.

Issue #1344, opened 2019-03-21
Zipkin trace headers contain multiple spans


coryflucas:
**Describe the bug**
When enabling the tracing services in versions > 0.50 the Zipkin headers `X-B3-Sampled`, `X-B3-Spanid`, and `X-B3-Traceid` all appear to have multiple values concatenated into them.

Example responses from httpbin container:

On 0.40.2:
```
$ curl http://localhost:8080/httpbin/get
{
  "args": {},
  "headers": {
    "Accept": "*/*",
    "Content-Length": "0",
    "Host": "localhost:8080",
    "User-Agent": "curl/7.54.0",
    "X-B3-Parentspanid": "5efc216a052d8f53",
    "X-B3-Sampled": "1",
    "X-B3-Spanid": "022c8f7c67c05af4",
    "X-B3-Traceid": "5efc216a052d8f53",
    "X-Envoy-Expected-Rq-Timeout-Ms": "3000",
    "X-Envoy-Original-Path": "/httpbin/get"
  },
  "origin": "172.21.0.3",
  "url": "http://localhost:8080/get"
}
```

On 0.51.2:
```
$ curl http://localhost:8080/httpbin/get
{
  "args": {},
  "headers": {
    "Accept": "*/*",
    "Content-Length": "0",
    "Host": "localhost:8080",
    "User-Agent": "curl/7.54.0",
    "X-B3-Parentspanid": "31708df6d4f1b3a8",
    "X-B3-Sampled": "1,1",
    "X-B3-Spanid": "31708df6d4f1b3a8,ba3032afcb429cc9",
    "X-B3-Traceid": "31708df6d4f1b3a8,31708df6d4f1b3a8",
    "X-Envoy-Expected-Rq-Timeout-Ms": "3000",
    "X-Envoy-Internal": "true",
    "X-Envoy-Original-Path": "/httpbin/get"
  },
  "origin": "172.21.0.1",
  "url": "http://localhost:8080/get"
}
```

Trace JSON from zipkin from 0.51.2:
```json
[
  {
    "traceId": "3d2e5cd9eadf6e66",
    "parentId": "3d2e5cd9eadf6e66",
    "id": "71f4e76b18114cbc",
    "kind": "CLIENT",
    "name": "router cluster_httpbin egress",
    "timestamp": 1553167127739170,
    "duration": 7395,
    "localEndpoint": {
      "serviceName": "ambassador-default",
      "ipv4": "172.21.0.4"
    },
    "tags": {
[...]

PR #1378, opened 2019-03-27
Add the ability to skip child span creation

esmet:
## Description
Allows an operator to disable child span creation for traces, which in practice prevents extra X-B3-* trace header values from propagating upstream.

## Related Issues
https://github.com/datawire/ambassador/issues/1344

## Testing
Manual testing. We use this patch for our internal use case and it works well.

## Todos
- [ ] Tests
- [ ] Documentation

## Other

### DIFF
--- a/ambassador/ambassador/envoy/v2/v2listener.py
+++ b/ambassador/ambassador/envoy/v2/v2listener.py
@@ -16,6 +16,7 @@ from typing import Any, Dict, List, Optional, Tuple, TYPE_CHECKING
 from typing import cast as typecast
 
 import json
+import os
 
 # from copy import deepcopy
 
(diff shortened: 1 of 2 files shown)


esmet:
Good call, this should be on the ambassador module config.

By not starting a child span, Envoy traces still appear, but they're sibling to the upstream request.

For example, with a gRPC AuthService and RateLimitService, my traces show one parent span created by Ambassador. Under that span, there is an extauth egress span, associated spans from my extauth service, a ratelimit service egress span, and then an upstream span plus everything associated with that.

Without this change, the X-B3-* headers are duplicated in the upstream request, which happens to confuse applications enough that they don't even produce traces (specifically, because I have some applications that do not understand "X-B3-Flags: 1,1")

coryflucas:
I'm bit sure this is the correct solution.  The child span was present prior to the 0.50 release, but the issue with the B3 headers was not.  I also don't think this would resolve the issue for if ambassador is passed B3 headers from an already created span (see my comment here: https://github.com/datawire/ambassador/issues/1344#issuecomment-476209298).

esmet:
@coryflucas I tried this on my deployment and indeed even with start_child_span=False, the issue you linked (1344) still isn't fixed. So, I think the real fix has to be deeper - possibly in Envoy.

esmet:
Update: This diff works around what I think is a bug in Envoy (fixed on Envoy master) so it's not clear how much long term value it has.

kflynn:
Closing this one, since #1344 is fixed as of 0.53.1. @esmet, I'd definitely consider a PR to configure this directly in the Ambassador config...

PR #1389, opened 2019-03-31
git-subtree hacks

LukeShu:
## TL;DR: Changes

 - Add some fancy hand-constructed merge commits made to look like `git subtree split --rejoin` created them, in order to get `git merge -Xsubtree=docs` algorithm (used by `make pull-docs`) to do the right thing.
 - `make push-docs`: On Thursday, I told @kflynn to run some commands manually instead of running `make push-docs`; update it to run those commands.
 - `make push-docs`: Pass `--rejoin` to `git subtree split`.
 - `make pull-docs`: Immediately do a `git subtree split` after the `git subtree merge`.
 - `make push-docs`: Enable pushing to a branch other than `master` by setting `PUSH_BRANCH=branchname`.

## Description

On Thursday (2019-03-28), @kflynn said that he couldn't `make push-docs`.  I told him to hack around it by running this instead:

```bash
git remote add ambassador-docs https://github.com/datawire/ambassador-docs
git fetch --all
git push git@github.com:datawire/ambassador-docs.git "$(git subtree split --onto=ambassador-docs/master --prefix=docs):master"
```

For comparison, `make push-docs` would run:

```bash
git subtree push --prefix=docs git@github.com:datawire/ambassador-docs.git master
```

which should be equivalent to

```bash
git push git@github.com:datawire/ambassador-docs.git "$(git subtree split --prefix=docs):master"
```

So, all that changed is adding `--onto=ambassador-docs/master`.

One of the core assumptions of `git subtree` is that `git subtree split` is deterministic; given the same input-history, it will produce the same output-history.  Being able to `git subtree push` relies on this; the newly-generated history must agree with the existing remote history.

However, because `docs/docs/` exists (the subtree contains a directory that has the same name as the subtree itself), when `git subtree split` encounters an already-split-off subtree-commit from a previous run, it mistakenly thinks it's an input-commit.  Even though it was given the same input-history, it behaves as if it is given different input-history, because it considers commits that it shouldn't to be part of that history.  This can be mitigated by signaling to `git subtree split` which existing commits are already-split subtree-commits, either with `git-subtree-split:` metadata in commit messages, or with the `--onto=` flag.

Because `ambassador.git` mostly does not include `git-subtree-split:` metadata, in order for things to work correctly, we use `--onto=` to signal this.  When we pass in a commit with `--onto`, it and *all* of its ancestors will be marked as already-split subtree-commits.

The reason that `ambassador.git` mostly does not include `git-subtree-split:` metadata is twofold:

 1. `make push-docs` used `git subtree push` for convenience, rather than calling `git subtree split` and `git push` separately.  Although `git subtree psuh` is a thin wrapper around `git subtree split`, it does not allow us to pass flags through to `git subtree split`, so we couldn't pass in the `--rejoin` flag, which would add that metadata.
 2. `git subtree merge` (used by `make pull-docs`) doesn't add the metadata, even though it totally could.

Going forward, we would like that metadata to be there.  So, start adding it by:

 1. Have `make push-docs` call `git subtree split` and `git push` separately, so that we can pass `--rejoin` to `git subtree split`.
 2. Have `make pull-docs` immediately do a `git subtree split --rejoin` after the `git subtree merge` to add the metadata that the merge "should have" added.

While I was trying to figure out what the hell was going on, I realized that there were a few bugs in `git subtree split` that - `git subtree split` has some bugs where for specific history-graph shapes it it gets confused about whether a commit is part of the already-split-off subtree, or is a commit that predates the subtree (backwards of the confusion described above).  We happen to have several of these shapes, so `ambassador-docs.git`'s history is really weird.  On Slack I stated that I though this was a source of the problem with `make pull-docs` and `make push-docs` not doing the right thing.  However, on further reflection, I don't think that's true.  I don't think this funny history really affects anything, and that the wrong behavior is entirely explained by the above issue.  But maybe it does... at this point there are enough independent bugs in play that I'm having a hard time reasoning about things.

Finally, while I'm at it, allow `make push-docs` to push to a branch other than `master` by setting `PUSH_BRANCH=branchname`.


LukeShu:
Note that this PR includes commits created by Flynn's run of `git subtree split`.  There are only 5 new commits in this PR that aren't already in `ambassador-docs.git`.

kflynn:
So... could we make this simpler by changing the name of `docs/docs`?

kflynn:
(I'm not necessarily proposing that we do the rename at this point, just asking the question.)

richarddli:
If we're going to change it, I'd prefer to change it to something like `documentation/docs` (i.e., not change the subdirectory, since it shows up in the URL. the top-level subdir doesn't matter.)

LukeShu:
> So... could we make this simpler by changing the name of `docs/docs`?

In the long-term, maybe? If we did the change today, it wouldn't make any difference for a while, because the git history would still have it at the confusing location.  The future where it actually does simplify things:
 - `git subtree split` has been patched to correctly identify historical already-split commits as already-split, based on `git-subtree-split` metadata (currently, it doesn't trickle down the tree, because it incorrectly assumes that parts of the tree are unreachable)
 - `ambassador.git` has enough commits with the `git-subtree-split` metadata that there's no way to reach one of those confusing historical commits without going through one of the commits with metadata.

And then, all it would simplify is that we no longer need to pass the `--onto` flag.

PR #1391, opened 2019-04-02
Update podAntiAffinity settings

neumanndaniel:
It is better to have the setting requiredDuringSchedulingIgnoredDuringExecution for the podAntiAffinity to ensure distribution of the Ambassador pods across different nodes, when also the externalTrafficPolicy is set to local.

If not intended to force pod distribution across nodes, then just close the PR and keep the configuration settings as they are.

### DIFF
--- a/docs/yaml/ambassador/ambassador-rbac-prometheus.yaml
+++ b/docs/yaml/ambassador/ambassador-rbac-prometheus.yaml
@@ -81,6 +81,17 @@ spec:
                 values:
                 - ambassador
             topologyKey: kubernetes.io/hostname
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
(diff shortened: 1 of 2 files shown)


richarddli:
By default, we set to 3 replicas. So if you're using a 1 node or 2 node setup (which is very common during evaluation), will the `requiredDuringSchedulingIgnoredDuringExecution` annotation work?

neumanndaniel:
Nope. It only works with at least 3 nodes. Otherwise you would have 1 or 2 pods in pending mode until you spin up a new node. 

As I said:

> If not intended to force pod distribution across nodes, then just close the PR and keep the configuration settings as they are.

So, you can close the PR cause I unterstand that the majority would run into issues when trying out Ambassador with the proposed config change above.

richarddli:
OK, thanks for the suggestion!

PR #1392, opened 2019-04-02
Add support for maglev LB besides ring_hash

containscafeine:


docs/reference/core/load-balancer.md
```diff
@@ -42,10 +42,13 @@ Configuring sticky sessions makes Ambassador route requests to the same backend
...
 - Source IP
 
+**NOTE:** Ambassador supports sticky sessions using 2 load balancing policies, `ring_hash` and `maglev`.
+
+
 ##### Cookie
 ```yaml
 load_balancer:
-  policy: ring_hash
+  policy: ring_hash/maglev
```
kflynn: This will probably have some folks thinking that the literal string `ring_hash/maglev` is what you should type. Maybe just pick one, and note separately that the other is supported too?

containscafeine: Makes sense, fixed :)

PR #1399, opened 2019-04-04
`case_sensitive` fix when `host_redirect` is true

reikani:
## Description
The `case_sensitive` field is ignored when `host_redirect` is set to true because no flattening occurs in `ambassador/ambassador/ir/irhttpmappinggroup.py`. To solve that, introduce another flattening step when `host_redirect` is there.

## Related Issues
https://github.com/datawire/ambassador/issues/699

## Testing
Simple manual tests-more to follow.

## Todos
- [ ] Tests
- [ ] Documentation

## Other
Work in progress. I would like to get that list coded out of there, something similar to `mapping.skip_key(k)` on L217.

### DIFF
--- a/ambassador/ambassador/ir/irhttpmappinggroup.py
+++ b/ambassador/ambassador/ir/irhttpmappinggroup.py
@@ -326,16 +326,8 @@ class IRHTTPMappingGroup (IRBaseMappingGroup):
             return list([ mapping.cluster for mapping in self.mappings ])
         else:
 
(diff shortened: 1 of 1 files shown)


reikani:
Changed it so it only flattens the `case_sensitive` field to minimize risk.

entone:
@kflynn wondering if anyone has had a chance to review this PR. It looks like maybe the Travis settings are causing it to fail CI.

PR #1404, opened 2019-04-06
clarify build process

richarddli:
Documentation to clarify the build/test process for Ambassador.


richarddli:
@containscafeine would appreciate a review, and if there's anything else specific for minikube that should be documented?

PR #1406, opened 2019-04-08
Update TLS documentation to call out HTTP2 configuration

cyrus-mc:
## Description

Add section to TLS documentation highlighting configuration required for HTTP2 negotiation.


docs/reference/core/tls.md
```diff
@@ -88,6 +88,23 @@ config:
...
+```yaml
+apiVersion: ambassador/v1
+kind:  Module
+name:  tls
+config:
+  server:
+    alpn_protocols: h2[, http/1.1]
+```
+
+Without setting setting alpn_protocols as shown above, HTTP2 will not be available via negotiation and will have to be explicitly requested by the client.
```
flands: Duplicate "setting"

PR #1393, opened 2019-04-02
Adding cluster_timeout_seconds as a configurable variable

peturgq:
## Description
I ran into an issue when trying to configure envoy to allow long-lived connections through ambassador and found out that the connection_timeout for the cluster is hard-coded to 3s, so I made a feature to allow configuration of it.
## Testing
A few sentences describing what testing you've done, e.g., manual tests, automated tests, deployed in production, etc.
Tested manually and verified that all was working and later deployed to production.
## Todos
- [automated tests] Tests
- [add to docs] Documentation

PR #1418, opened 2019-04-10
Adding documentation for cluster timeouts

peturgq:
## Description
Documentation for the cluster timeout feature.


richarddli:
Thanks! The term `cluster` isn't something we generally expose to Ambassador users; I'm thinking that it makes sense to create a `timeout.md` document (like we do for `rewrite`, `cors`, etc.) that talks about this in more detail. Is this something you could do?

peturgq:
I'll maybe take a look at it this evening/tomorrow.

richarddli:
Yes, OK, we need to do this especially now with #1426 too

peturgq:
Timeout PR:
https://github.com/emissary-ingress/emissary/pull/1393

PR #1425, opened 2019-04-11
Add support for idle_timeout_ms

AaronTriplett:
## Description
Adds support for Envoy's `idle_timeout` configuration with an `idle_timeout_ms` mapping attribute. When set, this value overrides the Envoy connection manager wide `stream_idle_timeout` for said route. 

`idle_timeout` is used to bound the amount of time a request’s stream may be idle. This is especially important for long-lived connections (ex: gRPC) that currently get terminated after 300 seconds of inactivity due to the default value of `stream_idle_timeout` in envoy.

Today, only values greater than 0 are supported (on the current envoy version) due to a bug in Envoy's config validation logic. In the future, setting this value to 0 should completely disable the route’s idle timeout, even if a connection manager stream idle timeout is configured. This has been fixed here: https://github.com/envoyproxy/envoy/pull/6500 and should automatically work once released and the ambassador-envoy-alpine image is updated.

Example mapping for 1 hour idle timeout:
```
---
      apiVersion: ambassador/v1
      kind: Mapping
      name: grpc_test_mapping
      grpc: True
      prefix: /cf.Echo/
      rewrite: /cf.Echo/
      timeout_ms: 0
      idle_timeout_ms: 3600000
      service: cfservice.intlgntsys-cui-conversation-framework-usw2-ppd-dev:9090
```

## Testing
- Open long-lived gRPC Stream from client to service through Ambassador/Envoy. 
- Let connection idle (no active RPC) for 5 minutes
- Observe connection termination promptly at 300 seconds / 5 minute mark (Envoy `stream_idle_timeout` default).
- Set `idle_timeout_ms` attribute on the gRPC service's Ambassador Mapping to 3600000 (1 hour)
- Open long-lived gRPC Stream from client to service through Ambassador/Envoy. 
- Let connection idle (no active RPC) for 60 minutes
- Observe connection termination promptly at 1 hour mark.

## Todos
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/ambassador/ambassador/envoy/v2/v2route.py
+++ b/ambassador/ambassador/envoy/v2/v2route.py
@@ -109,6 +109,11 @@ class V2Route(dict):
                 }
             }
 
+            idle_timeout_ms = group.get('idle_timeout_ms', None)
+
+            if idle_timeout_ms is not None:
+                route['idle_timeout'] = "%0.3fs" % (idle_timeout_ms / 1000.0)
+
             if group.get('rewrite', None):
                 route['prefix_rewrite'] = group['rewrite']
 
--- a/ambassador/ambassador/ir/irhttpmapping.py
+++ b/ambassador/ambassador/ir/irhttpmapping.py
@@ -83,6 +83,7 @@ class IRHTTPMapping (IRBaseMapping):
         "shadow": True,
         "cluster_timeout_ms": True,
         "timeout_ms": True,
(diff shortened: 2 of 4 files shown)


kflynn:
Looks good! Normally we'd need automated tests, too, but the test suite needs some work to handle timeouts right now, so I'm going to go ahead and land this one. Thanks!

PR #1429, opened 2019-04-15
Allow DNS resolvable names for endpoint routing

containscafeine:
Before this commit, if an Ambassador Mapping has a service set as
`qotm.default.svc.cluster.local`, then it is not resolved to
endpoints. This is an issue in cases where long resolvable
hostnames are used which span namespaces instead of short namespace
specific names.

This commit tries to resolve such a name and then does a reverse
lookup in Kubernetes services for the resolved IP to allow for
endpoint routing.

### DIFF
--- a/ambassador/ambassador/config/resourcefetcher.py
+++ b/ambassador/ambassador/config/resourcefetcher.py
@@ -308,6 +308,7 @@ class ServiceKind(ObjectKind):
                 service_info['ports'] = []
                 for port in ports:
                     service_info['ports'].append(port)
+        service_info['clusterIP'] = spec.get('clusterIP', None)
 
         metadata = self.object.get('metadata', None)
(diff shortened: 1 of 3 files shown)


flands:
@containscafeine what symptoms or problems is this PR attempting to address? What is the benefit of endpoint versus service? I ask because we have seen DNS resolution issues resulting in 503s and this sounds relevant. Can also discuss on Slack if easier.

richarddli:
@flands I discussed endpoint vs service in gory detail in the blog: https://blog.getambassador.io/session-affinity-load-balancing-controls-grpc-web-and-ambassador-0-52-2b916b396d0c. This bug is unlikely to be related to your 503s.

kflynn:
Closing this one -- it's superseded by all the Resolver work in 0.60.

PR #1471, opened 2019-04-26
Fix documentation example with TLS context

jonathanbeber:
Fix documentation example related to ambassador integration with istio when using mutual TLS. The name created to the example doesn't match with the name used to the mapping.


richarddli:
Thanks! Since this is a documentation change, do you mind closing this and reopening it against https://github.com/datawire/ambassador-docs so it gets published immediately? (We use `git subtree` to vendor in the docs into this repository for various technical reasons.)

jonathanbeber:
Sorry, @richarddli 

I was reading the doc and opened it on the github's interface without paying enough attention. I'll reopen it.

PR #1492, opened 2019-05-06
Do not deadlock on check_scout after observing an invalid envoy configuration

esmet:
## Description
Fixes an `app.check_scout` call to `self.check_scout` to avoid deadlocks on invalid envoy configuration.

## Related Issues
https://github.com/datawire/ambassador/issues/1491

## Testing
Manually tested by generating an invalid envoy configuration and observing that scout code gets called, instead of stopping just before the deadlock line of code.


kflynn:
Can't believe I missed this. 😦  Thanks!

PR #1494, opened 2019-05-06
Round each mapping's weight to an integer. 

esmet:
Spread out any remainder to achieve a weight sum of 100.

## Description
Round weights to integers to satisfy Envoy's requirement.

## Related Issues
https://github.com/datawire/ambassador/issues/1289

## Testing
Manually tested that two Mappings with weight 80 and weight 25 generate normalized envoy weights that are integers. 

## Todos
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/ambassador/ambassador/ir/irhttpmappinggroup.py
+++ b/ambassador/ambassador/ir/irhttpmappinggroup.py
@@ -323,14 +323,29 @@ class IRHTTPMappingGroup (IRBaseMappingGroup):
                 else:
                     total_weight += mapping.weight
 
-            # OK, once that's done normalize all the weights.
+            # OK, once that's done normalize all the weights. Be sure to round each
+            # weight into an integer, since the Envoy spec expects that. We may not
+            # sum to 100 exactly by doing this, so we'll need to fix that up later.
             if unspecified_mappings:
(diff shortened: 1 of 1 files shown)


ambassador/ambassador/ir/irhttpmappinggroup.py
```diff
@@ -323,14 +323,31 @@ def finalize(self, ir: 'IR', aconf: Config) -> List[IRCluster]:
...
                 for mapping in self.mappings:
-                    mapping.weight *= 100.0/total_weight
+                    mapping.weight = round(mapping.weight * (100.0/total_weight))
+
+            final_total = 0
+            for mapping in self.mappings:
+                final_total += mapping.weight
+
+            # Spread out the remainder evenly over the mappings. Technically, if
+            # rounding is perfect, this is never necessary. But to protect against
```
esmet: Now that I think about it, the very simple case of 33.3 + 33.3 + 33.3 would make this fixup actually necessary.

I'll fix the comment and test this by hand.

esmet: Cleaned this up and verified that this case works.

PR #1506, opened 2019-05-07
Test for weighted shadowing

kflynn:
Admittedly not a _great_ test, but we're not really testing how well Envoy can do random numbers for weighted things...


kflynn:
Merging in spite of a flake. Might need to tweak the limit on the weighted shadow test further.

PR #1507, opened 2019-05-07
Updated documentation for retries and circuit breakers

richarddli:
- Clarify how to use circuit breakers & retries (use cases, etc.)
- Separate docs on retries & circuit breakers (they're different features; we don't combine these)

Would like to verify:

1. That you can only put one of the various error codes into retries (as opposed to a list)
2. How do circuit breakers reset? As soon as you go below the trigger thresholds?


containscafeine:
> That you can only put one of the various error codes into retries (as opposed to a list)

I believe you're asking about `retriable_status_codes` field (https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto#envoy-api-field-route-retrypolicy-retriable-status-codes). We do not support it currently.

> How do circuit breakers reset? As soon as you go below the trigger thresholds?

Yep, I don't think there's an explicit way of resetting a broken circuit. It's triggered only when thresholds are reached.

docs/reference/circuit-breakers.md
```diff
@@ -0,0 +1,91 @@
...
+apiVersion: ambassador/v1
+kind:  Mapping
+name:  qotm_mapping
+prefix: /qotm/
+service: qotm
+```
+
+## Circuit breakers and automatic retries
+
+Circuit breakers are designed to be used in conjunction with [automatic retries](reference/retries). Here are some examples:
```
containscafeine:
```suggestion
Circuit breakers are best used in conjunction with [automatic retries](reference/retries). Here are some examples:
```

containscafeine: Basically, I want to indicate that it's a best practice instead of that circuit breakers are incomplete without retries.

PR #1531, opened 2019-05-13
Support reading idle_timeout_ms on the Ambassador module

esmet:
## Description
Allow an operator to set a default idle_timeout_ms on the Ambassador module that applies to all Mappings, except for those with an explicit override.

## Related Issues
-
## Testing
Tested manually, works for my use case so far.

## Todos
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/ambassador/ambassador/envoy/v2/v2route.py
+++ b/ambassador/ambassador/envoy/v2/v2route.py
@@ -91,10 +91,10 @@ class V2Route(dict):
                 }
             }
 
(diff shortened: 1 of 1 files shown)


esmet:
I'll try this and get back to you.

esmet:
This feature is not actually the one I want to PR. I'll submit another PR for `cluster_idle_timeout_ms` which is the timeout for upstream keepalives.

PR #1534, opened 2019-05-13
Bump ENVOY_COMMIT to include v1.10.0

### LINKED ISSUE #1522: Update to Envoy 1.10
We're pretty much already there, actually, but let's get to 1.10 and see what the process is like.

LukeShu:
## Description

I've rebased the Envoy commits on to v1.10.0, and updated the docker images here.  There should be no functional changes.

## Related Issues
Closes #1522 

## Other

I've also tagged the Envoy commit, so it won't go away with future rebases:

```bash
git tag "datawire-$(git describe --tags)"
git push --tags
```


LukeShu:
Oh, the Envoy branch name I'm using is `rebase/master`.

PR #1551, opened 2019-05-17
Tls version tests

gsagula:
## Description
This PR adds KAT tests and docs for min/max TLS feature.

## Related Issues
https://github.com/datawire/ambassador/issues/1521

## Testing
Error when downstream asks for a TLS version that's too low
Error when downstream asks for a TLS version that's too high
Check that Ambassador selects the correct protocol version based on the upstream connection.

## Todos
- [ x ] Tests
- [ x ] Documentation


gsagula:
Not sure why retry test is failing since I only changed the documentation. 
https://travis-ci.org/datawire/ambassador/builds/533886558#L1678

Rerunning it.

PR #1589, opened 2019-05-31
Fix with_request_body

kflynn:
Translate from Ambassador syntax to Envoy syntax correctly, and include the feature in regression tests!

### DIFF
--- a/ambassador/ambassador/envoy/v2/v2listener.py
+++ b/ambassador/ambassador/envoy/v2/v2listener.py
@@ -212,13 +212,24 @@ def v2filter_authv1(auth: IRAuth, v2config: 'V2Config'):
 
     assert auth.proto
 
-    body_info = auth.get('include_body')
+    raw_body_info: Optional[Dict[str, int]] = auth.get('include_body')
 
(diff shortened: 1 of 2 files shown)


ambassador/ambassador/envoy/v2/v2listener.py
```diff
@@ -212,14 +212,25 @@ def v2filter_authv1(auth: IRAuth, v2config: 'V2Config'):
...
     assert auth.proto
 
-    body_info = auth.get('include_body')
+    raw_body_info: Optional[Dict[str, int]] = auth.get('include_body')
 
-    if not body_info and auth.get('allow_request_body', False):
-        body_info = {
-            'max_request_bytes': 4096,
-            'allow_partial_message': True
+    if not raw_body_info and auth.get('allow_request_body', False):
```
gsagula: Why we still need `auth.get('allow_request_body', False)`?

kflynn: They might've given us an `AuthService` with neither `include_body` nor `allow_request_body`...

PR #1611, opened 2019-06-07
Apu/latest snapshot diag

Andrei-Predoiu:
## Description
Added Latest snapshot version to system_info

## Related Issues
None.

## Testing
There seems to be an issue with my environment, i can get it to build and push but it kubernetes it is stuck in a crash loop with:
"standard_init_linux.go:207: exec user process caused "no such file or directory""
It passes CI Just fine.

## Todos
- [ ] Tests


Andrei-Predoiu:
@richarddli I'm not sure who can take a look at this :)

PR #1618, opened 2019-06-12
Remove x_forwarded_proto_redirect infinite loop

KyleMartin901:
## Description
I am not a Python developer this PR is more just a this is how I fixed the issue I was experiencing as I am not sure of the carry on consequences of these changes. No tests have been added or changed as I wasn't sure on how to do that. Hopefully this PR will help you solve the issue described.

This commit removes a bug where when you set the ambassador module config x_forwarded_proto_redirect to true it would generate a configuration that caused an infinite redirect loop.

The main cause of this issue was that the if you set the x_forwarded_proto_redirect to true it would only create an envoy listener with the default route rather than still appending the desired custom routes specified in the Kubernetes services annotations.

## Changes
The default route now doesn't contain a redirect that does a [https_redirect](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto#route-redirectaction)
due to it causing an infinite loop of redirects as it would substitute the scheme portion of the URL for https and redirect to that URL even if that scheme was already https.

Removed the check for the `listener.redirect_listener` before building the routes so the routes would always be built and default route would be overwritten if routes have been defined whether or not
`listener.redirect_listener` had been set or not. Added a condition to set the `self.require_tls = None` only if `listener.redirect_listener` had not be set.

This change now means that Ambassador will add the `"require_tls": "EXTERNAL_ONLY"` configuration to the envoy virtual host if the Ambassador module `x_forwarded_proto_redirect` is set to true forcing all external requests to the gateway to use https and wont remove any routes.

## Related Issues
https://github.com/datawire/ambassador/issues/1571

## Testing
I have deployed this custom build to our systems and it works with our configuration with AWS and I have manually confirmed that the requests are redirected for the virtual host if x-forwarded-proto header is anything but `https`.

## Todos
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/ambassador/ambassador/envoy/v2/v2listener.py
+++ b/ambassador/ambassador/envoy/v2/v2listener.py
@@ -467,49 +467,44 @@ class V2Listener(dict):
 
         self.routes: List[dict] = [ {
                 'match': {
-                    'prefix': '/',
-                },
-                'redirect': {
-                    'https_redirect': True
+                    'prefix': '/'
                 }
             } ]
 
-        if listener.redirect_listener:
-            self.http_filters = [{'name': 'envoy.router'}]
-        else:
-            # Use the actual listener name & port number
-            self.name = "ambassador-listener-%s" % listener.service_port
+        # Use the actual listener name & port number
+        self.name = "ambassador-listener-%s" % listener.service_port
 
-            # Use a sane access log spec
-            self.access_log = [ {
-                'name': 'envoy.file_access_log',
-                'config': {
-                    'path': '/dev/fd/1',
-                    'format': 'ACCESS [%START_TIME%] \"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%\" %RESPONSE_CODE% %RESPONSE_FLAGS% %BYTES_RECEIVED% %BYTES_SENT% %DURATION% %RESP(X-ENVOY-UPSTREAM-SERVICE-TIME)% \"%REQ(X-FORWARDED-FOR)%\" \"%REQ(USER-AGENT)%\" \"%REQ(X-REQUEST-ID)%\" \"%REQ(:AUTHORITY)%\" \"%UPSTREAM_HOST%\"\n'
-                }
-            } ]
+        # Use a sane access log spec
+        self.access_log = [ {
+            'name': 'envoy.file_access_log',
+            'config': {
+                'path': '/dev/fd/1',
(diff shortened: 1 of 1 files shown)


KyleMartin901:
The tests failed because the test expected the default route to redirect which I removed because I believed it was a redirect loop.

I can change the default route to have the following config that still will redirect by default but will check the `X-Forwarded-Proto` header and if its https it will not redirect but anything else it will or would it be better to change the test to use the `x_forwarded_proto_redirect: true` config so it sets require_tls it at the virtual host level?

```
"match": {
    "prefix": "/",
    "headers": [
        {
            "name": "X-Forwarded-Proto",
            "exact_match": "https",
            "invert_match": true
        }
    ]
    },
    "redirect": {
        "https_redirect": true
}
```

containscafeine:
@KyleMartin901 thanks for the PR :+1: The approach seems reasonable.

However, when I try a `curl` with or without `"X-FORWARDED-PROTO: https"` I always get a 200. Do you mind posting a quick manual reproducer to verify the behavior?

KyleMartin901:
The route config I have is

```
"route_config": {
    "virtual_hosts": [
        {
            "domains": [
                "*"
            ],
            "name": "backend",
            "require_tls": "EXTERNAL_ONLY",
            "routes": [
                {
                    "match": {
                        "prefix": "/"
                    },
                    "redirect": {
                        "https_redirect": true,
                        "path_redirect": "/"
                    }
                }
            ]
        }
    ]
}
```

Where the service `nginx-hello.kmartin` is externally accessible via AWS ELB dns via `nginx-hello.example.com` and when I run curl against that URL I get the following

```
curl -kLi http://nginx-hello.example.com
HTTP/1.1 301 Moved Permanently
date: Wed, 12 Jun 2019 13:42:05 GMT
location: https://nginx-hello.example.com/
server: envoy
Content-Length: 0
Connection: keep-alive

HTTP/1.1 301 Moved Permanently
date: Wed, 12 Jun 2019 13:42:06 GMT
location: https://nginx-hello.example.com/
server: envoy
Content-Length: 0
Connection: keep-alive

HTTP/1.1 301 Moved Permanently
date: Wed, 12 Jun 2019 13:42:06 GMT
location: https://nginx-hello.example.com/
server: envoy
Content-Length: 0
Connection: keep-alive

HTTP/1.1 301 Moved Permanently
date: Wed, 12 Jun 2019 13:42:07 GMT
location: https://nginx-hello.example.com/
server: envoy
Content-Length: 0
Connection: keep-alive

HTTP/1.1 301 Moved Permanently
date: Wed, 12 Jun 2019 13:42:07 GMT
location: https://nginx-hello.example.com/
server: envoy
Content-Length: 0
Connection: keep-alive

HTTP/1.1 301 Moved Permanently
date: Wed, 12 Jun 2019 13:42:07 GMT
location: https://nginx-hello.example.com/
server: envoy
Content-Length: 0
Connection: keep-alive
```

kflynn:
Superseded by #1620 -- thanks @KyleMartin901!

containscafeine:
Superseded by #1620, commit included there. Thanks @KyleMartin901, you're awesome!

PR #1620, opened 2019-06-12
Fix XFP 

containscafeine:
Addendum to #1618


containscafeine:
Basically I've separated the concept of `require_tls` and `redirect_listener`, both do not imply each other anymore - because the former can mean XFP without being a redirect listener. I've re-enabled the tests as well.

kflynn:
Pulling `master` in to pick up the resolver-diagnostics fix.

PR #1637, opened 2019-06-18
Ignore IRCluster if invalid port specified

### LINKED ISSUE #1631: Crash on invalid port in Mapping
**Describe the bug**
When a Mapping has an invalid port that does not parse correctly as an integer, Ambassador crashes and does not restart into a healthy state until the Mapping has a valid value.

**To Reproduce**
Set the service on a mapping to `service: myservice:80.invalid`

**Expected behavior**
The expected behavior should be a clear error posted to the Ambassador diagnostics dashboard and for this Mapping to be ignored in config generation.

 - Ambassador: 0.61.0+
 - Kubernetes environment: On-prem BM
 - Version: Kubernetes 1.8.4

**Additional context**
We ran into this issue when an engineer wrote a Helm template for Mappings with a bug that put the service's namespace after the port instead of before it, causing Ambassador to crash in a development environment.

Ambassador had the following in its logs during pod restarts:

2019-06-14 19:31:45 diagd 0.50.0 [P45TAmbassadorEventWatcher] ERROR: ERROR ERROR ERROR Starting with configuration errors
2019-06-14 19:31:45 diagd 0.50.0 [P45TAmbassadorEventWatcher] ERROR: could not reconfigure: invalid literal for int() with base 10: '80.rtpapi'
2019-06-14 19:31:45 diagd 0.50.0 [P45TAmbassadorEventWatcher] ERROR: invalid literal for int() with base 10: '80.mynamespace'
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/ambassador-0.0.0.dev0-py3.6.egg/ambassador_diag/diagd.py", line 590, in run
    self.load_config_watt(rqueue, url)
  File "/usr/lib/python3.6/site-packages/ambassador-0.0.0.dev0-py3.6.egg/ambassador_diag/diagd.py", line 708, in load_config_watt
    self._load_ir(rqueue, aconf, fetcher, scc, snapshot)
  File "/usr/lib/python3.6/site-packages/ambassador-0.0.0.dev0-py3.6.egg/ambassador_diag/diagd.py", line 717, in _load_ir
    ir = IR(aconf, secret_handler=secret_handler)
  File "/usr/lib/python3.6/site-packages/ambassador-0.0.0.dev0-py3.6.egg/ambassador/ir/ir.py", line 195, in __init__
    MappingFactory.finalize(self, aconf)
  File "/usr/lib/python3.6/site-packages/ambassador-0.0.0.dev0-py3.6.egg/ambassador/ir/irmappingfactory.py", line 40, in finalize
    group.finalize(ir, aconf)
  File "/usr/lib/python3.6/site-packages/ambassador-0.0.0.dev0-py3.6.egg/ambassador/ir/irhttpmappinggroup.py", line 318, in finalize
    mapping.cluster = self.add_cluster_for_mapping(ir, aconf, mapping)
  File "/usr/lib/python3.6/site-packages/ambassador-0.0.0.dev0-py3.6.egg/ambassador/ir/irhttpmappinggroup.py", line 199, in add_cluster_for_mapping
    marker=marker)
  File "/usr/lib/python3.6/site-packages/ambassador-0.0.0.dev0-py3.6.egg/ambassador/ir/ircluster.py", line 172, in __init__
    port = p.port
  File "/usr/lib/python3.6/urllib/parse.py", line 169, in port
    port = int(port, 10)
ValueError: invalid literal for int() with base 10: '80.mynamespace'

containscafeine:
Fix #1631

### DIFF
--- a/ambassador/ambassador/ir/ircluster.py
+++ b/ambassador/ambassador/ir/ircluster.py
@@ -92,6 +92,9 @@ class IRCluster (IRResource):
         if marker:
             name_fields.append(marker)
 
+        # Set this flag to True if you discover something that's grave enough to warrant ignoring this cluster
+        self.ignore_cluster = False
+
         self.logger = ir.logger
 
         # Toss in the original service before we mess with it, too.
@@ -169,7 +172,13 @@ class IRCluster (IRResource):
         # p is read-only, so break stuff out.
 
         hostname = p.hostname
-        port = p.port
+        try:
+            port = p.port
+        except ValueError as e:
+            errors.append("found invalid port for service {}. Please specify a valid port between 0 and 65535 - {}".format(service, e))
+            errors.append("service {} cluster will be ignored, please re-configure".format(service))
+            self.ignore_cluster = True
+            port = 0
 
         # If the port is unset, fix it up.
         if not port:
@@ -278,6 +287,9 @@ class IRCluster (IRResource):
                 ir.post_error(error, resource=self)
 
     def setup(self, ir: 'IR', aconf: Config) -> bool:
+        if self.ignore_cluster:
+            return False
+
         # Resolve our actual targets.
         targets = ir.resolve_targets(self, self._resolver, self._hostname, self._port)
 
--- a/ambassador/ambassador/ir/ircluster.py
+++ b/ambassador/ambassador/ir/ircluster.py
@@ -175,8 +175,7 @@ class IRCluster (IRResource):
         try:
             port = p.port
         except ValueError as e:
-            errors.append("found invalid port for service {}. Please specify a valid port between 0 and 65535 - {}".format(service, e))
-            errors.append("service {} cluster will be ignored, please re-configure".format(service))
+            errors.append("found invalid port for service {}. Please specify a valid port between 0 and 65535 - {}. Service {} cluster will be ignored, please re-configure".format(service, e, service))
             self.ignore_cluster = True
             port = 0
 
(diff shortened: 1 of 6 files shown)


ambassador/ambassador/ir/ircluster.py
```diff
@@ -169,7 +172,13 @@ def __init__(self, ir: 'IR', aconf: Config,
         # p is read-only, so break stuff out.
 
         hostname = p.hostname
-        port = p.port
+        try:
```
kflynn: This doesn't feel like the right spot -- surely it's the `urlparse` that fails, not the port assignment?

containscafeine: Even I was surprised that it's the port assignment that fails. Posting the following from /usr/lib64/python3.7/urllib/parse.py -
```python
    @property
    def port(self):
        port = self._hostinfo[1]
        if port is not None:
            port = int(port, 10)
            if not ( 0 <= port <= 65535):
                raise ValueError("Port out of range 0-65535")
        return port
```

kflynn: ...oooookay then. :)

kflynn:
Landing this -- the `XFPRedirect` failure is due to an external service dying, and we've now removed the dependency. 

@containscafeine, we should talk about the "default errors" usage here -- I kinda tossed that in a little hackishly in the first place, and it's now at the point where we should probably make sure it makes a kind of sense.

PR #1638, opened 2019-06-19
Makefile: Improve Envoy workflow

LukeShu:
## Description

 - Rename `envoy` to `envoy-src`:
    * Running `make envoy` to grab the Envoy *sources* was confusing.
    * `./envoy-src/` and `./envoy-bin/` have better symmetry than `./envoy/` and `./envoy-bin/`.
    * Detect if `./envoy/` exists but `./envoy-src/` doesn't, and just rename it, to avoid re-downloading it and such; for if you had a copy from before this change.
 - rsync the Envoy sources in both directions: Both in to the image (like before) and out of the image (new).
    * This means that you can poke around in `./envoy-src/` to see build artifacts (but a lot of them will be symlinks to non-existent files in `~/.bazel/`).
    * This means I can tack on `--delete` to the the rsync commands; because cached build artifacts now exist on the host, --delete no longer means that the update-the-in-container rsync will delete the in-container build cache.
    * This also means that we can do the rsync more frequently, as it it is no longer the case that copying in the sources has the potential to stomp on a failed-build-state that you would like to inspect (see below)..
    * This has the downside that if you start a build, then keep hacking on the host's copy of the sources, when the build finishes the rsync-out will override your changes.
 - Also do the rsync stuff for `make envoy-shell` (and `make check-envoy`).
    * This is made possible by the rsync-both-directions change above.
    * Previously, you'd have to start a build with `make
      envoy-bin/envoy-static` to get it to copy in the sources before
      you could `make envoy-shell`.

## Testing

I've made these changes working from the macincloud, so I know that these changes work reasonably on macOS.

That said, I did get the GCC crash when trying to run `make check-envoy`.  The theory that it's an out-of-memory situation makes sense; the Docker VM on that box only has 2G RAM and 1G swap.

I have _not_ tried to run with these changes on GNU/Linux, but I have no reason to think that anything in this would not work there.

## Todos

Verify that `make go/apis/envoy` works.  I haven't successfully tested that, because I ran in to a python2 vs python3 issue on the macincloud.


LukeShu:
`make go/apis/envoy` was broken, then I spent a while trying to fix it.  There was breakage introduced by this PR, so I added a commit fixing that specific breakage.  But it was already broken, and I haven't fixed that breakage.  That can be a later unit of work.

I'm stoked that the workflow for Envoy that emerged resulted in `make go/apis/envoy` went un-noticed, since `make docker-update-base` is supposed to call it.  Oh well, that can be dealt with later.

gsagula:
@LukeShu I have tried other make commands such as `make docker-update-base` and I can confirm that it works.  I haven't tried `make go/apis/envoy` but I agree that we can deal with it later. Thanks!

PR #1648, opened 2019-06-26
Added try except for ACResource processing.

Andrei-Predoiu:
## Description
This is a fix for the following issue, It covers both points 1 and 2 of my proposed fix

https://github.com/datawire/ambassador/issues/1635

## Related Issues
https://github.com/datawire/ambassador/issues/1635

## Testing
I have tested it manually, deployed on several clusters and tried to reproduce the issue.
It doesn't seem possible to use the current kat setup to test for error handling in abassador so i was unable to write a mapping test.

## Other
Depending on what requirements ambassador will have for error reporting and handling.
The whole "Error" handling mechanism might need to be redesigned. I think using Exceptions for flow control ss adequate for now, especially since it's python.

### DIFF
--- a/ambassador/ambassador/config/resourcefetcher.py
+++ b/ambassador/ambassador/config/resourcefetcher.py
@@ -307,8 +307,11 @@ class ResourceFetcher:
         # Fine. Fine fine fine.
         serialization = dump_yaml(obj, default_flow_style=False)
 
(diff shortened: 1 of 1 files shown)


Andrei-Predoiu:
@kflynn Cleaned it up :)

kflynn:
Landing this -- the `XFPRedirect` failure is due to an external service dying, and we've now removed the dependency.

Issue #1478, opened 2019-04-30
Handle timeout when validating envoy config


thrucker:
**Describe the bug**
When the envoy config is updated, it is validated by running `envoy --config-path <path> --mode validate`. This currently has a fixed timeout of [5 seconds](https://github.com/datawire/ambassador/blob/9484c0b07465e1547cc57a8942d3374d4ef4cf66/ambassador/ambassador_diag/diagd.py#L874). However, when an actual timeout happens this is not properly handled and seems to crash the automatic reconfiguration of ambassador. 

This ultimately leads to a state where ambassador still runs and handles requests properly, but is not able to receive new route mappings anymore. In this case only a restart of ambassador seems to help.

**To Reproduce**
I'm not sure how one would reproduce such a timeout. At least in our case it seems to happen from time to time and results in the described behavior.

Looking into the code at https://github.com/datawire/ambassador/blob/9484c0b07465e1547cc57a8942d3374d4ef4cf66/ambassador/ambassador_diag/diagd.py#L873-L878 it seems that only `subprocess.CalledProcessError` is handled. According to the [python documentation](https://docs.python.org/3/library/subprocess.html#exceptions) a timeout will result in a `subprocess.TimeoutExpired` error.

Log Output:
```
2019-04-30 15:01:07 diagd 0.60.1 [P54TAmbassadorEventWatcher] ERROR: could not reconfigure: Command '['envoy', '--config-path', '/ambassador/snapshots/econf-tmp.json', '--mode', 'validate']' timed out after 5 seconds
2019-04-30 15:01:07 diagd 0.60.1 [P54TAmbassadorEventWatcher] ERROR: Command '['envoy', '--config-path', '/ambassador/snapshots/econf-tmp.json', '--mode', 'validate']' timed out after 5 seconds
Traceback (most recent call last):
  File "/usr/lib/python3.6/subprocess.py", line 425, in run
    stdout, stderr = process.communicate(input, timeout=timeout)
  File "/usr/lib/python3.6/subprocess.py", line 863, in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
  File "/usr/lib/python3.6/subprocess.py", line 1535, in _communicate
    self._check_timeout(endtime, orig_timeout)
  File "/usr/lib/python3.6/subprocess.py", line 891, in _check_timeout
    raise TimeoutExpired(self.args, orig_timeout)
[...]

PR #1652, opened 2019-06-26
handle TimeoutExpired on envoy validation, to keep receiving config updates

jwm:
## Description
When `envoy --mode validate` fails, Ambassador receives an unhandled `subprocess.TimeoutExpired` and stops picking up all future configuration changes.

This adds retries to Envoy config validation, in the fervent hope that a subsequent validation will not time out.

## Related Issues
#1478

## Testing
We were hitting `TimeoutExpired` a few times per day. We're running this branch in production, and are waiting for the timeout case to hit to make sure the updates handle it correctly.

## Todos
- [ ] Tests -- not sure if there are any automated tests for this. I'm not too familiar with Ambassador's code base, but looking at previous revisions that touched this code, it doesn't look like there are any tests to be updated.
- [X] Documentation -- think this change is pretty transparent, having Ambassador DTRT.

### DIFF
--- a/ambassador/ambassador_diag/diagd.py
+++ b/ambassador/ambassador_diag/diagd.py
@@ -111,7 +111,7 @@ class DiagApp (Flask):
     def setup(self, snapshot_path: str, bootstrap_path: str, ads_path: str,
               config_path: Optional[str], ambex_pid: int, kick: Optional[str],
               k8s=False, do_checks=True, no_envoy=False, reload=False, debug=False, verbose=False,
-              notices=None):
+              notices=None, validation_retries=5):
         self.estats = EnvoyStats()
         self.health_checks = do_checks
         self.no_envoy = no_envoy
@@ -121,6 +121,7 @@ class DiagApp (Flask):
         self.notices = Notices(self.notice_path)
         self.notices.reset()
         self.k8s = k8s
+        self.validation_retries = validation_retries
 
         # This will raise an exception and crash if you pass it a string. That's intentional.
         self.ambex_pid = int(ambex_pid)
@@ -732,7 +733,7 @@ class AmbassadorEventWatcher(threading.Thread):
 
         bootstrap_config, ads_config = econf.split_config()
 
-        if not self.validate_envoy_config(config=ads_config):
+        if not self.validate_envoy_config(config=ads_config, retries=self.app.validation_retries):
             self.logger.info("no updates were performed due to invalid envoy configuration, continuing with current configuration...")
             # Don't use app.check_scout; it will deadlock.
             self.check_scout("attempted bad update")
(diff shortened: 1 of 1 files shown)


jwm:
Also, WDYT about increasing the timeout here, maybe to `10s`? Feels like it would help Ambassador continue on with important management plane activity in the face of CPU contention.

Our current `envoy.json` is ~150k and takes ~750msec on average to validate, with spikes that correlate highly to load causing timeouts 5-10 times/day.

richarddli:
Thanks for tracking this down! How many services do you have and do you mind characterizing what type of load you're seeing? I'm a little surprised that traffic spikes can cause these issues, but find it completely believable. (And also FTR we're based in South Station, Boston, if you want to come over sometime for some swag!)

kflynn:
@jwm Let me second @richarddli's invitation -- if you're hitting this multiple times a day, I _definitely_ want to talk to you about your setup.

kflynn:
@jwm Code looks good. I'm fine with bumping the timeout (and/or making it configurable) -- how's it looking running at your shop?

jwm:
Hi @richarddli @kflynn, thanks for responding so quickly! Before this gig, I worked near Summer and A Streets. I'll have to make a point to drop by sometime. :-)

Our config renders into a ~dozen Envoy clusters and 2 listeners. Our primary listener has a dozen or so filter chain entries. `envoy.json` is about 150kbytes. Traffic volume is fairly light (10-20 QPS at peak).

One thing I just realized I should mention: we're running Ambassador as a `DaemonSet` with `Burstable` QoS, with a CPU request of `100m` and a limit of `500m`. I wouldn't be surprised at all that intermittent CPU starvation is a factor here, and I've been starting to think this is a suboptimal deployment strategy. We're just moving this workload to k8s, and there are a number of things we're shuffling around as we find our way.

The lightly trafficked environment I deployed these changes to hasn't hit a timeout yet. I'll move our production environment to it hopefully tonight. I'll let you know how it goes!

nicolasbelanger:
FYI... Got bitten by the same issue.  We were using 100m for both limits and requests cpu.  Bumping limit to 1 (1000m) and it's mostly problem solved.

jwm:
@richarddli @kflynn These changes have been fine here for the last few days, but we haven't triggered the configuration validation timeout since we bumped the CPU resource reservations on Ambassador pods. I can bump them down in one of our envs to double check, but I think the code is fine.

Beyond that, it looks like the [Travis build is failing](https://travis-ci.org/datawire/ambassador/builds/551367624#L1690):

```
Checking 336 url requirements... 322 not ready (http://ambassadoridtest-http: Get http://ambassadoridtest-http: net/http: request canceled (Client.Timeout exceeded while awaiting headers)) sleeping 10s...
```

I'll plan on setting up a local environment to run the test suite, but probably won't get to it in earnest until next week because of the holiday and trying to [figure out the `503 NR`s in Envoy we're experiencing](https://datawire-oss.slack.com/archives/CAULN7S76/p1562013430015900).

PR #1655, opened 2019-06-27
Configurable cluster_idle_timeout_ms to bound upstream keepalives

esmet:
## Description
Add cluster_idle_timeout_ms to the ambassador module and Mapping schema, so that operators can set upstream keepalive timeouts.

## Related Issues
https://github.com/datawire/ambassador/issues/1542

## Testing
We use this for our use case. It ensures that Envoy times out idle connections before the typical upstream, to avoid the http1 race condition of missing an upstream close event at the moment a connection is chosen for proxying.

## Todos
- [ ] Tests
- [ ] Documentation

This needs some docs and probably a unit test.

Open question: should we allow this to be overriden per Mapping? Or does that make merging Mappings into a cluster more complicated?

### DIFF
--- a/ambassador/ambassador/envoy/v2/v2cluster.py
+++ b/ambassador/ambassador/envoy/v2/v2cluster.py
@@ -35,11 +35,16 @@ class V2Cluster(dict):
             else:
                 dns_lookup_family = 'V6_ONLY'
 
+        cluster_idle_timeout_ms = cluster.ir.ambassador_module.get('cluster_idle_timeout_ms', 30000)
+
         fields = {
             'name': cluster.name,
             'type': cluster.type.upper(),
             'lb_policy': cluster.lb_type.upper(),
             'connect_timeout':"%0.3fs" % (float(cluster.connect_timeout_ms) / 1000.0),
+            'common_http_protocol_options': {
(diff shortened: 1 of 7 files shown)


ambassador/ambassador/envoy/v2/v2cluster.py
```diff
@@ -35,11 +35,16 @@ def __init__(self, config: 'V2Config', cluster: IRCluster) -> None:
             else:
                 dns_lookup_family = 'V6_ONLY'
 
+        cluster_idle_timeout_ms = cluster.ir.ambassador_module.get('cluster_idle_timeout_ms', 30000)
```
esmet: We might actually want a 60s default, which is what nginx does (http://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive_timeout)

kflynn:
I think that this should be configurable per-`Mapping`. It should be pretty apparent where you'll need to add code to split clusters -- if not, sing out!

kflynn:
Taking this one too; I'm hoping to support timeouts for testing shortly.

kflynn:
Whoops, jumped to gun there. @esmet, any joy on per-`Mapping` config here? 🙂

esmet:
> Whoops, jumped the gun there. @esmet, any joy on per-`Mapping` config here? 🙂

I've been slacking on this. I'll try it today.

PR #1656, opened 2019-06-27
Support selectors to filter which objects are processed by an Ambassador deployment.

esmet:
Support the ability to tell Watt which label selectors or field selectors to use. This can be used by an operator to cut down on the number of objects processed by an Ambassador deployment. 

## Related Issues
https://github.com/datawire/ambassador/issues/1292

## Testing
We use this for our testing environment use case to dramatically reduce the number of Service objects processed on a highly multi-tenant Kubernetes cluster. Each Ambassador deployment has a distinct ambassador_id, and they further provide Watt with a label selector such as `environment=someEnv` to limit Service results to only the set of things that can possibly contain mappings for that ambassador_id. Without this change, memory and CPU utilization on all Ambassadors is high.

## Todos
- [ ] Tests
- [ ] Documentation
This probably deserves to be mentioned in the documentation, especially near any existing documentation for running Ambassador in a multi-tenant cluster.

### DIFF
--- a/ambassador/entrypoint.sh
+++ b/ambassador/entrypoint.sh
@@ -200,11 +200,23 @@ if [[ -z "${AMBASSADOR_NO_KUBEWATCH}" ]]; then
         KUBEWATCH_SYNC_KINDS="$KUBEWATCH_SYNC_KINDS -s ConsulResolver -s KubernetesEndpointResolver -s KubernetesServiceResolver"
     fi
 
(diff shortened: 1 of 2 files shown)


kflynn:
For the record, I looked at this and like it. Planning to land it after 0.73 ships, so that we can do some experimentation with it.

kflynn:
We decided to change some things around with 0.73, so I'm going to go ahead and land this so we can start playing with it. Thanks!!

PR #1664, opened 2019-07-02
Allow configuring what '.' in a Secret name means

### LINKED ISSUE #1255: TLS custom secret with period in name does not work
**Describe the bug**
When your custom TLS secret name has a period (.) in the name Ambassador will fail to start Envoy and will error thinking the first part before the period is the secret name. Quoting the secret name in the annotation does not make a difference.

```
RichStatus BAD error=TLSContext server found no certificate in secret with in namespace period, ignoring...
```

Bad Module/tls configuration:

```yaml
  annotations:
    getambassador.io/config: |
      ---
      apiVersion: ambassador/v1
      kind:  Module
      name:  tls
      config:
        server:
          enabled: True
          secret: with.period
```

Good configuration:

```yaml
  annotations:
    getambassador.io/config: |
      ---
      apiVersion: ambassador/v1
      kind:  Module
      name:  tls
      config:
        server:
          enabled: True
          secret: withoutperiod
```

**To Reproduce**
Create a secret with a name of "with.period" and appropriate tls.key and tls.cert values
Apply the "Bad" example Module/tls annotation
Observe the error in Ambassador logs and the diagnostics UI

**Expected behavior**
The "with.period" secret should be read as specified in the Module/tls annotation and Ambassador should successfully start Envoy

 - Ambassador: 0.50.3
 - Kubernetes environment: AKS
 - Version: 1.12.5

**Additional context**
Secrets with periods in the name were working successfully in Ambassador 0.40.2

kflynn:
By default, if we see a `.` in a `Secret` name, we interpret the name as `secret.namespace`. However, that breaks some things that want to use `.` to mean... a `.`.

This PR allows you to disable that namespace crossing, so a `.` is just a `.`. You can do this in two ways:

1. Set `tls_secret_namespacing` in the `defaults` dict in the `ambassador` `Module` to set the behavior for all `TLSContext`s. By default, `tls_secret_namespacing` is `true`, but you can set it to `false` to change that:

```yaml
---
apiVersion: ambassador/v0
kind:  Module
name:  ambassador
config:
  defaults:
    tls_secret_namespacing: False
```

2. Set `secret_namespacing` to `false` to override the default for a specific `TLSContext`:

```yaml
---
apiVersion: ambassador/v1
kind: TLSContext
name: ...
secret: ...
secret_namespacing: False
```

FIxes #1255.


TLmaK0:
@kflynn thanks for this incredible tool. There is a little issue in the documentation about the `secret_namespacing` feature. It took to me some time to figure out why my config is not working. Looking in to the documentation there isn't information about this feature:
image
I think should also reflect the naming conventions in `secret` parameter.

PR #1669, opened 2019-07-05
Add Knative support

containscafeine:
TODO:

- [x] docs
- [x] assert that cluster ingress is always created when a kservice exists
- [x] add support for traffic split
- [ ] check support with eventing
- [x] see how CI behaves with running knative

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,7 +159,7 @@ KAT_BACKEND_RELEASE = 1.4.2-rc
 
 # Allow overriding which watt we use.
 WATT ?= watt
-WATT_VERSION ?= 0.5.1
+WATT_VERSION ?= 0.6.0
 
 # "make" by itself doesn't make the website. It takes too long and it doesn't
 # belong in the inner dev loop.
--- a/ambassador/ambassador/config/config.py
+++ b/ambassador/ambassador/config/config.py
@@ -63,6 +63,8 @@ class Config:
         'tracingservice': "tracing_configs",
     }
 
+    KnativeResources = ['ClusterIngress', 'kservice']
+
     SupportedVersions: ClassVar[Dict[str, str]] = {
         "v0": "is deprecated, consider upgrading",
         "v1": "ok",
@@ -203,6 +205,11 @@ class Config:
     # Often good_ambassador_id will be passed an ACResource, but sometimes
     # just a plain old dict.
     def good_ambassador_id(self, resource: dict):
+        resource_kind = resource.get('kind')
+        if resource_kind in self.KnativeResources:
+            self.logger.debug("Knative resource: {} does not require an Ambassador ID".format(resource_kind))
+            return True
+
         # Is an ambassador_id present in this object?
         allowed_ids: StringOrList = resource.get('ambassador_id', 'default')
 
@@ -213,7 +220,12 @@ class Config:
             if type(allowed_ids) != list:
                 allowed_ids = typecast(StringOrList, [ allowed_ids ])
 
-            return Config.ambassador_id in allowed_ids
+            if Config.ambassador_id in allowed_ids:
+                return True
+            else:
+                self.logger.debug("Ambassador ID {} does not exist in allowed IDs {}".format(Config.ambassador_id, allowed_ids))
+                self.logger.debug(resource)
+                return False
 
     def save_source(self, resource: ACResource) -> None:
         """
@@ -230,6 +242,7 @@ class Config:
         rcount = 0
 
         for resource in resources:
+            self.logger.debug("Trying to parse resource: {}".format(resource))
             rcount += 1
 
             if not self.good_ambassador_id(resource):
@@ -372,19 +385,25 @@ class Config:
             apiVersion = apiVersion.replace('getambassador.io/', 'ambassador/')
             resource.apiVersion = apiVersion
 
+        is_ambassador = False
+
         # Ditch the leading ambassador/ that really needs to be there.
         if apiVersion.startswith("ambassador/"):
+            is_ambassador = True
             apiVersion = apiVersion.split('/')[1]
+        elif apiVersion.startswith('serving.knative.dev') or apiVersion.startswith('networking.internal.knative.dev'):
+            pass
         else:
             return RichStatus.fromError("apiVersion %s unsupported" % apiVersion)
 
         version = apiVersion.lower()
 
         # Is this deprecated?
-        status = Config.SupportedVersions.get(version, 'is totally bogus')
+        if is_ambassador:
+            status = Config.SupportedVersions.get(version, 'is totally bogus')
 
-        if status != 'ok':
-            self.post_notice(f"apiVersion {originalApiVersion} {status}", resource=resource)
+            if status != 'ok':
+                self.post_notice(f"apiVersion {originalApiVersion} {status}", resource=resource)
 
         if resource.kind.lower() in Config.NoSchema:
             return RichStatus.OK(msg=f"no schema for {resource.kind} so calling it good")
--- a/ambassador/ambassador/config/config.py
+++ b/ambassador/ambassador/config/config.py
@@ -63,7 +63,7 @@ class Config:
         'tracingservice': "tracing_configs",
     }
 
-    KnativeResources = ['ClusterIngress', 'kservice']
+    KnativeResources = ['ClusterIngress']
 
     SupportedVersions: ClassVar[Dict[str, str]] = {
         "v0": "is deprecated, consider upgrading",
--- a/ambassador/ambassador/config/config.py
+++ b/ambassador/ambassador/config/config.py
@@ -63,7 +63,7 @@ class Config:
         'tracingservice': "tracing_configs",
     }
 
-    KnativeResources = ['ClusterIngress']
+    KnativeResources = {'ClusterIngress'}
 
     SupportedVersions: ClassVar[Dict[str, str]] = {
         "v0": "is deprecated, consider upgrading",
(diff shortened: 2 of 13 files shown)


ambassador/ambassador/config/config.py
```diff
@@ -63,6 +63,8 @@ class Config:
         'tracingservice': "tracing_configs",
     }
 
+    KnativeResources = ['ClusterIngress', 'kservice']
```
kflynn: How about a `set` instead of a `list` here?

ambassador/ambassador/config/config.py
```diff
@@ -230,6 +242,7 @@ def load_all(self, resources: Iterable[ACResource]) -> None:
         rcount = 0
 
         for resource in resources:
+            self.logger.debug("Trying to parse resource: {}".format(resource))
```
kflynn: FTR I've gotten _really_ fond of `f''` strings in Python3. 🙂

ambassador/ambassador/config/config.py
```diff
@@ -372,19 +385,25 @@ def validate_object(self, resource: ACResource) -> RichStatus:
             apiVersion = apiVersion.replace('getambassador.io/', 'ambassador/')
             resource.apiVersion = apiVersion
 
+        is_ambassador = False
+
         # Ditch the leading ambassador/ that really needs to be there.
         if apiVersion.startswith("ambassador/"):
+            is_ambassador = True
             apiVersion = apiVersion.split('/')[1]
+        elif apiVersion.startswith('serving.knative.dev') or apiVersion.startswith('networking.internal.knative.dev'):
```
kflynn: Should this be `elif apiVersion.endswith('.knative.dev')` ?

ambassador/ambassador/config/config.py
```diff
@@ -372,19 +385,25 @@ def validate_object(self, resource: ACResource) -> RichStatus:
...
 
         # Is this deprecated?
-        status = Config.SupportedVersions.get(version, 'is totally bogus')
+        if is_ambassador:
+            status = Config.SupportedVersions.get(version, 'is totally bogus')
 
-        if status != 'ok':
-            self.post_notice(f"apiVersion {originalApiVersion} {status}", resource=resource)
+            if status != 'ok':
+                self.post_notice(f"apiVersion {originalApiVersion} {status}", resource=resource)
```
kflynn: Whoa. I seriously wrote code so that a customer could see "apiVersion 99.3 is totally bogus"?? OMG. Let's, uh, fix that? 🙂

ambassador/ambassador/ir/ir.py
```diff
@@ -440,6 +443,68 @@ def add_mapping(self, aconf: Config, mapping: IRBaseMapping) -> Optional[IRBaseM
...
+
+                ci_http = ci_rule.get('http', None)
+                if ci_http is not None:
+                    ci_paths = ci_http.get('paths', [])
+                    for ci_path in ci_paths:
+                        ci_headers = ci_path.get('appendHeaders', [])
+
+                        ci_splits = ci_path.get('splits', [])
+                        for ci_split in ci_splits:
+                            mapping_identifier = ci_name + '-' + ''.join(random.sample(string.ascii_lowercase, 5))
```
kflynn: Hmmmmmm. Does this really _need_ to be nondeterministic? Is there something deterministic we can use instead? There are two reasons I ask:

1. It makes certain classes of transcoding tests much harder, and
2. If the name of the mapping changes on every reconfigure, it'll be much harder for Jane and for Envoy to follow which things are changing and which are not.

Makefile
```diff
@@ -189,7 +189,6 @@ clobber: clean
 	-rm -rf watt
 	-$(if $(filter-out -,$(ENVOY_COMMIT)),rm -rf envoy envoy-src)
 	-rm -rf docs/node_modules
-	-rm -rf .skip_test_warning	# reset the test warning too
```
kflynn: Hmmm, I almost missed this. Can you take this out of this PR, please? I'm not necessarily averse to this change as a separate thing, but let's not bury a change to this build semantic in this. 🙂

ambassador/ambassador/config/config.py
containscafeine: done

ambassador/ambassador/config/config.py
containscafeine: That's much cleaner, thanks :)

ambassador/ambassador/config/config.py
containscafeine: We could do that, but I wanted to explicit to parse only ClusterIngress or kservice here and no other Knative resource since Ambassador does not understand anything else

ambassador/ambassador/config/config.py
containscafeine: Hehe. Changed it to "is not supported"

```
            status = Config.SupportedVersions.get(version, 'is not supported')
```

ambassador/ambassador/ir/ir.py
containscafeine: makes sense, fixed.

ambassador/ambassador/config/config.py
kflynn: Good enough.

ambassador/ambassador/ir/ir.py
```diff
@@ -440,6 +443,69 @@ def add_mapping(self, aconf: Config, mapping: IRBaseMapping) -> Optional[IRBaseM
...
+
+                ci_http = ci_rule.get('http', None)
+                if ci_http is not None:
+                    ci_paths = ci_http.get('paths', [])
+                    for path_count, ci_path in enumerate(ci_paths):
+                        ci_headers = ci_path.get('appendHeaders', [])
+
+                        ci_splits = ci_path.get('splits', [])
+                        for split_count, ci_split in enumerate(ci_splits):
+                            unique_suffix = f"{rule_count}{path_count}"
```
kflynn: I feel like this should be `f'{rule_count}-{path_count}'` to avoid collisions between, say, rule 1, path 21 and rule 12, path 1...

ambassador/ambassador/ir/ir.py
```diff
@@ -11,7 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License
-
+import random
```
kflynn: This is no longer needed, correct?

ambassador/ambassador/ir/ir.py
containscafeine: makes sense, fixed

ambassador/ambassador/ir/ir.py
containscafeine: fixed

docs/user-guide/knative.md
```diff
@@ -0,0 +1,55 @@
...
+
+Knative requires a Kubernetes cluster v1.11 or newer with the MutatingAdmissionWebhook admission controller enabled. kubectl v1.10 is also required. This guide assumes that you’ve already created a Kubernetes cluster which you’re comfortable installing alpha software on.
+
+1. Install and configure Knative:
+
+   ```
+   kubectl apply -f https://github.com/knative/serving/releases/download/v0.7.1/serving.yaml
+   ```
+   
+   By default, Knative uses Istio of all ingress traffic. After applying the above manifests, edit the `config-network` ConfigMap to use Ambassador:
```
nbkrause: You actually don't need to patch the `config-network` ConfigMap to get this working anymore.

docs/user-guide/knative.md
```diff
@@ -0,0 +1,55 @@
...
+[Knative](https://knative.dev/) is is a popular Kubernetes-based platform for managing serverless workloads.
+
+Replacing Istio with Ambassador can potentially reduce the operational overhead of running Knative. Ambassador can watch for changes in Knative configuration in your Kubernetes cluster and set up routing accordingly.
+
+## Getting started
+
+#### Prerequisites
+
+Knative requires a Kubernetes cluster v1.11 or newer with the MutatingAdmissionWebhook admission controller enabled. kubectl v1.10 is also required. This guide assumes that you’ve already created a Kubernetes cluster which you’re comfortable installing alpha software on.
+
```
nbkrause: Add a 
`### Install`
line or something to separate this from the prerequisites

docs/user-guide/knative.md
```diff
@@ -0,0 +1,55 @@
+# Knative Integration
+
+[Knative](https://knative.dev/) is is a popular Kubernetes-based platform for managing serverless workloads.
+
+Replacing Istio with Ambassador can potentially reduce the operational overhead of running Knative. Ambassador can watch for changes in Knative configuration in your Kubernetes cluster and set up routing accordingly.
+
+## Getting started
```
nbkrause: I think it would be worthwhile to explain Knative at a very high level here. The reason I believe this is because the Ambassador integration is only for one portion of Knative (the `Knative Serving` portion) and we may get people here who do not know much enough about Knative to know this will not do everything they need.

docs/user-guide/knative.md
```diff
@@ -0,0 +1,55 @@
...
+        - image: gcr.io/knative-samples/helloworld-go
+          env:
+          - name: TARGET
+            value: "Ambassador is Awesome"
+   ```
+   
+5. Send a request to Ambassador with the hostname assigned to the Knative Service:
+   ```
+   curl -H “Host: <hostname>” <ambassador IP>
+   ```
```
nbkrause: It may be worth it to add a quick summary and any notes here. "Notes" could include any limitations or gotchas people should be aware of.

docs/user-guide/knative.md
containscafeine: fixed

docs/user-guide/knative.md
containscafeine: Fixed

docs/user-guide/knative.md
containscafeine: Makes sense. Added. LGTY?

docs/user-guide/knative.md
containscafeine: Added.

docs/user-guide/knative.md
nbkrause: Looks great!

PR #1675, opened 2019-07-10
Add Success rate Legend to overview html

wpears:
## Description
Adds a legend to the bottom of the overview template so that it's more obvious what the colors of the service row in the Route Table mean.

## Testing
Open overview page, view new legend table below Route Table.


kflynn:
Thanks!!

PR #1695, opened 2019-07-18
Tie weight to routes, refactor canary

containscafeine:


ambassador/ambassador/envoy/v2/v2route.py
```diff
@@ -24,14 +25,31 @@
...
         super().__init__()
 
         envoy_route = EnvoyRoute(group).envoy_route
 
+        mapping_prefix = mapping.get('prefix', None)
+        route_prefix = mapping_prefix if mapping_prefix is not None else group.get('prefix')
+
+        mapping_case_sensitive = mapping.get('case_sensitive', None)
+        case_sensitive = mapping_case_sensitive if mapping_case_sensitive is not None else group.get('case_sensitive', True)
+
```
kflynn: Wow. We really need to set up a method that gets from a mapping with a fallback to the group, huh?

ambassador/ambassador/envoy/v2/v2route.py
```diff
@@ -24,14 +25,31 @@
...
+        runtime_fraction = {
+            'default_value': {
+                'numerator': mapping.get('weight', 100),
+                'denominator': 'HUNDRED'
+            }
+        }
+
+        if len(mapping) > 0:
+            runtime_fraction['runtime_key'] = f'routing.traffic_shift.{mapping.cluster.name}'
+
```
kflynn: Hmmm! Does this mean that we can actually look at supporting the runtime settings now?

ambassador/ambassador/envoy/v2/v2route.py
```diff
@@ -44,7 +62,7 @@ def __init__(self, config: 'V2Config', group: IRHTTPMappingGroup) -> None:
         # `per_filter_config` is used for customization of an Envoy filter
         per_filter_config = {}
 
-        if group.get('bypass_auth', False):
+        if mapping.get('bypass_auth', False):
```
kflynn: Hmmm. This seems like a behavior change?

ambassador/ambassador/ir/irbasemappinggroup.py
```diff
@@ -9,4 +9,54 @@
 
 
 class IRBaseMappingGroup (IRResource):
-    pass
+    def normalize_weights_in_mappings(self):
```
kflynn: How much of this is moving code and how much is new?

ambassador/ambassador/envoy/v2/v2route.py
containscafeine: @kflynn so, I've added fallback to group to the fields that we are playing with in the group. So, just in case the mapping does not have field _x_ but the group has determined that the mapping _should_ have _x_, then we accept _x_.

priority(mapping) > priority(group)

Makes sense?

ambassador/ambassador/envoy/v2/v2route.py
containscafeine: Sure, what kind?

ambassador/ambassador/envoy/v2/v2route.py
containscafeine: Yeah, I've done this to fields that are not touched in the group. Other fields are `host_rewrite`, `auto_host_rewrite`, etc.

Till now `V2Route` was not aware of a mapping (mostly because of the weight), but now that weights are tied per mapping and not per group, I'm getting fields straight from the mapping.

ambassador/ambassador/ir/irbasemappinggroup.py
containscafeine: This is about 95% new - and is the meat of the PR.

The major difference is -
- till now we've wanted the weights to total to 100, e.g. 10 + 20 + 30 + 40.
- but now we want weights to increment to 100. So if we want assigned weights to be 10, 20, 30, 40 - we want feed the following to runtime fraction - 10, 30, 60, 100, i.e. 10, 10+20, 10+20+30, 10+20+30+40.

Like you had explained to me earlier, runtime fraction is going to calculate a number between 0 and 100, and whichever bracket it falls into, it's going to route to that.

So, if this normalizing algorithm goes wrong, weights are gonna break.

kflynn:
Looking sane, I think -- landing this after local testing as I continue wrestling with Travis.

PR #1713, opened 2019-07-26
be able to configure envoy for ipv6 support

vangheem:
## Description

To be able to run envoy with ipv6

## Related Issues
diag still not supported

## Testing

Deployed and ran on cluster

## Todos
- [ ] Tests? Do I need to write tests for this?
- [x] Documentation


kflynn:
Hey, so this looks really cool! (Sorry for the delay responding, had a bunch of testing issues to unravel recently.)

How are you setting up your cluster to support IPv6?

Simplest thing to do for `diagd` might just be to modify it such that if no `--host` is passed in, it binds to `INADDR_ANY` and catches both IPv4 and IPv6; have you tried that, by any chance?

Thanks again!

vangheem:
I have not even tried to setup diagd for it yet--we just need envoy to work correctly.

Do I need to rebase this so tests pass?

kflynn:
Yes, we'll need to get tests passing and we'll need to get `diagd` working before taking it.

I'm also really curious about how you're setting up Kubernetes for IPv6 because I want to start playing with this in-house. :wink:

vangheem:
I have no idea how to manually setup k8s for ipv6... I have an existing cluster that was setup for me that I'm doing testing on.

PR #1714, opened 2019-07-26
tls-context: added logic to bypass secret validation when tls context has no hostnames

gsagula:
Signed-off-by: Gabriel Linden Sagula <gsagula@gmail.com>

## Description
This PR modifies the TLS context object to not run certificate validation on contexts that do not contain a hostname.  

## Related Issues
https://github.com/datawire/ambassador/issues/1708

## Testing
Add one test that asserts not TLS error and another test that check validation errors.

## Todos
- Documentation

### DIFF
--- a/ambassador/ambassador/ir/irtlscontext.py
+++ b/ambassador/ambassador/ir/irtlscontext.py
@@ -140,10 +140,11 @@ class IRTLSContext(IRResource):
             is_valid = True
 
         # If we don't have secret info, it's worth logging.
-        if not self.secret_info:
+        if not self.secret_info and not self.hosts is None:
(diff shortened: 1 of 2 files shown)


ambassador/ambassador/ir/irtlscontext.py
```diff
@@ -140,11 +140,12 @@ def resolve(self) -> bool:
             is_valid = True
 
         # If we don't have secret info, it's worth logging.
-        if not self.secret_info:
+        if not self.secret_info and not self.hosts is None:
```
kflynn: You can probably just do

```
if self.hosts and not self.secret_info:
```

because for most things in Python, `if foo` will be `True` only if `if foo is not None` is `True`.

PR #1717, opened 2019-07-30
Add support for Kubernetes Ingress objects

containscafeine:
- [x] add more tests for TLS and default backend
- [x] add docs


containscafeine:
@kflynn the upgrade path is not going to be drop in - if you simply update you ambassador image with this one, watt crashes with ingress permission error and hence ambassador crashes as well.
This needs to be applied with the updated RBAC.

kflynn:
Superseded by #1836.

containscafeine:
@vichaos
> Would it possible to annotate ambassador mapping within the ingress resource.

Yes, #1839 takes care of that.

PR #1726, opened 2019-08-01
Capture pid from launch through an environment variable instead of a return status to avoid 8 bit overflow.

esmet:
## Description
Avoid 8 bit overflow when entrypoint.sh happens to use pids > 255

## Testing
Without the patch, 0.74.0+ reliably fails to kick ambex on my local build. With the patch, entrypoint.sh successfully kicks ambex and Envoy gets configured from XDS correctly.

### DIFF
--- a/ambassador/entrypoint.sh
+++ b/ambassador/entrypoint.sh
@@ -226,7 +226,9 @@ launch() {
 
     pids+=("${pid}:${cmd}")
 
-    return $pid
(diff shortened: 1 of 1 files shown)


esmet:
@kflynn may want to reject this in favor of a stdout/stderr capturing approach.

let this PR represent the issue and a naive fix, if all else fails.

kflynn:
Closing in favor of recent work -- thanks for finding this bug!!

PR #1740, opened 2019-08-06
Adding `max_request_headers_kb` configuration support.

ankurpshah:
## Description
Exposing `max_request-headers_kb` envoy parameter through the ambassador.

## Related Issues
- None

## Testing
- `max_request-headers_kb` configured in [Ambassador Global configuration](https://www.getambassador.io/reference/core/ambassador). For e.g.
```
---
apiVersion: ambassador/v1
kind: Module
name: ambassador
config:
  service_port: 4567
  max_request-headers_kb: 90
```

## Todos
- [X] Tests
- [X] Documentation


kflynn:
Hey, thanks for this! A couple of questions:

First, this looks like it'll enforce a 60KB limit on headers globally for all Ambassador installations, which is a behavioral change that's probably not desirable. If the user doesn't specify a size, I think we shouldn't apply any limit.

Second, this is definitely one that needs a test -- maybe set the limit to 1KB, then send a request through with longer headers than that. Should the request fail? or do the headers get truncated? or... what?

Thanks again!

Issue #1767, opened 2019-08-22
Circuit breakers need to be part of cluster resolution


kflynn:
Right now, if you create a global circuit breaker, then try to override that circuit breaker in a specific `Mapping`, the override will affect any other `Mapping` that resolves to the same cluster (because Envoy circuit breakers are associated with clusters, not with routes). The problem is that the presence of the overriding circuit breaker isn't taken into account during cluster resolution.

Issue #1702, opened 2019-07-23
KAT should track and save Kube events when things aren't working


kflynn:
There's probably a lot of information we can learn by saving Kubernetes event traces during the requirements phase. We can grab this stuff and then push it to S3 for now.

PR #1770, opened 2019-08-23
Much test tweaking

kflynn:
If you do this commit-by-commit it will likely work out best. The first five commits:

| ID     | What |
|------|:------|
| `0b86579` | Kill some needless debugging. |
| `76b65b1` | Undo the three-phase test run; label selectors manage this correctly. |
| `9e8e37d` | I can't take the spewage from `kubectl port-forward` any more. |
| `b3a7ba0` | Pull Kubernetes logs for tests where a check fails. |
| `33e9f75` | Save logs for all kat phases. Push a tarball of artifacts to S3 on test failure, rather than trying to dump them all to stdout. |

are about the test infrastructure. They address #1702 among other improvements.

The remaining commits:

| ID     | What |
|------|:------|
| `f2230c2` | Add a requirement to wait for Consul to actually be running. |
| `265b9ac` | Make certain that app.diag is ready before we allow check_ready to succeed. |
| `d5e3d10` | Clean up non-global CircuitBreakingTest, start diving into GlobalCircuitBreakingTest. |
| `d399ace` | Take circuit breakers into account when reconciling cluster names. Update t_clustertest to match, and to avoid httpstat.us. |
| `00cc3b9` | Ditch dead code. |

are about improving specific tests:

- they address #1767, so that multiple `Mapping`s to the same upstream with different circuit breaker settings do the right thing;
- they fix a race where an Ambassador pod could report ready before `diagd` could answer on the diag service;
- they fix a flake in the Consul test (Consul can take quite awhile to start); and
- they scrap a couple of lines we don't need any more.

We are still in a situation where Travis has nasty issues with Ambassador's CI; that's for the future. But this makes some things better.

### DIFF
--- a/kat/kat/harness.py
+++ b/kat/kat/harness.py
@@ -339,6 +339,33 @@ class Node(ABC):
     def requirements(self):
         yield from ()
 
+    # log_kube_artifacts writes various logs about our underlying Kubernetes objects to
+    # a place where the artifact publisher can find them. See run-tests.sh.
+    def log_kube_artifacts(self):
+        if not getattr(self, 'already_logged', False):
+            self.already_logged = True
+
+            print(f'logging kube artifacts for {self.path.k8s}')
+            sys.stdout.flush()
+
+            DEV = os.environ.get("AMBASSADOR_DEV", "0").lower() in ("1", "yes", "true")
+
+            log_path = f'/tmp/kat-logs-{self.path.k8s}'
+
+            if DEV:
+                os.system(f'docker logs {self.path.k8s} >{log_path} 2>&1')
+            else:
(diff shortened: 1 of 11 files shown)


kat/kat/harness.py
```diff
@@ -339,6 +339,33 @@ def matches(self, pattern):
     def requirements(self):
         yield from ()
 
+    # log_kube_artifacts writes various logs about our underlying Kubernetes objects to
+    # a place where the artifact publisher can find them. See run-tests.sh.
+    def log_kube_artifacts(self):
```
containscafeine: This is awesome, thanks! :heart:

PR #1772, opened 2019-08-23
Fix schema errors for Knative 0.8.0

kflynn:
Also toss in a five-second sleep after the requirements pass, to see if that has any effect in CI.

### DIFF
--- a/kat/kat/harness.py
+++ b/kat/kat/harness.py
@@ -1097,6 +1097,9 @@ class Runner:
 
         self._wait(selected)
 
+        print("Waiting 5s after requirements, just because...")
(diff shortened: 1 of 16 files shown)


kat/kat/harness.py
```diff
@@ -1097,6 +1097,9 @@ def _setup_k8s(self, selected):
 
         self._wait(selected)
 
+        print("Waiting 5s after requirements, just because...")
```
LukeShu: "just because" :rofl:

kflynn: We'll see if it seems to make anything better over time. 😂

PR #1788, opened 2019-09-03
Simple tweaks to clean up some annoyances that Abhay noticed

kflynn:
Nothing earth-shattering here, just make it a little smoother to work on things:

- you can run `venv/bin/pytest --collect-only` after a `make` but without `make shell`;
- `kubestatus` won't clutter `git status`
- the log collector will work on test failures even though `teleproxy` isn't running.

### DIFF
--- a/ambassador/tests/test_docker.py
+++ b/ambassador/tests/test_docker.py
@@ -1,10 +1,12 @@
-import sys
 import os
+
+import pytest
 
 import pexpect
 import requests
 
-DockerImage = os.environ["AMBASSADOR_DOCKER_IMAGE"]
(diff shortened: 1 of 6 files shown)


ambassador/tests/test_docker.py
```diff
@@ -51,6 +53,9 @@ def check_http(logfile) -> bool:
         return False
 
 def test_docker():
+    if not DockerImage:
+        pytest.fail('no DockerImage')
```
ark3: This failure message will make me `grep`, which will lead me to what I need to do (set the env var). But can we make the message more explicit so I can skip the search and reading?

kflynn: There's actually a better error message below. 🙄  Removed this one.

Issue #1779, opened 2019-08-27
`make docker-push` "succeeds" if it fails to talk to the registry


LukeShu:
**Describe the bug**

`make docker-push` exits with a `0` exit-code even if `docker push` fails to push the images.

**To Reproduce**
1. `DOCKER_REGISTRY=bogus.com:1234 make docker-push`
2. Observe a single line on stderr, then Make exiting successfully:

```console
$ DOCKER_REGISTRY=localhost:31000 make docker-push
Successfully tagged localhost:31000/ambassador:lukeshu-backtrace-04cc91fa-dirty
PUSH localhost:31000/ambassador:lukeshu-backtrace-04cc91fa-dirty
.Get http://localhost:31000/v2/: dial tcp [::1]:31000: connect: connection refused

$ echo $?
0
````

**Expected behavior**

I expected `make` to exit with a non-zero exit code, and for any targets that depend on `docker-push` to not be built.

**Versions (please complete the following information):**
 - Ambassador: 0.76.0 (3505747057de6dfff223812e7a77ad9dbb7cf5fd)

flands:
To be fair, there is little error handling in the Makefile today. Simple workaround is to just run `docker push` manually. In any case, will put up a quick PR for this specific issue.

LukeShu:
Wow, I didn't expect anyone other than me to work on this issue :)

I suspect all that's needed is to say `SHELL = bash -o pipefail`, but it's possible that this causes other issues.

PR #1791, opened 2019-09-05
Ensure docker-push targets fails on error

flands:
Ensure docker-push targets fails on error

## Description
Ensure docker-push targets fails on error

## Related Issues
#1779 

## Testing
manual tests and deployed in production


kflynn:
Come to think of it, shouldn't `set -o pipefail` be enough? for both pushes?

flands:
Learn something everyday! PR updated.

PR #1794, opened 2019-09-05
Make 128-bits traceids the default

Pluies:
## Description

Make traceids 128bit option the default

Several reasons for this change:

- The [b3 propagation spec][1] from OpenZipkin defines the traceid as _either_ 64 or 128 bits, so both values should be interchangeable
- 128-bit traceid are the [default in Istio 1.1+][2]
- The draft [W3C TraceContext spec][3] standardises on 128-bit; 64 bits is not supported

The OpenCensus libraries (e.g. [for nodejs][4]) use the TraceContext spec, so this change makes Ambassador compatible with them as well.

[1]: https://github.com/openzipkin/b3-propagation
[2]: https://github.com/istio/istio/issues/10732
[3]: https://w3c.github.io/trace-context/
[4]: https://github.com/census-instrumentation/opencensus-node/blob/v0.0.17/packages/opencensus-propagation-tracecontext/src/validators.ts#L67-L75

## Related Issues

n/a

## Testing

The default tracing test now checks that the generated traceids are 128 bits, and if the 128 option is set to false, that they are 64 bits.

## Todos
- [x] Tests
- [x] Documentation

## Other

### DIFF
--- a/ambassador/ambassador/envoy/v2/v2tracing.py
+++ b/ambassador/ambassador/envoy/v2/v2tracing.py
@@ -41,6 +41,9 @@ class V2Tracing(dict):
             # The collector_endpoint is mandatory now.
             if not driver_config.get('collector_endpoint'):
                 driver_config['collector_endpoint'] = '/api/v1/spans'
+            # Make 128-bit traceid the default
+            if not 'trace_id_128bit' in driver_config:
+                driver_config['trace_id_128bit'] = True
 
         self['http'] = {
             "name": name,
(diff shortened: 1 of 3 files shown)


kflynn:
Hmmmmm. I don't think this'll work -- `trace.pb.go` is a generated file, IIRC. @LukeShu, can you confirm?

Feels like what you might want is to change _Ambassador's_ default, which should be in the `TracingService` module, and all but certainly warrants a test to make sure that switching it and such works.

Pluies:
@kflynn that makes sense, thanks for the quick feedback - I'll build this locally and add a test case. :)

Pluies:
@kflynn there we go - you were right, my original commit did nothing.

Here is a much better version, with tests, and updated formatting in the commit message.

Please let me know if there's anything else I can do to help move this forward 👍

containscafeine:
Thanks for the contribution :)

PR #1796, opened 2019-09-05
Switch to version tags with a leading 'v'

kflynn:
This way, Go code using the protobuf code we generate can reference `v0.77.0` or the like, instead of being forced to use a commit hash.

(Yes, Go is silly.)

Note that the version numbers in the diag UI will stay the same (e.g. `0.77.0` instead of `v0.77.0`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,9 +69,9 @@ IS_PRIVATE ?= $(findstring private,$(_git_remote_urls))
 # Note that for everything except RC builds, VERSION will be set to the version
 # we'd use for a GA build. This is by design.
 ifneq ($(GIT_TAG_SANITIZED),)
-VERSION = $(shell printf "$(GIT_TAG_SANITIZED)" | sed -e 's/-.*//g')
+VERSION = $(shell printf "$(GIT_TAG_SANITIZED)" | sed -e 's/-.*//g' | sed -e 's/^[vV]//')
 else
-VERSION = $(GIT_VERSION)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -69,9 +69,9 @@ IS_PRIVATE ?= $(findstring private,$(_git_remote_urls))
 # Note that for everything except RC builds, VERSION will be set to the version
 # we'd use for a GA build. This is by design.
 ifneq ($(GIT_TAG_SANITIZED),)
-VERSION = $(shell printf "$(GIT_TAG_SANITIZED)" | sed -e 's/-.*//g')
+VERSION = $(shell printf "$(GIT_TAG_SANITIZED)" | sed -e 's/-.*//g' | sed -e 's/^[vV]//')
 else
```
LukeShu: I'd have `$(patsubst v%,%, …)`.

LukeShu: Actually, `VERSION = $(patsubst v%,%,$(firstword $(subst -, ,$(GIT_TAG_SANITIZED)))`

Makefile
```diff
@@ -69,9 +69,9 @@ IS_PRIVATE ?= $(findstring private,$(_git_remote_urls))
...
-VERSION = $(GIT_VERSION)
+VERSION = $(shell printf "$(GIT_VERSION)" | sed -e 's/^[vV]//')
```
LukeShu: Likewise.

Makefile
kflynn: I'm going to land this as-is and then make the change after 0.77.0.

PR #1800, opened 2019-09-08
Add AMBASSADOR_WATCH_INTERVAL_MS to configure the minimum time between snapshots

esmet:
## Description
Add the ability to configure the minimum time between pushing snapshots. Note that this also includes the time it takes to process a snapshot, so it must be significantly larger than that in order to meaningfully reduce the work spent processing a large stream of updates.

## Related Issues
List related issues.

## Testing
Only tested by hand so far

## Todos
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/python/entrypoint.sh
+++ b/python/entrypoint.sh
@@ -381,12 +381,12 @@ if [[ -z "${AMBASSADOR_NO_KUBEWATCH}" ]]; then
 
     AMBASSADOR_FIELD_SELECTOR_ARG=""
     if [ -n "$AMBASSADOR_FIELD_SELECTOR" ] ; then
-	    AMBASSADOR_FIELD_SELECTOR_ARG="--fields $AMBASSADOR_FIELD_SELECTOR"
+        AMBASSADOR_FIELD_SELECTOR_ARG="--fields $AMBASSADOR_FIELD_SELECTOR"
     fi
 
     AMBASSADOR_LABEL_SELECTOR_ARG=""
     if [ -n "$AMBASSADOR_LABEL_SELECTOR" ] ; then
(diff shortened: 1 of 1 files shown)


ambassador/entrypoint.sh
```diff
@@ -392,6 +392,7 @@ if [[ -z "${AMBASSADOR_NO_KUBEWATCH}" ]]; then
            ${KUBEWATCH_SYNC_KINDS} \
            ${AMBASSADOR_FIELD_SELECTOR_ARG} \
            ${AMBASSADOR_LABEL_SELECTOR_ARG} \
+           --interval ${AMBASSADOR_WATCH_INTERVAL_MS:-"250"}ms \
```
esmet: 250 is the default in `watt` so I left it here.

SecurityInsanity: For watt it's worth, you don't need to quote the 250 here.

esmet: Is that a "watt" pun? If so, excellent.

esmet: Rebased, and removed quotes. Let me know if you have any other comments.

python/entrypoint.sh
```diff
@@ -396,6 +396,7 @@ if [[ -z "${AMBASSADOR_NO_KUBEWATCH}" ]]; then
            ${KUBEWATCH_SYNC_KINDS} \
            ${AMBASSADOR_FIELD_SELECTOR_ARG} \
            ${AMBASSADOR_LABEL_SELECTOR_ARG} \
+           --interval ${AMBASSADOR_WATCH_INTERVAL_MS:-250}ms \
```
SecurityInsanity: Taking a deeper look at this I don't see any tests for watt, that test this interval does what it's supposed too. If we're going to add it in this entrypoint, we should really test this.

PR #1802, opened 2019-09-08
Reject mapping groups that have no non-shadow mappings

esmet:
## Description
Ambassador currently generates an invalid envoy config if a shadow mapping has no associated non-shadow mapping.

## Related Issues
https://github.com/datawire/ambassador/issues/1493

## Testing
Tested by hand. Needs a real test.

## Todos
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/ambassador/ambassador/ir/irhttpmappinggroup.py
+++ b/ambassador/ambassador/ir/irhttpmappinggroup.py
@@ -208,14 +208,22 @@ class IRHTTPMappingGroup (IRBaseMappingGroup):
         # ...and return it. Done.
         return stored
 
-    def finalize(self, ir: 'IR', aconf: Config) -> List[IRCluster]:
+
+    def finalize(self, ir: 'IR', aconf: Config) -> bool:
+        if len(self.mappings) == 0:
+            errstr = "rejecting mapping group with no non-shadow mappings"
+            err = RichStatus.fromError(errstr)
+            aconf.post_error(err, self)
+            self.ir.logger.debug(errstr)
+            return False
+
         """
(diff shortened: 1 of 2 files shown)


ambassador/ambassador/ir/irmappingfactory.py
```diff
@@ -36,5 +36,12 @@ def finalize(cls, ir: 'IR', aconf: Config) -> None:
         # OK. We've created whatever IRMappings we need. Time to create the clusters
         # they need.
 
+        to_delete = []
         for group in ir.groups.values():
-            group.finalize(ir, aconf)
```
esmet: this code is just the most straightforward solution. feedback welcome on the way it should be done.

ambassador/ambassador/ir/irhttpmappinggroup.py
```diff
@@ -208,14 +208,22 @@ def add_cluster_for_mapping(ir: 'IR', aconf: Config, mapping: IRBaseMapping,
...
+            self.ir.logger.debug(errstr)
+            return False
+
         """
         Finalize a MappingGroup based on the attributes of its Mappings. Core elements get lifted into
         the Group so we can more easily build Envoy routes; host-redirect and shadow get handled, etc.
 
         :param ir: the IR we're working from
         :param aconf: the Config we're working from
-        :return: a list of the IRClusters this Group uses
```
esmet: it looked like the caller wasn't using this return value, so I felt OK changing it

ambassador/ambassador/ir/irmappingfactory.py
kflynn: Pretty sure I've done similar things elsewhere, actually. Looks sane.

PR #1816, opened 2019-09-12
Allow envoy's access log file location to be configurable

suever:
## Description

As Ambassador currently configures it, Envoy will log all requests to STDOUT. This PR introduces a `envoy_log_path` configuration option that allows the user to specify any filepath that Envoy will use for this logging. 

Potential use cases:

- Logging to a mounted `volume`
- Logging to `/dev/null` to suppress all output

## Related Issues
https://github.com/datawire/ambassador/issues/1819

## Testing

Not sure the best testing that needs to be done here. We use this in production and any issues would ultimately be issues with Envoy rather than Ambassador itself. Let me know if there is something specific that we need to test.

## Todos
- [ ] Tests
- [x] Documentation


suever:
> A good way to test this is probably to have a test that sets it to `/tmp/ambassador/ambassador.log`, then use `kubectl exec $pod cat /tmp/ambassador/ambassador.log` at the end of the test to check its contents. The `check` method of a `kat` test could do this.

I have added a test like you mentioned [here](https://github.com/datawire/ambassador/pull/1816/commits/98ac059a3c9fc6bd6a8c23c4a65ba89504fe8e4f). Feel free to make any suggestions to the test. I contemplated adding a `Query` to a specific endpoint and then check the access logs for that unique endpoint, but settled on checking for the ambassador healthchecks. Let me know if you prefer otherwise.

PR #1818, opened 2019-09-12
Allow users to specify filters for Envoy to use for access logs

suever:
## Description
Envoy has a feature-rich [access log filter](https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/accesslog/v2/accesslog.proto#envoy-api-msg-config-filter-accesslog-v2-accesslogfilter) specification that allows users to display the most important access logs only. This PR allows users to specify these access log filters via the ambassador configuration YAML.

For example, we are using this in production to log anything with a 5xx status code or < 200 status code:

```
    getambassador.io/config: |
      ---
      apiVersion: ambassador/v1
      kind: Module
      name: ambassador
      ambassador_id: primary-ambassador
      config:
        # Filters for access logging. Spec is found here:
        # https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/accesslog/v2/accesslog.proto#envoy-api-msg-config-filter-accesslog-v2-accesslogfilter
        envoy_log_filter:
          or_filter:
            filters:
            - status_code_filter:
                comparison:
                  op: LE
                  value:
                    default_value: 199
                    runtime_key: 'null'
            - status_code_filter:
                comparison:
                  op: GE
                  value:
                    default_value: 500
                    runtime_key: 'null'
```

By default, there are no filters applied and all requests are logged (the current behavior).

With respect to documentation, this is a fairly advanced topic and most of the information for the filter spec is within Envoy's documentation so in this PR the documentation change only links out to that for users to lookup available configuration of the filter.

## Related Issues
https://github.com/datawire/ambassador/issues/1819

## Testing
This PR does not front-end validation and instead blindly passes the provided configuration through to Envoy so not sure how much testing we need in Ambassador itself unless we want to have a tighter integration with validation within Ambassador.

## Todos
- [ ] Tests
- [x] Documentation


SecurityInsanity:
Just to leave a comment for those not in the ambassador open source slack channel about the update of this PR.  We are very uncomfortable passing in raw envoy through ambassador, and envoys config for this is _quite verbose_. As such the current status is: "figuring out a better configuration format" (perhaps something with infix expressions?).

PR #1827, opened 2019-09-13
Propagate KAT_ variables to tests

containscafeine:
Recently introduced KAT_CLIENT_DOCKER_IMAGE and
KAT_SERVER_DOCKER_IMAGE environment variables are not propagated
correctly leading to failures when running tests locally.

This commit fixes that so `make test` does not fail.


LukeShu:
Ahh, I had been doing `eval "$(make export-vars)"` (like CI does), so they were getting in to my environment that way.

PR #1832, opened 2019-09-16
Allow TLS context in RateLimitService

ppeble:
## Description

Adds ability to set TLS context when defining `RateLimitService`. This is already available with `AuthService` and operates in the same way.

## Testing

Deployed and tested manually local setup with `istio` using strict `mTLS` configuration.

Added automated testing for v1 format and for TLS.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -422,10 +422,10 @@ which will build the test services, then push them to `$(DOCKER_REGISTRY)`.
 **Using locally-built tests:**
 
 To use a single test service that you've built locally, set the environment variable
-`TEST_SVC_$svc` to point to the image you've just built and pushed, e.g.
+`TEST_SERVICE_$svc` to point to the image you've just built and pushed, e.g.
 
 ```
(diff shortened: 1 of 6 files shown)


ppeble:
> Changes look great, thank you! I think we need a `RateLimitTLSTest` as well, that's just like the `RateLimitV1Test` but actually _uses_ the new `tls` element.

I agree in principle but I couldn't figure out how to do this, to be entirely honest. I originally had a test that set the `tls` property but I couldn't figure out how to actually check that TLS was used correctly. For the existing `AuthService` tests you can check the `extauth` headers to make sure it was set correctly but I don't see anything similar for `RateLimitService`.

Would you be able to point me towards how I could do that? I'd be happy to take another crack at it if you were able to provide me a handhold.

ppeble:
> Looks good with one last thing -- the `RatelimitV0Test` and the `RatelimitV1Test` probably shouldn't share the `rate-limit` service either. Maybe `rate-limit-v0` and `rate-limit-v1` or something?
> 
> (I feel like the test platform should detect this and throw an error -- will have to look into that.)

Sure, not a problem. Let me make that change and rerun.

PR #1834, opened 2019-09-17
Bump to Envoy commit 18b1f5acc8d75e992f81d540bbb0d05f8abfe244

kflynn:
That's after Envoy v1.11.1 on the v1.12 path, after the fix for [Envoy issue 8025](https://github.com/envoyproxy/envoy/issues/8025).

Note that this _disables_ `retry_policy` in an `AuthService`. This was added in [Ambassador PR 1632]( https://github.com/datawire/ambassador/pull/1632/files), but as it happens, Envoy didn't actually support `retry_policy` -- it just wasn't generating an error, because Envoy's default behavior was to ignore unknown fields.

That default was flipped in Envoy commit 0418a855d9f9e37ec70b4c6d1942688fc8bb5751, so now, absent the `--allow-unknown-static-fields` command-line argument, Envoy will reject `retry_policy` and other unknown fields. 

`status_on_failure` and `failure_mode_allow` in an `AuthService` are still supported, and in many use cases they're all that's needed. If `retry_policy` is needed, it'll need to be set on a `Mapping`.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -285,9 +285,32 @@ That's it! Now simply run `make clean docker-push test` for the first time. In t
 Updating Ambassador's Envoy
 ---------------------------
 
-Ambassador currently relies on a custom Envoy build. This build lives in `https://github.com/datawire/envoy`, which is a fork of `https://github.com/envoyproxy/envoy`, and it'll need to be updated at least as often as Envoy releases happen. To do that:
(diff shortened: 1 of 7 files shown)


kflynn:
(I updated the PR description to explain how it worked. `git bisect` FTW! also, thanks to the Envoy folks for pushing a Docker image for every single commit...)

PR #1837, opened 2019-09-18
First very simple config-health reporting

kflynn:
Getting an Ambassador running can be tricky. This PR updates the diag UI to show you if Ambassador sees issues in your configuration that we expect will cause you problems:

- First, we always show you your `ambassador_id` and namespace information.
- We also always show some important environment variables, if they're set.
- We check your configuration for a few high-level things:
   - do you have any valid `Mapping`s?
   - do you have any valid contexts for TLS termination?
   - does your configuration show errors? (things like "oh, you should switch from a `v0` config to a `v1` config don't count)?
- If any of those checks fail, we show that in the diag UI.
   - alternately, you can set `debug_mode: true` in your `ambassador` `Module`, and we'll always show the results of the checks.

Lots of plans for this area in the near future; this is kind of just starting out.


ambassador/ambassador_diag/diagd.py
```diff
@@ -965,68 +1030,92 @@ def chime(self):
...
-            some_mappings = False
+                    if pfx:
+                        if not pfx.startswith('/ambassador/v0') or not name.startswith('internal_'):
+                            mapping_count += 1
 
-            for group in ir.groups.values():
-                if group and (group.location != '--internal--'):
-                    some_mappings = True
-                    break
+        if error_count:
```
containscafeine: Do we want this instead?

```suggestion
        if error_count > 0:
```

PR #1843, opened 2019-09-19
Return Response object in /metrics endpoint for Prometheus

jhsiaomei:
## Description
A `/metrics` endpoint was previously added to expose envoy stats for Prometheus without having to configure a sidecar (see: https://github.com/datawire/ambassador/pull/1591). However, this endpoint was returning a response that stripped all of the headers and formatting that envoy provided. This work updates the response to return the envoy stats in a way that Prometheus expects.

## Related Issues
https://github.com/datawire/ambassador/issues/1627

## Testing
We pushed the docker image with the fix to our internal ECR, then used the helm chart to deploy this test docker image instead of the official tagged version. Our DataDog instance was able to hit the `/metrics` endpoint and grab all of the stats successfully.


kflynn:
Looks good, let's land this!

PR #1828, opened 2019-09-13
Datadog apm

stefansedich:
## Description
Adding support for tracing using DataDog APM, this should add more options for using and visualizing traces with Ambassador.

There is still one outstanding issue here in that we should be logging the trace and span IDs as part of the Ambassador logs so that we can get full log correlation [for the Ambassador logs to traces](https://docs.datadoghq.com/tracing/advanced/connect_logs_and_traces/), I plan to tackle this as a seperate PR in the future as well as adding support for generic environment expansion as disucssed with @kflynn 

Note on documentation too, it is an early cut and am happy to expand with screenshots etc if required but I kept it light on purpose and linked to DataDog documentation where appropriate to reduce that burden on the Ambassador end, open for input.

## Related Issues
None

## Testing
This has been manually tested in my staging environment and the changes are fairly trivial. I was unsure how we would do any trivial e2e testing here.

It could be worth a unit test to test the env var expansion aspect but it is such a small surface area it did not feel hugely valuable. Happy to be coninced otherwise.

## Todos
- [ ] Tests (unsure if e2e is desired or trivial)
- [X] Documentation

PR #1846, opened 2019-09-19
Manifest environment variable expansion

stefansedich:
## Description
Adding environment variable expansion support for manifests (CRD/Anootations), it will allow you to for example expand $HOST_IP into manifests which is currently used to solve the issue for #1828 in a more generic way.

## Related Issues
#1828

## Testing
Have written an initial test, I plan to add one more to cover another scenario but as I cannot for the life of me get tests running locally opening a PR to have CI do it for me.

## Todos
- [X] Tests
- [ ] Documentation

### DIFF
--- a/ambassador/tests/t_lua_scripts.py
+++ b/ambassador/tests/t_lua_scripts.py
@@ -7,7 +7,11 @@ class LuaTest(AmbassadorTest):
 
     def init(self):
         self.target = HTTP()
-        self.env = ["LUA_SCRIPTS_ENABLED=Processed"]
+
+        self.manifest_envs = """
+    - name: LUA_SCRIPTS_ENABLED
+      value: Processed
+"""
 
     def manifests(self) -> str:
         return super().manifests() + self.format('''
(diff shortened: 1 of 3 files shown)


ambassador/tests/t_lua_scripts.py
```diff
@@ -20,7 +25,7 @@ def manifests(self) -> str:
   config:
     lua_scripts: |
       function envoy_on_response(response_handle)
-        response_handle: headers():add("Lua-Scripts-Enabled", "Processed")
+        response_handle: headers():add("Lua-Scripts-Enabled", "$LUA_SCRIPTS_ENABLED")
```
kflynn: How about swap one of these to be the `${}` form so we have coverage of that syntax too?

stefansedich: I actually had that originally but landed on a KeyError as format was picking it up, I guess I could pass the right args to format to not blow up, let me see.

stefansedich: Ok easy I made the other test use the ${} style, no need to mess with anything!

PR #1850, opened 2019-09-20
Make `make push-docs` also push to the local repo

kflynn:
It currently only pushes to the docs repo, and leaves a couple of commits lying unpushed locally.

I really don't know a good way to test this, sadly, until our next release...


LukeShu:
Not that I'm the one running `make push-docs`, but
```console
$ grep -A1 push .config/git/config 
[push]
	default = nothing
```
so, uh, a bare `git push` isn't gonna work on my laptop

kflynn:
Withdrawing, since I think we want to make this PR-centered and lock down pushes directly to `master`.

PR #1852, opened 2019-09-20
fixup envoy build

LukeShu:
## Description

This should fix the "slow" issues that macOS users (@kflynn) see and the "doesn't work" issues that minikube-dockerd users (@containscafeine) see.

Please provide feedback.

As a frame of reference, a no-op everything-is-already-up-to-date `make -j1 envoy-bin/envoy-static-stripped` takes 6 seconds for me if `make docker-base-images` hasn't been done, and about 4 seconds if it has.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,3 +6,4 @@
 !watt
 !kubestatus
 !envoy-bin/
+envoy-bin/*-static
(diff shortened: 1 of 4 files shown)


LukeShu:
@rhs I use your rsync trickery

stefansedich:
@LukeShu awesome this fixes the issues I had yesterday with remote dockerd in minikube!

PR #1859, opened 2019-09-25
Rework test service support

kflynn:
- Support building and releasing the test services from the Makefile. 
- Push them to `quay.io` instead of the `dwflynn` repo *cough*.
- Allow overriding them from the the environment when running the tests.


Makefile
```diff
@@ -122,6 +122,41 @@ ENVOY_FILE ?= envoy-bin/envoy-static-stripped
...
+docker.tag.release = $(AMBASSADOR_DOCKER_TAG)
+docker.tag.local = $(AMBASSADOR_DOCKER_TAG)
+
+TEST_SERVICE_VERSION = 0.0.2
+
+# ...then set overrides for the test services.
+test-%.docker.tag.release: docker.tag.release = quay.io/datawire/test_services:$(notdir $*)-$(TEST_SERVICE_VERSION)
+
+LOCAL_REPO = $(if $(filter-out -,$(DOCKER_REGISTRY)),$(DOCKER_REGISTRY)/test_services,test_services)
+test-%.docker.tag.local: docker.tag.local = $(LOCAL_REPO):$(notdir $*)-$(GIT_DESCRIPTION)
```
LukeShu: Is `.tag.local` used for things other than the test services? Why not just write
```Makefile
docker.tag.local = $(LOCAL_REPO):$(notdir $*)-$(GIT_DESCRIPTION)
```
instead of doing an override?

Makefile
```diff
@@ -416,6 +452,23 @@ base-go.docker: Dockerfile.base-go $(var.)BASE_GO_IMAGE $(WRITE_IFCHANGED)
...
+	$(MOVE_IFCHANGED) $@.tmp $@
+
+test-services: $(TEST_SERVICE_IMAGES) $(TEST_SERVICE_LOCAL_TAGS) $(TEST_SERVICE_LOCAL_PUSHES) FORCE
+test-services-release: $(TEST_SERVICE_IMAGES) $(TEST_SERVICE_RELEASE_TAGS) $(TEST_SERVICE_RELEASE_PUSHES) FORCE
+
+# XXX: Why doesn't just test-%.docker.push: test-%.docker.push.local work??
+#
+# This three-element form of $(addsuffix ...) is kind of an implicit foreach. 
+# We're generating a rule for each word in $(TEST_SERVICE_IMAGES).
+$(addsuffix .push,$(TEST_SERVICE_IMAGES)): %.push: %.push.local
```
LukeShu: Add a
```Makefile
.PHONY: $(addsuffix .push,$(TEST_SERVICE_IMAGES))
```

Makefile
```diff
@@ -416,6 +452,23 @@ base-go.docker: Dockerfile.base-go $(var.)BASE_GO_IMAGE $(WRITE_IFCHANGED)
...
+test-%.docker: test-services/%/Dockerfile $(MOVE_IFCHANGED) FORCE
+	docker build --quiet --iidfile=$@.tmp test-services/$*
+	$(MOVE_IFCHANGED) $@.tmp $@
+
+test-auth-tls.docker: test-services/auth/Dockerfile $(MOVE_IFCHANGED) FORCE
+	docker build --quiet --build-arg TLS=--tls --iidfile=$@.tmp test-services/auth
+	$(MOVE_IFCHANGED) $@.tmp $@
+
+test-services: $(TEST_SERVICE_IMAGES) $(TEST_SERVICE_LOCAL_TAGS) $(TEST_SERVICE_LOCAL_PUSHES) FORCE
+test-services-release: $(TEST_SERVICE_IMAGES) $(TEST_SERVICE_RELEASE_TAGS) $(TEST_SERVICE_RELEASE_PUSHES) FORCE
```
LukeShu: Add a
```Makefile
.PHONY: test-services test-services-release
```

ambassador/tests/t_circuitbreaker.py
```diff
@@ -61,8 +61,9 @@ def manifests(self) -> str:
 """
 
         return self.format(RBAC_CLUSTER_SCOPE + AMBASSADOR, image=os.environ["AMBASSADOR_DOCKER_IMAGE"],
-                           envs=envs, extra_ports="", capabilities_block="") + STATSD_MANIFEST.format(name='cbstatsd-sink',
-                                                                                                      target=self.__class__.TARGET_CLUSTER)
+                           envs=envs, extra_ports="", capabilities_block="") + \
+               STATSD_MANIFEST.format(name='cbstatsd-sink', image=self.test_image['stats'],
+                                      target=self.__class__.TARGET_CLUSTER)
```
LukeShu: Is `image=self.test_image['stats']` & `{image}` clearer than `self=self` & `{self.test_image[stats]}` that's used elsewhere?  Consistency is nice.

kat/kat/harness.py
```diff
@@ -26,6 +26,26 @@
...
+
+            image = os.environ.get(f'TEST_SVC_{key}', f'{default_registry}:test-{svc}-{default_version}')
+
+            self.images[svc] = image
+
+    def __getitem__(self, key: str) -> str:
+        return self.images[key]
+
+
+GLOBAL_TEST_IMAGE = TestImage()
```
LukeShu: Is this better than just returning the `images` dict?  Having a class that implements `__getitem__` as just wrapping a dict feels funny to me (but I'm not a pythonista).

LukeShu: hmm, I wish I could tel GitHub to include the entire `TestImage` class as context for that comment :/

Makefile
kflynn: I _think_ it's going to get used for other things shortly.

ambassador/tests/t_circuitbreaker.py
kflynn: Right now, there are separate `manifest`s for the statsd things that are handled totally differently from way most `manifest`s are handled. That whole thing needs to be untangled (almost certainly by ripping it out and replacing it with the standard mechanism) but I don't want to lump that in here. 😐

kat/kat/harness.py
kflynn: Hmmmm... so basically what's going on here is that, for whatever reason, I'm more comfortable having a singleton class instance as a global than I am with inlining the code in `TestImage.__init__` at global scope and having it create a global dict. But I didn't want to create a new `TestImage` for every `Node` because there can be a lot of `Node`s...

It's not exactly a slam-dunk either way. 😐

Makefile
kflynn: Done.

Makefile
kflynn: Switched from `FORCE` to `.PHONY`.

Issue #1854, opened 2019-09-21
No one notices when `go/apis/envoy/` is stale


LukeShu:
**Describe the bug**

When we upgrade Envoy, we need to (should?) update `go/apis/envoy/`.  `make docker-update-base` should do this.  It didn't happen (did it error and we didn't care? did we just run other things and bypass the "main" entry point?).

It seems that trying to re-generate `go/apis/envoy/` fails right now (it looks like we need to fiddle with the Protobuf mappings), and yeah, that's a problem, but is pretty minor.  To me, the bigger problem is that we didn't notice.  If we'd said "it fails to re-generate right now, but we'd like to ship a release, so we're not worrying about that for now" and made the conscious decision to ignore it, that would be fine by me; but we accidentally made that decision because no one noticed.

So I think the solution is to have CI do a re-generate, and make sure that git says the tree is clean afterward.  Similar to how Pro does.

LukeShu:
notstale

PR #1866, opened 2019-09-27
Update go/apis/envoy

LukeShu:
## Description

This fixes our ability to update `go/apis/envoy/`, and updates it (#1854).  It does not do anything to address the problem that no one noticed that it wasn't updated, which is what #1854 is really about, so I'm not having it close that issue.

Once we update our `ENVOY_COMMIT` to include envoyproxy/envoy@32e6018f10f9d6de58fd6c3d6844a68a5be7723b the mechanism by which this happens will change, and it looks like for the better.  So um, I wouldn't spend too much time worrying about the "I have no idea why" comment.


Makefile
```diff
@@ -739,28 +741,46 @@ venv/bin/protoc-gen-validate: go.mod $(FLOCK) | venv/bin/activate
 
 # Search path for .proto files
 gomoddir = $(shell $(FLOCK) go.mod go list $1/... >/dev/null 2>/dev/null; $(FLOCK) go.mod go list -m -f='{{.Dir}}' $1)
-# This list is based 'imports=()' in https://github.com/envoyproxy/go-control-plane/blob/master/build/generate_protos.sh
+# This list is based 'imports=()' in https://github.com/envoyproxy/go-control-plane/blob/0e75602d5e36e96eafbe053999c0569edec9fe07/build/generate_protos.sh
```
kflynn: "This list is based on" maybe?

Makefile
```diff
@@ -739,28 +741,46 @@ venv/bin/protoc-gen-validate: go.mod $(FLOCK) | venv/bin/activate
...
-imports += $(call gomoddir,istio.io/gogo-genproto)
+imports += $(call gomoddir,github.com/googleapis/googleapis)
 imports += $(call gomoddir,istio.io/gogo-genproto)/prometheus
+imports += $(call gomoddir,istio.io/gogo-genproto)
 
 # Map from .proto files to Go package names
-# This list is based 'mappings=()' in https://github.com/envoyproxy/go-control-plane/blob/master/build/generate_protos.sh
-mappings += gogoproto/gogo.proto=github.com/gogo/protobuf/gogoproto
-mappings += google/api/annotations.proto=github.com/gogo/googleapis/google/api
+# This list is based 'mappings=()' in https://github.com/envoyproxy/go-control-plane/blob/0e75602d5e36e96eafbe053999c0569edec9fe07/build/generate_protos.sh
```
kflynn: Another missing "on" ?

LukeShu:
Silly me assumed that just because `protoc` ran without error, that the Go it output would compile! (And didn't bother to check before pushing, shame on me.)

Should be fixed now. (Along with the grammatical errors.)

PR #1868, opened 2019-09-28
Propagate LB status to Ingress resources

### LINKED ISSUE #1851: Update Ingress status with LB external-ip to better support external-dns
**Please describe your use case / problem.**
With the new support for Ingress it should be possible to utilize extrenal-dns to automatically create the required CNAME for an ingresses host such that it maps to the Ambassador service.

If you view an ingress configured to use Ambassador as ingress, you will notice the address is not configured, nor does it ever update.

```
NAME   HOSTS                             ADDRESS   PORTS   AGE
test        test.dev.intacct-ai.com                         80            3m36s
```

**Describe the solution you'd like**
Once Ambassador picks up an Ingress object it should find the correct backing service and then use this to update the status of the ingress to reflect the external-ip of the service, this should enable external-dns to pick up the Ingress and configure things as expected.

**Describe alternatives you've considered**
Unsure if there is another way to achieve this, looking at the nginx and traefik controllers it appears to be how they are handling it.

**Additional context**
https://github.com/containous/traefik/blob/a20a6636b4582e4e7b562fc01f64567775fc8a9a/pkg/provider/kubernetes/crd/client.go#L295

containscafeine:
This commit propagates Ambassador service's status to Ingress
resources.

Fix #1851

### DIFF
--- a/ambassador/ambassador/config/resourcefetcher.py
+++ b/ambassador/ambassador/config/resourcefetcher.py
@@ -411,6 +411,11 @@ class ResourceFetcher:
 
         ambassador_id = annotations.get('getambassador.io/ambassador-id', 'default')
 
+        # We don't want to deal with non-matching Ambassador IDs
+        if ambassador_id != Config.ambassador_id:
+            self.logger.info(f"Ingress {ingress_name} does not have Ambassador ID {Config.ambassador_id}, ignoring...")
+            return None
+
         self.logger.info(f"Handling Ingress {ingress_name}...")
 
         ingress_tls = ingress_spec.get('tls', [])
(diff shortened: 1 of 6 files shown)


ambassador/ambassador/config/resourcefetcher.py
```diff
@@ -687,6 +694,15 @@ def handle_k8s_service(self, k8s_object: AnyDict) -> HandlerResult:
                 'namespace': resource_namespace,
                 'ports': ports
             }
+
+            selector = spec.get('selector', {})
+            for key, value in selector.items():
+                if key == 'service' and value == 'ambassador' and resource_name == 'ambassador':
```
containscafeine: @kflynn need to talk to you if you think this is the right way to look for Ambassador's service

kflynn: Uhhhh... I don't have anything better. Is there any way you could imagine allowing the user to force a particular service to be the right one?

stefansedich: Glossing over this one thing I can see is that my services are not all called `ambassador` for example I have `ambassador-private` and `ambassador-public` and I don't think this logic would locate the service properly.

I will get it fired up tonight after work and look in more detail, but that was my first observation.

containscafeine: @stefansedich you're correct - we ship the ambassador pod with the label `service: ambassador` and that's the selector that ambassador service looks for. But when we do the opposite and look for a service with the selector `service: ambassador` there is ambassador and ambassador-admin, which is why I tied it the name to `ambassador`. But clearly that's not ideal.

stefansedich: Does it currently handle multiple ambassador instances with an ambsassador_id set not running in a single namespace?

I have not ran this yet but to me it looks like it might store the last ambassador service it matches in `ambassador_service_raw` and then when processing Ingress they would all get linked to this one, however it might not be the ambassador instance it expects?

containscafeine: nope, Ambassador IDs are not handled yet.

stefansedich: @containscafeine I see a couple options here, we can do the following:

1. Label the service with a `getambassador.io/ambassador-id` if set
2. In `handle_k8s_service` if we see a service has the `app.kubernetes.io/name=ambassador` label set add it to a new dictionary, something like: `self.ambassador_services[ambassador_id] = x` where ambassador_id comes from either the label or we default to `default`.
3. In `handle_k8s_ingress` attempt to locate the correct service but looking at the `self.ambassador_services` dictionary for the correct instance using either the ambassador_id set in the ingress annotation or default.

Now the second idea after digging into Traefik quickly is to populate the service name into some configuration variable defining the service and just use this to the load the service.

https://github.com/containous/traefik/blob/2471f893e7d8b20cb9ee7051fbc97eaef1b023d7/docs/content/providers/kubernetes-ingress.md#ingressendpoint
https://github.com/helm/charts/blob/4a8d788986ee3700ef5eb4e178ef92fa5d605ca6/stable/traefik/templates/configmap.yaml#L180

containscafeine:
Test added in #1897

Eronarn:
Hi @kflynn, @containscafeine - I tried this out, as we're running into difficulties using Ambassador in combination with External DNS and wanted to see if this new support would facilitate an integration. Unfortunately, I think there's an error in the implementation, in this block of code: https://github.com/datawire/ambassador/commit/64ee83a1e9700fe76b3e7da7ab3477b8c078d1b5#diff-42a97a8bbfa211f2bd32e08b0d12c73aR510-R524

By my reading, `parsed_ambassador_annotations` will be non-`None` if the Ingress was annotated with `getambassador.io/config`, which will then cause an early `return`. This mirrors what I'm seeing on the cluster with `0.83.0`: the Ingresses create Mappings in Ambassador as expected, but there are no log lines related to status (from the new diff) at all.

Unfortunately, the provided test doesn't provide an annotation with that key, so it's not actually exercising this code path.

containscafeine:
@Eronarn thanks for bringing this up. You were spot on about the return statement blocking the ingress update. I've sent #1907 with a test to fix this.

Arttii:
I ran into an issue when deploying Eclipse Che, where it relies on the Ingress status field not being empty, but we are running on Baremetal with just a ClusterIP and hence the status remains empty. The nginx ingress controller has some support for [this](https://kubernetes.github.io/ingress-nginx/deploy/baremetal/) via `--report-node-internal-ip-address`. Is there a way to get something similar with Ambassador or I am doing something wrong?

PR #1870, opened 2019-09-30
fixed the knative manifests

rhs:
## Description
The tests bomb out if the cluster is recent enough to enable knative. These changes seem necessary in order to allow the tests to just run. They do not seem to pass yet, there are one or two intermittent failures in the knative tests.


kflynn:
Ah, crap, I broke that and didn't realize it because I was working with Kubernaut. We really need to update Kubernaut...

PR #1877, opened 2019-09-30
Schema fixes _and tests_

### LINKED ISSUE #1869: WARNING: corrupt schema at /ambassador/.cache/.../v1/AuthService.schema
**Describe the bug**

I'm getting this error on some pods after upgrading ambassador from 0.77 to 0.81 and they are going into a crash loop. It doesn't happen on all pods though.

I tried rolling back and the crash loop is happening on the 0.77 pods now as well. They don't have the schema error though, so maybe it's not the source of the crash loops.

```
2019/09/30 09:35:53 aggregator: watch hook stderr: 2019-09-30 09:35:52 watch-hook WARNING: corrupt schema at /ambassador/.cache/ambassador-0.0.0.dev0-py3.6.egg-tmp/schemas/v1/AuthService.schema, skipping (Expecting property name enclosed in double quotes: line 52 column 5 (char 1725))
2019/09/30 09:35:53 aggregator: watch hook stderr: 2019-09-30 09:35:52 watch-hook ERROR: ERROR ERROR ERROR Starting with configuration errors
```

The error is due to a dangling comma:
```
        "failure_mode_allow": { "type": "boolean" },
    },
```

I've compared the good vs bad version of the AuthService.schema on the pods and the bad one is missing a section about retry right after that:

```
--- AuthService.schema	2019-09-30 05:27:44.919391121 -0400
+++ AuthService.schema-good	2019-09-30 05:31:31.508702606 -0400
@@ -49,6 +49,18 @@
             }
         },
         "failure_mode_allow": { "type": "boolean" },
+        "retry_policy": {
+            "type": "object",
+            "properties": {
+                "retry_on":  {
+                    "type": "string",
+                    "enum": ["5xx", "gateway-error", "connect-failure", "retriable-4xx", "refused-stream", "retriable-status-codes"]
+                },
+                "num_retries": { "type": "integer" },
+                "per_try_timeout": { "type": "string" }
+            },
+            "additionalProperties": false
+        }
     },
     "required": [ "apiVersion", "kind", "name", "auth_service" ],
     "additionalProperties": false
```

 - Ambassador: 0.77.0, 0.81.0
 - Kubernetes environment: kops
 - Version: 1.13.8

kflynn:
Fix AuthService schema... and build in a test for all the schemas forevermore.

Fixes #1869.

### DIFF
--- a/ambassador/schemas/v1/AuthService.schema
+++ b/ambassador/schemas/v1/AuthService.schema
@@ -48,7 +48,7 @@
                 "code": { "type": "integer" }
             }
         },
-        "failure_mode_allow": { "type": "boolean" },
+        "failure_mode_allow": { "type": "boolean" }
     },
     "required": [ "apiVersion", "kind", "name", "auth_service" ],
     "additionalProperties": false
--- a/ambassador/tests/test_schemas.py
+++ b/ambassador/tests/test_schemas.py
@@ -1,0 +1,77 @@
+from typing import Optional
+
+import json
+import logging
+import os
+import sys
+
+import jsonschema
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s test %(levelname)s: %(message)s",
+    datefmt='%Y-%m-%d %H:%M:%S'
+)
+
+logger = logging.getLogger("ambassador")
+
+
+def test_schemas():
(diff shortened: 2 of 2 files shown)


kflynn:
The test does indeed catch both a schema with a JSON syntax error (which is #1869's case) and a schema that's syntactically-correct JSON, but not valid per JSONSchema's rules.

PR #1873, opened 2019-09-30
Run envoy tests in docker in CI

containscafeine:

PR #1878, opened 2019-10-01
Run envoy tests in CI

concaf:
Overrides #1873


Makefile
```diff
@@ -423,7 +425,7 @@ envoy-bin/envoy-static: $(ENVOY_BASH.deps) FORCE | envoy-bin
 
 check-envoy: $(ENVOY_BASH.deps)
 	$(call ENVOY_BASH.cmd, \
-	    docker exec --workdir=/root/envoy $$(cat envoy-build-container.txt) bazel test --verbose_failures -c dbg --test_env=ENVOY_IP_TEST_VERSIONS=v4only //test/...; \
+	    docker exec $$(cat envoy-build-container.txt) /bin/sh -c 'cd /root/envoy && bazel test --verbose_failures -c dbg --test_env=ENVOY_IP_TEST_VERSIONS=v4only //test/...'; \
```
containscafeine: Surprisingly, `--workdir` fails with certain versions of docker even with compatible APIs, hence this change.

Makefile
```diff
@@ -454,6 +456,13 @@ base-go.docker: Dockerfile.base-go $(var.)BASE_GO_IMAGE $(WRITE_IFCHANGED)
 	fi
 	@docker image inspect $(BASE_GO_IMAGE) --format='{{.Id}}' | $(WRITE_IFCHANGED) $@
 
+envoy-tests-image:
```
kflynn: Should this be recast to use `docker.mk`, now that we have it?

releng/travis-script.sh
```diff
@@ -76,6 +76,40 @@ esac
 set -o xtrace
 make print-vars
 
+case "$TRAVIS_EVENT_TYPE" in
```
kflynn: Any reason this isn't just `if [[ "$TRAVIS_EVENT_TYPE" == "cron" ]]` ?

Also, should the cronjob continue to run the build when it's a cronjob, or exit after the Envoy checks?

containscafeine: > Any reason this isn't just if [[ "$TRAVIS_EVENT_TYPE" == "cron" ]] ?

Well, just in case we want to add other events for `TRAVIS_EVENT_TYPE` later on, one of `push, pull_request, api, cron`.

> Also, should the cronjob continue to run the build when it's a cronjob, or exit after the Envoy checks?

> Also, should the cronjob continue to run the build when it's a cronjob, or exit after the Envoy checks?

Makes sense, I've made it exit after envoy checks.

Makefile
containscafeine: I don't fully understand how that works, but I've made a few changes to accommodate this, lmk :)

Dockerfile.envoy-tests
```diff
@@ -0,0 +1,12 @@
+FROM fedora:30
```
kflynn: Hmmmm. So this makes _three_ Linux distributions we're using while wrangling Ambassador... 😂

releng/travis-script.sh
kflynn: Where? I'm not seeing that. 😕

containscafeine: Sorry about that :\ I had made the changes to the main `make test` instead of here. But anyway, fixed it now. It should now exit after running Envoy tests.

releng/travis-script.sh
```diff
@@ -71,6 +71,47 @@ esac
...
+    cron)
+        printf "========\nRunning Envoy tests...\n"
+
+        # Delete VM if it already exists
+        gcloud beta compute instances delete envoy-tests-ambassador --zone us-east1-b --quiet || true
```
SecurityInsanity: What are our thoughts about running tests in the gcp environment from the local env?

---

I'm thinking something like:

1. Turn this into a makefile task (or a script makefile calls)
2. Make the gcp instance name be something like: `envoy-tests-ambassador-$(date +%s)`
3. Have the cleanup action list all running hosts, and check if they've been powered on for more than 24 hours. If so shutdown. (Or maybe using gcp's cloud scheduler to do it automatically?)

---

Do you think it'd be worthwhile to test this part of the gcloud script? Or too much work for not much benefit?

.travis.yml
```diff
@@ -52,3 +52,6 @@ script:
 after_script:
   # release the kubernaut claim
   - ./releng/travis-cleanup.sh
+
+before_install:
+  - openssl aes-256-cbc -K $encrypted_29dfbda8eb61_key -iv $encrypted_29dfbda8eb61_iv -in datawireio-d9aadf7d8d9f.json.enc -out datawireio-d9aadf7d8d9f.json -d
```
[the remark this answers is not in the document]
SecurityInsanity: Also I think we should mark the decrypted file in .gitignore so we don't accidentally commit it if someone is messing locally.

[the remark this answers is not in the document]
SecurityInsanity: FWIW @kflynn has confirmed, there is no `openssl chacha20` for OSX.

PR #1880, opened 2019-10-02
Code reorganization

inercia:
## Description

* Code reorganization, trying to follow the Standard Go Project Layout (https://github.com/golang-standards/project-layout), even when there is a lot of Python 
* Moved `/kat/*` to `/ambassador`
* Silence some mypy errors
* Moved `/multi/*` to `/ambassador`
* Moved /ambassador to `/python`
* Moved all the generated files for envoy to `/pkg/api/envoy`

## Testing

Not neccessary: our regular e2e tests should pass.


Makefile
```diff
@@ -493,37 +492,41 @@ docker-push-base-images:
...
+ambassador.docker: Dockerfile base-go.docker base-py.docker $(ENVOY_FILE) $(WATT) $(KUBESTATUS) $(WRITE_IFCHANGED) python/ambassador/VERSION.py FORCE
 	docker build --build-arg ENVOY_FILE=$(ENVOY_FILE) --build-arg BASE_GO_IMAGE=$(BASE_GO_IMAGE) --build-arg BASE_PY_IMAGE=$(BASE_PY_IMAGE) $(DOCKER_OPTS) -t $(AMBASSADOR_DOCKER_IMAGE) .
 	@docker image inspect $(AMBASSADOR_DOCKER_IMAGE) --format='{{.Id}}' | $(WRITE_IFCHANGED) $@
 
-kat-client-docker-image: kat-client.docker
-.PHONY: kat-client-docker-image
-kat-client.docker: kat-client-docker-image/Dockerfile base-py.docker kat-client-docker-image/teleproxy kat-client-docker-image/kat_client $(WRITE_IFCHANGED) $(var.)KAT_CLIENT_DOCKER_IMAGE
-	docker build --build-arg BASE_PY_IMAGE=$(BASE_PY_IMAGE) $(DOCKER_OPTS) -t $(KAT_CLIENT_DOCKER_IMAGE) kat-client-docker-image
+build/kat/client: kat-client.docker
+.PHONY: build/kat/client
```
LukeShu: This is a little weird that `build/THING` is being used both for real files and for `.PHONY` targets.

cmd/ambex/README.md
```diff
@@ -0,0 +1,157 @@
+Ambex: Ambassador Experimental ADS service
```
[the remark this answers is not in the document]
kflynn: Admittedly `ambex` doesn't have much history. 🙂  But it _would_ be nice to keep what it has.

That's not to say that I actually know how to do that, mind you. :wink:

Makefile
kflynn: What would you recommend? `build/kat/client.docker` everywhere instead of `build/kat/client`? Just using `kat-client` as a `.PHONY`? 

I can see how it's a little weird, I'm just not sure what's better here...

Makefile
```diff
@@ -748,56 +747,56 @@ venv/bin/protoc-gen-grpc-web: $(var.)GRPC_WEB_VERSION $(var.)GRPC_WEB_PLATFORM |
...
-pkg/apis/kat/echo.pb.go: kat/api/echo.proto venv/bin/protoc venv/bin/protoc-gen-gogofast
+pkg/apis/kat/echo.pb.go: ambassador/api/kat/echo.proto venv/bin/protoc venv/bin/protoc-gen-gogofast
 	./venv/bin/protoc \
-		--proto_path=$(CURDIR)/kat/api \
+		--proto_path=$(CURDIR)/ambassador/api/kat \
 		--plugin=$(CURDIR)/venv/bin/protoc-gen-gogofast --gogofast_out=plugins=grpc:$(@D) \
 		$(CURDIR)/$<
 
-kat/sandbox/grpc_web/echo_grpc_web_pb.js: kat/api/echo.proto venv/bin/protoc venv/bin/protoc-gen-grpc-web
+tools/sandbox/grpc_web/echo_grpc_web_pb.js: ambassador/api/kat/echo.proto venv/bin/protoc venv/bin/protoc-gen-grpc-web
```
kflynn: Hm. Do we really use the sandbox any more? I just realized I don't actually know what it's really for. :wink:

Makefile
```diff
@@ -810,7 +809,7 @@ venv/bin/ambassador: venv/bin/activate ambassador/requirements.txt
 	@releng/install-py.sh dev install ambassador/requirements.txt
 	@releng/fix_kube_client
 
-venv/bin/activate: dev-requirements.txt multi/requirements.txt kat/requirements.txt
+venv/bin/activate: dev-requirements.txt multi/requirements.txt
```
kflynn: Hm. So I see that you've moved the contents of `kat/requirements` into `ambassador/requirements`, and that's prompting me to think a bit more. 

I think that `kat` should actually probably stay out of the `ambassador` directory, because `ambassador` is the stuff we ship as part of the delivered image. Delivering `kat` to an end user, rather than a developer, is probably unnecessary.

ambassador/kat/harness.py
```diff
@@ -183,7 +183,7 @@ def sanitize(obj):
 
 
 def abstract_test(cls: type):
-    cls.abstract_test = True
+    cls.abstract_test = True  # type: ignore
```
kflynn: I just now realized that all of these errors are happening because `kat` is now under `ambassador`. D'oh! 🙂 

In general, we should clean up the hinting that's resulting in the errors, rather than just telling `mypy` to ignore them. That could happen in a separate PR right after this one, if necessary.

Dockerfile
```diff
@@ -48,10 +48,10 @@ RUN touch -t 197001010000 /usr/local/bin/envoy
 FROM $BASE_PY_IMAGE as base-py
 
 # Install the application itself
-COPY multi/ multi
 COPY ambassador/ ambassador
+RUN rm -rf ./multi
```
kflynn: Hmmm. Do we need this `rm -rf` at all now?

Makefile
```diff
@@ -845,69 +848,69 @@ venv/bin/protoc-gen-grpc-web: $(var.)GRPC_WEB_VERSION $(var.)GRPC_WEB_PLATFORM |
 	curl -o $@ -L --fail https://github.com/grpc/grpc-web/releases/download/$(GRPC_WEB_VERSION)/protoc-gen-grpc-web-$(GRPC_WEB_VERSION)-$(GRPC_WEB_PLATFORM)
 	chmod 755 $@
 
-go/apis/kat/echo.pb.go: kat-apis/echo.proto venv/bin/protoc venv/bin/protoc-gen-gogofast
+pkg/apis/kat/echo.pb.go: python/api/kat/echo.proto venv/bin/protoc venv/bin/protoc-gen-gogofast
```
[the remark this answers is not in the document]
inercia: I agree with you that all the Go code should be moved there, but we have decided not to move things to a top-level `/api` until we can move the `/go/apis/envoy/*` code too (and that should come for the 1.0 release). So if `echo.proto` is only used from Go, what about leaving it in `/go/apis/kat`?

Makefile
inercia: Yes, I forgot to remove the `.PHONY` targets. Would that be ok?

Dockerfile.base-go
```diff
@@ -37,10 +37,15 @@ RUN apk --no-cache add bash curl python3 jq
...
 RUN chmod 755 /usr/bin/kubectl
+
+WORKDIR /go/src/app
+COPY cmd ./cmd
+COPY pkg ./cmd
+
+RUN go get -d -v ./...
+RUN go install -v ./...
+
+
```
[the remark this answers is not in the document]
inercia: I will fix it, thanks.

.gitignore
```diff
@@ -20,11 +20,14 @@ teleproxy
...
 
-/kat-server-docker-image/kat-server
-/kat-sandbox/grpc_auth/docker-compose.yml
-/kat-sandbox/grpc_web/docker-compose.yaml
-/kat-sandbox/grpc_web/*_pb.js
-/kat-sandbox/http_auth/docker-compose.yml
+# binaries
+/cmd/ambex/ambex
+
+/build/kat/server/kat-server
```
kflynn: Should the KAT backend be in `cmd` as well?

Makefile
```diff
@@ -113,7 +113,7 @@ ENVOY_FILE ?= envoy-bin/envoy-static-stripped
   BASE_ENVOY_RELVER ?= 3
   # Increment BASE_GO_RELVER on changes to `Dockerfile.base-go`
   BASE_GO_RELVER    ?= 16
-  # Increment BASE_PY_RELVER on changes to `Dockerfile.base-py`, `releng/*`, `multi/requirements.txt`, `ambassador/requirements.txt`
+  # Increment BASE_PY_RELVER on changes to `Dockerfile.base-py`, `releng/*`, `python/requirements.txt`
   BASE_PY_RELVER    ?= 16
```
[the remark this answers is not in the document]
inercia: Ok, I will do that.

cmd/ambex/README.md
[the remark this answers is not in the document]
inercia: Sure, I can try to merge the Ambex history here.

Dockerfile.base-go
```diff
@@ -37,10 +37,15 @@ RUN apk --no-cache add bash curl python3 jq
...
 # Grab kubectl
 RUN curl -o /usr/bin/kubectl -L https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
 RUN chmod 755 /usr/bin/kubectl
+
+WORKDIR /go/src/app
+COPY cmd ./cmd
+COPY pkg ./cmd
+
+RUN go get -d -v ./...
+RUN go install -v ./...
```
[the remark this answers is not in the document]
inercia: Right, this code is from an _old_ PoC.... I 'll remove it.

Dockerfile
```diff
@@ -92,12 +92,12 @@ RUN chgrp -R 0 ${AMBASSADOR_ROOT} && \
     chmod -R g=u ${AMBASSADOR_ROOT} /etc/passwd
 
 # COPY the entrypoint and Python-kubewatch and make them runnable.
-COPY ambassador/entrypoint.sh .
-COPY ambassador/grab-snapshots.py .
-COPY ambassador/kick_ads.sh .
-COPY ambassador/kubewatch.py .
-COPY ambassador/post_update.py .
-COPY ambassador/watch_hook.py .
+COPY python/entrypoint.sh .
```
kflynn: The wholesale rename of `ambassador` to `python` produces a couple of really odd effects. This is one of thing: `entrypoint.sh` isn't Python code, so it seems strange that it would be in a directory named `python`. 

Another effect - and I think this is more of a real issue - is that we're almost certainly shipping more than we should be now. `kat` actually is not code that we need to ship to end users -- developers need it, but it doesn't need to be in the Docker images we ship. (Though having `mypy` type-check `kat` is a huge win...)

Dockerfile.base-go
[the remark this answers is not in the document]
kflynn: Hey! Are you calling me a space alien?? 🙂

python/kat/harness.py
```diff
@@ -203,7 +203,7 @@ def sanitize(obj):
 
 
 def abstract_test(cls: type):
-    cls.abstract_test = True
+    cls.abstract_test = True  # type: ignore
```
kflynn: Having `mypy` check these now is great! Of course that means we should actually fix the type errors rather than ignoring them. 🙂 

I'd be OK with that being a PR right after this one, if need be.

python/requirements.txt
```diff
@@ -9,3 +9,6 @@ scout.py==0.5.0
 semantic-version==2.6.0
 kubernetes==8.0.0
 click==7.0
+pytest
```
kflynn: Note that `pytest` and `pyyaml` are _development_ dependencies, but they should not be _runtime_ dependencies, because we shouldn't be shipping `kat` in the Docker image at all. So they should really be in `dev-requirements.txt` instead of here.

Dockerfile
inercia: I agree that `entrypoint.sh` should be in a difference place. We are following the [Go Standard Project](https://github.com/golang-standards/project-layout) layout. Maybe we should move it to `/build` or `/deployments` or /configs`...

inercia: Regarding `kat`, we decided that, even when this package is used only by developers, it didn' hurt to include it in the final image...

.gitignore
inercia: The `kat` code is located in `/cmd`, but the binary is left in `/build/kat/server/kat-server`. You can see the build rule:

`
build/kat/server/kat-server: $(wildcard cmd/kat-server/* cmd/kat-server/*/*) pkg/apis/kat/echo.pb.go
	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o $@ ./cmd/kat-server
`

ambassador/kat/harness.py
inercia: Yeah, I know, but I didn't have enough knowledge for fixing these warnings. I will try to do a proper fix in a subsequent PR...

go/apis/envoy/admin/v2alpha/clusters.pb.go
```diff
@@ -505,15 +505,15 @@ func (m *Clusters) MarshalToSizedBuffer(dAtA []byte) (int, error) {
 		for iNdEx := len(m.ClusterStatuses) - 1; iNdEx >= 0; iNdEx-- {
 			{
 				size, err := m.ClusterStatuses[iNdEx].MarshalToSizedBuffer(dAtA[:i])
-				if err != nil {
-					return 0, err
-				}
+			if err != nil {
+				return 0, err
+			}
```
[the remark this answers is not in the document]
inercia: Done. I have regenerated all those files.

cmd/ambex/README.md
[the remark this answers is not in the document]
inercia: Merged.

Dockerfile.base-go
```diff
@@ -24,7 +24,7 @@
 # Makefile, then run "make docker-update-base" to build and push the
 # new image.
 
-FROM frolvlad/alpine-glibc:alpine-3.9
+FROM golang:1.13-alpine
```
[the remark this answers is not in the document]
inercia: You are right, I have rolled back to the previous base image.

python/kat/harness.py
inercia: :+1:

Makefile
inercia: I don't know. I can remove it if it is not neccessary.

python/requirements.txt
inercia: Done.

inercia:
I have removed Ambex from this PR. We can try  to get the Ambex history in a different PR.

Makefile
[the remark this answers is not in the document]
inercia: As discussed in our meeting, I have moved all the envoy files to `/pkg/api/envoy`, and all the `proto` files to `/api/*`

Makefile
LukeShu: It's from kat-backend.git, so I guess it's useful for an inner loop when hacking on kat-client or kat-server?

kflynn:
Landed as `shared/edge_stack` for further dev work, so closing this PR. Thanks!!

PR #1886, opened 2019-10-02
Only add diagd to AMBASSADOR_DEBUG when enabled for a test

suever:
## Description
Previously in the tests, it was not possible for a test to suppress the `diagd` debug logs since they were always set regardless of the value of `AmbassadorTest.debug_diagd`. This PR respects the `AmbassadorTest.debug_diagd` value when modifying the `manifest_envs` such that a subclass of `AmbassadorTest` (such as the ones we need to write for [this PR](https://github.com/datawire/ambassador/pull/1817)) can disable `AMBASSADOR_DEBUG` completely.

In order to maintain the existing behavior, the default value is changed to `True`

## Testing
The expected behavior is that all current tests pass.

### DIFF
--- a/ambassador/tests/abstract_tests.py
+++ b/ambassador/tests/abstract_tests.py
@@ -79,13 +79,14 @@ class AmbassadorTest(Test):
     extra_ports: Optional[List[int]] = None
     debug_diagd: bool = False
     manifest_envs = ""
-    
+
     env = []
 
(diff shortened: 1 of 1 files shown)


kflynn:
I'm not actually sure we need that level of backward compatibility -- I might change the default back to `False` later. :wink: Thanks!

PR #1890, opened 2019-10-04
Support window_bits parameter for gzip filter

Pluies:
## Description

The window_bits parameter is exposed by Envoy (see [docs](https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/http/gzip/v2/gzip.proto)), and according to our testing is one of the most important settings for the resulting compressed size.

On averaging a few JS & CSS files, we get pretty consistently (to a few percent):
- gzip, no extra configuration: 100 (baseline)
- gzip, compression_level=BEST: ~99
- gzip, compression_level=BEST, window_size=15: ~80

So setting this gives us 20% extra compression!

The tradeoff in memory usage is documented by Envoy, whose documentation I've lifted in this PR.

## Related Issues

Background: https://github.com/envoyproxy/envoy/issues/8448

## Testing

Tested by building locally and seeing the gzipped size go down, see https://github.com/Pluies/gzip-comparison

## Todos
- [x] Documentation


Pluies:
Investigating why the tests didn't go through. 👍

Pluies:
Looks like the failing tests were just flakiness, everything is good now.
Let me know if this looks good @kflynn 👍

kflynn:
Looks great, thanks! Could you add this to one of the existing tests in `t_gzip.py`? I don't think we need to test that Envoy actually changes the window size, but I think that we should be testing that Ambassador at least configures Envoy correctly...

Pluies:
Thanks Flynn, I'll add a test :)

Pluies:
Actually, after looking into it... I'm not sure what to add to `t_gzip.py`. I was planning to model this new test after one of the other options, like `compression_level`, `memory_level`, `compression_strategy`; but I can't see any tests for these.
Given that tests are integration-level tests, we shouldn't see any difference in the output, besides better compression. But envoy's compression performance doesn't really sound like something we should test at the Ambassador level, right?
What would be the best approach to test in this case @kflynn ?

Pluies:
(For what it's worth, I've tested this manually by confirming that a build from this commit does accept the `window_bits` parameter, and the gains in compression size over a few files are present and identical to what pure Envoy gives us for the same parameter :) )

kflynn:
I just meant to add the `window_bits` parameter to one of the existing tests, so that we have something that'll fail if we somehow generate the wrong Envoy config in the future. 🙂

Pluies:
Ah ok, makes sense! https://github.com/datawire/ambassador/pull/1890/commits/4a88ac8155d0f68b2cce0a34ed182eafe0e33c92 should do the trick 👌

Pluies:
FWIW this build is breaking on the same test as the current master branch (https://github.com/datawire/ambassador/commit/50fd69d11e9af044f1137df59d0563f17dcf1b48):
```
tests/t_circuitbreaker.py:220: in check

    if 'X-Envoy-Overloaded' in result.headers:

E   TypeError: argument of type 'NoneType' is not iterable
```

I assume this is just a flaky test rather than an issue with this test; editing my commit to re-trigger a CI build...

kflynn:
Yeah, we've been seeing that circuit breaker test flaking an awful lot recently. I'm going to go ahead and take this, and we're going to dig into what's going on.

Pluies:
Thank you very much @kflynn 🙏

PR #1891, opened 2019-10-04
Import ambex.git

LukeShu:
## Description

This imports `ambex.git` (commit t8690fa0ab7fb684105d882c360222e4d71aa562a) to `go/ambex/`, and builds it from source there, instead of downloading a binary in `Dockerfile.base-go`.

I think that it's worth preserving the history of ambex.git (unlike in #1880), and that because that's a large number of commits, for code-review purposes, it's worth doing in a separate PR from #1880.

## Testing

I ended up manually building the new base-go image with `docker build  -t quay.io/datawire/ambassador-base:go-17 --cache-from=quay.io/datawire/ambassador-base:go-16 -f Dockerfile.base-go .` (the same as the Makefile would, except I added `--cache-from`) to avoid triggering a Python 3.7.3→3.7.4 update, which seems to be problematic (SIGSEGV, maybe related to PyYAML being built for 3.7.3, but I don't want to spend more time investigating it this week).


inercia:
I was wondering if it wouldn't be better to follow the Go Standard Project layout and have ambex imported in `cmd/ambex`...

LukeShu:
I was going to let #1880 do that rename.

inercia:
> I was going to let #1880 do that rename.

:open_mouth: :grin:

PR #1892, opened 2019-10-04
Envoy/base-image build stuff

LukeShu:
## Description

It's amazing what improvements I realize can be made when I spend a day looking at Flynn's screen while he works.

Adds two workflow-variables to the `Makefile`: `YES_I_AM_OK_WITH_COMPILING_ENVOY` and `YES_I_AM_UPDATING_THE_BASE_IMAGES` which must be set to a non-empty value to enable those parts of the Makefile; otherwise an error is printed, because those are things you normally don't want to do.

Also other things, but that was the most disruptive.

I guess dbe4127 isn't related to Envoy or the base images, but I included it in this PR anyway.


kflynn:
Need to dig into why the `GlobalCircuitBreakingTest` is flaking more often again. 😛

PR #1901, opened 2019-10-08
Reorganization: code layout

kflynn:
Props to @inercia for doing all the heavy lifting here! 

We’ve felt for awhile that the organization of the Ambassador sources was far less clear than it could be. This is especially true since big chunks of Ambassador are actually written in Go, and they generally don’t currently live in the Ambassador repo. This causes a ton of problems when we need to change them, and we wanted to make that better.

This PR is our first cut at making it better: we're reorganizing the code to basically match the [Standard Go Project Layout](https://github.com/golang-standards/project-layout). Please note that this PR is a reorganization, not a rewrite. :slightly_smiling_face: We do expect to be seeing more Golang in Ambassador in the future, but that’s not what’s happening right now.


kflynn:
I'm landing this -- CI is flaking like a flaky thing, but I get good runs of the flaking tests manually. Time to do some digging into CI tomorrow.

PR #1903, opened 2019-10-09
Minor cleanups

inercia:
## Description

Some minor fixes after the recent code reorganization...

## Related Issues

#1901


kflynn:
How about merge `master` back into this and let's see if all's well...

PR #1908, opened 2019-10-10
Minor CI tweaks

kflynn:
- Change the `debug_diagd` default for the tests to False. There's just too much output to make sense of when it's default True.
- Drop the circuitbreaker tests to 400 requests rather than 500 or 1000. I'm of two minds about this one; more digging is indicated.
- Make the Knative tests XFail, rather than being skipped, when they're not supported. (They're not supported in Kubernaut CI, and skipping all the tests was making the Knative part fail, which makes CI fail.)
- Make the Ingress tests XFail on MacOS, where we can't run the (built-for-Linux) `kubestatus` command.


kflynn:
Hmmmmmm. Gotta sort out what's up with this stuck "waiting for status to be reported"...

PR #1911, opened 2019-10-10
Bump watt to the latest commit

containscafeine:
This commit updates teleproxy to the latest commit because we
need the fix for rapid apply-delete bug - see
https://github.com/datawire/teleproxy/pull/197

go get github.com/datawire/teleproxy@62878af1fb91eebb080593a2185147e3bb543d9f
was run and the resulting diff has been committed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	git.lukeshu.com/go/libsystemd v0.5.3
 	github.com/Masterminds/sprig v2.17.1+incompatible
 	github.com/datawire/libk8s v0.0.0-20190923150809-3b461b0ee981
-	github.com/datawire/teleproxy v0.7.3-0.20190923151525-f9c9898fc956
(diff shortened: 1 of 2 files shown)


rhs:
BTW, I *believe* this PR should be unnecessary now. The watt that ambassador now uses is the one from the ambassador repo which was imported from a teleproxy version that already contained this change.

containscafeine:
@rhs I see go.mod has teleproxy pointing to https://github.com/datawire/teleproxy/commit/f9c9898fc956 which is 17 days ago - I want https://github.com/datawire/teleproxy/commit/568282a1fca691c684a658fc8854148731663221 which is 2 days ago. What am I missing here?

kflynn:
I'm going to close this -- we're now building `watt` from source rather than pulling in a binary. At Some Point™ we'll clean out `go.mod`, but that's blocked behind other work right now, and it doesn't need to be done right this second.

PR #1917, opened 2019-10-11
Write a `k8sregistryctl` command, use it in CI

LukeShu:
This adds a `k8sregistryctl` command for managing private in-cluster registries.  It adds the functionality of `k3sctl` that is missing when you bring-your-own-cluster.

This is intended to replace `docker-cluster.mk` such that it is palatable to @rhs, and is useful to outside contributors (and useful with kubernaut in CI).

## Todos
- [ ] Documentation


LukeShu:
It doesn't work on macOS.  But I've been using this on GNU/Linux for several days now, and I like it.

PR #1720, opened 2019-07-31
Set a capability to allow binding to low ports

swalberg:
## Description

The cap_net_bind_service capability allows a process to bind to a low
port without root. While Kubernetes lets you allow a capability in the
pod spec, the effective capabilities are wiped when the root -> user
transition is made.

This change `setcap`'s the binary so that it can assert that capability. e.g.
```
 securityContext:                                                                                                                                  |~
   capabilities:                                                                                                                                   |~
     add:                                                                                                                                          |~
       - NET_BIND_SERVICE
``` 
(there is no `CAP_` at the beginning; K8s seems to add that)

If you don't want to bind to a low port, don't add the capability and
we're back to the previous behavior. If you want to bind to a low port,
add the capability and configure Ambassador appropriately.

## Related Issues

This is a feature addition -- I want to be able to run this as a non root user but still bind to a low port.

## Testing

Manual testing.

## Todos
- [ ] Tests
- [ ] Documentation

PR #1920, opened 2019-10-14
Revert capabilities on envoy binary

swalberg:
Some environments don't like capabilities enabled binaries. This rolls
back #1720 which adds the net_bind_service capability directly to Envoy.

I will work on a different version that detects if the capability would
cause a problem, and to use a wrapper if so.

I have left the parts of the test suite in that let us specify the
capabilities block in the pod template since I will use them in the next
version of this.


kflynn:
Many thanks!!!

Issue #1125, opened 2019-01-18
Ensure tls_context has all the features of TLS Module, including redirect_cleartext_from


klamontagne:
_Ambassador Helm Chart `0.40.2`, ambassador container `quay.io/datawire/ambassador:0.50.0-rc5`, GKE deployment_

Upgrading to 0.50, I wish to use the SNI feature. I add a TLSContext to my service (through helm):
```yaml
---
apiVersion: ambassador/v1
kind:  Mapping
name:  myapp_mapping
prefix: /
host: myapp.mycompany.com
service: myapp.default:3000

---
apiVersion: ambassador/v1
kind: TLSContext
name: myapp-tls
hosts:
- myapp.mycompany.com
secret: mycompany-cert
```
`mycompany-cert` being a [cert-manager](https://github.com/jetstack/cert-manager) cert for `*.mycompany.com`

The ambassador service also has an annotation to redirect cleartext requests:
```yaml
---
apiVersion: ambassador/v1
kind: Module
name: tls
config:
  server:
    enabled: True
    redirect_cleartext_from: 80
    secret: ambassador-certs
```
Here, `ambassador-certs` is a cert-manager self-signed certificate for the name `ambassador`

If the server in the `tls` module is enabled, the ambassador log is flooded with:
```
[000033][warning][config] [bazel-out/k8-dbg/bin/source/common/config/_virtual_includes/grpc_mux_subscription_lib/common/config/grpc_mux_subscription_impl.h:70] gRPC config for type.googleapis.com/envoy.api.v2.Listener rejected: Error adding/updating listener ambassador-listener-443: error adding listener '0.0.0.0:443': partial wildcards are not supported in "server_names"
```

If I remove or disable the `tls` Module, the log clears up. However, I don't know where else to add the `redirect_cleartext_from` directive. It is not an accepted value in a TLSContext:
```
SYNC: ... <RichStatus not a valid TLSContext: Additional properties are not allowed ('redirect_cleartext_from' was unexpected)
```

I would like to redirect cleartext with the TLSContext.

klamontagne:
Possibly related: #981

kflynn:
Estimating at 2 because we should see what else is missing and fix it all, rather than simply adding `redirect_cleartext_from`.

PR #1928, opened 2019-10-15
port over tls module fields to tlscontext

SecurityInsanity:
#  Description

this commit adds in all fields documented in: `docs/reference/core/tls.md` into TlsContext, it turns out this was only missing redirect_cleartext_from. however I want to make it clear this should be all of them at this point. So TlsContext should be the thing that is used from this point forward.

## Related Issues

#1125 

## Testing

- I've added: `redirect_cleartext_from` to the Tls tests, this should validate that it doesn't affect people using https directly. Than I added a test to make sure when manually requesting http you get a redirect.

## Todos
- [x] Tests
- [x] Documentation (N/A)


kflynn:
Looks good, thanks! and sorry for the delay landing this one.

PR #1934, opened 2019-10-16
Builder tweaks

kflynn:
- Update to Alpine 3.10 (and thus Python 3.7), and update to the Envoy from Ambassador 0.83.0.
- Add the `kubectl` binary
- Update to PyYAML 5.1, and build it in the `builder` container, so that we can use the C parser and dumper rather than the pure Python implementations.

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -54,6 +54,18 @@ COPY --from=envoy /usr/local/bin/envoy /usr/local/bin/envoy
 
 ENV KUBECONFIG=/buildroot/kubeconfig.yaml
 
+# The YAML parser is... special. First, we need to install libyaml and Cython. No,
+# I don't know why it's called yaml-dev.
+RUN apk --no-cache add yaml-dev cython
+
+# After that, we can download !*@&#!*@&# PyYAML...
+RUN curl -O -L http://pyyaml.org/download/pyyaml/PyYAML-5.1.tar.gz
+RUN tar xzf PyYAML-5.1.tar.gz
+
+# ...and then install it.
+RUN cd PyYAML-5.1 && python3 setup.py --with-libyaml install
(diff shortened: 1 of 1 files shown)


builder/Dockerfile
```diff
@@ -53,6 +53,18 @@ COPY --from=envoy /usr/local/bin/envoy /usr/local/bin/envoy
 
 ENV KUBECONFIG=/buildroot/kubeconfig.yaml
 
+# The YAML parser is... special. First, we need to install libyaml and Cython. No,
+# I don't know why it's called yaml-dev.
+RUN apk --no-cache add yaml-dev cython
```
LukeShu: Can this be part of the main `apk add`?

builder/Dockerfile
```diff
@@ -53,6 +53,18 @@ COPY --from=envoy /usr/local/bin/envoy /usr/local/bin/envoy
...
+
+# After that, we can download !*@&#!*@&# PyYAML...
+RUN curl -O -L http://pyyaml.org/download/pyyaml/PyYAML-5.1.tar.gz
+RUN tar xzf PyYAML-5.1.tar.gz
```
LukeShu: I think the reason that CI is unhappy is that PyYAML is getting extracted over our sources in `/buildroot`.  Perhaps do this in `/tmp/pyyaml`?

builder/Dockerfile
kflynn: Done.

builder/Dockerfile
kflynn: Oh, nice catch! Done, and I folded the download/build/install `RUN`s together for you too...

kflynn:
Another one I'm going to take even with the failing `watt` test.

PR #1939, opened 2019-10-17
Increase default regex_max_size to 200

containscafeine:
The regex that we generate for the simplest knative service has
an RE2 maximum size of more than 100, which is Envoy's default.
This causes Ambassador to either crash or ignore the new config.

This commit increases the default regex_max_size to 150 which
should be enough for our generated regex.

### DIFF
--- a/python/ambassador/envoy/v2/v2route.py
+++ b/python/ambassador/envoy/v2/v2route.py
@@ -33,7 +33,7 @@ def regex_matcher(config: 'V2Config', regex: str, key="regex", safe_key=None) ->
         # 'safe' is the default. You must explicitly say "unsafe" to get the unsafe
         # regex matcher.
         if re_type != 'unsafe':
-            max_size = int(config.ir.ambassador_module.get('regex_max_size', 100))
+            max_size = int(config.ir.ambassador_module.get('regex_max_size', 200))
 
             if not safe_key:
                 safe_key = "safe_" + key


containscafeine:
@kflynn @SecurityInsanity while this fixes the crash, but I'm not aware if this change has any other implications - does it?

richarddli:
I have a vague memory that someone on the OSS Slack needed 160 chars. Is there any harm to bumping this to, say, 200?

SecurityInsanity:
So this limit isn't a character based limit. It's a "regex program size" limit. This is essentially meant to be a "rough evaluation of cost to evaluate this regex in memory". Although IIRC the 100 from envoy wasn't chosen too _scientifically_ (I believe it was just chosen on what google saw as a good default), being on the same default as envoy is _nice_.

Since there is a config option can we just bump the knative yaml to use 150 where it's needed, and ensure the option is documented? Bumping the cost globally can remove safeguards of shotting yourself in the foot.

If we really feel it's worth it _I'm honestly not incredibly opposed_ to doing it this way. I'd just prefer if we could use the config option.

SecurityInsanity:
Yeah I have seen those, but I still worry about the global bumping. There's always going to be unique cases where more is needed. Essentially I don't wanna "knee-jerk" remove the protection, for just a fresh change. Right now I don't see it in `docs/`, I do think we should document it. 

Than I wonder how many will need to change it at all? Are the slack messages perhaps people using knative? If we see a lot of people changing it, perhaps we can bring up changing the default in envoy too that a regex that works in envoy natively works in ambassador?

containscafeine:
@SecurityInsanity I've added docs for configuring this in https://github.com/datawire/ambassador/pull/1938 - and there are non-knative users hitting this as well. 

As @kflynn, you and I discussed on Slack, I've bumped this to 200.

PR #1940, opened 2019-10-17
Make regex_max_size in Ambassador module work

psalaberria002:
Without this entry it does not work and it defaults to 100.


psalaberria002:
Closing in favor of https://github.com/datawire/ambassador/pull/1938

PR #1943, opened 2019-10-17
Merge prep, part 2

LukeShu:
## Description
Prepare for the #1941 merge

## Testing
It seems to work on my laptop?  But CI was unhappy about something before I resolved a merge-conflict with master, so maybe something is wrong?


kflynn:
Taking with flakes.

For those following along who are wondering why, the reason is that we need to get the build machinery back up to snuff to be in a good position to let someone dig into the flakes...

PR #1944, opened 2019-10-17
Update TLS docs

kflynn:
…now that we support `redirect_cleartext_from` in a `TLSContext`.


kflynn:
Not waiting for CI on a doc-only change. ;)

Issue #1945, opened 2019-10-17
The CircuitBreaking and GlobalCircuitBreaking tests need to be de-flaked.


kflynn:
These tests have gotten awfully bad, and they need someone to dig in and make them un-flaky.

PR #1946, opened 2019-10-17
Make the circuitbreaking tests XFail instead of Fail

kflynn:
The circuit breaker tests are just too excessively flaky right now, so make them XFail instead of Fail when something goes wrong. Obviously this isn't a good idea, and they need attention: this is tracked in #1945.


LukeShu:
LGTM, to the degree that I'm qualified to say that.

PR #1948, opened 2019-10-18
fix release?

LukeShu:
## Description

Chase down irregularities with things changing or becoming dirty.

## Testing
I've tagged this as -ea6, to test the release machinery.  The tests have flaked 3 times now.  It's running again, but I'm going to bed. (A watched pot never turns green?)


kflynn:
Merging since -ea10 is the same commit as this, and we got a clean run there. 🙂

PR #1952, opened 2019-10-18
Fix a typo.

kflynn:
Whoops -- we tried to push to a nonexistent repo.


kflynn:
Not waiting for CI here, since this affects only RC builds.

PR #1955, opened 2019-10-18
Oops. Can't update AWS before the artifacts are pushed!

kflynn:
If you do, it's wrong. Sigh.


kflynn:
Closing in favor of #1956

Issue #1958, opened 2019-10-19
Upgrade from 0.83 to 0.84 fails, first issues with egg cache, than with snapshots


dmayle:
**Describe the bug**
Attempting to upgrade in place to 0.84 fails.  New pods

**To Reproduce**
1. Install a basic installation following "Get Started..."
2. kubectl apply -f ambassador-rbac.yaml # Version with 0.84
3. See failures in pod logs

**Expected behavior**
Pods should come up

**Versions (please complete the following information):**
 - Ambassador: 0.84
 - Kubernetes environment: bare metal
 - Version 1.15

**Additional context**
Initally, pods crash on a permission error for /ambassador/.cache, because diagd needs unpacking and the container can't write to that directory.  I set the environment variable PYTHON_EGG_CACHE=/tmp just to see if that would fix it (via the deployment with kubectl), and it moves forward, but then fails because "/ambassador/snapshots/snapshot-tmp.yaml" is not found.  I haven't gotten further than that

containscafeine:
Thanks for reporting this. While we try to reproduce it, can you try setting the following env var in your Ambassador deployment -  `AMBASSADOR_CONFIG_BASE_DIR: /tmp/ambassador` and see if it works.

kflynn:
@containscafeine I actually have a fix in CI. 🙂

containscafeine:
Fixed in #1959

PR #1959, opened 2019-10-20
Fix permissions of /ambassador to allow non-root without overriding AMBASSADOR_CONFIG_BASE_DIR

kflynn:
Thanks to @dmayle for the catch!

Ambassador uses `AMBASSADOR_CONFIG_BASE_DIR` to decide where on disk it will write things it needs to write. By default, it's `/ambassador`, and in our large chunk of build system work recently, the permissions there didn't get set correctly, so only `root` could write there.

We didn't catch this because, while our end-to-end tests don't run as `root`, they also use a read-only root filesystem, so they override `AMBASSADOR_CONFIG_BASE_DIR` to point into a memory filesystem on `/tmp` (mimicking a customer configuration). To fix _that_ problem, this PR modifies `test_docker.py` to run as non-root without overriding `AMBASSADOR_CONFIG_BASE_DIR`.

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -124,6 +124,11 @@ COPY --from=artifacts /buildroot/ambassador/demo/services /ambassador/demo-servi
 
 WORKDIR /ambassador
 
+# Fix permissions to allow correctly running as a non root user
+RUN chgrp -R 0 /ambassador && \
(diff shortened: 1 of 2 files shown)


kflynn:
Fixes #1958.

PR #1960, opened 2019-10-20
Update RELEASING.md to use a pull-request-based process

kflynn:
It used to be based around just committing to master. That actually doesn't work for most people.


kflynn:
(No need for a CI run for a doc update.)

PR #1801, opened 2019-09-08
Support a configurable log format

esmet:
## Description
Allows an operator to set the text log format. Useful for creating custom logs, since the ones out of the box may or may not contain all of the useful information for someone's particular use case.

## Testing
Tested by hand

## Todos
- [ ] Tests
- [ ] Documentation

PR #1969, opened 2019-10-22
Support a configurable log format

inercia:
## Description

From #1801 

Allows an operator to set the text log format. Useful for creating custom logs, since the ones out of the box may or may not contain all of the useful information for someone's particular use case.

## Testing
Tested by hand

## Todos
- [X] Tests
- [X] Documentation.


docs/reference/core/ambassador.md
```diff
@@ -91,6 +91,10 @@ spec:
   # envoy_log_path defines the path of log envoy will use. By default this is standard output
   # envoy_log_path: /dev/fd/1
 
+  # envoy_log_format defines the envoy log line format.
+  # see https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log for a complete list of operators
+  # envoy_log_format: "ACCESS [%START_TIME%] \"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%\" %RESPONSE_CODE% %RESPONSE_FLAGS% %BYTES_RECEIVED% %BYTES_SENT% %DURATION% %RESP(X-ENVOY-UPSTREAM-SERVICE-TIME)% \"%REQ(X-FORWARDED-FOR)%\" \"%REQ(U*    545 SER-AGENT)%\" \"%REQ(X-REQUEST-ID)%\" \"%REQ(:AUTHORITY)%\" \"%UPSTREAM_HOST%\"\n"
```
kflynn: Hmmmm... it looks like we always add the `\n`, right? So the user shouldn't include it? (Sorry, just noticed this.)

inercia: Good point. Fixing...

PR #1970, opened 2019-10-22
builder.sh: Let `compile` work if stdin isn't a TTY

kflynn:
Cherry-picking @LukeShu's 20a13dc22ed3aa9e28386f61e6fa272863025f9d for the upcoming releases.


kflynn:
Merging even with test flake.

PR #1971, opened 2019-10-22
Makefile tweaks

kflynn:
- Don't untag Docker images on `make clean` -- save that for `make clobber`
- Switch to short tracebacks for `pytest` runs
- Outdent two comments to prevent a MacOS error(???!!!!)


kflynn:
Taking this one too. The Makefile has succeeded long before the flakes happen.

PR #1977, opened 2019-10-24
Add a capabilities wrapper

swalberg:
This wrapper is given the CAP_NET_BIND_SERVICE capability so that it can
run Envoy with that bit set. This will let Envoy run on a low port as a
non privileged user.

Envoy needs to be able to inherit the capability, but I've verified that
this does not suffer from the same problem of not being able to run when
that capability is explicitly dropped.

In a separate PR I will update the entrypoint so that it calls the
wrapper if the capability is available, and Envoy otherwise

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -46,7 +46,7 @@ ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local
 # build it locally -- just using pip won't work.
 
 RUN apk --no-cache add bash curl jq rsync python3 python3-dev build-base libffi-dev openssl-dev sudo \
-                       iptables docker openssh-client libcap yaml-dev cython && \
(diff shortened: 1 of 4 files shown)


swalberg:
A few notes... I'll do the magic to use the wrapper or not depending on if the capabilities is available in another PR, but locally I yolo'ed the wrapper into the entrypoint and it works great:
```
bash-5.0$ netstat -anp | grep 80
tcp        0      0 127.0.0.1:8001          0.0.0.0:*               LISTEN      -
tcp        0      0 0.0.0.0:80              0.0.0.0:*               LISTEN      -
```

I also checked that if I drop the capability that Envoy still runs.

This might be because of my Go experience, but the `x/sys` module needed to be installed from git rather than whatever `go mod` does. The syscalls and structs for the capabilities are about 4 months old and the last release seems to be just over that.

swalberg:
@kflynn this is the one we were talking about on Slack.

PR #1991, opened 2019-10-29
Support injecting an HTML banner in diag UI

alexgervais:
## Description
Injecting an HTML banner in diag UI from a remote location.

## Related Issues
None

## Testing
Manual tests with remote endpoint.

## Todos
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/python/ambassador_diag/diagd.py
+++ b/python/ambassador_diag/diagd.py
@@ -29,6 +29,7 @@ import signal
 import threading
 import time
 import uuid
+import requests
 
 from pkg_resources import Requirement, resource_filename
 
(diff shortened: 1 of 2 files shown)


python/ambassador_diag/diagd.py
```diff
@@ -460,10 +463,23 @@ def show_overview(reqid=None):
...
+    app.logger.info(app.banner_endpoint)
+    if app.banner_endpoint:
+        try:
+            response = requests.get(app.banner_endpoint)
+            app.logger.info(response)
+            if response.status_code == 200:
+                banner_content = response.text
+        except Exception as e:
+            app.logger.error("could not get banner_content: %s" % e)
+            app.logger.exception(e)
```
kflynn: Some of the other things (like `envoy_stats`) are polled periodically rather than at every call to the overview. Would that make more sense? or would a push make sense?

Also, should the detail view show this as well?

alexgervais: No polling for now, as we expect the banner content might get more dynamic over time, and only for the overview.

kflynn:
Looks good!

PR #1994, opened 2019-10-30
support for keepalive envoy section

KowalczykBartek:
## Description
Because of how k8s is deployed and how networking works on such environment (on k8s clusters, there is more firewalls/iptables/ipvs/conntrack stuff than trees in the forest),some stale connection appear - and because of that 503 are returned from the ambassador/envoy to the client. Adding support for keep_alive should help :) 

## Testing
TODO - not testing, I have no test environment / tests on local cluster doesn't work on my computer  

## Todos
- [ ] Tests
- [x] Documentation

### DIFF
--- a/docs/reference/core/ambassador.md
+++ b/docs/reference/core/ambassador.md
@@ -123,6 +123,15 @@ spec:
   #   policy: round_robin/least_request/ring_hash/maglev
   #   ...
 
+  # keepalive sets the global keepalive settings.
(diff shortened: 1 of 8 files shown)


KowalczykBartek:
what have been done 
- per service / global configuration for keepalive 
- tests on local minikube: with global / with per service / with global and per service / no configuration 
- documentation

PR #1999, opened 2019-11-01
Add protobuf stuff for HostStatus

LukeShu:
From high-priority to low-priority:

 * Add `HostStatus` protobuf specification
 * Have `pkg/api/ambassador/` get `.UnmarshalJSON()` methods that do the right thing (even if the way we get them to do the right thing is hacky).
 * Add `Host` protobuf specification
 * Have `make generate` work properly even if `$(OSS_HOME)` != `$(CURDIR)`
 * Revive the `Makefile` bits to generate `pkg/api/envoy/`, since I wanted to use some of those bits for `pkg/api/ambassador/` (but not the bits to generate `api/envoy/`, or the other Envoy build stuff)
 * Upgrade protoc 3.5.1→3.8.0

The low-priority stuff is done as supporting changes to the high-priority stuff.


LukeShu:
A slightly abridged Slack conversation so that more than just Flynn and I can see it:
```
<flynn> OK, I don’t really understand #1999 — what’s the `nojson` bit?
<lukeshu> things in most `.proto` get a `.UnmarshalJSON()` method in their generated Go.  I arranged it so that things defined in `*_nojson.proto` don't get a generated `.UnmarshalJSON()`.
<lukeshu> why do I want Host to not get a generated `.UnamrshalJSON()?`
<lukeshu> A `Host` contains 3 things:
<lukeshu> - metadata
<lukeshu> - spec
<lukeshu> - status
<lukeshu> The generated `.UnmarshalJSON()` calls out to `protobuf/jsonpb`, so that an ordinary call to `encoding/json.Unmarshal` will call `→ obj.UnmarshalJSON() → jsonpb.Unmarshal()`, and it will do the "right thing" if you use the standard `encoding/json` instead of using `protobuf/jsonpb`.
<lukeshu> Where "right thing" means "the thing specified by proto3"
<lukeshu> But k8s.io uses proto2.  And proto2 didn't have a Timestamp primitive, so Kubernetes invented the hack that is https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#Time
<lukeshu> And protobuf/jsonpb chockes on metav1.Time
<lukeshu> Now, doing "the thing specified by proto3" is actually really important for Enums, because the default `encoding/json` thing won't work for them at all.
<flynn> :face_vomiting:
<lukeshu> But k8s.io enums don't actually get emitted as protobuf enums. Additionally, they have defined all of the appropriate UnmarshalJSON methods such that `encoding/json` already does the right thing for them.
<lukeshu> So, in our `Host`, we want to call `encoding/json` for the `metadata`, but `protobuf/jsonpb` for `spec` and `status`.  We don't particularly care which is called for the `Host` itself.
<flynn> That’s… disgusting.
<lukeshu> So, call `encoding/json` for the `Host` itself (i.e. don't generate a `.UnmarshalJSON()` for it), but then do generate `.UnmarshalJSON()` methods for `HostSpec` and `HostStatus` that will have `encoding/json` call in to `protobuf/jsonpb`.
<lukeshu> And this works because the types that we generate `.UnmarshalJSON()` for don't contain any problematic k8s.io types.
<lukeshu> So I guess the shorter form of the answer is:
<lukeshu>  - broadly: "We need to avoid generating `.UnmarshalJSON()` methods for anything that makes use of a problematic k8s.io type; we do this by putting it in a `*_nojson.proto` file."
<lukeshu>  - specifically: "`Host` contains k8s.io meta/v1.ObjectMeta, and ObjectMeta contains meta/v1.Time, and meta/v1.Time is problematic"

<flynn> Mostly I worry about fragility…
<lukeshu> Yeah, the situation with k8s.io protobuf is fragile, but I'm not sure what a better route would be.
<lukeshu> I think a decent rule of thumb is
<lukeshu> - `ambassador.Thing` goes in a `_nojson.proto`
<lukeshu> - `ambassador.Thing` only contains `k8s.TypeMeta`, `k8s.ObjectMeta`, `ambassador.ThingSpec`, and possibly `ambassador.ThingStatus`
<lukeshu> - `ambassador.ThingSpec` and `ambassador.ThingStatus` go in a `.proto`
<lukeshu> - `ambassador.ThingSpec` and `ambassador.ThingStatus` cannot use any k8s type that includes `k8s.Time`, `k8s.Duration`, or other types that are problematic
<lukeshu> (and hope beyond all hope that we never want to use a k8s type that happens to include `k8s.Time`)

<lukeshu> I think all of the root problematic types are
<lukeshu> 
<lukeshu>     $ cd ~/src/github.com/kubernetes/kubernetes/staging/src/k8s.io && git grep ') UnmarshalJSON(' ':!*_test.go' |grep /v|sed -r 's,(.*)/[^/]*:func \(\S* (\S*)\) UnmarshalJSON.*,k8s.io/\1.\2,'
<lukeshu>     k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1.*JSONSchemaPropsOrBool
<lukeshu>     k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1.*JSONSchemaPropsOrStringArray
<lukeshu>     k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1.*JSONSchemaPropsOrArray
<lukeshu>     k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1.*JSON
<lukeshu>     k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1.*JSONSchemaPropsOrBool
<lukeshu>     k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1.*JSONSchemaPropsOrStringArray
<lukeshu>     k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1.*JSONSchemaPropsOrArray
<lukeshu>     k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1.*JSON
<lukeshu>     k8s.io/apimachinery/pkg/apis/meta/v1.*Duration
<lukeshu>     k8s.io/apimachinery/pkg/apis/meta/v1.*GroupVersion
<lukeshu>     k8s.io/apimachinery/pkg/apis/meta/v1.*FieldsV1
<lukeshu>     k8s.io/apimachinery/pkg/apis/meta/v1.*MicroTime
<lukeshu>     k8s.io/apimachinery/pkg/apis/meta/v1.*Time
<lukeshu>     k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.*Unstructured
<lukeshu>     k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.*UnstructuredList

<flynn> We could stop using the K8s types inside Host entirely.
<flynn> (Not sure that’s better, just brainstorming.)
<lukeshu> So the alternative is to just not specify `Host` in protobuf, and write the struct definition in Go.
<lukeshu> (1) That Go would be identical to the `protoc` output now
<lukeshu> (2) We'd still have to be careful to avoid the wrong k8s types in the rest of our protobuf specs.
<lukeshu> So I'm not sure that it's better, and that it's just less... obvious that the whole thing is fragile.
```

PR #2028, opened 2019-11-12
Add `edgectl license`

ark3:


cmd/edgectl/aes_license.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	return nil
+}
+
+const secretManifest = `
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ambassador-edge-stack
+  namespace: ambassador
```
alexgervais: The `namespace` should be a variable and passed in the Sprintf template.

ark3: Good catch. Thank you!

PR #2032, opened 2019-11-13
URLs should not contain newlines.

kflynn:
If they do, MacOS (at least) will decide to URL-encode the whole thing, which makes anchors in URLs completely fail.


LukeShu:
LGTM.  And I can empirically confirm that it doesn't break anything on GNU/Linux with Firefox.

PR #2033, opened 2019-11-13
builder.sh: Only embed the apro.git version number in Go binaries

LukeShu:
Per our discussion.

### DIFF
--- a/builder/builder.sh
+++ b/builder/builder.sh
@@ -279,7 +279,7 @@ case "${cmd}" in
                     printf "${CYN}==> ${GRN}Building ${BLU}${module}${GRN} go code${END}\n"
                     echo_on
                     mkdir -p /buildroot/bin
-                    (cd ${MODDIR} && go build -trimpath ${BUILD_VERSION:+ -ldflags "-X main.Version=$BUILD_VERSION" } -o /buildroot/bin ./cmd/...) || exit 1
+                    (cd ${MODDIR} && CGO_ENABLED=0 go build -trimpath ${BUILD_VERSION:+ -ldflags "-X main.Version=$BUILD_VERSION" } -o /buildroot/bin ./cmd/...) || exit 1
                     if [ -e ${MODDIR}/post-compile.sh ]; then (cd ${MODDIR} && bash post-compile.sh); fi
                     echo_off
                 fi
--- a/builder/builder.sh
+++ b/builder/builder.sh
@@ -272,14 +272,14 @@ case "${cmd}" in
         # This runs inside the builder image
         for MODDIR in $(find-modules); do
             module=$(basename ${MODDIR})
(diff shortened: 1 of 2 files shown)


ark3:
Okay, so this will use apro.version for all go binaries, if it exists, or otherwise leave binaries as (no version). Is that right?

LukeShu:
> Okay, so this will use apro.version for all go binaries, if it exists, or otherwise leave binaries as (no version). Is that right?

That's correct.

ark3:
Okay, so OSS builds will be versionless. I guess that's fine. I think we should remove Edge Control build/push/tag/promote from OSS CI, then.

kflynn:
Concur re not building `edgectl` from A/OSS CI.

LukeShu:
I'm about to add another commit.

ark3:
This loses basically all the module-generic aspects of the new build system.

LukeShu:
> This loses basically all the module-generic aspects of the new build system.

kiiiiinda? There always has and continues to be a bit that is specific to the python ambassador.

LukeShu:
I think what's left is to revert b42e55c and d655c26 and to re-create them in apro.git?

ark3:
Yes, that sounds right. How shall we proceed?

LukeShu:
Oh, also, `VERSION.py` needs hacked up to parse `apro.version` instead of (in addition to?) `ambassador.version`.

LukeShu:
```console
$ find / -name '*.version'
/usr/lib/python3.7/site-packages/ambassador-0.0.0.dev0-py3.7.egg/ambassador.version
/buildroot/ambassador/python/ambassador.version
/buildroot/ambassador/python/apro.version
```
Unfortunately, it seems that `/usr/lib/python3.7/site-packages/ambassador-0.0.0.dev0-py3.7.egg/ambassador.version` is the file that it's reading, so with the `builder.sh` on this branch, it isn't as simple as `s/ambassador/apro/` in the filename in `VERSION.py`

LukeShu:
Other than that, this PR seems to do sane things.

 - @ark3 Do you want to lift b42e55c and d655c26 over to apro.git? You surely know what those scripts are doing better than I do.
 - @kflynn Do you want to figure out something sane around the info in `apro.version` making it in to `VERSION.py`?

ark3:
Yes, I'll add the appropriate stuff to apro.git to build Edge Control for Linux and MacOS. Thank you.

LukeShu:
Flynn says hacking up VERSIONS.py is a rabbit hole, and to have APro's `post-compile.sh` overwrite `ambassador.version` for now.

LukeShu:
https://github.com/datawire/apro/pull/461

LukeShu:
I'll wait for CI to finish on both, then click merge on both?

PR #2044, opened 2019-11-18
Pass #1 at linting everything

LukeShu:
1. Add `make lint-oss` (named that instead of `make lint` for now to avoid conflicting with the apro.git Makefile), which assumes that `mypy`, `golangci-lint`, and `shellcheck` are already in your `PATH`.
2. Get the `shellcheck` portion of `make lint-oss` to come back clean.
3. While I'm in the `Makefile`, move all of the `make generate` stuff to `build-aux-local/generate.mk`

The mypy *almost* comes back clean.  golangci-lint is very far from coming back clean.


LukeShu:
Just resolved merge-conflicts.  I haven't tested this, since I'm having problems with the builder on this network?

PR #2048, opened 2019-11-21
Keep watching CRDs after bootstrap

inercia:
## Description

Some Custom Resource Definitions can be missing from the cluster (like after an upgrade, where users have changed only the image). In that case we should not fail on bootstrap but

1) install a watcher for `customresourcedefinition`s, and then
2) retry to watch the all the pending Custom Resources


inercia:
@kflynn I remember you mentioned we should also watch RBAC, just in case we do not have permissions for watching these resources. Should I add a watch for `"rbac.authorization.k8s.io/v1"` in this PR?

kflynn:
Whoops, this was long ago subsumed by #2071.

PR #2056, opened 2019-11-25
Simple ridethrough if the LogService CRD isn't present

inercia:
# Description

Ignore the `LogService` CRD missing.

```
[...]
2019-11-25 09:13:45 diagd 0.86.0-rc5-268-g06ed6a14-dirty [P305TAmbassadorEventWatcher] INFO: ...update finished, rc 0
2019-11-25 09:13:45 diagd 0.86.0-rc5-268-g06ed6a14-dirty [P305TAmbassadorEventWatcher] INFO: configuration updated from snapshot 1
2019-11-25 09:13:45 diagd 0.86.0-rc5-268-g06ed6a14-dirty [P305TAmbassadorEventWatcher] INFO: starting Envoy status updater
2019-11-25 09:13:45 diagd 0.86.0-rc5-268-g06ed6a14-dirty [P305TAmbassadorEventWatcher] INFO: error  Ambassador could not find the Host CRD definition. Please visit https://www.getambassador.io/reference/core/crds/ for more information. You can continue using Ambassador via Kubernetes annotations, any configuration via CRDs will be ignored...
2019-11-25 09:13:45 diagd 0.86.0-rc5-268-g06ed6a14-dirty [P305TAmbassadorEventWatcher] INFO: error  Ambassador could not find the LogService CRD definition. Please visit https://www.getambassador.io/reference/core/crds/ for more information. You can continue using Ambassador via Kubernetes annotations, any configuration via CRDs will be ignored...
2019-11-25 09:13:46 diagd 0.86.0-rc5-268-g06ed6a14-dirty [P305TAmbassadorEventWatcher] INFO: Scout reports {"latest_version": "0.85.0", "application": "ambassador", "cached": false, "timestamp": 1574673226.31647}
2019-11-25 09:13:46 diagd 0.86.0-rc5-268-g06ed6a14-dirty [P305TAmbassadorEventWatcher] INFO: Scout notices: []
```

Fix datawire/apro#514

## Related Issues

Copycat of https://github.com/datawire/ambassador/commit/dce9e3f03954725ff3ec9bee8acda3d723e06415

PR #2063, opened 2019-11-26
Add Host CRD error to default

concaf:
This PR adds the missing host CRD error to the default errors list that we have in our tests. This is required because #2055 shows the missing CRD error to the users but our tests don't account for it, which means a couple of test (Empty and Plain) have been failing due to this.


inercia:
I think I will need something similar for fixing tests for #2056, right?

concaf:
@inercia yep

PR #2071, opened 2019-12-01
Handle missing CRDs at runtime

kflynn:
Two things about this:

1. This is using the old, more static, error-reporting code. I'm not going to fix that until @ark3's error-path code is merged.

2. If you delete a CRD type once we're watching for it, then re-add it, you'll need to restart the Ambassador pod. We should tackle this Soon™.

3. Props to @inercia for writing the beginning of this, and to @LukeShu for help with the catcher!

### DIFF
--- a/cmd/watt/aggregator.go
+++ b/cmd/watt/aggregator.go
@@ -17,7 +17,7 @@ import (
 
 type WatchHook func(p *supervisor.Process, snapshot string) WatchSet
 
-type Aggregator struct {
+type aggregator struct {
 	// Input channel used to tell us about kubernetes state.
 	KubernetesEvents chan k8sEvent
 	// Input channel used to tell us about consul endpoints.
@@ -47,8 +47,8 @@ type Aggregator struct {
 }
 
 func NewAggregator(snapshots chan<- string, k8sWatches chan<- []KubernetesWatchSpec, consulWatches chan<- []ConsulWatchSpec,
-	watchHook WatchHook, limiter limiter.Limiter) *Aggregator {
-	return &Aggregator{
+	watchHook WatchHook, limiter limiter.Limiter) *aggregator {
+	return &aggregator{
 		KubernetesEvents:    make(chan k8sEvent),
 		ConsulEvents:        make(chan consulEvent),
 		k8sWatches:          k8sWatches,
@@ -63,7 +63,7 @@ func NewAggregator(snapshots chan<- string, k8sWatches chan<- []KubernetesWatchS
 	}
 }
 
-func (a *Aggregator) Work(p *supervisor.Process) error {
+func (a *aggregator) Work(p *supervisor.Process) error {
 	p.Ready()
 
 	for {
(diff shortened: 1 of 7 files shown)


rhs:
Why are we trying to solve the CRD problem by modifying watt? Can't we solve this problem more simply/easily by asking watt to watch for what CRDs exist and then modifying the watch hook to request only those resources that do in fact exist? In other words, do the equivalent of:

```
$ kubectl get -w crds
NAME                                           CREATED AT
authservices.getambassador.io                  2019-11-21T05:58:25Z
consulresolvers.getambassador.io               2019-11-21T05:58:26Z
filterpolicies.getambassador.io                2019-11-21T05:58:29Z
filters.getambassador.io                       2019-11-21T05:58:29Z
hosts.getambassador.io                         2019-11-21T05:58:26Z
kubernetesendpointresolvers.getambassador.io   2019-11-21T05:58:27Z
kubernetesserviceresolvers.getambassador.io    2019-11-21T05:58:27Z
logservices.getambassador.io                   2019-11-21T05:58:27Z
managedcertificates.networking.gke.io          2019-10-22T16:37:23Z
mappings.getambassador.io                      2019-11-21T05:58:28Z
modules.getambassador.io                       2019-11-21T05:58:28Z
ratelimits.getambassador.io                    2019-11-21T05:58:29Z
ratelimitservices.getambassador.io             2019-11-21T05:58:28Z
tcpmappings.getambassador.io                   2019-11-21T05:58:28Z
tlscontexts.getambassador.io                   2019-11-21T05:58:29Z
tracingservices.getambassador.io               2019-11-21T05:58:29Z
updateinfos.nodemanagement.gke.io              2019-10-22T16:37:23Z
...
```

kflynn:
Because if you only watch CRDs and rely on the watch hook, missing RBAC leaves you completely DOA... and that's especially problematic if you're running in single namespace mode and don't have permission to watch CRDs cluster-wide, since watching for CRDs within your single namespace just doesn't work.

(Watching for CRDs as you suggest was the first thing I tried, in fact, and it doesn't pass regressions.)

rhs:
If we were to solve the RBAC problem do you think it would work/result in a smaller watt change? I'm just trying to minimize the scope of the watt change since those are riskier changes. I'm gonna try reading the code change again this morning when I'm more awake, but last night as I was going through the code, the scope of the change was enough to trigger my risk warning, hence my questions. ;-)

cmd/watt/aggregator.go
```diff
@@ -30,7 +30,7 @@ type aggregator struct {
 	snapshots chan<- string
 	// We won't consider ourselves "bootstrapped" until we hear
 	// about all these kinds.
-	requiredKinds       []string
+	requiredKinds       map[string]bool
```
rhs: What's the new semantics associated with this field? Could you update the comment to reflect that?

cmd/watt/aggregator.go
```diff
@@ -17,7 +17,7 @@ import (
 
 type WatchHook func(p *supervisor.Process, snapshot string) WatchSet
 
-type aggregator struct {
+type Aggregator struct {
```
rhs: Is this changed to uppercase for a reason? (I think in go that means it's now public, but I don't see any public usage of it.)

kflynn: It does mean it's now public, and it's public because `kubebootstrap` needs it now (to update the `requiredTypes` field).

cmd/watt/aggregator.go
kflynn: Done.

cmd/watt/aggregator.go
kflynn: (cf `kubewatchman.go:109`)

PR #2074, opened 2019-12-02
Include TLSContexts in the diag snapshot (and in mockery)

kflynn:
This should make it much easier for Edge Stack's admin UI to work with TLSContexts.

### DIFF
--- a/python/ambassador/diagnostics/diagnostics.py
+++ b/python/ambassador/diagnostics/diagnostics.py
@@ -585,7 +585,8 @@ class Diagnostics:
             'errors': self.errors,
             'notices': self.notices,
             'groups': { key: value.as_dict() for key, value in self.groups.items() },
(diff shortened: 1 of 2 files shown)


python/ambassador/diagnostics/diagnostics.py
```diff
@@ -585,7 +585,8 @@ def as_dict(self) -> dict:
             'errors': self.errors,
             'notices': self.notices,
             'groups': { key: value.as_dict() for key, value in self.groups.items() },
-            'clusters': { key: value.as_dict() for key, value in self.clusters.items() }
+            'clusters': { key: value.as_dict() for key, value in self.clusters.items() },
+            'tlscontexts': [ x.as_dict() for x in self.ir.tls_contexts.values() ]
```
ark3: This is why I like to add that trailing comma as I go.

PR #2075, opened 2019-12-02
Updated cluster-health stats

kflynn:
Switch to using upstream_rq_completed as the base for cluster-health calculations.


kflynn:
Thanks! That might have been it. I'm gonna land this one, and we can talk more tomorrow.

PR #2080, opened 2019-12-04
Avoid Linkerd sidecar injection like Consul/Istio

ark3:
As described by Alvaro in his results for the LinkerD forklift upgrade scenario (F4), we need to add an annotation to Ambassador's Deployment to avoid it being injected with LinkerD's sidecar.

Related:
https://github.com/datawire/getambassador.io/pull/122
https://github.com/datawire/apro/pull/604


ark3:
Turns out this set of PRs is wrong. Istio and Consul allow Ambassador to initiate mTLS connections to upstream services by grabbing a certificate from a Kubernetes Secret. LinkerD does not work this way, so Ambassador must rely on LinkerD for mTLS connections to upstream services. This means we want LinkerD to inject its sidecar into Ambassador's pods.

PR #2082, opened 2019-12-05
upgrading envoy, this time with extra edgeyness

SecurityInsanity:
## Description

this upgrades envoy to v1.12.1, however this time we do it ontop
of shared/edgy, instead of having it on master like we had at one
point.

## Testing

`make check envoy`
`make test`

all pass

## Todos
- [x] Tests
- [ ] Documentation


SecurityInsanity:
@kflynn you'll want to remember to check the goldfiles here. Due to things we talked about in person.

kflynn:
I did, in fact, force a full goldfile check. Although CI passing is rather a nice check too.

PR #2084, opened 2019-12-06
Fixing knative and statsd env var parsing, reporting actual state of feature

alexgervais:


kflynn:
Change looks good, the fact that we need it makes me sad. :wink:

PR #2087, opened 2019-12-07
Hide the Fallback TLSContext from old OSS

LukeShu:
This is the counterpart to https://github.com/datawire/apro/pull/631

stand-alone title: Recognize a magic `"_automatic_"` namespace

## Description

This allows the init-config to have dynamic namespaces; `_automatic_` will be set to AMBASSADOR_NAMESPACE.  Because `_` is forbidden in Kubernetes resource names, the name `_automatic_` can't come from a real Kubernetes resource; only from init-config.


LukeShu:
I have no idea what's going on with CI... I can't imagine it's related to anything in this PR.

LukeShu:
Revert using POD_NAMESPACE instead of AMBASSADOR_NAMESPACE, per a conversation with Flynn.

PR #2089, opened 2019-12-09
Tweaks to make init-config work all the time

kflynn:
Resources in `init-config` were being mishandled if your `ambassador_id` or `AMBASSADOR_CONFIG_BASE_DIR` were changed from their defaults. Obviously that's a problem -- this fixes it.

### DIFF
--- a/python/ambassador/config/config.py
+++ b/python/ambassador/config/config.py
@@ -234,6 +234,14 @@ class Config:
         # Is an ambassador_id present in this object?
         allowed_ids: StringOrList = resource.get('ambassador_id', 'default')
 
+        # If we find the array [ '_automatic_' ] then allow it, so that hardcoded resources
+        # can have a useful effect. This is mostly for init-config, but could be used for
+        # other things, too.
+
+        if allowed_ids == [ "_automatic_" ]:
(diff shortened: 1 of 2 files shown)


python/ambassador/config/config.py
```diff
@@ -234,6 +234,14 @@ def good_ambassador_id(self, resource: dict):
...
         allowed_ids: StringOrList = resource.get('ambassador_id', 'default')
 
+        # If we find the array [ '_automatic_' ] then allow it, so that hardcoded resources
+        # can have a useful effect. This is mostly for init-config, but could be used for
+        # other things, too.
+
+        if allowed_ids == [ "_automatic_" ]:
+            self.logger.debug(f"ambassador_id {allowed_ids} always accepted")
+            return True
+
```
LukeShu: (to type words that I said to Flynn in-person):

Given that this is strict (compared to say `if "_automatic_" in allowed_ids:`) because it's a hack, and isn't something that we want users to think they can type: Would it be better to have the `_automatic_` namespace logic also munge the resource's `ambassador_id`?  This would have the advantage that a user literally _can't_ type it; `_` is illegal in a real namespace.

kflynn:
Merging for now, but I think I'll switch this per @LukeShu's excellent suggestion later.

PR #2092, opened 2019-12-10
 Add fallback TLSContext, if needed, during config load

ark3:
Only adds the fallback (self-signed) context if

- there is no termination context available (a TLSContext with Hosts defined)
- Ambassador is running as part of the Edge Stack


ark3:
Manual testing of AES for M1 succeeds for me. Plus automated testing of OSS via CI, of course.

khussey:
Thanks @ark3 I'm +1 to merge.

PR #2093, opened 2019-12-10
Be smarter with kubestatus

kflynn:
This was annoying.

- Use `concurrent.futures` to run up to five `kubestatus` commands at a time
- (Try to) keep track of things for which we've already set the status, and don't do it again.

The second one is particularly important, at least on GKE, to prevent a pathological cycle between `watt` and `diagd`. 😛


kflynn:
Whoops, I should've let @rhs actually approve this with a review. He approved it in person.

PR #2094, opened 2019-12-10
Mention that it can take a few minutes to see the result of license change

alexgervais:
> Please wait a minute or two for your license to be refreshed in your running Ambassador Edge Stack.


alexgervais:
Addressing https://github.com/datawire/apro/issues/606

PR #2097, opened 2019-12-10
Merge shared/edgy in to master

LukeShu:

PR #2098, opened 2019-12-10
Activate fallback context only if its secret is available

ark3:
I'll rebase this once #2097 lands


ark3:
Oops, I'm a dummy. I need to create a new PR to land this onto master.

PR #2104, opened 2019-12-11
Allow Edge Control to build on Windows

ark3:


khussey:
@ark3 can you comment on the testing that was done with these changes?

ark3:
The usual automated testing of Edge Control on Linux (via CI) ran, plus testing of `edgectl version` on all supported platforms via AES CI that I'll PR next.

The primary goal is to make sure I haven't broken Edge Control for Mac or Linux; I think CI covers that decently, but we don't have automated tests around the AES-specific features. I re-tested them manually.

I'm going to test on a real Windows machine tomorrow, but I want CI going for this stuff ASAP.

rhs:
Looks good to me!

PR #2110, opened 2019-12-12
Add a consistent name to dtest's k3s and registry

ark3:
This change makes it easier to see what's going on when your run `docker ps`.

This was originally #2004.


rhs:
Looks good to me!

ark3:
Sweet. Let's land this then. I had wanted :eyes: from @rhs just in case.

PR #2112, opened 2019-12-13
Coalesce k8sEvent

alexgervais:
Performance improvement. Coalescing `k8sEvent`s, we don't generate a new snapshot and pay the high cost of sending a notification to the python process when dealing with multiple incoming Kubernetes events.
Each event still needs to process `a.setKubernetesResources` so we don't miss any watch status update.


rhs:
Shouldn't you add some comments? You can probably crib most of the comments from the snapshot coalescing code.

cmd/watt/aggregator.go
```diff
@@ -60,14 +60,97 @@ func NewAggregator(snapshots chan<- string, k8sWatches chan<- []KubernetesWatchS
...
+		case kubernetesEventProcessor <- potentialKubernetesEventSignal:
+			// if we aren't currently blocked in
+			// a.maybeNotify() then the above goroutine will be
+			// reading from the kubernetesEventProcessor channel and we
+			// will send the current potentialKubernetesEventSignal
+			// value over the kubernetesEventProcessor channel to be
+			// processed
+			select {
+			case potentialKubernetesEvent := <-a.KubernetesEvents:
+				a.setKubernetesResources(potentialKubernetesEvent)
```
kflynn: Next time you're here, a comment about "here's the blocking read I was talking about!" wouldn't be amiss. Not going to hold for that though.

PR #2113, opened 2019-12-13
Fixing flaky npm install perm

alexgervais:
Don't let NPM try to outsmart us by switching user and group when installing apps. It should run in the current user context (root).


LukeShu:
I'm kinda flabbergasted that this was an "only sometimes flake" issue, if this is the fix.  But, if it seems to fix it, LGTM.

PR #2117, opened 2019-12-16
Split tests to support Edge Stack

kflynn:
This was a royal pain, and there's more to do, but this tweaks things up enough to support starting to run the KAT tests under Edge Stack.

- There's a lot of stuff here to separate manifests and goldfiles depending on whether we're running in Edge Stack or OSS. 
- There's also easier support for `xfail`ing a test, so that (among other things) it's straightforward to `xfail` a test on Edge Stack.
- At present, we don't allow running in local mode for Edge Stack. Coming soon.
- At present, most tests disable Edge Stack's automatic redirect-cleartext-from, because most of the tests assume that they can speak cleartext without redirecting.
- At present, several tests (mostly around TLS) are `xfail`ed on Edge Stack as we sort out fallback TLS and such.

Obviously we need to get rid of the `xfail`s and support local mode. This is just a starting point.


LukeShu:
I assume there's going to be a sibling PR in apro.git to add the gold and manifests directories?

LukeShu:
Ah, the counterpart is https://github.com/datawire/apro/pull/668

LukeShu:
The xfail logic seems... funny to me.

LGTM
 1. Assuming nothing sneaky happened, and the `manifests/` directory is verbatim from `manifests.py`
 2. Pending my approval on https://github.com/datawire/apro/pull/668, which I haven't done yet.

PR #2120, opened 2019-12-17
Scout reports aes app and version

alexgervais:
## Description
When running in the context of AES, scout tags the app as `aes` and will report the same version as diag UI and Edge Console.
The behaviour change is required because we build Ambassador from a detached head and the current version parsing assumes we are running a local build.

## Testing
For AES, scout will report: `aes`, `1.0.0-ea5-18-g56976d54-dirty`
For AOSS, scout will report: `ambassador`, `1.0.0-local+33.gcb0a71ac5.dirty`


alexgervais:
@kflynn please have a look. The changes seem simple enough (and there are effectively no changes of AOSS) but given your knowledge of scout and version-update-checks, I might not have the full picture.

alexgervais:
Indeed, I tested both builds.

kflynn:
👍

PR #2123, opened 2019-12-19
Mockery fixes

kflynn:
Instead of running the `watch_hook` as an external Python program from `mockery`, just `import` it. This is prep for not having `watt` need to invoke it either.

Also, clean up some logging and typing stuff.


rhs:
Looks good to me! Leaving to @khussey to merge.

PR #2124, opened 2019-12-19
TLS fixes

kflynn:
*Note well: this is built on my `mockery-fixes` branch, PR #2123.*

This fixes two nasty bugs in TLS/SNI handling:

1. For some time now, we've split the world into SNI routes and non-SNI routes. An SNI route is built from a `Mapping` that specifies a `host` requirement that matches a `host` in a `TLSContext` -- in other words, we expect SNI to be in play. A non-SNI route is anything else. This is fine, except that SNI routes were assembiled with a lower precedence than any non-SNI route, and that's a _huge_ problem for Edge Stack -- SNI routes were coming in _after_ Edge Stack's fallback route! Honestly, I'm astonished that we didn't have OSS users getting burned by this.

   The fix is to ditch the SNI-vs-non-SNI distinction: instead, just keep one list of routes, but remember inline which need SNI.

2. If you manually specify a `TLSContext` that matches `host: "*"`, that has to take the place of Edge Stack's fallback context -- if you try to include both, Envoy will scream because they have duplicate `filter_chain_match`es. So drop the fallback context in that case.

Given that the above are fixed, we can stop `XFail`ing most of the TLS tests in Edge Stack.


rhs:
Looks good to me! Leaving to @khussey to merge.

PR #2125, opened 2019-12-19
Add GCLOUD_CONFIG to run pytest

concaf:
This commit adds support for a GCLOUD_CONFIG variable for users
who want to run their kat tests in GKE. This is required because
the .kube/config file that gcloud generates requires to run
gcloud commands to retrieve authentication token and token expiry,
which means that the builder image needs to have access to the
gcloud command and cluster.


rhs:
Looks good to me!

PR #2126, opened 2019-12-19
Made Envoy listener idle_timeout configurable

neufeldtech:
## Description
We currently have Ambassador deployed on-premise behind a firewall and fronted by a CDN. We encountered an issue where on low-traffic services, our clients were receiving some 503 first byte timeouts. 

The issue is present because when the firewall chose to remove idle tcp sessions from its session table, it would not send a FIN or RST to the CDN. On the next request, the CDN thinks that it still has a tcp connection open and attempts to re-use the now-closed TCP connection, resulting in the request getting 'black-holed', and leads to an HTTP 503.

Because Envoy doesn't ever attempt to close idle tcp connections by default, we're left with Envoy thinking that sessions are open, but our firewall has closed them after a 30 minute timeout (without notifying the CDN that the connection is closed), and the CDN still thinks that the connection is open.

Our "options" for remediating this scenario were:
- Configure tcp idle timeout on the CDN side to control connection idle time
  - This turned out to not be an option that we can readily control on our CDN. They have 'dynamic' connection times.
- Make the firewall send a FIN or RST to the CDN when it removes a session from its session table
  - We investigated this as well, but on our firewall this is not an option that is configurable.
- Have Envoy close idle tcp connections gracefully after a configurable timeout (in our case, some timer that is lower than our firewall idle session timeout)
  - We believe that this is the most appropriate way to solve the issue, and gives us control when tuning idle timeouts across each network device in the path, so this PR addresses making the `idle_timeout` Envoy setting configurable on the Envoy http listener.

## Related Issues
N/A

## Testing
Manually tested via the following process:

- Configure `idle_timeout: 15s` in Ambassador module
- Open connection via netcat, send a request, wait for envoy to close the connection after 15s
```bash
time nc localhost 80                                                                                                                  (docker-desktop/default)
GET / HTTP/1.1
Host: localhost

HTTP/1.1 200 OK
content-length: 2002
content-disposition: inline; filename="index.html"
accept-ranges: bytes
etag: "334cde9c1e9a17acbc4b7fc05810191319a69eec"
content-type: text/html; charset=utf-8
vary: Accept-Encoding
date: Thu, 19 Dec 2019 16:54:58 GMT
x-envoy-upstream-service-time: 1
server: envoy

<!doctype html>......</html>
nc localhost 80  0.00s user 0.01s system 0% cpu 22.138 total
```
Regression test performed as follows:
- Set `idle_timeout: 0s` in Ambassador module
- Open connection via netcat, send a request, wait
- Confirm that Envoy _never_ closes the connection, and that one can send another request some minutes later

Default test performed as follows:
- Ensure `idle_timeout` param is not defined in Ambassador module
- Open connection via netcat, send a request, wait
- Confirm that Envoy _never_ closes the connection, and that one can send another request some minutes later

I plan to deploy this change to our Staging and Production environments after we've ensured that our current configuration is compatible with the latest Ambassador (currently we are running 0.73)

## Todos
- [ ] Tests
- [x] Documentation


neufeldtech:
I can't seem to find the way to update the `json` acceptance files that the tests are failing on. Could someone point me in the correct direction on how to do this?

CamelCaseNotation:
@kflynn bump, this seems wildly useful. Any reason it won't get merged after the tests pass? I assume the default behavior is kept in place by having `idle_timeout: 0s` be the default.

python/ambassador/ir/irambassador.py
```diff
@@ -53,7 +53,8 @@ class IRAmbassador (IRResource):
         'use_proxy_proto',
         'use_remote_address',
         'x_forwarded_proto_redirect',
-        'xff_num_trusted_hops'
+        'xff_num_trusted_hops',
+        'idle_timeout'
```
kflynn: Minor request: keep this list alphabetized, please? 🙂

python/ambassador/envoy/v2/v2listener.py
```diff
@@ -670,6 +670,9 @@ def __init__(self, config: 'V2Config', listener: IRListener) -> None:
         if 'xff_num_trusted_hops' in config.ir.ambassador_module:
             base_http_config["xff_num_trusted_hops"] = config.ir.ambassador_module.xff_num_trusted_hops
 
+        if 'idle_timeout' in config.ir.ambassador_module:
```
kflynn: This will indeed break basically every test out there, because the tests check for changes in the generated `envoy.json`. I think if you do

```
idle_timeout = config.ir.ambassador_module.get('idle_timeout', '0s')

if idle_timeout != '0s':
  ...
```

then you'll find that tests pass, except for the test you've surely added to change the `idle_timeout` and make sure it works? 🙂

neufeldtech:
Opened #2164 as the new PR

PR #2130, opened 2019-12-19
Use proper executable name for Windows

ark3:
I let APro CI, which is the only thing that uses the modified script, test this for me.


ark3:
We should rationalize all our shell scripts at some point, follow a variable naming policy, run them all through `shellcheck` and `shfmt`, etc. But this isn't the place for that...

I considered letting `go build` pick the name, and then looking at what was there, but that was harder for a shell n00b like me, plus I'm not convinced it's _better_.

rhs:
Looks good to me! Will leave to @khussey to merge.

PR #2131, opened 2019-12-20
Drop duplicate resources on the way into the resourcefetcher.

kflynn:
In some cases (related to overlapping watch terms) `watt` can hand duplicates of the same resource to the Python side of Ambassador. This causes problems.

While the right longer-term answer is to make sure `watt` doesn't do that, changing `watt` is a pretty high-risk sort of thing for right now. Filtering the duplicates out is a lot safer.


rhs:
Looks good to me!

PR #2133, opened 2019-12-23
Cluster resolution relative to the resource's namespace, not the ambassador pod namespace

alexgervais:
## Description
Cluster resolution relative to the resource's namespace, not the ambassador pod namespace.

> We're going to change things such that the upstream service is assumed to be in the same namespace as the element referring to the upstream service so that e.g. a `Mapping` in namespace `foo` will look for its `service` in namespace `foo` unless the service explicitly mentions a different namespace.

> Since this is very definitely a breaking change, we will also support a way to turn it off from the ambassador `Module`: if you set `upstream_ambassador_namespace: true`, then Ambassador will follow the old behavior.

## Related Issues
None.

## Testing
- AmbassadorIDTest is configured with `upstream_ambassador_namespace: true`.
- All other tests `econf.json` were changed to reflect the added namespace to the `address` field.
- TODO: Validate ConsulTest and the need for `upstream_ambassador_namespace: true` if operating a `ConsulResolver`.

## Todos
- [ ] Tests
- [ ] Documentation


alexgervais:
Closing in favour of #2141

PR #2141, opened 2020-01-02
Service resolution relative to the resource's namespace, not the ambassador pod namespace, using Kubernetes Resolvers for Envoy cluster resolution

alexgervais:
## Description
Alternative solution to https://github.com/datawire/ambassador/pull/2133, where the service resolution is delegated to the underlying Resolver (`KubernetesEndpointResolver`, `KubernetesServiceResolver`, `ConsulResolver`) instead of being globally applied to the service name.

---

Cluster resolution relative to the resource's namespace, not the ambassador pod namespace.

> We're going to change things such that the upstream service is assumed to be in the same namespace as the element referring to the upstream service so that e.g. a `Mapping` in namespace `foo` will look for its `service` in namespace `foo` unless the service explicitly mentions a different namespace.

> Since this is very definitely a breaking change, we will also support a way to turn it off from the ambassador `Module`: if you set `use_ambassador_namespace_for_service_resolution: true`, then Ambassador will follow the old behavior.

## Related Issues
None.

## Testing
- AmbassadorIDTest is configured with `use_ambassador_namespace_for_service_resolution: true`.
- All other tests `econf.json` were changed to reflect the added namespace to the `address` field.
- Manual testing also highlighted the fact the namespace was previously only set on `IRBaseMapping`.

## Todos
- [x] Tests
- [x] Documentation


alexgervais:
For reference, AES tests passed in https://github.com/datawire/apro/pull/759 as well as manual smoke tests locally.

CamelCaseNotation:
Woot woot! Let's get this merged in! :D

docs/reference/core/ambassador.md
```diff
@@ -41,6 +41,7 @@ spec:
 | `server_name: envoy` | By default Envoy sets server_name response header to `envoy`. Override it with this variable |  |
 | `service_port: 8080` | If present, service_port will be the port Ambassador listens on for microservice access. If not present, Ambassador will use 8443 if TLS is configured, 8080 otherwise. |  |
 | `statsd` | Configures Ambassador statistics. These values can be set in the Ambassador module or in an environment variable. For more information, see the [Statistics reference](../../statistics#exposing-statistics-via-statsd). |  |
+| `upstream_ambassador_namespace` | Controls whether Ambassador will resolve upstream services assuming they are in the same namespace as the element referring to them, e.g. a Mapping in namespace `foo` will look for its service in namespace `foo`. If `true`, Ambassador will resolve the upstream services assuming they are in the same namespace as Ambassador, unless the service explicitly mentions a different namespace. | `upstream_ambassador_namespace: false` |
```
alexgervais: We're changing things such that the upstream service is assumed to be in the same namespace as the element referring to the upstream service so that e.g. a Mapping in namespace foo will look for its service in namespace foo unless the service explicitly mentions a different namespace.

Since this is very definitely a breaking change, we also support a way to turn it off from the ambassador Module: if you set `upstream_ambassador_namespace: true`, then Ambassador will follow the old behavior. The **default will be false**, meaning that Ambassador 1.0 will default to the **new** behavior.

cc. @richarddli , @kelseyevans I think this breaking change is important to highlight in the release blog post.

rhs: This name sounds to me like the value would be a namespace not a boolean.

alexgervais: I propose using `use_upstream_ambassador_namespace` as this would also fit the pattern of other Module configuration options.

rhs: How about `use_ambassador_namespace_for_service_resolution` just to be clear that you're not talking about "upstream ambassadors".

PR #2143, opened 2020-01-03
make {pull,push}-docs: Expect to use a fixed git-subtree

LukeShu:
## Description

This adds a preflight-check to `make pull-docs` and `make push-docs` that checks whether the installed `git subtree` behaves reasonably.  If it doesn't, then it errors and prints instructions for installing a better `git subtree`.

This lets us drop all of the hacks in `make pull-docs`, and most of the hacks in `make push-docs` (we still need to say `--onto`).

This has substantial devex improvement that `make push-docs` on my laptop now takes about 7 seconds instead of about 40 minutes.  Oh, and also the progress output is actually kinda helpful, and isn't the worse-than-nothing garbage that was there before.  Here's a screencast of that gifrecord_2020-01-03_16:21:10 190445364

Also while I'm at it, for the time-being, have PUSH_BRANCH and PULL_BRANCH default to `early-access` instead of `master`.

## Testing

I did test `make pull-docs` and `make push-docs` (recorded above).  Not in the recording is me inspecting the results to make sure that sane things happened.

### DIFF
--- a/build-aux-local/docs.mk
+++ b/build-aux-local/docs.mk
@@ -1,23 +1,45 @@
 # This is the branch from ambassador-docs.git to pull for "make pull-docs".
 # Override if you need to.
-PULL_BRANCH ?= master
+PULL_BRANCH ?= early-access
+
+# This is the branch from ambassador-docs.git to push to for "make push-docs".
+# Override if you need to.
+PUSH_BRANCH ?= early-access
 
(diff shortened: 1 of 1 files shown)


LukeShu:
I would be super duper happy if we can land this before I cut an EA at 4pm.

LukeShu:
Yeah, eventually. :)

Issue #739, opened 2018-08-22
Request bodies that go through an external authentication service go missing once they hit their destination after version 0.34.1


FuzzOli87:
**Describe the bug**
After updating to Ambassador 0.38.0 We started seeing issues with some of our services returning bad request errors, turns out that bodies were missing from POST requests. After some experimentation, it turns out that this behavior starts after 0.34.1(0.34.2 switched over to envoy 1.6) 

My setup is running with Istio 1.0.0 which might be causing an issue due to Istio version mismatching.

**To Reproduce**
1. Deploy Ambassador and Istio 
2. Make a POST request that goes through an external auth service
3. Destination service will not get the body

**Expected behavior**
Destination service should receive the body of the original request.

**Versions (please complete the following information):**
 - Ambassador: 0.38.0, 0.34.2
 - Kubernetes environment: AWS via KOPS
 - Version: 1.9.8

feitnomore:
Any news on this one? Is this still happening on new versions?

PR #2150, opened 2020-01-07
Add a patch_client optional query param to /ambassador/v0/diag/ to send a partial representation in JSON Patch format [apro #739]

alexgervais:
## Description
Improving the performance of JSON diag snapshots fetched from AES.
Added a `patch_client` optional query param to /ambassador/v0/diag/ to send a partial representation in [JSON Patch](http://jsonpatch.com/) format.

## Related Issues
None.

## Testing
Manual tests with `curl` and AES Edge Policy Console.

## Todos
- [x] Tests
- [ ] Documentation


builder/requirements.txt
```diff
@@ -13,6 +13,8 @@ scout.py==0.5.0
 semantic-version==2.6.0
 kubernetes==8.0.0
 click==7.0
+jsonpatch==1.24
```
alexgervais: BSD License, https://pypi.org/project/jsonpatch/

builder/requirements.txt
```diff
@@ -13,6 +13,8 @@ scout.py==0.5.0
...
+expiringdict==1.2.0
```
alexgervais: Apache 2 License, https://pypi.org/project/expiringdict/

python/ambassador_diag/diagd.py
```diff
@@ -56,6 +60,7 @@
 __version__ = Version
 
 boot_time = datetime.datetime.now()
+tvars_cache = ExpiringDict(max_len=10, max_age_seconds=60)
```
alexgervais: We need to keep previous representations of `tvars` in memory to diff between each client request.
We could theoretically support 10 clients, refreshing at least every 60 seconds -- or they'll be evicted.
This cache will be memory intensive!

python/ambassador_diag/diagd.py
```diff
@@ -481,11 +486,24 @@ def show_overview(reqid=None):
...
 
+    patch_client = request.args.get('patch_client', None)
     if request.args.get('json', None):
         key = request.args.get('filter', None)
 
         if key:
             return jsonify(tvars.get(key, None))
+        elif patch_client:
+            # Get the previous full representation
+            cached_tvars_json = tvars_cache.get(patch_client, "{}")
```
kflynn: Is there a reason to save the JSON serialization of the dict, rather than the dict itself? It seems that saving the dict itself would save you a `json.loads()` later...

alexgervais: True, let me test that real quick.

python/ambassador_diag/diagd.py
```diff
@@ -481,11 +486,24 @@ def show_overview(reqid=None):
...
             return jsonify(tvars.get(key, None))
+        elif patch_client:
+            # Get the previous full representation
+            cached_tvars_json = tvars_cache.get(patch_client, dict())
+            # Serialize the tvars into a json-string using the same jsonify Flask serializer, then load the json object
+            response_content = json.loads(flask_json.dumps(tvars))
+            # Diff between the previous representation and the current full representation  (http://jsonpatch.com/)
+            patch = jsonpatch.make_patch(cached_tvars_json, response_content)
+            # Save the current full representation in memory
+            tvars_cache[patch_client] = response_content
```
kflynn: Same comment here. If you save `tvars` instead of `response_content` you should be able to just directly do

```
cached_tvars = tvars_cache.get(patch_client, {})
patch = jsonpatch.make_patch(cached_tvars, tvars)
tvars_cache[patch_client] = tvars
```

no?

alexgervais: I experimented with that, but the library is picky and uses a different json encoder than the jsonify one and would spit out serialization errors:
```
TypeError: Object of type datetime is not JSON serializable
```

I played on the safe-side and reused the Flask json encoder for comparing the resuling serialized output vs the dict content.

kflynn: Oh FFS. OK, good enough for now.

PR #2152, opened 2020-01-08
Fix Knative integration in AES [apro #769]

concaf:
Prior to this commit, the mappings being generated from Knative Ingress resources did not have a namespace assigned to them. Also, the namespace was being fetched from the mapping dict without `.get()` method. This PR fixes both of these issues to fix Knative integration.


python/ambassador_diag/diagd.py
```diff
@@ -1051,7 +1051,7 @@ def _load_ir(self, rqueue: queue.Queue, aconf: Config, fetcher: ResourceFetcher,
 
         if mappings:
             for mapping_name, mapping in mappings.items():
-                app.kubestatus.mark_live("Mapping", mapping_name, mapping.namespace)
+                app.kubestatus.mark_live("Mapping", mapping_name, mapping.get('namespace'))
```
kflynn: This should have a default -- probably `mapping.get('namespace', Config.ambassador_namespace)`?

kflynn:
BTW nice catch! and mea culpa for breaking it. 🙁

python/ambassador_diag/diagd.py
concaf: Fixed

PR #2153, opened 2020-01-08
Add test for Knative 0.11.0, remove tests for older versions [apro #769]

concaf:
Add test for Knative 0.11.0


concaf:
@inercia requested a review from you because you're the only one with the hardware to run Knative tests right now :P

concaf:
Wow, the git diff looks very weird because of all the YAML. Let me explain what I did.

I added the 0.11.0 tests on the very same lines of older Knative tests.
Then I updated the CRDs to 0.11.0.
Then I removed older tests and their manifest files.

The diff looks very weird because the new and old manifests have a lot in common, so maybe look at the commits separately!?

PR #2157, opened 2020-01-08
Support Host.requestPolicy for TLS and HTTP [apro #721]

kflynn:
This is a whole lot of Listener rework to:

- support `requestPolicy`
- never ever reply with 301s to ACME challenges
- give Edge Stack a fighting chance of warning users about scary browser warnings
- ditch some of the evilness about SNI in the past so that _we_ have a fighting chance of being able to reason about SNI

This works by linking everything back to the `Host` resource: we create an `IRListener` for each `Host`/port combination, and save what to do for secure requests and insecure requests. Once we scan the `Host`s, we look for any `TLSContexts` that weren't associated with a `Host` and synthesize `IRListener`s as if they were `Host`s, too.

If we have no `Host`s and no `TLSContext`s, we synthesize as if we had a `Host` for `*`.

**VERY IMPORTANT**: we don't use `require_tls` any more. Instead we use explicit header matching and the `redirect` action. 

**VERY IMPORTANT**: Ambassador was previously often pretty lax about SNI header matching, and this change makes it considerably more strict. If you use `Host`s in OSS-land, and you don't give a `Host` matching `*`, you'll need to provide SNI for anything to work.

Best to review this commit-by-commit, to expect this description to change quickly, and to plan to just reread the following classes:

- `IRListener` 
- `V2VirtualHost`
- `V2Listener`

Testing: OSS CI passes. Edge Stack hand testing for ACME and the 404 page, so far.


kflynn:
This is for https://github.com/datawire/apro/issues/721 and https://github.com/datawire/apro/issues/735.

python/ambassador/envoy/v2/v2listener.py
```diff
@@ -512,380 +618,389 @@ def __init__(self, config: 'V2Config', listener: IRListener) -> None:
...
+                            logger.info(
+                                f"V2Listeners: {listener.name} {vhostname} {variant}: Drop due to Reject action")
+                            action = "Drop"
+                        else:
+                            logger.info(
+                                f"V2Listeners: {listener.name} {vhostname} {variant}: Accept as {action}")
+
+                        if action != 'Reject':
+                            vhost.append_route(route)
+                        # elif action == 'Redirect':
```
concaf: Do we need this?

python/ambassador_diag/diagd.py
```diff
@@ -1340,6 +1340,12 @@ def validate_envoy_config(self, config, retries) -> bool:
             self.logger.info("successfully validated the resulting envoy configuration, continuing...")
             return True
 
+        try:
+            decoded_error = odict['output'].decode('utf-8')
+            odict['output'] = decoded_error
+        except:
+            pass
```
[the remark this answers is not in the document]
kflynn: Yes. What happens here is that `odict['output']` should be a `bytes` object, but if you dump such a thing into the logs and it has embedded newlines (as the Envoy error outputs tend to), it's hard to read. So if we can decode it, great! but if anything at all goes wrong, we haven't overwritten `odict['output']` so we don't lose the error.

python/ambassador/envoy/v2/v2listener.py
kflynn: It's actually about to come back (in a slightly modified fashion) for Edge Stack. So stand by. 🙂

kflynn:
@alexgervais The ACME hole-punching won't synthesize a `Mapping` for `/.well-known/acme-challenge/` if you've already supplied one, so that should be fine.

api/getambassador.io/v2/Host.proto
```diff
@@ -96,3 +113,24 @@ message ACMEProviderSpec {
   // This is normally set automatically
   string registration = 4;
 }
+
+message RequestPolicy {
+  // How shall we handle insecure requests?
```
rhs: We should explain how we decide whether a given request is secure or not. I believe we had spoken about  separating the idea of a request being secure/insecure from the way it happens to arrive, i.e. classifying a request as insecure should be based on X-Forwarded-Proto, not port, so that if e.g. an ELB terminates TLS and sends it to ambassador over a cleartext connection, it is able to still be considered secure if that's what the user wants.

api/getambassador.io/v2/Host.proto
```diff
@@ -1,5 +1,19 @@
 /**
- * Host defines a way that an Ambassador will be visible to the 
+ * === HACK HACK HACK ===
+ *
+ * The existence of Host_nojson.proto is because if we bring in all the
+ * k8s.io types that we would want to, the generated *.pb.json.go files try
+ * to import github.com/gogo/protobuf/jsonpb, and that turns out to crash on
```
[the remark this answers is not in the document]
kflynn: Not gonna argue there.

api/getambassador.io/v2/Host.proto
```diff
@@ -96,3 +113,24 @@ message ACMEProviderSpec {
...
+
+  // Later we may define a 'secure' section too.
+}
+
+message InsecureRequestPolicy {
+  // What action should be taken for an insecure request?
+  InsecureRequestAction action = 1;
+
+  // Is there an additional insecure port we should listen on?
+  uint32 additionalPort = 2;
```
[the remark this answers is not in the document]
kflynn: I'm assuming that the `additionalPort` will be rarely used, and that when it is used, it'll be by Julian -- that's why it's associated with a `Host`, and why it's explicitly "this is an insecure port" rather than "this is a cleartext port".

We do need it, to support the scenario of a split L4 load balancer -- this was spelled out in the doc about the `requestPolicy`, remember? 🙂

api/getambassador.io/v2/Host.proto
kflynn: In fact, it _is_ completely based on `X-Forwarded-Proto`! with the single exception that a  `requestPolicy.insecure.additionalPort` is defined to always be insecure.

PR #2158, opened 2020-01-08
Remove "serialization" keys from UI snapshots (served via patch) [apro #741]

ark3:
Assume only the Admin UI asks for patch-based updates. Recursively drop all
"serialization" keys. This avoids leaking secrets and generally makes the
snapshot a lot smaller without losing information that the Admin UI cares about.


ark3:
How would that happen? The only way to reach this endpoint from the outside is via the Go code, right? And that code always submits a patch client ID.

I guess users can choose to expose that endpoint directly. Perhaps some variant of our sample YAML does that. But we can't just drop the serialization everywhere always; it exists for a reason. Perhaps we should figure that out.

alexgervais:
@ark3 Anybody who exposes their `/ambassador/v0/diag/` endpoint could see their secrets leak. :/

ark3:
Okay, then we need to figure out why those serializations exist, how they are used, etc. I'll get on it.

ark3:
I think serialization manipulation should be in a separate PR; it touches a different part of the code. I think we should land this in the meantime. Thoughts?

@alexgervais Can I ask to re-review? Apparently my rebase push came in just after your review. Sorry about that!

PR #2159, opened 2020-01-09
Pass DOCKER_NETWORK in the builder

inercia:
## Description

* Allow to pass the `DOCKER_NETWORK` in the builder, so the builder containers are connected to the same network as your local kuberenetes cluster when using something like k3d.
* Minor fixes.

### DIFF
--- a/builder/builder.sh
+++ b/builder/builder.sh
@@ -33,8 +33,8 @@ DOCKER_RUN=${DOCKER_RUN:-docker run}
 # note: use your local k3d/microk8s/kind network for running tests
 DOCKER_NETWORK=${DOCKER_NETWORK:-${BUILDER_NAME}}
 
-builder() { docker ps -q -f label=builder -f label=${BUILDER_NAME}; }
-builder_network() { docker network ls -q -f name=${DOCKER_NETWORK}; }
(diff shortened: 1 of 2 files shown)


builder/builder.sh
```diff
@@ -282,7 +282,7 @@ case "${cmd}" in
         # This runs inside the builder image
         for MODDIR in $(find-modules); do
             module=$(basename ${MODDIR})
-            eval "$(grep BUILD_VERSION apro.version 2>/dev/null)" # this will `eval ''` for OSS-only builds, leaving BUILD_VERSION unset; don't embed the version-number in OSS Go binaries
+            eval "$(grep BUILD_VERSION apro.version 2>/dev/null)" # this will `eval ''` for OSS-only builds, leaving BUILD_VERSION unset; dont embed the version-number in OSS Go binaries
```
concaf:
```suggestion
            eval "$(grep BUILD_VERSION apro.version 2>/dev/null)" # this will `eval ''` for OSS-only builds, leaving BUILD_VERSION unset; don't embed the version-number in OSS Go binaries
```

inercia: This is for a very silly reason: that `'` breaks my syntax highlighting! :sweat_smile:

PR #2162, opened 2020-01-09
Fix Knative manifest parsing in tests and Ambassador ID exemption [apro #769]

concaf:
This PR fixes 2 things.

1. Knative manifests have CRDs in 2 separate places - there is separate CRDs file, plus CRDs are present in the serving manifest as well. This PR removes CRDs from the serving manifest as a result. Also, comments and examples that have no impact on the configuration have been removed. Most of this has been done because these extra bits in the manifests have strings like `status: {}` or `{name: example}` which confuses our YAML parses because post parsing we format the YAML as a string for variable substitution and that gets broken causing an `IndexError`. FTR, the code that breaks is here - https://github.com/datawire/ambassador/blob/8bd96c648b0b2de2b6c19354b8b23950c8eed98f/python/kat/harness.py#L542

2. All Knative ClusterIngress and Ingress objects are exempt from having an Ambassador ID because these resources are auto-generated. We already had code for this, but it was broken. This PR fixes that as well.


python/tests/manifests/knative_serving_0.11.0.yaml
```diff
@@ -27,10 +28,6 @@ rules:
   - list
   - watch
 
----
```
concaf: Removing whitespace and noise here

python/tests/manifests/knative_serving_0.11.0.yaml
```diff
@@ -329,408 +326,6 @@ subjects:
   name: controller
   namespace: knative-serving
 
----
-
----
-
----
-apiVersion: apiextensions.k8s.io/v1beta1
```
concaf: Removing CRDs which have already been applied from CRD manifest

python/tests/manifests/knative_serving_0.11.0.yaml
```diff
@@ -1292,7 +885,7 @@ data:
     # container name, if none is specified.  This field supports
     # Go templating and is supplied with the ObjectMeta of the
     # enclosing Service or Configuration, so values such as
-    # {{.Name}} are also valid.
```
concaf: Removing characters in comments which break our parsing

python/tests/manifests/knative_serving_0.11.0.yaml
```diff
@@ -1442,30 +1033,6 @@ data:
     # These sample configuration options may be copied out of
     # this example block and unindented to be in the data block
     # to actually change the configuration.
-
-    # Common configuration for all Knative codebase
```
concaf: Removing sample configuration that has no impact

python/tests/manifests/knative_serving_0.11.0.yaml
```diff
@@ -1485,122 +1052,6 @@ metadata:
 ---
 apiVersion: v1
 data:
-  _example: |
```
concaf: Removing examples

python/ambassador/config/config.py
```diff
@@ -74,7 +74,7 @@ class Config:
         'logservice': "log_services",
     }
 
-    KnativeResources = { 'ClusterIngress', 'KnativeIngress' }
```
concaf: This allows Ambassador ID to be ignored

Issue #2155, opened 2020-01-08
Allow configuring idle_timeout Envoy parameter on HTTP Connection Manager


neufeldtech:
**Please describe your use case / problem.**
We currently have Ambassador deployed on-premise behind a firewall and fronted by a CDN. We encountered an issue where on low-traffic services, our clients were receiving some 503 first byte timeouts. 

The issue is present because when the firewall chose to remove idle tcp sessions from its session table, it would not send a FIN or RST to the CDN. On the next request, the CDN thinks that it still has a tcp connection open and attempts to re-use the now-closed TCP connection, resulting in the request getting 'black-holed', and leads to an HTTP 503.

Because Envoy doesn't ever attempt to close idle tcp connections by default, we're left with Envoy thinking that sessions are open, but our firewall has closed them after a 30 minute timeout (without notifying the CDN that the connection is closed), and the CDN still thinks that the connection is open.

Our "options" for remediating this scenario were:
- Configure tcp idle timeout on the CDN side to control connection idle time
  - This turned out to not be an option that we can readily control on our CDN. They have 'dynamic' connection times.
- Make the firewall send a FIN or RST to the CDN when it removes a session from its session table
  - We investigated this as well, but on our firewall this is not an option that is configurable.
- Have Envoy close idle tcp connections gracefully after a configurable timeout (in our case, some timer that is lower than our firewall idle session timeout)
  - We believe that this is the most appropriate way to solve the issue, and gives us control when tuning idle timeouts across each network device in the path, so this issue requests making the `idle_timeout` Envoy setting configurable on the Envoy http listener.

**Describe the solution you'd like**
Would like to be able to have the `idle_timeout` on the downstream HTTP Listener configurable for the reason mentioned above.

**Describe alternatives you've considered**
See description above.

**Additional context**
Looks like #1738 originally brought this up, but was closed as stale

I originally opened #2126 for this, but because of Github shenanigans the PR got closed when I was updating my fork. Created this issue for tracking, then will link PR's to this when I have one ready.

neufeldtech:
Closing - implemented in #2164

Issue #1738, opened 2019-08-06
Add support for `idle_timeout` Envoy parameter for HTTP connection manager 


ankurpshah:
Currently, there is no way to override/configure [HTTP connection manager's `idle_timeout`](https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/network/http_connection_manager/v2/http_connection_manager.proto.html?highlight=idle_timeout#envoy-api-field-config-filter-network-http-connection-manager-v2-httpconnectionmanager-drain-timeout) parameter of Envoy. Please add support for same.

AaronTriplett:
I'll likely pick this up in the next few weeks if no one else gets to it. In the meantime, route/mapping level `idle_timeout_ms` is the way to go.

khussey:
Duplicate of #2155.

PR #2164, opened 2020-01-09
Allow configuring idle_timeout Envoy parameter on HTTP Connection Manager

neufeldtech:
## Description
We currently have Ambassador deployed on-premise behind a firewall and fronted by a CDN. We encountered an issue where on low-traffic services, our clients were receiving some 503 first byte timeouts. 

The issue is present because when the firewall chose to remove idle tcp sessions from its session table, it would not send a FIN or RST to the CDN. On the next request, the CDN thinks that it still has a tcp connection open and attempts to re-use the now-closed TCP connection, resulting in the request getting 'black-holed', and leads to an HTTP 503.

Because Envoy doesn't ever attempt to close idle tcp connections by default, we're left with Envoy thinking that sessions are open, but our firewall has closed them after a 30 minute timeout (without notifying the CDN that the connection is closed), and the CDN still thinks that the connection is open.

Our "options" for remediating this scenario were:
- Configure tcp idle timeout on the CDN side to control connection idle time
  - This turned out to not be an option that we can readily control on our CDN. They have 'dynamic' connection times.
- Make the firewall send a FIN or RST to the CDN when it removes a session from its session table
  - We investigated this as well, but on our firewall this is not an option that is configurable.
- Have Envoy close idle tcp connections gracefully after a configurable timeout (in our case, some timer that is lower than our firewall idle session timeout)
  - We believe that this is the most appropriate way to solve the issue, and gives us control when tuning idle timeouts across each network device in the path, so this PR addresses making the `idle_timeout` Envoy setting configurable on the Envoy http listener.

## Related Issues
#2155 
#1738 
#2126 

## Testing
- Manual testing via netcat with ambassador running locally in docker-for-desktop Kubernetes cluster.
- Will run in our staging environment once we upgrade our stack to Ambassador v1.0

### Manual testing:
Default configuration - observe that the TCP connection can remain open for (at least) 2 mins and 30 seconds. I sent a request, waited a few mins, then was able to send an additional request over the same connection.

```bash
$ time nc localhost 80
GET / HTTP/1.1
Host: localhost
HTTP/1.1 200 OK
content-length: 2002
content-disposition: inline; filename="index.html"
accept-ranges: bytes
etag: "334cde9c1e9a17acbc4b7fc05810191319a69eec"
content-type: text/html; charset=utf-8
vary: Accept-Encoding
date: Thu, 09 Jan 2020 20:15:36 GMT
x-envoy-upstream-service-time: 2
server: envoy
<!doctype html>....</html>
GET / HTTP/1.1
Host: localhost
HTTP/1.1 200 OK
content-length: 2002
content-disposition: inline; filename="index.html"
accept-ranges: bytes
etag: "334cde9c1e9a17acbc4b7fc05810191319a69eec"
content-type: text/html; charset=utf-8
vary: Accept-Encoding
date: Thu, 09 Jan 2020 20:18:03 GMT
x-envoy-upstream-service-time: 4
server: envoy
<!doctype html>....</html>

nc localhost 80  0.00s user 0.01s system 0% cpu 2:31.62 total
```

Now with `idle_timeout: 30s` setting configured - observe that the TCP connection is closed by envoy after 30 seconds of being idle. I send one request, then it gets gracefully closed ~30s later because there is no further activity.

```bash
$ time nc localhost 80
GET / HTTP/1.1
Host: localhost
HTTP/1.1 200 OK
content-length: 2002
content-disposition: inline; filename="index.html"
accept-ranges: bytes
etag: "334cde9c1e9a17acbc4b7fc05810191319a69eec"
content-type: text/html; charset=utf-8
vary: Accept-Encoding
date: Thu, 09 Jan 2020 20:21:05 GMT
x-envoy-upstream-service-time: 1
server: envoy
<!doctype html>....</html>
nc localhost 80  0.00s user 0.01s system 0% cpu 37.362 total
```

## Todos
- [x] Tests
- [x] Documentation

## Other


neufeldtech:
@kflynn  Please review when you are able

docs/reference/core/ambassador.md
```diff
@@ -35,6 +35,7 @@ spec:
 | `envoy_log_format` | Defines the envoy log line format. See [this page](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log) for a complete list of operators | See [this page](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log#default-format-string) for the standard log format. |
 | `envoy_log_path` | Defines the path of log envoy will use. By default this is standard output | `envoy_log_path: /dev/fd/1` |
 | `envoy_log_type` | Defines the type of log envoy will use, currently only support json or text | `envoy_log_type: text` |
+| `idle_timeout` | Controls how Envoy configures the tcp idle timeout on the http listener. Default is no timeout (TCP connection may remain idle indefinitely). NOTE: only s (seconds) is supported for the duration suffix | `idle_timeout: 300s` |
```
kflynn: Everywhere else, we do timeouts like `idle_timeout_ms` with an integer value in milliseconds. I kinda hate to say it, but let's change to that here as well?

neufeldtech: Thanks for the feedback.

I'll take a look at changing the value to accept an integer rather than a duration.

Regarding the name change from `idle_timeout`... If we change it to `idle_timeout_ms`, then this directly collides with the [idle_timeout_ms](https://www.getambassador.io/reference/timeouts/#idle-timeout-idle_timeout_ms) field on the upstream connection side.

Any suggestions or proposals on what to name it so that we can land somewhere close to the envoy naming, but still have it distinct from the existing attribute. Technically, this property is on the http listener, so maybe something like `listener_idle_timeout_ms` ? Are there other examples of similarly-named attributes that have to exist both on the upstream and the downstream?

kflynn: Oh, crap, I missed that collision. 😐 

We do sometimes use the names of Envoy things to differentiate, yes. I'd vote for probably `client_idle_timeout_ms`? but `listener_...` is OK with me too.

PR #2185, opened 2020-01-13
Make kubestatus logging quieter [ambassador 2186]

kflynn:
We log _way_ too much crap that around `kubestatus` that was only meant as debugging logs.

CHANGELOG: reduce log clutter around status updates

### DIFF
--- a/python/ambassador_diag/diagd.py
+++ b/python/ambassador_diag/diagd.py
@@ -703,7 +703,7 @@ class KubeStatus:
     def mark_live(self, kind: str, name: str, namespace: str) -> None:
         key = f"{kind}/{name}.{namespace}"
 
-        print(f"KubeStatus MASTER {os.getpid()}: mark_live {key}")
+        # print(f"KubeStatus MASTER {os.getpid()}: mark_live {key}")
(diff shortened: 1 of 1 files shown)


python/ambassador_diag/diagd.py
```diff
@@ -1099,7 +1100,7 @@ def _load_ir(self, rqueue: queue.Queue, aconf: Config, fetcher: ResourceFetcher,
                 kind, namespace, update = app.ir.k8s_status_updates[name]
                 text = json.dumps(update)
 
-                self.logger.info(f"K8s status update: {kind} {resource_name}.{namespace}, {text}...")
+                # self.logger.info(f"K8s status update: {kind} {resource_name}.{namespace}, {text}...")
```
alexgervais: Do we want to keep at least 1 line of log?

ark3: There's still "KubeStatus UPDATE <pid>" and "KubeStatus DONE"

kflynn: We have output when the kubestatus command actually runs, still.

PR #2190, opened 2020-01-13
Force the eviction tvars cached elements with max_age_seconds [apro #806]

alexgervais:
## Description
Force the eviction tvars cached elements with max_age_seconds. Should improve stability when running low on memory.

## Related Issues
https://github.com/datawire/apro/issues/806
https://github.com/datawire/apro/issues/827

## Testing
Manually tested the eviction and memory footprint.

## Todos
- [x] Tests
- [ ] Documentation

### DIFF
--- a/python/ambassador_diag/diagd.py
+++ b/python/ambassador_diag/diagd.py
@@ -198,6 +198,11 @@ def standard_handler(f):
 
         app.logger.debug("%s handler %s" % (prefix, func_name))
 
+        # getting elements in the `tvars_cache` will make sure eviction happens on `max_age_seconds` TTL
+        # for removed patch_client rather than waiting to fill `max_len`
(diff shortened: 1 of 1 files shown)


python/ambassador_diag/diagd.py
```diff
@@ -465,6 +467,11 @@ def check_ready():
 def show_overview(reqid=None):
     app.logger.debug("OV %s - showing overview" % reqid)
 
+    # getting elements in the `tvars_cache` will make sure eviction happens on `max_age_seconds` TTL
```
kflynn: Maybe do this in the detail case as well? or even in the `standard_handler`...

alexgervais: Great suggestion, thanks! I moved the eviction to the `standard_handler`.

PR #2194, opened 2020-01-14
Update Helm docs

kflynn:
Helm docs. Oh yeah.

### DIFF
--- a/docs/user-guide/helm.md
+++ b/docs/user-guide/helm.md
@@ -1,87 +1,100 @@
 # Installing Ambassador with Helm
 
-&nbsp;
-&nbsp;
-&nbsp;
-
-| Warning! |
-| :---------------------------:|
-|This installation method is not supported for Early Access release. Check back soon! |
-
-&nbsp;
-&nbsp;
-&nbsp;
-&nbsp;
-
 [Helm](https://helm.sh) is a package manager for Kubernetes that automates the release and management of software on Kubernetes. The Ambassador Edge Stack can be installed via a Helm chart with a few simple steps, depending on if you are deploying for the first time, or upgrading from an existing installation.
 
 ## Prerequisites
 
-The Ambassador Edge Stack Helm chart is hosted in the
-[helm/charts](https://github.com/helm/charts) repository. Add this repo to your
-helm client with:
+The Ambassador Edge Stack Helm chart is hosted by Datawire and published at https://getambassador.io.
+Start by adding this repo to your helm client with:
 
 ```bash
-helm repo add stable https://kubernetes-charts.storage.googleapis.com/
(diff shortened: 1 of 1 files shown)


docs/user-guide/helm.md
```diff
@@ -1,87 +1,100 @@
...
 
 ```
-kubectl apply -f https://www.getambassador.io/early-access/yaml/ambassador/ambassador-edge-stack-resources.yaml # FIXME: this URL is broken
+helm install --name ambassador --namespace ambassador datawire/ambassador
 ```
 
-The Helm chart deploys The Ambassador Edge Stack but does not create the Edge Stack resources necessary for authentication, rate limiting, automatic HTTPS, etc.
+This will install the necessary deployments, RBAC, Custom Resource Definitions, etc. for the Ambassador Edge Stack to route traffic. Details on how to configure Ambassador using the Helm chart can be found in the Helm chart [README](https://github.com/datawire/ambassador-chart/tree/master).
+
+**Note:** The Helm chart deploys the Ambassador Edge Stack, but it does not create the Edge Stack resources necessary for authentication, rate limiting, automatic HTTPS, etc.
```
scoyle391: I would link to where the user can create those things

docs/user-guide/helm.md
```diff
@@ -1,87 +1,100 @@
...
 
-## Upgrade an Existing Installation
+## Upgrading an Existing Edge Stack Installation
 
-Upgrading your existing Ambassador Edge Stack is as simple as upgrading any other Helm release. Complete the following steps:
+**Note:** If your existing installation is not already running Ambassador **Edge Stack** as opposed to Ambassador Open Source, **do not use these instructions**. See "Migrating to Ambassador Edge Stack" below.
```
scoyle391: If your existing installation is not already running the Ambassador **Edge Stack** (as opposed to the Ambassador API Gateway), **do not use these instructions**.

docs/user-guide/helm.md
```diff
@@ -1,87 +1,100 @@
...
+helm upgrade ambassador datawire/ambassador
 ```
 
 This will upgrade the image and deploy and other necessary resources for the Ambassador Edge Stack. 
 
-**Note:** It is strongly encouraged for you to move your Ambassador release to the `ambassador` namespace with the following command: 
-`helm upgrade --namespace ambassador ambassador stable/ambassador`
+## Migrating to Ambassador Edge Stack
+
+If you have an existing Ambassador Open Source installation, but are not yet running Ambassador Edge Stack, the upgrade process is somewhat different than above.
```
scoyle391: Ambassador API Gateway (we want to stay away from calling it the OSS bc it's confusing - richard's request)

kflynn: 👍

PR #2212, opened 2020-01-18
Move Knative test to an independent test

### LINKED ISSUE #2196: Make knative test an independent test
Right now the Knative test is tested as an `AmbassadorTest` in kat. This presents some interesting challenges because it does not fit the interface very well. For example, we need to apply Knative manifests first and kservice second, but kat does not give us that. And, our queries to kservice need to timeout after a while because the pod takes some time to start after the first query, this is also not easy with Kat.

So, for reasons like these we need to make it independent - in the likes of watt scaling and docker test.

concaf:
This commit moves testing Knative from Kat to an independent test.
This is done because we need apply Knative and Ambassador manifests
sequentially and we need to wait for the serverless pod to come up.
Broader benefit of this is that the Knative test should not flake
now.

Fix #2196


kflynn:
(Obviously the tests do need to actually pass before landing this though. :wink:)

inercia:
Getting an error error on the KAT tests:

```
STDERR None
Running command ['kubectl', 'wait', '--timeout=90s', '--for=condition=Ready', 'pod', '-l', 'app=activator', '-n', 'knative-serving']
STDOUT 
STDERR None
```

concaf:
@inercia can you try with kube version 14/15.

@kflynn -
> Looks good! I assume that the utils.py functions are copies from before?

Yeah, it's mostly stuff that's common between knative and watt scaling tests.

Also, I've updated the PR (see last couple of commits). The reason this was failing in CI was that the code that determines if the cluster is compatible with Knative was not reachable, so I fixed it.

concaf:
I've tested this with 1.14 and 1.15 on GKE, I think we can move ahead with this.

Issue #881, opened 2018-10-12
EA doesn't seem to like `alpn_protocols: h2`.


nbkrause:
**Describe the bug**
EA doesn't seem to like `alpn_protocols: h2`. Connections timeout when `alpn_protocols: h2` is present in the TLS module. Ambassador does not crash and does not log any error or requests. 
cURL error:
```bash
curl -v https://<ambassador-ex-ip>/httpbin/
*   Trying 35.227.87.184...
* TCP_NODELAY set
* Connected to nkrause-ambassador.k736.net (35.227.87.184) port 443 (#0)
* Operation timed out after 300244 milliseconds with 0 out of 0 bytes received
* Closing connection 0
curl: (28) Operation timed out after 300244 milliseconds with 0 out of 0 bytes received
```

**To Reproduce**
Configure tls with `alpn_protocols: h2`
```yaml
      ---
      apiVersion: ambassador/v0
      kind:  Module
      name:  tls
      ambassador_id: ambassador-tls
      config:
        server:
          enabled: true
          secret: user-secret
          redirect_cleartext_from: 80
          alpn_protocols: h2
```

**Expected behavior**
Support for `alpn_protocols: h2`. 

**Versions (please complete the following information):**
 - Ambassador: 0.50.0-ea1
 - Kubernetes environment GKE
 - Version 1.9.7

kflynn:
Bumping to EA4 -- sorry for the delay!

kflynn:
Fixed in EA6.

PR #2222, opened 2020-01-22
Adding custom tvars dict filter for webui [apro #881]

alexgervais:
## Description
Customer filter of the diagd json output, specially catered to the `webui` required fields.

## Related Issues
Relates to https://github.com/datawire/apro/issues/881

## Testing
With 200 mappings:
```
~ curl --output /dev/null "http://localhost:8877/ambassador/v0/diag/?json=true"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 4762k  100 4762k    0     0  1283k      0  0:00:03  0:00:03 --:--:-- 1283k
```
```
~ curl --output /dev/null "http://localhost:8877/ambassador/v0/diag/?json=true&filter=webui"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 50985  100 50985    0     0   160k      0 --:--:-- --:--:-- --:--:--  160k
```

## Todos
- [x] Tests
- [ ] Documentation

## Other


ark3:
This is a great improvement! But I'm concerned it will be difficult to maintain. I'm sure you've tested the Web UI to make sure everything works with these filters _now_, but how do we keep the UI and the filters in sync? Would it be ~possible~ practical to keep the filter specification closer to where the data is consumed?

alexgervais:
@ark3 great valid feedback and concern. I too have a hard time reasoning about the schema-less diag output. I can see value in moving the field selector to `webui` and passing them explicitly in the http request, but I'm not sure it would make future evolutions easier... adding fields to `diag` or consuming extra fields in `webui` means understanding the json representation which is the non-trival task.

alexgervais:
More thoughts on having the frontend pass in the filter query: Since the requested structure has deeply nested elements and filters, I think a good path forward for the webui is to consume this data from a graphql endpoint, rather than a simple "rest" call.
This would mean a structured dataset with a schema and a common filter API.

alexgervais:
@kflynn I'm actually keeping `precedence ` in `route_info`. The listed keys are the ones that are kept:

```
unwanted_keys = set(d) - set(keys_to_keep)
for unwanted_key in unwanted_keys: del d[unwanted_key]
```

Issue #2227, opened 2020-01-23
`docker run --demo` doesn't work with Edge Stack


kflynn:
If you run 

```
docker run -it -p 8080:8080 --name=ambassador --rm quay.io/datawire/aes:1.0.0 --demo
```

per https://www.getambassador.io/about/quickstart/, it blows up:

```
time="2020-01-23 15:10:01" level=error msg="error reading license key: open /root/.config/ambassador/license-key: no such file or directory"
time="2020-01-23 15:10:01" level=info msg="Ambassador Edge Stack configuation loaded"
time="2020-01-23 15:10:01" level=error msg="shut down with error: the server is currently unable to handle the request"
```

There are also errors from the dev portal, though those aren't the thing that causes the crash:

```
2020-01-23 15:10:01 demo-auth 0.0.1 DEBUG: =>> <Request 'http://localhost:8080/auth/v0//api?timeout=32s' [GET]>
2020-01-23 15:10:01 demo-auth 0.0.1 DEBUG: =>>   Content-Length: 0
2020-01-23 15:10:01 demo-auth 0.0.1 DEBUG: =>>   Host: localhost:8080
2020-01-23 15:10:01 demo-auth 0.0.1 DEBUG: =>>   User-Agent: amb-sidecar/v0.0.0 (linux/amd64) kubernetes/$Format
2020-01-23 15:10:01 demo-auth 0.0.1 DEBUG: =>>   X-Envoy-Expected-Rq-Timeout-Ms: 5000
2020-01-23 15:10:01 demo-auth 0.0.1 DEBUG: =>>   X-Envoy-Internal: true
2020-01-23 15:10:01 demo-auth 0.0.1 DEBUG: =>>   X-Forwarded-For: 172.17.0.2,172.17.0.2
2020-01-23 15:10:01 demo-auth 0.0.1 DEBUG: =>>   X-Forwarded-Proto: http
2020-01-23 15:10:01 demo-auth 0.0.1 DEBUG: Requested path is /api
2020-01-23 15:10:01 demo-auth 0.0.1 DEBUG: =<< <Response 36 bytes [200 OK]>
2020-01-23 15:10:01 demo-auth 0.0.1 DEBUG: =<<   Content-Length: 36
2020-01-23 15:10:01 demo-auth 0.0.1 DEBUG: =<<   Content-Type: text/html; charset=utf-8
[...]

PR #2229, opened 2020-01-23
Make sure that test_docker gives the test container a chance to crash

kflynn:
The `test_docker` container currently crashes when run under Edge Stack (cf #2227) and the test wasn't catching that, because it was quick enough on the draw that it got a "successful" result before the container crashed. Slow it down to catch that.

This should pass OSS KAT, but fail Edge Stack KAT until #2227 is fixed.

### DIFF
--- a/python/tests/test_docker.py
+++ b/python/tests/test_docker.py
@@ -2,6 +2,7 @@ import os
 
 import pexpect
 import requests
+import time
 
 DockerImage = os.environ.get("AMBASSADOR_DOCKER_IMAGE", None)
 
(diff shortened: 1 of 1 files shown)


python/tests/test_docker.py
```diff
@@ -84,6 +88,9 @@ def test_docker():
             logfile.write('No $AMBASSADOR_DOCKER_IMAGE??\n')
         else:
             if docker_start(logfile):
+                logfile.write("Demo started, sleeping 10 seconds just in case...")
+                time.sleep(10)
```
alexgervais: Maybe we could have 2 checks? Fail fast first, then wait and see if the container remains healthy?

Issue #2198, opened 2020-01-14
Mappings created from Ingress resources collide across namespaces


micahlmartin:
**Describe the bug**

I have multiple ingress resources defined in separate namespaces. They're all pointing to different services and different domains. For some reason ambassador is failing because it thinks they are duplicate.

```
Jan 14 16:01:38 ambassador-55f8dc9499-pqlvc aes ERROR diagd 1.0.0 [P104TAEW] ERROR: scs-0-0.scs-pr-6044.1: <RichStatus BAD error=<Mapping scs-0-0.scs-pr-6044.1> defines Mapping scs-0-0, which is already defined by scs-0-0.scs-cffa046.1 hostname=ambassador-55f8dc9499-pqlvc version=1.0.0>
Jan 14 16:01:38 ambassador-55f8dc9499-pqlvc aes ERROR diagd 1.0.0 [P104TAEW] ERROR: scs-0.scs-pr-6044.1: <RichStatus BAD error=<TLSContext scs-0.scs-pr-6044.1> defines TLSContext scs-0, which is already defined by scs-0.scs-cffa046.1 hostname=ambassador-55f8dc9499-pqlvc version=1.0.0>
Jan 14 16:01:38 ambassador-55f8dc9499-pqlvc aes ERROR diagd 1.0.0 [P104TAEW] ERROR: scs-0.scs-pr-6056.1: <RichStatus BAD error=<TLSContext scs-0.scs-pr-6056.1> defines TLSContext scs-0, which is already defined by scs-0.scs-pr-6044.1 hostname=ambassador-55f8dc9499-pqlvc version=1.0.0>
Jan 14 16:01:38 ambassador-55f8dc9499-pqlvc aes ERROR diagd 1.0.0 [P104TAEW] ERROR: scs-1-0.scs-pr-6056.1: <RichStatus BAD error=<Mapping scs-1-0.scs-pr-6056.1> defines Mapping scs-1-0, which is already defined by scs-1-0.scs-pr-6044.1 hostname=ambassador-55f8dc9499-pqlvc version=1.0.0>
[...]

PR #2230, opened 2020-01-23
Change resource name to store if duplicate found

### LINKED ISSUE #2226: Cannot create mappings with same name in different namespaces
**Describe the bug**
Ambassador does not allow 2 mappings with the same name with different/same config in different namespaces. An error pops up saying -
```
CONFIGURATION ERRORS
httpbin.ingress2.1: <Mapping httpbin.ingress2.1> defines Mapping httpbin, which is already defined by httpbin.ingress1.1
```

**To Reproduce**
1. Create 2 mappings with the same name (but not necessarily same config) in different namespaces. Example:

Create namespaces:

`kubectl create ns ingress1`
`kubectl create ns ingress2`
 
and then create the following mappings:

```yaml
kind: Mapping
metadata:
  name: httpbin
  namespace: ingress2
spec:
  host_rewrite: hb.com
  prefix: /hb2/
  service: http://hb.com
```

and 

```yaml
kind: Mapping
metadata:
  name: httpbin
  namespace: ingress1
spec:
  host_rewrite: httpbin.org
  prefix: /hb/
  service: http://httpbin.org                                                                                                                                                                                                                                 
```

**Expected behavior**
Both mappings should work fine

 - Ambassador: 1.0.0 and master and since the beginning of time??

concaf:
Change resource name to store if duplicate found
    
    Currently, resources stored in Ambassador are stored as
    {resource_name: resource object, ...} treating resource name as
    a unique identifier to look up resources of a given kind. However,
    this causes issues becuase this does not allow different resources
    in different namespace to have the same name, hence resulting in
    errors or ignoring these resources completely.
    
    In this commit, if two (or more) resources have the same name (in
    different namespaces), then the duplicate resources are stored
    as {name.namespace: resource} instead of just {name: resource}.
    
    An ideal solution would be to idenitfy all resources with
    {
      name: resource_name,
      namespace: resource_namespace
    }
    instead of flat strings, but that comes later.

Fix #2226 #2198

### DIFF
--- a/python/ambassador/config/config.py
+++ b/python/ambassador/config/config.py
@@ -593,10 +593,18 @@ class Config:
         storage = self.config.setdefault(storage_name, {})
 
         if resource.name in storage:
-            # Oooops.
-            self.post_error("%s defines %s %s, which is already defined by %s" %
-                            (resource, resource.kind, resource.name, storage[resource.name].location),
-                            resource=resource)
+            if resource.namespace == storage[resource.name].get('namespace'):
+                # If the name and namespace, both match, then it's definitely an error.
+                # Oooops.
+                self.post_error("%s defines %s %s, which is already defined by %s" %
+                                (resource, resource.kind, resource.name, storage[resource.name].location),
+                                resource=resource)
+            else:
+                # Here, we deal with the case when multiple resources have the same name but they exist in different
+                # namespaces. Our current data structure to store resources is a flat string. Till we move to
+                # identifying resources with both, name and namespace, we change names of any subsequent resources with
+                # the same name here.
+                resource.name = f'{resource.name}.{resource.namespace}'
 
         if allow_log:
             self.logger.debug("%s: saving %s %s" %
--- a/python/ambassador/config/resourcefetcher.py
+++ b/python/ambassador/config/resourcefetcher.py
@@ -605,7 +605,7 @@ class ResourceFetcher:
             ingress_status = self.ambassador_service_raw.get('status', {})
             ingress_status_update = (k8s_object.get('kind'), ingress_namespace, ingress_status)
             self.logger.info(f"Updating Ingress {ingress_name} status to {ingress_status_update}")
-            self.aconf.k8s_status_updates[ingress_name] = ingress_status_update
+            self.aconf.k8s_status_updates[f'{ingress_name}.{ingress_namespace}'] = ingress_status_update
 
         if parsed_ambassador_annotations is not None:
             # Copy metadata_labels to parsed annotations, if need be.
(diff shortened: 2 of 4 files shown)


concaf:
This PR now also fixes updating multiple Ingress' status if they happen to have the same same in different namespaces.

concaf:
Fixes #2226

python/ambassador/config/config.py
```diff
@@ -593,10 +593,18 @@ def safe_store(self, storage_name: str, resource: ACResource, allow_log: bool=Tr
...
+                # Oooops.
+                self.post_error("%s defines %s %s, which is already defined by %s" %
+                                (resource, resource.kind, resource.name, storage[resource.name].location),
+                                resource=resource)
+            else:
+                # Here, we deal with the case when multiple resources have the same name but they exist in different
+                # namespaces. Our current data structure to store resources is a flat string. Till we move to
+                # identifying resources with both, name and namespace, we change names of any subsequent resources with
+                # the same name here.
+                resource.name = f'{resource.name}.{resource.namespace}'
```
alexgervais: Is there a reason why changing the resource name done in an `else` block? It currently feels like we handle an edge-case, yet I'm pretty sure all `resource.name` could include the namespace from the moment they are handled.
Following the logic here, it seems given resources: `{name: something, namespace: a}` and `{name: something, namespace: b}`, we would end up with 2 resources named: `something` and `something.b` where we don't associate the namespace to the first resource we put in storage.

concaf: You're correct. This is what I tried to explain at https://github.com/datawire/ambassador/pull/2230#issue-366506982 that we need to identify all resources with both name and namespace, and that's what I tried to do in the first go - but it turned out to be a much more intrusive change for the following reasons -
- there were test failures everywhere
- there are parts of code that rely on the resource name coming from aconf.config instead of aconf.config['name'], so that approach broke it

So, yes, this is more or less handling of an edge case without breaking the status quo.

alexgervais:
@concaf tests LGTM:
```
~ k get ingress -A
NAMESPACE     NAME   HOSTS                                                                                              ADDRESS          PORTS     AGE
scs-cffa046   scs    scs-cffa046.dev.mydomain.com,admin.scs-cffa046.dev.mydomain.com,api.scs-cffa046.dev.mydomain.com   104.198.45.191   80, 443   9m34s
scs-pr-6044   scs    scs-pr-6044.dev.mydomain.com,admin.scs-pr-6044.dev.mydomain.com,api.scs-pr-6044.dev.mydomain.com   104.198.45.191   80, 443   9m33s
scs-pr-6056   scs    scs-pr-6056.dev.mydomain.com,admin-scs-pr-6056.dev.mydomain.com,api-scs-pr-6056.dev.mydomain.com   104.198.45.191   80, 443   9m33s
```
Screen Shot 2020-01-24 at 09 54 35

python/ambassador/config/config.py
kflynn: We'll definitely need to handle the larger case, but, yeah, got it.

PR #2232, opened 2020-01-23
Allow TCPMappings to be reached on HTTPS listener

texascloud:
Add 'transport_protocol':'tls' to filter chain match of folded TCPMapping

Ignores any TCPMapping that does not specify a host (SNI) while also setting
its port to ambassador's service_port

Currently a limitation is a TCPMapping who has specified a host which
relies on a TLSContext for providing certs, but that TLSContext doesn't
specify the exact host value (because it is using wildcards), then the
TCPMapping filter chain when folded will accidentally include the
application_protocols portion of the filter chain match, which checks
that the protocol is HTTP/1.1 or HTTP2, thus preventing a match from
occurring when connecting via TCP using SNI of the "host" value which
the TCPMapping specified.

## Description
The motivation behind this PR is to allow TLS connections with SNI to TCP Services to be connected to over port 443. Without this change, that is not possible.

Please see the limitation in commit message. I would like feedback on getting around this limitation. I have a feeling it requires conditionally adding the `application_protocols` field to a V2Listener's filter chain match.

## Testing

I have manually verified these changes work on a dev cluster. 

## Todos
- [x] Tests
- [x] Documentation

## Other

### DIFF
--- a/python/ambassador/envoy/v2/v2listener.py
+++ b/python/ambassador/envoy/v2/v2listener.py
@@ -563,6 +563,9 @@ class V2VirtualHost(dict):
 
         if enable_sni and self._ctx and self._hostname and (self._hostname != '*'):
             match["server_names"] = [ self._hostname ]
+        
+        # Necessary if a TCPMapping is going to use the same listener as HTTPS
+        match["application_protocols"] = [ "http/1.1", "h2" ]
 
         self["filter_chain_match"] = match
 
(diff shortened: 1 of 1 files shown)


CamelCaseNotation:
Could I get some help in triaging this build? Where to start? Or how can I run these tests within the shell after running `make shell`?

kflynn:
Hey @texascloud, sorry for the delay here -- I was heads-down for awhile on the 1.0 transition. 🙁  Can you take a crack at conflict resolution here, and then I can help sort out how we can get this landed?

Alternately, feel free to ping me on [our Slack](https://d6e.co/slack) if you have questions -- thanks for the contribution!

PR #2242, opened 2020-01-24
Switch to -ea.N and -rc.N

### LINKED ISSUE #2211: Adjust all of the machinery to use -ea.N and -rc.N instead of -eaN and -rcN
We agreed that after 1.0.0, we would follow strict SemVer.  This means that our pre-release suffixes need to change from `-ea${N}` and `-rc${N}` to including a dot: `-ea.${N}` and `-rc.${N}`.  While the existing dotless versions are valid SemVer pre-release identifiers; they don't represent the information that we intend them to (thanks to OSS Sack user "yawniek" for pointing this out): They compare lexicographically, not numerically, which becomes a problem if we get in to the double-digits.  So while we expect that they sort as -ea8 < -ea9 < -ea10 < -ea11, SemVer actually says that they sort as -ea10 < -ea11 < -ea8 < -ea9, because '1' sorts before '8'.

There are a number of regexes that will need to be updated.

kflynn:
Instead of `-eaN` and `-rcN`, use `-ea.N` and `-rc.N`. 

Fixes #2211 for the `ambassador.git` side of the world.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -13,13 +13,13 @@ Note. PRs will pile up on `master`. **Don't accept PRs for which CI doesn't show
 When we get to the stage of creating a release, all the PRs that we want to merge will have been merged
 and the CI will be green.
 
-1. Once `master` has all the release drivers, tag `master` with an RC tag, e.g. `v0.77.0-rc1`. **This version tag must start with a 'v'.** For example:
-    git tag v0.77.0-rc1 master
+1. Once `master` has all the release drivers, tag `master` with an RC tag, e.g. `v0.77.0-rc.1`. **This version tag must start with a 'v'.** For example:
+    git tag v0.77.0-rc.1 master
     git push --tags origin master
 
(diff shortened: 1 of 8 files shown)


LukeShu:
I know RELEASING.md is hella out-of-date, but maybe adjust the examples in it?

LukeShu:
GitHub won't let me attach a comment to the line, but the

```Makefile
if ! [[ "$$rc_latest" == "$(RELEASE_VERSION)"-rc* ]]; then \
```

on builder.mk line 235 should be changed to


```Makefile
if ! [[ "$$rc_latest" == "$(RELEASE_VERSION)"-rc.* ]]; then \
```

LukeShu:
And update the comment on builder.sh line 134

LukeShu:
And the check on builder.sh line 260

LukeShu:
and the comment on builder/module_version.sh line 45

LukeShu:
Does python/ambassador/ambscout.py need adjusted?

kflynn:
Thanks for the catches! and yes, `ambscout.py` does have a check that I missed (because instead of `-(ea|rc)\d+` it was `-[A-Za-z_+]\d+` -- we should probably narrow that later).

PR #2246, opened 2020-01-27
Correct CI for apictl-key

kflynn:
Move build_push_cli.sh and module_version.sh to apro.git: they're not actually used from here.


LukeShu:
Looks good, pending https://github.com/datawire/apro/pull/940

PR #2247, opened 2020-01-27
Stupid Mac tricks

kflynn:
Turns out MacOS doesn't like empty alternatives in bash regexes. ?????


builder/builder.mk
```diff
@@ -186,7 +186,10 @@ release/promote-oss/.main:
 	@[[ "$(RELEASE_VERSION)"      =~ ^[0-9]+\.[0-9]+\.[0-9]+(-.*)?$$ ]]
 	@[[ '$(PROMOTE_FROM_VERSION)' =~ ^[0-9]+\.[0-9]+\.[0-9]+(-.*)?$$ ]]
 	@[[ '$(PROMOTE_TO_VERSION)'   =~ ^[0-9]+\.[0-9]+\.[0-9]+(-.*)?$$ ]]
-	@[[ '$(PROMOTE_CHANNEL)' =~ ^(|early|test)$$ ]]
+	@case "$(PROMOTE_CHANNEL)" in \
+		""|early|test) true ;; \
+		*) echo "Unknown PROMOTE_CHANNEL $(PROMOTE_CHANNEL)" >&2 ;; \
```
LukeShu: Stick an `exit 1` or something in here to prevent it from continuing?

PR #2255, opened 2020-01-29
Tell Travis to only build commits on master and shared/*

kflynn:
Running on branches _and_ PRs seems just silly.


kflynn:
No sense waiting for CI to finish, I see that it started correctly.

PR #2259, opened 2020-01-30
`make pull-docs`

LukeShu:
## Description

Sync the docs from ambassador-docs.git to here, by running `make pull-docs`.  There are no other changes; just syncing the docs.  Review is therefore mostly just a rubber stamp.

We normally do this at release-time.  I guess it is helpful to @scoyle391 to also do it now, and @ark3 wanted it to be a separate PR from her main PR?


ark3:
It's helpful to all of us, because this way all the `shared/*` branches will get these changes _for free_, they won't contend when landing back on master, etc. Thank you!

PR #2262, opened 2020-01-30
Add minimal end-to-end `edgectl install` with lots of FIXME and TODO comments

ark3:


cmd/edgectl/aes_install.go
```diff
@@ -0,0 +1,290 @@
...
+		// FIXME: Time out at some point...
+		time.Sleep(500 * time.Millisecond)
+		resp, err := http.Get("http://" + ipAddress + "/.well-known/acme-challenge/")
+		if err != nil {
+			fmt.Printf("Waiting for Ambassador (get): %#v\n", err)
+			continue
+		}
+		_, _ = ioutil.ReadAll(resp.Body)
+		resp.Body.Close()
+		if resp.StatusCode != 404 {
```
alexgervais: I would not perform these StatusCode and Header checks on the client-side. If we can open a socket and get any kind of response other than the handled `err`, we should be good to proceed.

cmd/edgectl/aes_install.go
```diff
@@ -0,0 +1,290 @@
...
+		fmt.Println("Please enter an email address. We'll use this email address to notify you prior to domain and certification expiration [None]:", emailAddress)
+		fmt.Println("FIXME: let the user enter an address")
+		// Create a Host resource
+		hostResource := fmt.Sprintf(hostManifest, hostname, namespace, hostname, emailAddress)
+		kargs, err := i.kubeinfo.GetKubectlArray("apply", "-f", "-")
+		if err != nil {
+			return errors.Wrapf(err, "cluster access for install AES")
+		}
+		fmt.Println("\n$ kubectl apply -f - < [Host Resource]")
+		cmd := exec.Command("kubectl", kargs...)
```
alexgervais: Should we use a client wrapper here, in case `kubectl` is not in the user's path?

cmd/edgectl/aes_install.go
```diff
@@ -0,0 +1,290 @@
...
+
+	return res
+}
+
+func aesInstall(cmd *cobra.Command, args []string) error {
+	metrics := NewMetrics()
+	_ = metrics.Report("install")
+
+	// Display version information
+	fmt.Printf("-> Installing the Ambassador Edge Stack %s\n", Version)
```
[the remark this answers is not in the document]
ark3: Yes, that's correct. We could also insist that you have the latest Edge Control. That would allow us to make the installation process more complicated in the future, assuming this install method gains traction. Anyhow, yes, there should be a FIXME or TODO here.

cmd/edgectl/aes_install.go
ark3: Isn't `upstream request timeout` coming from Envoy? That'd be `err == nil` and `StatusCode == 503` (or whatever), right?

cmd/edgectl/aes_install.go
ark3: Oh definitely. This and `edgectl license` should both avoid using `kubectl` if possible. But that's far lower priority (in my opinion) than getting all the story features and required robustness in place. Do you disagree? Using `kubectl` in the short run can save us a lot of time figuring out the details of how to do the same things it does for us.

cmd/edgectl/aes_install.go
alexgervais: I think you right, that would make much more sense!

cmd/edgectl/aes_install.go
alexgervais: If `edgectl license` is already dependent on `kubectl`, then let's keep on iterating fast here and revisit this later.

ark3:
I added some comments. What do you think, @alexgervais ?

PR #2269, opened 2020-02-03
Replace no-op Metrics with real pushes to beta Metriton

ark3:


cmd/edgectl/scout.go
```diff
@@ -0,0 +1,108 @@
...
+			installID = string(installIDBytes)
+			// Validate UUID format
+			if _, err := uuid.Parse(installID); err == nil {
+				newInstall = false
+				return nil
+			} // else the value is not a UUID
+		} // else file doesn't exist, etc.
+
+		// Try to create a new install ID
+		installID = uuid.New().String()
```
alexgervais: Instead of generating a UUID for identifying the install end-user, should we instead try to replicate what Ambassador does and marshall a UUID given the unique target installation namespace?

https://github.com/datawire/ambassador/blob/6a613b46ab10b060ef92b0d9546b22531188fa75/python/kubewatch.py#L178

This way, if a single edgectl binary is used to install multiple clusters, it'll result in multiple `installId`.

cmd/edgectl/scout.go
```diff
@@ -0,0 +1,108 @@
...
+	}()
+	if err != nil {
+		return nil, err
+	}
+
+	return &Scout{mode, installID, newInstall}, nil
+}
+
+func (s *Scout) Report(action string, meta ...ScoutMeta) error {
+	if os.Getenv("SCOUT_DISABLE") != "" {
```
alexgervais: 👍

cmd/edgectl/scout.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	data := map[string]interface{}{
+		"application": "edgectl",
+		"install_id":  s.installID,
+		"version":     Version,
+		"metadata":    metadata,
+	}
+
+	// Copy-pasta from apro.git/lib/metriton/phonehome.go
```
alexgervais: That comment seems inappropriate for a public repo :/

cmd/edgectl/scout.go
ark3: The story specifically asks for 

1. Edge Control to have its own ID so that it can post something right away, before it establishes connectivity to a cluster
2. The installed Ambassador's cluster_id to be sent in an extra_trait

The second item is not implemented yet.

cmd/edgectl/scout.go
ark3: Yes, you're right.

cmd/edgectl/scout.go
alexgervais: Sending the Ambassador cluster_id in an extra_trait seems reasonable. As the code is working, let's ship it and review the collected events with @richarddli

PR #2275, opened 2020-02-05
Add cluster ID reporting to installer metrics

ark3:
Also report metrics errors and fix the Metriton URLs

### DIFF
--- a/cmd/edgectl/aes_install.go
+++ b/cmd/edgectl/aes_install.go
@@ -269,6 +269,7 @@ func (i *Installer) CaptureKubectl(name string, args ...string) (res string, err
 		err = errors.Wrap(err, name)
 	}
 	res = string(resAsBytes)
+	fmt.Println(res)
 	return
 }
 
--- a/cmd/edgectl/aes_install.go
+++ b/cmd/edgectl/aes_install.go
@@ -80,6 +80,15 @@ func aesInstall(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Grab Ambassador's install ID as the cluster ID we'll send going forward
+	for {
+		if clusterID, err := i.CaptureKubectl("get cluster ID", "-n", "ambassador", "exec", "deploy/ambassador", "python3", "kubewatch.py"); err == nil {
+			metrics.SetClusterID(strings.TrimSpace(clusterID))
+			break
+		}
+		time.Sleep(500 * time.Millisecond) // FIXME: Time out at some point...
+	}
+
 	_ = metrics.Report("deploy") // TODO: Send cluster type and Helm version
 
 	ipAddress := ""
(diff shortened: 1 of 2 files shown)


cmd/edgectl/scout.go
```diff
@@ -64,7 +66,14 @@ func NewScout(mode string) (*Scout, error) {
 		return nil, err
 	}
 
-	return &Scout{mode, installID, newInstall}, nil
+	return &Scout{mode, installID, newInstall, ""}, nil
+}
+
+func (s *Scout) SetClusterID(clusterID string) {
+	if s.clusterID != "" {
+		panic(fmt.Sprintf("Trying to replace cluster ID %q with %q", s.clusterID, clusterID))
```
alexgervais: `Trying` should not be capitalized
(PS. Sorry I hate making these kind of comments in PRs)

cmd/edgectl/aes_install.go
```diff
@@ -269,6 +278,7 @@ func (i *Installer) CaptureKubectl(name string, args ...string) (res string, err
 		err = errors.Wrap(err, name)
 	}
 	res = string(resAsBytes)
+	fmt.Println(res)
```
alexgervais: That's just for debugging I presume?

cmd/edgectl/aes_install.go
```diff
@@ -80,6 +80,15 @@ func aesInstall(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Grab Ambassador's install ID as the cluster ID we'll send going forward
+	for {
+		if clusterID, err := i.CaptureKubectl("get cluster ID", "-n", "ambassador", "exec", "deploy/ambassador", "python3", "kubewatch.py"); err == nil {
```
alexgervais: Would it be safer to reproduce the logic from `kubewatch.py` rather than trying to exec into a running pod? Reproducing the logic simply means reading metadata information about the namespace.

Also, should the namespace `-n ambassador` be a variable?

cmd/edgectl/aes_install.go
ark3: All the `kubectl` output is for debugging. I'm adding this one just for consistency. Part of the "Manage output" task (later this week) is to be able to redirect that output separately from other output, then make smart decisions about what goes where.

cmd/edgectl/aes_install.go
ark3: No, because part of the point here is that the Ambassador pod is running (though not necessarily Ready), which I'm considering good enough to for the "deploy" metrics report. I will point that out in a comment.

Also, no. The installer always installs into the ambassador namespace, always creates one pod, always has AES managing the whole cluster, etc. This is not intended to be a configurable installation, rather an opinionated one. The story document doesn't say this clearly, but that from talking to Rafi, that was his intent.

alexgervais: Note from Slack discussion:

> The installer always installs into the ambassador namespace, always creates one pod, always has AES managing the whole cluster, etc. 

This assumption also true for detecting existing Helm installs, in the sense that an existing install in a different namespace will be ignored.

PR #2277, opened 2020-02-05
Correctly handle istio secrets

ppeble:
## Description

Change to correctly consume `istio.io/key-and-cert` secret type so that we no longer are required to mount the secret as a volume as specified [here](https://www.getambassador.io/reference/tls/mtls/#mutual-tls-mtls).

As background, the mounting solution works but [we discovered](https://www.activecampaign.com/) that when `istio` secrets change (say, with an istio upgrade) it seems that Ambassador was not correctly picking up the new `key-and-cert` secret. The solution when using the mounted volume method was to reboot Ambassador. With this change Ambassador correctly reads the secret like any other secret and therefore will pick up any changes.

## Related Issues

* https://github.com/datawire/ambassador/issues/1475
* https://github.com/datawire/ambassador/issues/1255

## Testing

I've performed manual testing of on `istio` versions 1.2 and 1.3. I do not have a more recent (1.4) version to test against but nothing in the documentation leads me to believe that there is a change to the `istio/key-and-cert` secret format.

I have added a test to verify that the key is correctly consumed into a `TLSContext`. I am open to guidance on how to add additional tests but to be quite honest I am overwhelmed trying to learn how `kat` works after a few months away from it. 😬


python/ambassador/utils.py
```diff
@@ -348,19 +362,23 @@ def from_dict(cls, resource: 'IRResource',
                 return None
 
             cert = None
+        elif secret_type == 'istio.io/key-and-cert':
+            resource.ir.logger.error(f"{resource.kind} {resource.name}: found data but handler for istio key not finished yet")
```
ppeble: I wasn't sure if this was necessary but it didn't feel right to simply ignore this type. I didn't find that I had to add this but I'm also not exactly sure how this method is used. 🤔

kflynn: `secret_info_from_k8s` turns out to be unused, so we're good to go. ;)

PR #2278, opened 2020-02-05
Docs for apro#970

LukeShu:
Docs changes for https://github.com/datawire/apro/pull/970

### DIFF
--- a/docs/reference/filter-reference.md
+++ b/docs/reference/filter-reference.md
@@ -349,7 +349,6 @@ spec:
 
     # Settings for grantType=="AuthorizationCode"
     clientURL:             "url-string"      # required
-    stateTTL:              "duration-string" # optional; default is "5m"
     clientID:              "string"          # required
     # A client secret must be specified.
     # This can be done by including the raw secret as a string in "secret",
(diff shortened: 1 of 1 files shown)


docs/reference/filter-reference.md
```diff
@@ -389,7 +388,6 @@ Settings that are only valid when `grantType: "AuthorizationCode"`:
    * As a Kubernetes `generic` Secret, named by `secretName`/`secretNamespace`.  The Kubernetes secret must of
      the `generic` type, with the value stored under the key`oauth2-client-secret`.  If `secretNamespace` is not given, it defaults to the namespace of the Filter resource.
    * **Note**: It is invalid to set both `secret` and `secretName`.
- - `stateTTL`: (You decide this) How long the Ambassador Edge Stack will wait for the user to submit credentials to the identity provider and receive a response to that effect from the identity provider
```
scoyle391: "how long" can a wait be? like 56743854398 hours? or 60 minutes? etc. Please specify the range and unit of time.

LukeShu: This line is getting removed, not added.  But even if it weren't:
 - it could be any range of time; the default is specified above to be 5 minutes
 - it's stated above to be a "duration-string", which just a few lines down are explained as:

> `"duration-string"` strings are parsed as a sequence of decimal numbers, each with optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h". See [Go `time.ParseDuration`](https://golang.org/pkg/time/#ParseDuration).

scoyle391: Oh my bad. LGTM!

PR #2279, opened 2020-02-06
Fix Host support choosing cleartext or TLS

kflynn:
CHANGELOG: Properly support using a `Host` to specify cleartext or non-ACME TLS.
CHANGELOG: Be more forgiving about mismatched SNI connections.

This is probably best reviewable commit-by-commit. There are several small changes here that add up to substantial effects:

- (commit abf295fd) Edge Stack was creating the fallback `TLSContext` whenever no termination contexts were present, irrespective of what any `Host`s said. If you think this through you'll realize that it means that there would _always_ be a termination context present, preventing running a cleartext-only Edge Stack.

    We now only create the fallback context if there are no termination contexts _and no `Host`s_, which is to say, when Edge Stack is running basically unconfigured. If a `Host` is present that says to run cleartext, it will be honored.

- (commit 21823535) Don't look at the `acmeProvider` when deciding whether or not to enable TLS termination for a `Host` -- instead, only look at whether a `tlsSecret` is present. The effect here is that if you say `acmeProvider.authority: none` but manually provide a `tlsSecret`, you'll still get TLS termination.

    (When Edge Stack ACME is in use, the ACME client will fill in the `tlsSecret` for you if you don't set it, so there will always be a `tlsSecret` if termination should happen. If you _do_ set a `tlsSecret` with Edge Stack ACME in use, the ACME client will happily use the secret name you provide.)

- (commits cd509607 and a63b529a) Rationalize Envoy `VHost` entries such that we'll always have a `VHost` for `*`. If the configuration specifies a `Host` with `hostname: "*"`, we'll use that -- otherwise, grab the first `VHost` entry and make it have `hostname: "*"`. 

    This seems odd, but is probably the least surprising thing to do: if you connect to a port and offer a strange SNI domain, it's probably less surprising to people to get back a certificate for something that should be listening there, rather than having Envoy simply drop the connection with a TLS error. If need be, we can create a setting that allows disabling this more forgiving behavior later.

Still to come: doc tweaks, and the port-override mechanism suggested by @nbkrause to fully support L4 split LBs. Those will be separate PRs.

Testing: KAT passes for both API Gateway and Edge Stack, including dropping the XFail for `HostCleartext`.


python/ambassador/ir/irhost.py
```diff
@@ -167,3 +160,41 @@ def load_all(cls, ir: 'IR', aconf: Config) -> None:
...
+            # we have any termination contexts.
+            host_count = len(ir.get_hosts() or [])
+            contexts = ir.get_tls_contexts() or []
+
+            found_termination_context = False
+            for ctx in contexts:
+                if ctx.get('hosts'):  # not None and not the empty list
+                    found_termination_context = True
+
+            ir.logger.info(f"HostFactory: FTC {found_termination_context}, host_count {host_count}")
```
LukeShu: I don't like that name.  Not enough buzz-words.  Try to fit in "bean" or "monoid" or "functor".

python/ambassador/ir/irtlscontext.py
```diff
@@ -382,29 +382,38 @@ def load_all(cls, ir: 'IR', aconf: Config) -> None:
...
+        #         )
+        #
+        #         if not os.environ.get('AMBASSADOR_NO_TLS_REDIRECT', None):
+        #             new_ctx['redirect_cleartext_from'] = 8080
+        #
+        #         ctx = IRTLSContext(ir, aconf, **new_ctx)
+        #
+        #         assert ctx.is_active()
+        #         if ctx.resolve_secret(tls_name):
+        #             ir.save_tls_context(ctx)
```
LukeShu: Is there a reason to keep this present & commented out?  It looks to me like it's sufficiently similar to the code that appeared in irhost.py to just remove it.

python/ambassador/ir/irhost.py
```diff
@@ -60,52 +60,45 @@ def setup(self, ir: 'IR', aconf: Config) -> bool:
...
+                            namespace=self.namespace,
+                            location=self.location,
+                            hosts=[ self.hostname or self.name ],
+                            secret=tls_name
+                        )
 
-                            if not os.environ.get('AMBASSADOR_NO_TLS_REDIRECT', None):
-                                new_ctx['redirect_cleartext_from'] = 8080
+                        # if not os.environ.get('AMBASSADOR_NO_TLS_REDIRECT', None):
+                        #     new_ctx['redirect_cleartext_from'] = 8080
```
LukeShu: What was that ever for?  Just as a hack for KAT?

python/ambassador/ir/irhost.py
kflynn: `HostBeanFactoryMonoidFunctorFactory`?

LukeShu:
I'm having a hard time understanding what's going on in a63b529ae225272e436e1d11d76bd258a053d101.  Perhaps it will make more sense after I put some food in me.

python/ambassador/ir/irhost.py
LukeShu: Exactly!

python/ambassador/ir/irtlscontext.py
kflynn: None whatsoever, I just missed it in my "clean up the world" pass. Deleting.

python/ambassador/ir/irhost.py
kflynn: Many of the KAT tests did a poor job of handling the automagic redirect on 8080 that used to be there. This was a crutch that I expect to be deleting shortly (I have to do a full test run to make sure it's really no longer used).

kflynn:
An important note for https://github.com/datawire/ambassador/commit/a63b529ae225272e436e1d11d76bd258a053d101 is that `enable_sni` used to keep track of whether we would use a `VHost` of "*" (if only one unique domain was present) or a `VHost` with an actual domain name in it (if more than one unique domain was present). So much of that commit is ripping out the now-dead `enable_sni` logic, and then there's the small chunk of "if we don't have a `VHost "*"`, then overwrite the first `VHost`'s `hostname` that takes its place.

python/ambassador/envoy/v2/v2listener.py
```diff
@@ -954,46 +954,34 @@ def generate(cls, config: 'V2Config') -> None:
...
+
+                # This insecure vhost can _only_ have a hostname of "*", since (by definition)
+                # there is no SNI associated with it.
+                #
+                # Also, no, it is not a bug to have action=None. There is no secure action
                 # for this vhost.
-                listener.make_vhost(name="forced-8080",
-                                    hostname="*",
+                listener.make_vhost(name=vhostname,
+                                    hostname=vhostname,
```
LukeShu: The comment makes me think that this should be `hostname="*"`.  What's going on?

kflynn: The comment is wrong, but we also need a check to make sure that the `VHost` doesn't already exist. I'll add that shortly.

PR #2287, opened 2020-02-10
ambex: Upgrade Envoy go-control-plane

LukeShu:
So I was spending my Sunday experimenting with Go tooling stuff, and was using ambassador.git as a test-bed, and the weird go-control-plane situation with ambex was causing me headaches, so... here's a PR.

## Description

Upgrade the go-control-plane used by `ambex`.  But instead of importing `github.com/envoyproxy/go-control-plane`, vendor a modified copy.

This is because:

 - `github.com/envoyproxy/go-control-plane/pkg` (hand-written code) needs to make use of `github.com/envoyproxy/go-control-plane/envoy` (`protoc`-generated code),
 - but it's important that we use `github.com/datawire/ambassador/pkg/api/envoy` instead of `github.com/envoyproxy/go-control-plane/envoy`

and it turns out that it's not terribly hard to just have `make generate` vendor a modified copy of go-control-plane.

## Related Issues

https://github.com/datawire/ambassador/issues/2146

## Testing

`go test ./pkg/envoy-control-plane/... && go build ./cmd/ambex`

##  Other

To the reviewer: The interesting stuff is in `build-aux-local/generate.mk`.  You can probably ignore reading any of the `pkg/envoy-control-plane/` file diffs.


LukeShu:
Are the KAT tests actually testing Ambex' ability to talk to Envoy, or do I need to do some "disable gold" test?

LukeShu:
Do I need to do something special to make sure KAT actually tests ambex's ability to talk to Envoy? (disable gold?).  Or is CI already testing that?

LukeShu:
OK, after some shenanigans with GKE giving me junk nodes, I verified that `make pytest KAT_RUN_MODE=envoy` passes on kubernaut.

PR #2306, opened 2020-02-13
python/tests/t_*: Use the "super().manifests()"-at-the end idiom

LukeShu:
In today's Ambassador-U, Flynn explained that the proper idiom is to
include new manifests at the beginning; eg.

        return """
    MY-YAML
    """ + super().manifests()

and not

        return super().manifests() + """
    MY-YAML
    """

pretty much because "namespaces".

But a bunch of the existing tests do it the discouraged way.  This
increases the odds that someone won't just figure-it-out looking at
whats's there.  So change it.


python/tests/t_consul.py
```diff
@@ -24,6 +24,8 @@ def init(self):
         self.k8s_target = HTTP(name="k8s")
 
     def manifests(self) -> str:
+        # Unlike usual, super().manifests() must come before our added
+        # manifests, because of some magic with ServiceAccounts?
```
LukeShu: NB

python/tests/t_consul.py
```diff
@@ -24,6 +24,8 @@ def init(self):
...
         return super().manifests() + self.format("""
```
LukeShu: NB

python/tests/t_shadow.py
```diff
@@ -19,9 +19,7 @@ def init(self) -> None:
         self.options = None
 
     def manifests(self) -> str:
-        s = super().manifests() or ""
-
-        return s + """
+        return super().manifests() + """
 ---
```
LukeShu: NB

kflynn: Why did you not swap the order here? Do the tests fail if so?

LukeShu: Um.  Good question.  Because my Emacs macro didn't catch it properly?

kflynn: :joy:

I'll expect a change then... 🙂

LukeShu:
v2:
 - rebase on to master
 - fix t_shadow.py per Flynn's comment
 - also adjusted a new test in t_tls.py

`make pytest KAT_RUN_MODE=envoy` is running now...

PR #2309, opened 2020-02-13
Watch ambassadorinstallations.getambassador.io

concaf:
This commit makes Ambassador watch for
ambassadorinstallations.getambassador.io CRD if it exists in the
cluster.

### DIFF
--- a/python/kubewatch.py
+++ b/python/kubewatch.py
@@ -245,6 +245,13 @@ def main(debug):
                          ' To enable Ingress support, configure RBAC to allow Ambassador to read Ingress resources,' +
                          ' then restart the Ambassador pod.')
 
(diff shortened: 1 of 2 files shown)


bnfb:
@concaf Why is this needed? Why should Ambassador watch the AmbassadorOperator CRD? Is this just so that it will show the data in the debug page? If so, then I don't think it's needed because _Ambassador_ doesn't need to update the Envoy configuration when the AmbassadorInstallation CRD changes: only the operator needs to know about that CRD, correct...?

bnfb:
Re
> Why is this needed? 
See Slack conversation https://datawire.slack.com/archives/CT0TVAL0M/p1581949417006300 for the answer.

PR #2311, opened 2020-02-14
Supporting work for merging the app-sidecar image

LukeShu:
This is supporting work for https://github.com/datawire/apro/pull/1001

It has the side-effect that it should also help shrink the images.

### DIFF
--- a/docs/docs/dev-guide/service-preview.md
+++ b/docs/docs/dev-guide/service-preview.md
@@ -100,7 +100,8 @@ In this quick start, we're going to preview a change we make to the backend serv
               value: "8080"
             - name: AMBASSADOR_LICENSE_KEY
               value: eJcbGciOiaIUzI1NiIsInR5cCkpXVCJ9.eCI6Im5rcmF1c2UiLCJleHAiOjE1Nzg0MTg4ODZ9.S_6-zdPyy4z1N4Jmo5e4A7fME4CbQVL_13ikw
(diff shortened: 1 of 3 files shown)


docs/docs/dev-guide/service-preview.md
```diff
@@ -166,4 +167,4 @@ In this quick start, we're going to preview a change we make to the backend serv
 
 Service Preview will match HTTP headers based on the headers that are seen by the *sidecar* and not the edge gateway. Matches are made on the whole header, e.g., a match rule of `dev` will not match in the example above, while `/backend/dev` will match.
 
-While any HTTP header will match, in practice, using host-based routing (i.e., the `:authority` header), a custom HTTP header (e.g., the `x-service-preview` header used above), or an authentication header is recommended.
\ No newline at end of file
+While any HTTP header will match, in practice, using host-based routing (i.e., the `:authority` header), a custom HTTP header (e.g., the `x-service-preview` header used above), or an authentication header is recommended.
```
LukeShu: The file was missing a trailing newline, and so Emacs (like any decent text editor) appended one when I saved.  But of course GitHub can't display what that difference is in a reasonable way, can it?

kflynn:
After approving this, I thought a bit more, and realized that I actually don't like moving the installation of `ambex` et al out of the Dockerfile into `post-compile.sh`. However, after talking with @LukeShu... it's really more of a larger-issue sort of thing with the build system as a whole. I'm going to hold my nose and click merge, now that I've filed an internal issue for the larger thing.

Issue #2302, opened 2020-02-12
`use_ambassador_namespace_for_service_resolution` and `add_linkerd_headers` do not work together


ricky26:
**Describe the bug**
When both of the features associated with these flags are used, the `l5d-dst-override` header is set to the name of the service, without the namespace suffix, and load-balancing falls back to k8s' connection-balancing.

**To Reproduce**
1. Install ambassador
2. Install linkerd
3. Create a mapping in another namespace, to a service in the same service.
4. Observe on the linkerd dashboard, that the service does not appear meshed.

**Expected behavior**
The `l5d-dst-override` header must be set to the qualified service name if it's in another namespace.

**Versions (please complete the following information):**
 - Ambassador: 1.1.0
 - Kubernetes environment: GKE 1.14.10-gke.17
 - Linkerd: 2.6.1

**Additional context**
It looks like the namespace suffix is added at service resolution point but the l5d headers are populated sooner. Since the namespace is implicit based on the resource, would it make sense to move it earlier? I'm happy to attempt a PR.

PR #2313, opened 2020-02-14
Make sure correct namespace is attached to l5d-dst-override header

ricky26:
## Description
This PR moves namespace qualification of service names earlier (during IR mapping construction).

I figured this was better than repeating the logic from the endpoint generation (after all, the address we're referring to can happily contain the namespace, it shouldn't affect resolution elsewhere). Please give me a shout if you think this should be done another way, I'm happy to rework and clean up. :)

I've committed all of the gold output changes, since this PR does change the behaviour in some cases and I thought it warranted discussion.
- e.g. in the case of TCP mappings, the address generated in the endpoints would always contain the resolved service name with namespace, whereas if it matches the ambassador namespace, I now leave it unqualified.

## Related Issues
- #2302 - resolved by this PR because the `l5d-dst-override` header is now always populated with a qualified service name.

## Testing
I've added an additional Python test which ensures that the `l5d-dst-override` header contains the namespace when it is not fully-qualified and in a different namespace to ambassador.

## Todos
- [x] Tests


ricky26:
I noticed too late that I hadn't rebased onto master - I'll do that when I get a moment!

kflynn:
Hey @riky26 -- think you can tackle conflict resolution here? I'd suggest:

- rebase onto `master`, but ignore the `python/tests/gold` files
- `make clean push pytest-only KAT_RUN_MODE=envoy`
- Assuming that passes, `make pytest-gold`

Then we can get this landed. Thanks! and sorry for the delay!

ricky26:
@kflynn okay, I've done the rebase.

The error from Travis is: `./releng/travis-script.sh: line 69: DOCKER_BUILD_PASSWORD: unbound variable`. This looks like a CI issue?

ricky26:
@kflynn any update on being able to merge this?

kflynn:
Augh, how did I miss this? 🙁  Mea culpa! If you can get the conflicts resolved, I'll get this landed. (If not I'll take a crack at it. 🙂 )

ricky26:
I'm happy to have a go at the conflicts. Mostly it looks like the test result changes, which I can just regenerate. :)

kflynn:
Awesome, thanks! 🙂

ricky26:
@kflynn okay, done! (And I've just noticed that my first commit has a typo in... not sure I want to anger the CI gods at this point though...)

kflynn:
Thanks! 🙂

PR #2316, opened 2020-02-14
Add email address prompt to the installer

ark3:


cmd/edgectl/aes_install.go
```diff
@@ -350,3 +404,9 @@ spec:
   acmeProvider:
     email: %s
 `
+
+// FIXME: Mention that this will be shared with Let's Encrypt?
```
alexgervais: @richarddli

richarddli: If it will be shared, then we should mention it.

PR #2327, opened 2020-02-19
Support disabling Edge Policy Console

kflynn:
Dynamically add Edge Policy Console mappings -- but only if diag isn't disabled.

### DIFF
--- a/docs/reference/core/ambassador.md
+++ b/docs/reference/core/ambassador.md
@@ -70,8 +70,8 @@ cors:
 
 `diagnostics` configures Ambassador's diagnostics services.
 
-- Both API Gateway and Edge Stack provide low-level diagnostics at `/ambassador/v0/diag/`.
-- Ambassador Edge Stack also provides the higher-level Edge Policy Console at `/edge_stack/admin/`.
+- Both the API Gateway and the Edge Stack provide low-level diagnostics at `/ambassador/v0/diag/`.
+- The Ambassador Edge Stack also provides the higher-level Edge Policy Console at `/edge_stack/admin/`.
 
 By default, both services are enabled: 
 
--- a/docs/reference/core/ambassador.md
+++ b/docs/reference/core/ambassador.md
@@ -68,12 +68,25 @@ cors:
   ...
 ```
 
-The `diagnostics` service (at /ambassador/v0/diag/) defaults on, but you can disable the API route. It will remain accessible on diag_port.
+`diagnostics` configures Ambassador's diagnostics services.
+
+- Both API Gateway and Edge Stack provide low-level diagnostics at `/ambassador/v0/diag/`.
+- Ambassador Edge Stack also provides the higher-level Edge Policy Console at `/edge_stack/admin/`.
+
+By default, both services are enabled: 
 
 ```
 diagnostics:
   enabled: true
 ```
+
+Setting `diagnostics.enabled` to `false` will disable the routes for both services (they will remain accessible from inside the Ambassador pod on port 8877):
+
+```
+diagnostics:
+  enabled: false
+```
+
 
 `keepalive` sets the global keepalive settings.
 Ambassador will use for all mappings unless overridden in a
--- a/python/ambassador/envoy/v2/v2route.py
+++ b/python/ambassador/envoy/v2/v2route.py
@@ -92,7 +92,19 @@ class V2Route(dict):
         if envoy_route == 'prefix':
             match['prefix'] = route_prefix
         else:
-            match.update(regex_matcher(config, route_prefix))
+            # Cheat.
+            if config.ir.edge_stack_allowed and (self.get('_precedence', 0) == -1000000):
+                # Force the safe_regex engine.
+                match.update({
+                    "safe_regex": {
+                        "google_re2": {
+                            "max_program_size": 200,
+                        },
+                        "regex": route_prefix
+                    }
+                })
+            else:
+                match.update(regex_matcher(config, route_prefix))
 
         headers = self.generate_headers(config, group)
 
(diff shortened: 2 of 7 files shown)


python/ambassador/ir/irambassador.py
```diff
@@ -276,6 +276,33 @@ def add_mappings(self, ir: 'IR', aconf: Config):
                 mapping.referenced_by(self)
                 ir.add_mapping(aconf, mapping)
 
+        if ir.edge_stack_allowed:
+            if self.diagnostics and self.diagnostics.get("enabled", False):
```
alexgervais: `False` by default? would this break existing installs?

kflynn: Hmmm. So this is mimicking the logic that we used for the old diag service - cf irambassador.py line 188 - where, basically, if you have no `diagnostics` key, then we enable it, but if you list `diagnostics` then you need to be explicit about `enabled`. We could change that...

docs/reference/core/ambassador.md
```diff
@@ -68,13 +68,26 @@ cors:
   ...
 ```
 
-The `diagnostics` service (at /ambassador/v0/diag/) defaults on, but you can disable the API route. It will remain accessible on diag_port.
+`diagnostics` configures Ambassador's diagnostics services.
+
+- Both API Gateway and Edge Stack provide low-level diagnostics at `/ambassador/v0/diag/`.
```
scoyle391: Both the API Gateway.

docs/reference/core/ambassador.md
```diff
@@ -68,13 +68,26 @@ cors:
...
+- Ambassador Edge Stack also provides the higher-level Edge Policy Console at `/edge_stack/admin/`.
```
scoyle391: The Ambassador Edge Stack

docs/reference/core/ambassador.md
LukeShu: and the Edge Stack

python/ambassador/envoy/v2/v2route.py
```diff
@@ -92,7 +92,19 @@ def __init__(self, config: 'V2Config', group: IRHTTPMappingGroup, mapping: IRBas
...
+            if config.ir.edge_stack_allowed and (self.get('_precedence', 0) == -1000000):
+                # Force the safe_regex engine.
+                match.update({
+                    "safe_regex": {
+                        "google_re2": {
+                            "max_program_size": 200,
+                        },
+                        "regex": route_prefix
+                    }
+                })
```
LukeShu: What's the reason to cheat here?  The regex is simple enough that the unsafe portions of ecmascript regex shouldn't come in to play, if that's the regex engine that the ambassador module sets it to use.

python/ambassador/ir/irambassador.py
LukeShu: On Monday in #dev-general, I described this by writing
```yaml

diag_port: int           # optional, default is 8877
liveness_probe:          # optional, default is {enabled: true}
  enabled: bool            # optional, default is false
  prefix:  "path"          # optional, default is "/ambassador/v0/check_alive"
  rewrite: "path"          # optional, default is "/ambassador/v0/check_alive"
  service: "host[:port]"   # optional, default is f"127.0.0.1:{diag_port}"
readiness_probe:         # optional, default is {enabled: true}
  enabled: bool            # optional, default is false
  prefix:  "path"          # optional, default is "/ambassador/v0/check_ready"
  rewrite: "path"          # optional, default is "/ambassador/v0/check_ready"
  service: "host[:port]"   # optional, default is f"127.0.0.1:{diag_port}"
diagnostics:             # optional, default is {enabled: true}
  enabled: bool            # optional, default is false
  prefix:  "path"          # optional, default is "/ambassador/v0/"
  rewrite: "path"          # optional, default is "/ambassador/v0/"
  service: "host[:port]"   # optional, default is f"127.0.0.1:{diag_port}"
```
The `false` default here only comes in to play if the `{enabled: true}` default in `__init__()` on line 101 doesn't get applied (because other fields in `.diagnostics` are set).

LukeShu: On Monday in #dev-general, I described this by writing
```yaml

diag_port: int           # optional, default is 8877
liveness_probe:          # optional, default is {enabled: true}
  enabled: bool            # optional, default is false
  prefix:  "path"          # optional, default is "/ambassador/v0/check_alive"
  rewrite: "path"          # optional, default is "/ambassador/v0/check_alive"
  service: "host[:port]"   # optional, default is f"127.0.0.1:{diag_port}"
readiness_probe:         # optional, default is {enabled: true}
  enabled: bool            # optional, default is false
  prefix:  "path"          # optional, default is "/ambassador/v0/check_ready"
  rewrite: "path"          # optional, default is "/ambassador/v0/check_ready"
  service: "host[:port]"   # optional, default is f"127.0.0.1:{diag_port}"
diagnostics:             # optional, default is {enabled: true}
  enabled: bool            # optional, default is false
  prefix:  "path"          # optional, default is "/ambassador/v0/"
  rewrite: "path"          # optional, default is "/ambassador/v0/"
  service: "host[:port]"   # optional, default is f"127.0.0.1:{diag_port}"
```
The `false` default here only comes in to play if the `{enabled: true}` default in `__init__()` on line 101 doesn't get applied (because other fields in `.diagnostics` are set).

python/ambassador/envoy/v2/v2route.py
kflynn: The cheating here isn't out of concern around safety: it's because the dictionary keys for safe-regex prefixes are different than they are for unsafe-regex prefixes, and it's easier to force it on setting than to look up both keys later.

(A more right answer would be to refactor things to remember the fallback route _a priori_, and carry that all the way through to the `V2Route`. I don't want to add more private members to `V2Route` though, and I don't want to refactor that class for more elegance right now.)

kflynn:
> Because (I believe) a customer also requested this feature, this should also probably be mentioned in /about/edge-policy-console, shouldn't it?

Done.

PR #2328, opened 2020-02-19
Prepare for Envoy 1.13 upgrade; migrate from now-disabled deprecated fields

LukeShu:
## Description

This is everything (so far?) that is a part of upgrading to Envoy 1.13, short of doing the actual upgrade and things that depend on the actual upgrade.  The biggest part of that is dealing with Envoy disallowing many previously deprecated fields (datawire/apro#713).

It probably makes sense to review this commit-by-commit.

## Related Issues

PR https://github.com/datawire/ambassador/pull/2318
Issue datawire/apro#713
Issue datawire/apro#998


python/ambassador/envoy/v2/v2listener.py
```diff
@@ -788,9 +791,8 @@ def __init__(self, config: 'V2Config', service_port: int) -> None:
         if self.config.ir.tracing:
             self.base_http_config["generate_request_id"] = True
 
-            self.base_http_config["tracing"] = {
-                "operation_name": "egress"
-            }
+            self.base_http_config["tracing"] = {}
+            self.traffic_direction = "OUTBOUND"
```
LukeShu: This seems wrong-ish to me, but I think it's what preserves existing behavior.

kflynn: I think you're correct. @alexgervais might have some thoughts as to whether we should support configuring this.

alexgervais: LGTM
At some point, `"operation_name": "egress"` was required for tracing to work on upstream calls. Having a quick look at the docs, and it would seem `traffic_direction` could be omitted entirely. Anyways, I'd be happy to test tracing manually before shipping.

PR #2336, opened 2020-02-20
Edgectl Intercept using mappings

kflynn:
Neuter a lot of the old sidecar stuff and use an actual Ambassador to do routing, with `edgectl intercept` managing the mappings for the intercepts.

It's probably simplest to review the whole thing at once, rather than commit-by-commit.

### DIFF
--- a/python/ambassador/ir/ir.py
+++ b/python/ambassador/ir/ir.py
@@ -35,6 +35,7 @@ from .irfilter import IRFilter
 from .ircluster import IRCluster
 from .irbasemappinggroup import IRBaseMappingGroup
 from .irbasemapping import IRBaseMapping
+from .irhttpmapping import IRHTTPMapping
 from .irhost import IRHost, HostFactory
 from .irmappingfactory import MappingFactory
 from .irratelimit import IRRateLimit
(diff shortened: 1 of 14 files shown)


python/ambassador/ir/ir.py
```diff
@@ -426,12 +429,19 @@ def agent_finalize(self, aconf) -> None:
         # routed insecure. !*@&#*!@&#* We need per-mapping security settings.
         #
         # XXX What if they already have a mapping with this name?
+
+        ctx_name = None
+
+        if self.agent_orgination_ctx:
+            ctx_name = self.agent_origination_ctx.name
```
ark3: These are spelled differently. Or my eyes are _really_ bad today.

python/ambassador/ir/ir.py
```diff
@@ -417,14 +421,18 @@ def agent_finalize(self, aconf) -> None:
 
         self.logger.info(f"Intercept agent active for {self.agent_service}:{agent_port}, adding fallback mapping")
 
+        # XXX OMG this is a crock. Don't use precedence -1000000 for this, because otherwise Edge
+        # Control might decide it's the Edge Policy Console fallback mapping and force it to be
+        # routed insecure. !*@&#*!@&#* We need per-mapping security settings.
```
ark3: "Edge Control" -> "Edge Stack"?

python/ambassador/ir/ir.py
kflynn: You were right, as it happens. FFS.

python/ambassador/ir/ir.py
kflynn: Done.

PR #2338, opened 2020-02-21
More backports from the Envoy 1.13 adventure

LukeShu:
## Description

These are some changes that I made when working on the Envoy 1.13 upgrade that I super duper want to land before we build another Envoy (regardless of whether it's a 1.13.x or still a 1.12.x).

## Testing

 - [x] `make update-base` is currently still running (so I expect CI to fail until that finishes)
 - [x] verify that the binary in the new Docker image matches what we've been shipping (result: it's not, but within expected difference of GCC vs Clang)
 - [x] `make pytest KAT_RUN_MODE=envoy`
 - [ ] `make check-envoy`

## Todos
- [ ] Tests (see above)

## Other

This is targeted at `rel/v1.2.0`; but I can re-target at a different branch if necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ branches:
   only:
     - master
     - /^shared\/.*/
+    - /^rel\/v.*/
     - /^v[0-9]+\.[0-9]+\.[0-9]+(-.*)?$/
 
 services:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,20 +24,25 @@ for all users, and includes all the functionality of the Ambassador API Gateway 
 to the additional capabilities mentioned above. Due to popular demand, we’re offering a free
 tier of our core features as part of the Ambassador Edge Stack, designed for startups.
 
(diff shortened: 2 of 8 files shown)


LukeShu:
> verify that the binary in the new Docker image matches what we've been shipping

So it looks like what we've been shipping was built with GCC instead of clang.  So there are lots of small differences.

LukeShu:
So I'm actually expecting the ratelimit tests from `make check-envoy` to fail, because https://github.com/datawire/envoy/commit/35c9d50ed17baa22eb98ccb8435cb5b579a93364 doesn't update the tests.  The RL is affectively tested by KAT, so nothing to worry about.

LukeShu:
`out of 518 tests: 235 tests pass, 1 fails to build, 1 fails locally and 281 were skipped.`

The failures are the RL thing.  The skips are `There were tests whose specified size is too big.` because apparently 160GB isn't enough disk space.

LukeShu:
Eh, I'm not worried; KAT says it's fine and it's the same C++ as we've been shipping.

PR #2352, opened 2020-02-25
More custom-build-name stuff

kflynn:
Extend the custom names into the `SNAPSHOT` and `REPO` variables, to fully isolate images across builds.

**NOTE WELL** that if you set `BUILDER_NAME=foo`, then instead of getting an image named e.g. `ambassador:latest`, you'll now get `foo:latest`. `make export` will show you the correct information, of course.


rhs:
Could you add the variable to the output of `make env`?

kflynn:
Updated the help. Help changes might be easier commit-by-commit?

PR #2356, opened 2020-02-26
Reorganize the layers in the ambassador image

ark3:
Reorganize the layers in the ambassador image

Goals
- Less duplication of data
- Fewer layers to push during most dev work

Changes
- Move kubectl (~50MB) way earlier
- Move setcap stuff as early as possible because it duplicates the
  binary it's modifying
- Order layers by how likely we are to modify them during development
  From most stable to most variable:
  - System dependencies and libraries
  - Our Python dependencies
  - Our Envoy
  - Our Go binaries
  - Our Python code
  - Stuff that changes on every build (to be fixed later)

  We probably change Python and Go code equally often, but the Go blob
  is much bigger, so I put it earlier.
- Add comments about the contents of each set of layers

The biggest benefits come from moving `kubectl` and `envoy` out of the critical path of layers that get pushed when we change Python or Go code. That's roughly 70MB less pushing every change.

A future big win (to be had in a separate PR) is to reduce the size of the Go binaries layer by combining the binaries. That is a separate task I may get to this week. Possibly.

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -105,38 +105,45 @@ FROM ${artifacts} as artifacts
 
 FROM ${base} as ambassador
 
+# External stuff that should change infrequently
 RUN apk --no-cache add bash curl python3 libcap
-
 RUN ln -s /usr/bin/python3 /usr/bin/python
-
-COPY --from=envoy /usr/local/bin/envoy-static-stripped /usr/local/bin/envoy
-COPY --from=artifacts /usr/lib/python3.7/site-packages /usr/lib/python3.7/site-packages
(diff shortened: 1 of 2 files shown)


ark3:
The original has this output every time for me:
```
Step 40/41 : RUN setcap cap_net_bind_service=p /usr/local/bin/wrapper; setcap cap_net_bind_service=ei /usr/local/bin/envoy
 ---> Running in 849ecb288baa
NOTE: Under Linux, effective file capabilities must either be empty, or
      exactly match the union of selected permitted and inheritable bits.
Failed to set capabilities on file `/usr/local/bin/wrapper' (Invalid argument)
```

I see the same output in the corresponding spot in this version too.

PR #2380, opened 2020-03-04
Add tcp mapping to the overview.html page

jfrabaute:
## Description
Add tcp mapping to the overview.html page

## Related Issues
The overview page shows only the http mappings, so adding the tcp mapping is useful when there are some defined.

It's very basic right now, it probably needs improvements, polishing. It's a first iteration.
I don't know if it can be merged as is as a first basic step or if it's better to have something more polished for merging.

### DIFF
--- a/python/templates/overview.html
+++ b/python/templates/overview.html
@@ -277,6 +277,59 @@ limitations under the License
 
       <div class="row">
         <div class="col-12">
+          Ambassador TCP Mappings
+
+          <div class="row">
+            <div class="col-12">
+              <table cellpadding="2em" width="100%">
+                <thead>
+                  <td><b>Name</b></td>
+                  <td><b>Port</b></td>
+                  <td><b>Source</b></td>
+                  <td><b>Service</b></td>
+                </thead>
+                <tbody>
+                  {% for group_id, group in groups.items() if group.kind == "IRTCPMappingGroup" %}
+                  <tr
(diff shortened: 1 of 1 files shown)


kflynn:
So far this looks pretty reasonable! Were you going to drop the `GROUP ` prefix first, or... ?

jfrabaute:
> So far this looks pretty reasonable! Were you going to drop the `GROUP ` prefix first, or... ?

Yes, sorry, I didn't send a change for this yet. I'll do it on monday.

jfrabaute:
> So far this looks pretty reasonable! Were you going to drop the `GROUP ` prefix first, or... ?

Done.

jfrabaute:
> * I think it'd be better to put the TCP Mappings group up above the service color chart.

Done

> * I _think_ you should be able to use the same color information for the status of the service?

I couldn't find this information. For the routes, there is a `_hcolor` field, but it's not present in the groups for tcp mappings.

> * `test-mapping-2`'s `host` isn't displayed, and it probably needs to be. You could do this the way we show headers on HTTP `Mapping`s?

Done

I've also changed the `Services` column to display the backend service.
I was displaying the `group.location` before, which I moved to a new column `Source`.

Screenshot:
image

jfrabaute:
Any chance to have this reviewed?

jfrabaute:
Rebased on top of latest master.

PR #2402, opened 2020-03-11
Clean up behavior around edgectl connect and namespaces

ark3:
- Namespace, where teleproxy runs, is whatever the context says by default
- Manager namespace, where the manager runs, is ambassador by default
- Both can be set independently

### DIFF
--- a/cmd/edgectl/intercept.go
+++ b/cmd/edgectl/intercept.go
@@ -110,7 +110,7 @@ func (d *Daemon) ListIntercepts(_ *supervisor.Process, out *Emitter) error {
 	for idx, cept := range d.intercepts {
 		ii := cept.ii
 		out.Printf("%4d. %s\n", idx+1, ii.Name)
-		out.Send("local_intercept", ii.Name)
+		out.Send(fmt.Sprintf("local_intercept.%d", idx+1), ii.Name)
 		key := "local_intercept." + ii.Name
 		out.Printf("      Intercepting requests to %s when\n", ii.Deployment)
 		out.Send(key, ii.Deployment)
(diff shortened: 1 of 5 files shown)


cmd/edgectl/intercept.go
```diff
@@ -262,8 +262,6 @@ func (cept *Intercept) removeMapping(p *supervisor.Process) error {
 	if cept.mappingExists {
 		p.Logf("%v: Deleting mapping in namespace %v", cept.ii.Name, cept.ii.Namespace)
 		delete := cept.cluster.GetKubectlCmd(p, "delete", "-n", cept.ii.Namespace, "mapping", fmt.Sprintf("%s-mapping", cept.ii.Name))
-		delete.Stdout = os.Stdout
-		delete.Stderr = os.Stderr
```
kflynn: Hmmm. So what happens if this fails?

cmd/edgectl/intercept.go
```diff
@@ -326,7 +324,6 @@ func MakeIntercept(p *supervisor.Process, out *Emitter, tm *TrafficManager, clus
 	}
 
 	manifest, err := json.MarshalIndent(&mapping, "", "  ")
-
```
kflynn: Dude. Vertical whitespace is not a bad thing. :joy:

cmd/edgectl/notify.go
```diff
@@ -7,10 +7,21 @@ import (
 	"github.com/datawire/ambassador/pkg/supervisor"
 )
 
-var notifyRAI *RunAsInfo
+var (
+	notifyRAI     *RunAsInfo
+	notifyEnabled = false
```
kflynn: Did I miss a way to turn this back on?

cmd/edgectl/intercept.go
ark3: The output gets logged anyhow (by `supervisor.Command`). This was causing it all to be doubled, half the lines with the logging headers and half without.

cmd/edgectl/intercept.go
ark3: I blame `gofmt`! I didn't remove those on purpose.

cmd/edgectl/notify.go
ark3: Oh, good point. Not everything in my tree got committed; I should have left this out too. That said, notifications don't work on Linux (any longer), so I'm not too fussed about leaving it off until the next RC. Is that okay?

cmd/edgectl/intercept.go
kflynn: Ahhhh gotcha.

cmd/edgectl/intercept.go
kflynn: *!@&*!#& but whatever. :wink:

cmd/edgectl/notify.go
kflynn: I'm OK with leaving it for the next RC. Let's talk about it later; I think the notifications could be very useful if tweaked a bit.

PR #2409, opened 2020-03-13
Update Intercept documentation and Changelog

ark3:


docs/reference/edge-control.md
```diff
@@ -101,8 +99,12 @@ Depending on the type of cluster, your operations team may be involved. If you o
 ### Traffic Manager
```
richarddli: Could we add a description of what Traffic Manager does?

docs/reference/edge-control.md
```diff
@@ -1,12 +1,10 @@
 # Edge Control
 
-Edge Control is the architecture for `edgectl` and allows you to provide the functionality for application users and ease of use for your cluster modification. With Edge Control, developers can safely share a single development cluster, and enable the "always-on" experience originally achieved with custom workarounds. For your users, Edge Control allows you to provide basic and advanced functionality to your application with minimal cluster-side modifications.
+Edge Control is the command-line tool for installing and managing the Ambassador Edge Stack. And Edge Control's outbound and intercept features allow developers to preview changes to their services while sharing a single development cluster.
 
-You can use Edge Control for developing new services, and debugging existing services.
+**New Service**: If you are a developer and you want to write a new service, it depends on existing services running in your cluster. You can use the `edgectl connect` command to set up outbound connectivity from your laptop to your cluster. This allows the work-in-progress implementation of your new service to connect to existing services from your laptop.
```
richarddli: I'm not sure I follow the framing of connect for new and intercept for existing? I can see a new service where you may want to route traffic to it, and I can also see an existing service that wants to connect to services in the cluster.

So it seems to me that we should just frame this as:

* If you have a service that depends on other in-cluster services, use `edgectl connect` set up routing from your laptop to the cluster ...
* If you want to test your service with traffic from the other cluster, use `edgectl intercept` ...

PR #2428, opened 2020-03-17
Improve Knative support

### LINKED ISSUE #2140: Setting timeout_ms for knative services
Hello,

I created a test minikube cluster following this document: https://knative.dev/docs/install/knative-with-ambassador/ which is very similar to this https://www.getambassador.io/user-guide/knative. Now when I try to call my knative service first couple requests always fail with "upstream request timeout" because knative needs time to scale up deployment for my service. In normal use of mappings that can be addressed by setting timeout_ms, but I can't find any references on how to do it for knative services where everything happens under the hood.

Thanks.

impl:
## Description
This change improves support for Knative:

* We move the CRD processing from the IR to the resource fetcher, and emit Mappings to the IR directly. This mirrors how Kubernetes Ingresses are currently processed.
* We make sure to check the current generation of the CRD against the observed generation; the current implementation will reprocess the CRD every second because the timestamp in the conditions changes.
* We map each host by itself instead of using a regular expression, as the regular expressions were not being escaped and I wasn't comfortable bringing in a dependency on pyre2 to do the escaping.
* We inform Knative of the internal route to the Ambassador service, if present. This should improve performance and allows cluster-local Knative services to resolve.
* We handle the path and timeout properties of the ingress paths in the CRDs.

I made these changes specifically to support our use case, so please let me know if anything seems out of place. Of course this is my first time hacking on this code as well; I am very open to any and all feedback!

## Related Issues
This incidentally fixes #2140.

## Testing
This has been tested using the integration tests in the repository and validated locally and in a testing cluster. We are planning to use this in a production environment soon.

## Tasks That Must Be Done

## Other

### DIFF
--- a/python/ambassador/config/config.py
+++ b/python/ambassador/config/config.py
@@ -75,8 +75,6 @@ class Config:
         'logservice': "log_services",
     }
 
-    KnativeResources = { 'clusteringress', 'knativeingress' }
-
     SupportedVersions: ClassVar[Dict[str, str]] = {
         "v0": "is deprecated, consider upgrading",
         "v1": "ok",
@@ -221,20 +219,9 @@ class Config:
     def as_json(self):
         return json.dumps(self.as_dict(), sort_keys=True, indent=4)
 
-    def ambassador_id_required(self, resource_kind: str) -> bool:
-        if resource_kind in self.KnativeResources:
-            return False
-
-        return True
-
     # Often good_ambassador_id will be passed an ACResource, but sometimes
     # just a plain old dict.
     def good_ambassador_id(self, resource: dict):
-        resource_kind = resource.get('kind', '').lower()
-        if not self.ambassador_id_required(resource_kind):
-            self.logger.debug(f"Resource: {resource_kind} does not require an Ambassador ID")
-            return True
-
         # Is an ambassador_id present in this object?
         #
         # NOTE WELL: when we update the status of a Host (or a Mapping?) then reserialization
--- a/python/ambassador/config/config.py
+++ b/python/ambassador/config/config.py
@@ -16,6 +16,7 @@ import socket
 from typing import Any, Callable, ClassVar, Dict, Iterable, List, Optional, Tuple, Union
 from typing import cast as typecast
 
+import collections
 import importlib
 import json
 import logging
@@ -101,6 +102,8 @@ class Config:
     breakers: Dict[str, ACResource]
     outliers: Dict[str, ACResource]
 
+    counters: Dict[str, int]
+
     # rkey => ACResource
     sources: Dict[str, ACResource]
 
@@ -159,6 +162,8 @@ class Config:
 
         self.breakers = {}
         self.outliers = {}
+
+        self.counters = collections.defaultdict(lambda: 0)
 
         self.sources = {}
 
@@ -222,6 +227,8 @@ class Config:
     # Often good_ambassador_id will be passed an ACResource, but sometimes
     # just a plain old dict.
     def good_ambassador_id(self, resource: dict):
+        resource_kind = resource.get('kind', '')
+
         # Is an ambassador_id present in this object?
         #
         # NOTE WELL: when we update the status of a Host (or a Mapping?) then reserialization
@@ -255,6 +262,12 @@ class Config:
 
                 self.logger.debug(f"{rkey}: {resource_kind} {name} has IDs {allowed_ids}, no match with {Config.ambassador_id}")
                 return False
+
+    def incr_count(self, key: str) -> None:
+        self.counters[key] += 1
+
+    def get_count(self, key: str) -> int:
+        return self.counters.get(key, 0)
 
     def save_source(self, resource: ACResource) -> None:
         """
--- a/python/ambassador/config/resourcefetcher.py
+++ b/python/ambassador/config/resourcefetcher.py
@@ -1,11 +1,15 @@
 from typing import Any, Dict, List, Optional, Tuple, TYPE_CHECKING
 # from typing import cast as typecast
 
+import datetime
+import itertools
 import json
 import logging
 import os
 import yaml
 import re
+
+import durationpy
 
 from .config import Config
 from .acresource import ACResource
@@ -49,6 +53,8 @@ CRDTypes = frozenset([
     'ingresses.networking.internal.knative.dev'
 ])
 k8sLabelMatcher = re.compile(r'([\w\-_./]+)=\"(.+)\"')
+
+AMBASSADOR_KNATIVE_INGRESS_CLASS = 'ambassador.ingress.networking.knative.dev'
 
 class ResourceFetcher:
     def __init__(self, logger: logging.Logger, aconf: 'Config',
@@ -339,37 +345,17 @@ class ResourceFetcher:
         namespace = metadata.get('namespace') or 'default'
         metadata_labels: Optional[Dict[str, str]] = metadata.get('labels')
         generation = metadata.get('generation', 1)
-        annotations = metadata.get('annotations', {})
 
         if not self.check_k8s_dup(kind, namespace, name):
             return
 
         spec = obj.get('spec') or {}
+        status = obj.get('status') or {}
 
         # Replace a sentinel value with the namespace of this ambassador pod.
         # This allows hard-coded initialization resources to have a useful namespace.
         if namespace == "_automatic_":
             namespace = Config.ambassador_namespace
-
-        if not apiVersion:
-            # I think this is impossible.
-            self.logger.debug(f'{self.location}: ignoring K8s {kind} CRD, no apiVersion')
-            return
-
-        # We do not want to confuse Knative's Ingress with Kubernetes' Ingress
-        if apiVersion.startswith('networking.internal.knative.dev') and kind.lower() == 'ingress':
-            self.logger.debug(f"Renaming kind {kind} to KnativeIngress")
-            kind = 'KnativeIngress'
-
-            # Let's not parse KnativeIngress if it's not meant for us.
-            # We only need to ignore KnativeIngress iff networking.knative.dev/ingress.class is present in annotation.
-            # If it's not there, then we accept all ingress classes.
-            if 'networking.knative.dev/ingress.class' in annotations:
-                if annotations.get('networking.knative.dev/ingress.class').lower() != 'ambassador.ingress.networking.knative.dev':
-                    self.logger.debug(f'Ignoring KnativeIngress {name}; set networking.knative.dev/ingress.class '
-                                      f'annotation to ambassador.ingress.networking.knative.dev for ambassador to '
-                                      f'parse it.')
-                    return
 
         if not name:
             self.logger.debug(f'{self.location}: ignoring K8s {kind} CRD, no name')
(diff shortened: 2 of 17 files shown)


python/ambassador/config/resourcefetcher.py
```diff
@@ -376,6 +367,154 @@ def handle_k8s_crd(self, obj: dict) -> None:
...
+                status['loadBalancer'] = load_balancer
+                status['privateLoadBalancer'] = load_balancer
+
+            status_update = (f"{kind.lower()}.networking.internal.knative.dev", namespace, status)
+
+            self.logger.info(f"Updating Knative {kind} {name} status to {status_update}")
+            self.aconf.k8s_status_updates[f"{name}.{namespace}"] = status_update
+        else:
+            self.logger.debug(f"Not reconciling Knative {kind} {name}: observed and current generations are in sync")
+
```
kflynn: This looks like it's mostly the old `cluster_ingresses_to_mapping` function? I'm curious why you inlined it, and what the changes between the two really are? (I can split it out and run `diff`, but that won't tell me what you were thinking. :wink:)

(Obviously one change is that `cluster_ingresses_to_mappings` wasn't in this file...)

python/ambassador/ir/ir.py
```diff
@@ -976,12 +863,6 @@ def features(self) -> Dict[str, Any]:
         od['endpoint_routing_envoy_maglev_count'] = endpoint_routing_envoy_maglev_count
         od['endpoint_routing_envoy_lr_count'] = endpoint_routing_envoy_lr_count
 
-        cluster_ingresses = self.aconf.get_config("ClusterIngress")
-        od['cluster_ingress_count'] = len(cluster_ingresses.keys()) if cluster_ingresses else 0
-
-        knative_ingresses = self.aconf.get_config("KnativeIngress")
-        od['knative_ingress_count'] = len(knative_ingresses.keys()) if knative_ingresses else 0
-
```
kflynn: It would definitely be a good thing to preserve this count somehow. Maybe just keep a countin the `aconf`?

python/ambassador/config/resourcefetcher.py
```diff
@@ -310,38 +316,18 @@ def handle_k8s_crd(self, obj: dict) -> None:
...
 
         # Replace a sentinel value with the namespace of this ambassador pod.
         # This allows hard-coded initialization resources to have a useful namespace.
         if namespace == "_automatic_":
             namespace = Config.ambassador_namespace
 
-        if not apiVersion:
-            # I think this is impossible.
-            self.logger.debug(f'{self.location}: ignoring K8s {kind} CRD, no apiVersion')
-            return
```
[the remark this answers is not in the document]
impl: It was actually duplicated! It's still present in the code, a few lines below this. :stuck_out_tongue:

python/ambassador/config/resourcefetcher.py
```diff
@@ -350,11 +336,16 @@ def handle_k8s_crd(self, obj: dict) -> None:
...
+        if apiVersion.startswith('networking.internal.knative.dev'):
+            self.handle_k8s_knative_crd(kind, namespace, name, metadata, spec, status)
+            return
+
         # if not spec:
         #     self.logger.debug(f'{self.location}: ignoring K8s {kind} CRD {name}: no spec')
         #     return
 
         # We use this resource identifier as a key into self.k8s_services, and of course for logging .
+        annotations = metadata.get('annotations', {})
```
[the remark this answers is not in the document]
impl: I think that is accurate! Will remove.

python/ambassador/ir/ir.py
impl: Sounds good. I'll figure out how to bring it back. :smile:

python/ambassador/config/resourcefetcher.py
impl: I'd be happy to move it to a separate function if it clarifies things. It was moved to this file so I could access the `status` of the resource initially, but then I also found the placement to be more sensible because treating external CRDs like IR (instead of emitting IR) didn't feel like the best fit.

Aside from that, the main changes are:
* We generate a Mapping per-host-and-service (what Knative calls a "split") combination (previously we were emitting one per split only, with the host as a regex).
* We handle the path and timeout from the Knative Ingress and pass it on to the Mapping (not previously handled at all).
* We used to write out an update to the status of Knative CRD for each split, which didn't make sense to me as it was updating the same key, but now it's just generated once.
* We set the load balancer endpoint if we can resolve it, same as with Kubernetes ingresses.
* The status update is only written if the generation or load balancer endpoint changes.

python/ambassador/config/resourcefetcher.py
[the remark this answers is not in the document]
kflynn: D'oh! Thanks. :)

python/ambassador/config/resourcefetcher.py
kflynn: Thanks for the summary! All of that sounds really good. Let's talk about structure on Slack a bit -- to me this does feel like multiple functions, but I want to make sure that I'm not just needlessly complicating the world. 🙂

python/ambassador/ir/ir.py
kflynn: Thanks!

python/ambassador/config/config.py
```diff
@@ -218,19 +221,10 @@ def as_dict(self) -> Dict[str, Any]:
...
-        if resource_kind in self.KnativeResources:
-            return False
-
-        return True
-
     # Often good_ambassador_id will be passed an ACResource, but sometimes
     # just a plain old dict.
     def good_ambassador_id(self, resource: dict):
-        resource_kind = resource.get('kind', '').lower()
-        if not self.ambassador_id_required(resource_kind):
```
impl, line 231: I removed this check because it seemed like it prevented this method from working with regular Ambassador resources. But I can add it back if that behavior is intentional.

python/ambassador/config/resourcefetcher.py
```diff
@@ -151,8 +164,7 @@ def load_from_filesystem(self, config_dir_path, recurse: bool=False,
             self.finalize()
 
     def parse_yaml(self, serialization: str, k8s=False, rkey: Optional[str]=None,
-                   filename: Optional[str]=None, finalize: bool=True, namespace: Optional[str]=None,
-                   metadata_labels: Optional[Dict[str, str]]=None) -> None:
+                   filename: Optional[str]=None, finalize: bool=True, metadata_labels: Optional[Dict[str, str]]=None) -> None:
```
impl, line 167: I removed the namespace arg from this method and its callees because it wasn't called with it anywhere that I could find and propagating it to the new resource processing code felt convoluted.

python/ambassador/config/resourcefetcher.py
```diff
@@ -261,63 +271,59 @@ def check_k8s_dup(self, kind: str, namespace: str, name: str) -> bool:
...
+            # can't process it.
             return
 
-        metadata = obj.get('metadata') or {}
-        name = metadata.get('name') or '(no name?)'
-        namespace = metadata.get('namespace') or 'default'
+        with self.manager.locations.push_reset():
+            if self.k8s_processor.try_process(obj):
+                # Nothing else to do.
+                return
```
impl, line 287: My plan is to fast-follow this PR with a few more PRs that move Kubernetes object handling from the fetcher to individual KubernetesProcessor subclasses. So this will eventually be the end of this method.

python/ambassador/config/resourceprocessor.py
```diff
@@ -0,0 +1,671 @@
...
+        # statement. At the conclusion of the statement block, the location will
+        # automatically be popped from the stack.
+        @contextlib.contextmanager
+        def popper():
+            yield current
+            self.pop()
+
+        return popper()
+
+    def push_reset(self) -> ContextManager[Location]:
```
impl, line 69: I wasn't 100% sure if the justification for this method existing is correct but it's how parse_object works when used inside the Kubernetes handler. Happy to accept suggestions for a better name/docstring.

python/ambassador/config/resourceprocessor.py
```diff
@@ -0,0 +1,671 @@
...
+        ir_obj['namespace'] = namespace
+        ir_obj['kind'] = kind
+        ir_obj['generation'] = generation
+        ir_obj['labels'] = labels or {}
+
+        return cls(ir_obj, rkey)
+
+
+@dataclasses.dataclass(frozen=True)
+class KubernetesGVK:
```
impl, line 111: I'm not sure it's been used elsewhere in Ambassador code but this is a pretty common acronym in controller-runtime. Happy to use an alternative if you have one.

python/ambassador/config/resourceprocessor.py
```diff
@@ -0,0 +1,671 @@
...
+    @property
+    def spec(self) -> Dict[str, Any]:
+        return self.get('spec', {})
+
+    @property
+    def status(self) -> Dict[str, Any]:
+        return self.get('status', {})
+
+    def as_resource_emission(self) -> ResourceEmission:
+        if self.gvk.api_group != 'getambassador.io':
```
impl, line 234: I believe this is true now (everything stored in aconf most have an apiVersion of getambassador.io/v{0,1,2}) but want to surface it for discussion if that isn't true.

python/ambassador/config/resourceprocessor.py
```diff
@@ -0,0 +1,671 @@
...
+            'LogService',
+            'Mapping',
+            'Module',
+            'RateLimitService',
+            'TCPMapping',
+            'TLSContext',
+            'TracingService',
+        ]
+
+        return frozenset([KubernetesGVK.for_ambassador(kind) for kind in kinds])
```
impl, line 420: Right now this will only handle Kubernetes objects that have v2 as their Ambassador API version. Will we need to also support v0 and v1? (Note this only applies to actual Kubernetes objects, not annotations.)

python/ambassador/utils.py
```diff
@@ -59,7 +59,7 @@
 yaml_logged_dumper = False
 
 
-def parse_yaml(serialization: str, **kwargs) -> Any:
+def parse_yaml(serialization: str) -> Any:
```
impl, line 62: I removed the kwargs because they weren't being passed to the underlying parser and seemed to be confusingly misused in the resource fetcher.

python/tests/test_docker.py
```diff
@@ -107,4 +109,4 @@ def test_docker():
     assert test_status, 'test failed'
 
 if __name__ == '__main__':
-    test_docker()
+    pytest.main(sys.argv)
```
impl, line 112: Instead of calling the test methods directly you can just let pytest take care of them and then you don't have to worry about having one test entrypoint per file.

python/ambassador/config/resourceprocessor.py
```diff
@@ -0,0 +1,671 @@
...
+    @property
+    def aconf(self) -> Config:
+        return self.manager.aconf
+
+    @property
+    def logger(self) -> logging.Logger:
+        return self.manager.logger
+
+
+class AmbassadorProcessor (ManagedKubernetesProcessor):
```
impl, line 399: Note that this isn't actually used yet (though it is tested). I intend to switch over to it in a subsequent PR.

concaf:
FYI: also testing for regressions at https://github.com/knative/serving/pull/7716

python/ambassador/config/resourcefetcher.py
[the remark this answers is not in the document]
concaf: Just glanced through the code and could not find this check below :( Can you please point me to it @impl? (sorry, being a bit paranoid here)

kflynn:
Taking this -- all the build tests actually passed, but the script exited with status 1. 😛  We're suspecting that the cleanup script is erroring out. I'll be looking at that shortly.

PR #2432, opened 2020-03-18
builder: Use docker cp instead of rsync when possible

ark3:
... because `docker cp` is much faster than `rsync`, especially with a remote Docker host.

Reviewer: Please check my use/non-use of quotation marks.


ark3:
Yes

kflynn:
Cool. :)

PR #2433, opened 2020-03-18
Run the Ingress conformance tests in Travis

inercia:
## Description

Try to run the Ingress Conformance Tests in Travis

### DIFF
--- a/builder/builder.mk
+++ b/builder/builder.mk
@@ -24,6 +24,35 @@ BUILDER = BUILDER_NAME=$(BUILDER_NAME) $(abspath $(BUILDER_HOME)/builder.sh)
 DBUILD = $(abspath $(BUILDER_HOME)/dbuild.sh)
 COPY_GOLD = $(abspath $(BUILDER_HOME)/copy-gold.sh)
 
+# the image used for running the Ingress v1 tests with KIND.
+# the current, official image does not support Ingress v1, so we must build our own image with k8s 1.18.
+# build this image with:
+# 1. checkout the Kuberentes sources in a directory like "~/sources/kubernetes"
+# 2. kind build node-image --kube-root ~/sources/kubernetes
+# 3. docker tag kindest/node:latest quay.io/datawire/kindest-node:latest
+# 4. docker push quay.io/datawire/kindest-node:latest
+# This will not be necessary once the KIND images are built for a Kubernetes 1.18 and support Ingress v1beta1 improvements.
+# KIND_IMAGE ?= kindest/node:latest
+KIND_IMAGE ?= quay.io/datawire/kindest-node:latest
+KIND_KUBECONFIG = /tmp/kind-kubeconfig
+
+# The ingress conformance tests directory
+# build this image with:
+# 1. checkout https://github.com/kubernetes-sigs/ingress-controller-conformance
+# 2. cd ingress-controller-conformance && make image
+# 3. docker tag ingress-controller-conformance:latest quay.io/datawire/ingress-controller-conformance:latest
+# 4. docker push quay.io/datawire/ingress-controller-conformance:latest
+INGRESS_TEST_IMAGE ?= quay.io/datawire/ingress-controller-conformance:latest
+
+# local ports for the Ingress conformance tests
+INGRESS_TEST_LOCAL_PLAIN_PORT = 8000
+INGRESS_TEST_LOCAL_TLS_PORT = 8443
+
+# directory with the manifests for loading Ambassador for running the Ingress Conformance tests
+# NOTE: these manifests can be slightly different to the regular ones asd they include
+INGRESS_TEST_MANIF_DIR = $(BUILDER_HOME)/../docs/yaml/ambassador/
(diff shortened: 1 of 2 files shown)


builder/builder.mk
```diff
@@ -172,7 +185,63 @@ gotest: test-ready
...
+	@[ -n "$(AMB_IMAGE)" ] || { printf "$(RED)ERROR: no AMB_IMAGE defined$(END)\n"; exit 1; }
+	@[ -n "$(INGRESS_CONFORMANCE_TESTS_DIR)" ] ||  { printf "$(RED)ERROR: no INGRESS_CONFORMANCE_TESTS_DIR defined$(END)\n"; exit 1; }
+	@[ -d "$(INGRESS_CONFORMANCE_TESTS_DIR)" ] ||  { printf "$(RED)ERROR: $(INGRESS_CONFORMANCE_TESTS_DIR) does not seem a valid directory$(END)\n"; exit 1; }
+
+	@# make sure the `ingress-controller-conformance` executable has been built
+	@[ -x $(INGRESS_CONFORMANCE_TESTS_DIR)/ingress-controller-conformance ] || make -C $(INGRESS_CONFORMANCE_TESTS_DIR)
+
+	@printf "$(CYN)==> $(GRN)Creating cluster with KIND$(END)\n"
+	kind delete cluster 2>/dev/null || /bin/true
+	kind create cluster --image $(KIND_IMAGE)
```
alexgervais: We should probably create clusters with unique names, to prevent race conditions or 2 simultaneous runs sharing the same KIND cluster.

releng/travis-install.sh
```diff
@@ -18,6 +18,11 @@ GO_VERSION=1.13
 HELM_VERSION=2.9.1
 KUBECTL_VERSION=1.10.2
 KUBERNAUT_VERSION=2018.10.24-d46c1f1
+KIND_VERSION=v0.7.0  # see https://github.com/kubernetes-sigs/kind/releases
+
+# The ingress conformance tests (as a Zip, from GitHub)
+INGRESS_CONFORMANCE_TESTS_ZIP="https://github.com/alexgervais/ingress-controller-conformance/archive/ambassador-ref.zip"
```
alexgervais: We should use a binary package as soon as it's available, perhaps from the task in this story.

inercia: I agree. We could also build a Docker image with the Conformance tests and run them with a simple `docker run`. But any of these solutions would require some automatic build or CI on this repo, and that could take some time (and probably some discussions in the community).

builder/builder.mk
inercia: Not really important in Travis, as two Travis jobs will not share the same Docker daemon (and the KIND cluster is created there). The only problematic thing could be in local environments, but I'm not sure this would be a real problem... :thinking:

builder/builder.mk
```diff
@@ -24,6 +24,19 @@ BUILDER = BUILDER_NAME=$(BUILDER_NAME) $(abspath $(BUILDER_HOME)/builder.sh)
 DBUILD = $(abspath $(BUILDER_HOME)/dbuild.sh)
 COPY_GOLD = $(abspath $(BUILDER_HOME)/copy-gold.sh)
 
+# the image used for running the Ingress v1 tests with KIND
+# build this image with:
+# 1. checkout the Kuberentes sources in a directory like "~/sources/kubernetes"
+# 2. kind build node-image --kube-root ~/sources/kubernetes
+# 3. docker tag kindest/node:latest quay.io/datawire/kindest-node:latest
+# 4. docker push quay.io/datawire/kindest-node:latest
+# This will not be necessary once the KIND images are built for a Kuberentes that support Ingress v1
```
alexgervais: Just being picky, but the comment should say "... once the KIND images are built for a Kubernetes 1.18 and support Ingress v1beta1 improvements"

builder/builder.mk
```diff
@@ -172,7 +185,65 @@ gotest: test-ready
...
+	@printf "$(CYN)==> $(GRN)Creating cluster with KIND$(END)\n"
+	kind delete cluster 2>/dev/null || /bin/true
+	kind create cluster --image $(KIND_IMAGE)
+
+	@printf "$(CYN)==> $(GRN)Showing some cluster info:$(END)\n"
+	@kubectl cluster-info
+	@kubectl version
+
+	@printf "$(CYN)==> $(GRN)Loading Ambassador (from the Ingress conformance tests) with image=$(AMB_IMAGE)$(END)\n"
+	@for f in $(INGRESS_CONFORMANCE_TESTS_DIR)/examples/ambassador/*.yaml ; do \
```
alexgervais: Instead of relying on yaml samples from the ingress-conformance project (which we don't fully manage and could be a pain to update), can we use the manifests from ambassador.git/templates/ambassador?

inercia: Working on that.

releng/travis-install.sh
inercia: Submitted https://github.com/kubernetes-sigs/ingress-controller-conformance/pull/34 for packaging the conformance tests in a container.

PR #2467, opened 2020-03-19
Add cluster_tag to mapping to enable controlling the uniqueness of the cluster name

stefansedich:
## Description
Currently when looking at metrics they are grouped to the clusters which makes it not possible to segregate metrics at an endpoint or mapping level.

This change adds a `cluster_tag` parameter to mappings so that a unique prefix will be applied to the cluster names, for example given the following mappings:

```
---
kind: Mapping
metadata:
  name: mapping-a
spec:
  prefix: /a
  service: my-service.default
  cluster_tag: foo
---
kind: Mapping
metadata:
  name: mapping-b
spec:
  prefix: /b
  service: my-service.default
  cluster_tag: bah
```

Previously we would end up with one cluster, with this change we would see two which would provide a way to segregate things like metrics based on our unique endpoints.

```
cluster_foo_my_service_default_default
cluster_bah_my_service_default_default
```

The only downside of this approach is that I used the existing `marker` logic to append the tag, which results in the marker being added as a prefix, I personally would prefer a suffix but doing it in this way enabled me to re-use all the existing logic and make minimal changes.

## Related Issues
https://datawire-oss.slack.com/archives/CB46TNG83/p1578300808033800

## Testing
Initial manual testing and currently adding real tests.

## Tasks That Must Be Done


stefansedich:
Should be ready for review, am awaiting new from @kflynn on the new docs changes and process before doing any documentation.

stefansedich:
@kflynn just wrote a test to cover this scenario looks like it works as expected.

kflynn:
The CI "failure" is a bug with the doc-preview part of the CI run. D'oh! Merging. 🙂

PR #2471, opened 2020-03-20
Fix the location of the realm in the JWT Filter reference

nbkrause:


khussey:
@nbkrause documentation changes need to be carefully coordinated with what's happening on the shared/story-doc-improvement branch at this point.

nbkrause:
@khussey Thanks for the review. I made the change you requested. I left the original location changed as well since I believe I remember you saying those old pages will still be linked for some time. Please let me know if you need me to change that as well.

PR #2473, opened 2020-03-22
Installer: Check for an existing installation, skip applying

ark3:
Please comment/advise on FIXME and TODO comments...

The diff is easiest to read with whitespace changes ignored.

Also: This is only very lightly tested.

### DIFF
--- a/cmd/edgectl/aes_install.go
+++ b/cmd/edgectl/aes_install.go
@@ -408,7 +408,7 @@ func (i *Installer) Perform(kcontext string) error {
 	matches := aesVersionRE.FindStringSubmatch(aesManifests)
 	if len(matches) != 2 {
 		i.log.Printf("matches is %+v", matches)
-		// FIXME i.Report("something")
+		i.Report("fail_bad_manifests")
 		return errors.Errorf("Failed to parse downloaded manifests. Is there a proxy server interfering with HTTP downloads?")
 	}
 	aesVersion := matches[1]
@@ -475,14 +475,13 @@ func (i *Installer) Perform(kcontext string) error {
 			i.ShowWrapped(abortExisting)
 			i.show.Println()
 			i.ShowWrapped(seeDocs)
-			// FIXME i.Report("Something")
+			i.Report("fail_existing_aes", ScoutMeta{"installing", aesVersion}, ScoutMeta{"found", installedVersion})
 			return errors.Errorf("existing AES %s found when installing AES %s", installedVersion, aesVersion)
 		default:
 			i.ShowWrapped(abortCRDs)
 			i.show.Println()
 			i.ShowWrapped(seeDocs)
-			// FIXME: Explain how to remove existing CRDs? Or do something smarter?
-			// FIXME i.Report("Something")
+			i.Report("fail_existing_crds")
 			return errors.New("CRDs found")
 		}
 	}
@@ -908,6 +907,7 @@ This tool does not support upgrades/downgrades at this time.
 Aborting the installer to avoid corrupting an existing installation of AES.
 `
 
+// TODO: Explain how to remove existing CRDs? Or do something smarter?
 const abortCRDs = `
 -> Found Ambassador CRDs in your cluster, but no AES installation.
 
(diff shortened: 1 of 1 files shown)


cmd/edgectl/aes_install.go
```diff
@@ -319,6 +319,50 @@ func (i *Installer) CheckACMEIsDone(hostname string) error {
...
 }
 
+// GetAESCRDs returns the names of the AES CRDs available in the cluster (and
+// logs them as a side effect)
+func (i *Installer) GetAESCRDs() ([]string, error) {
+	crds, err := i.CaptureKubectl("get AES crds", "", "get", "crds", "-lproduct=aes", "-o", "name")
+	if err != nil {
+		return nil, err
+	}
+	res := make([]string, 0, 15)
```
alexgervais: Is there a reason to use 15 cap? It seems brittle and could cause issues for misconfigured/duplicated CRDs and won't play nicely if we ever reach that number of CRDs.

cmd/edgectl/aes_install.go
```diff
@@ -809,6 +901,19 @@ Timed out waiting for the load balancer's IP address for the AES Service.
 
 const tryAgain = "If this appears to be a transient failure, please try running the installer again. It is safe to run the installer repeatedly on a cluster."
 
+const abortExisting = `
+This tool does not support upgrades/downgrades at this time.
+
+Aborting the installer to avoid corrupting an existing installation of AES.
+`
+
+// TODO: Explain how to remove existing CRDs? Or do something smarter?
```
alexgervais: Yes, I think we should have instructions for removing the existing CRDs and resuming using the installer 👍

cmd/edgectl/aes_install.go
ark3: Have I misunderstood how that works? I thought I was just suggesting an initial capacity for the slice, but that `append` will increase it (by allocating a new array) as needed.

cmd/edgectl/aes_install.go
ark3: I couldn't come up with those instructions quickly, so I punted to a TODO. Thoughts? 

- `kubectl delete ambassador-crds` doesn't work
- `kubectl delete crd -l product=aes` only works if the CRDs are installed using our YAML. 
- `kubectl delete $(kubectl get crd -o name | grep "[.]getambassador[.]io$")` doesn't work on Windows

cmd/edgectl/aes_install.go
alexgervais: You are right, append will reallocate the array.

cmd/edgectl/aes_install.go
alexgervais: I would do `kubectl delete crd -l product=aes`. Assuming the listed CRDs are also visible in the terminal output, _most_ users should be able to figure it out.

ark3: Oh, the other issue is that deleting the CRDs also deletes and CRs of those types. Need to warn about that!

alexgervais:
Validated:
✔️ Empty cluster install
✔️ Completing the install after previously having installed the correct AES version
✔️ Handling "upgrade" path
```
edgectl install
-> Installing the Ambassador Edge Stack 1.3.0.
-> Found an existing installation of Ambassador Edge Stack 1.2.2.

This tool does not support upgrades/downgrades at this time.

Aborting the installer to avoid corrupting an existing installation of AES.

See https://www.getambassador.io/user-guide/getting-started/

Full logs at /var/folders/n6/w_qkrxw148j_l7h_45z35c0r0000gn/T/edgectl-install-20200323-150118.log

Error: install: existing AES 1.2.2 found when installing AES 1.3.0
```
✔️ Handling installing just the CRDs
```
edgectl install     
-> Installing the Ambassador Edge Stack 1.3.0.
-> Found Ambassador CRDs in your cluster, but no AES installation.

You can manually remove installed CRDs if you are confident they are not in use
by any installation.
Removing the CRDs will cause your existing Ambassador Mappings and other
resources to be deleted as well.

$ kubectl delete crd -l product=aes

Aborting the installer to avoid corrupting an existing (but undetected)
installation.

See https://www.getambassador.io/user-guide/getting-started/

Full logs at /var/folders/n6/w_qkrxw148j_l7h_45z35c0r0000gn/T/edgectl-install-20200323-150558.log

Error: install: CRDs found
```

PR #2480, opened 2020-03-24
Updated TLS documentation

nbkrause:
Updates to the TLS documentation for clarification.

Many old references have been removed and some have been kept. These changes can be reverted if we think we need to keep these old references around but I think they confuse the documentation.

TODO: Create document for serving HTTP and HTTPS on the same domain.


docs/docs/latest/howtos/client-cert-validation.md
```diff
@@ -0,0 +1,108 @@
...
+Sometimes, for additional security or authentication purposes, you will want
+the server to validate who the client is before establishing an encryopted 
+connection.
+
+To support this, Ambassador can be configured to use a provided CA certificate 
+to validate certificates sent from your clients. This allows for client-side 
+mTLS where both Ambassador and the client provide and validate each other's 
+certificates.
+
+## Prerequesites
```
richarddli: Prerequisites

docs/docs/latest/topics/running/tls/mtls.md
```diff
@@ -1,117 +1,76 @@
...
+For this reason, many organizations rely on a service mesh for their
+service-to-service authentication and encryption. 
+
+Ambassador integrates with multiple service meshes and makes it easy to
+configure mTLS to upstream services for all of them. Click the links below to 
+see how to configure Ambassador to do mTLS with any of these service meshes:
+
+- [Consul Connect](../../howtos/consul#encrypted-tls)
+
+- [Istio](../../howtos/istio#istio-mutual-tls)
```
richarddli: Did these Istio & Consul docs get copied over?

richarddli:
Regarding serving HTTP/HTTPS from same domain -- is that a separate document or can we cover that as an option in the TLS explanation docs?

PR #2484, opened 2020-03-24
Updating links to getting started page

khussey:
Updating links to reference new location for the getting started page.


khussey:
The 4 "bad" links being reported here are not broken, but the URLs are temporarily returning "too many requests".

kflynn:
I'm also told that these broken links are OK for this PR. :joy:

Issue #2466, opened 2020-03-19
edgectl install failing post helm install


lbpage:
**Describe the bug**
edgectl install failing post helm install

**To Reproduce**
1. helm install ambassador
2. install edgectl locally
3. run edgectl install
4. fail logs generated

**Expected behavior**
successful edgestack installation

**Versions (please complete the following information):**
 - Ambassador: 1.3.0
 - Kubernetes environment: azure aks
 - Version client/server: v1.17.0
 - helm: v3.1.1

**Additional context**
output logs gist
https://gist.github.com/lbpage/98aef229f524dd741b174b9a858936ad
https://gist.github.com/lbpage/c2a219a7865f69b0f180ab9dcfdfe1a6

devkws:
I had the same issue, and I swore this worked earlier in the week.... strange.

Anyways, I just used edgectl to install and everything works without the helm steps.

https://www.getambassador.io/user-guide/getting-started/

PR #2492, opened 2020-03-26
Use Helm for installing AES

inercia:
## Description

Use the Helm repo for running `edgectl install`

## Related Issues

#2466

## Testing

1 - Basic installation

* install with `edgectl install`
* After a successful installation, check that the Helm release is there:
```console
$ helm list -n ambassador 
NAME      	NAMESPACE 	REVISION	UPDATED                                 	STATUS  	CHART           	APP VERSION
ambassador	ambassador	1       	2020-04-08 14:03:12.893838399 +0200 CEST	deployed	ambassador-6.2.4	1.3.2      
```
* Check that the `app.kubernetes.io/managed-by` label is properly set:
```console
$ kubectl get deployments -n ambassador ambassador -o yaml | grep managed-by
    app.kubernetes.io/managed-by: edgectl
        app.kubernetes.io/managed-by: edgectl
```

2 - Installation after Helm

* Install AES with Helm. For example (in local env, with `NodePort`) with :
```console
$ kubectl create namespace ambassador
namespace/ambassador created
$ helm install ambassador --namespace ambassador datawire/ambassador --set service.type=NodePort --wait
# ...
```
* Check that `edgectl install` detects the Helm installation and continues with the rest of the setup:
```console
 $ ./edgectl install                

Installing the Ambassador Edge Stack

Please enter an email address for us to notify you before your TLS certificate
and domain name expire. In order to acquire the TLS certificate, we share this
email with Let’s Encrypt.
Email address [alvaro.saurin@*******]: 

========================================================================
Beginning Ambassador Edge Stack Installation

-> Checking latest version of Ambassador Edge Stack available...
-> Found an existing installation of Ambassador Edge Stack 1.3.2 [helm].
-> Ambassador was installed with Helm: will continue setup...
-> Checking the AES pod deployment
...
```

3 - Installation after an old Helm chart

* Same as previous scenario. Check that `edgectl install` aborts with a message saying it does not support upgrades.

4 - Installation after applying manifests from `https://www.getambassador.io/yaml/*.yaml`

* Same as previous scenario. Check that `edgectl install` aborts with a message saying it does not support upgrades.


cmd/edgectl/aes_install.go
```diff
@@ -365,6 +367,8 @@ func (i *Installer) GetInstalledImageVersion() (string, error) {
 
 // Perform is the main function for the installer
 func (i *Installer) Perform(kcontext string) error {
+	ctx := context.TODO()
```
ark3: There's `i.ctx` that will be canceled if the user hits Ctrl-C.

cmd/edgectl/aes_install.go
```diff
@@ -382,40 +386,11 @@ func (i *Installer) Perform(kcontext string) error {
...
-	}
-
-	// Download AES manifests
-	crdManifests, err := getManifest(fmt.Sprintf("https://%s/yaml/aes-crds.yaml", manifestsDomain))
+	namespace, err := i.kubeinfo.Namespace()
 	if err != nil {
-		i.Report("fail_no_internet", ScoutMeta{"err", err.Error()})
-		return errors.Wrap(err, "download AES CRD manifests")
+		i.log.Printf("%w", err)
+		os.Exit(1)
```
ark3: Why? Returning the error will result in an orderly shutdown.

cmd/edgectl/aes_install.go
```diff
@@ -747,11 +760,13 @@ func (i *Installer) Quit() {
 	i.cancel()
 }
 
-func (i *Installer) ShowWrapped(text string) {
-	text = strings.Trim(text, "\n")                  // Drop leading and trailing newlines
-	for _, para := range strings.Split(text, "\n") { // Preserve newlines in the text
-		for _, line := range doWordWrap(para, "", 79) { // But wrap text too
-			i.show.Println(line)
+func (i *Installer) ShowWrapped(texts ...string) {
+	for _, text := range texts {
```
ark3: Brilliant!

cmd/edgectl/aes_install.go
inercia: :+1:

cmd/edgectl/aes_install.go
inercia: Yeah, well, just copy&pasted. Fixing...

cmd/edgectl/aes_install.go
```diff
@@ -20,15 +20,23 @@ import (
...
 	k8sClientCoreV1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
+
+	"github.com/datawire/ambassador/pkg/helm"
+	"github.com/datawire/ambassador/pkg/k8s"
+	"github.com/datawire/ambassador/pkg/supervisor"
 )
 
+// clusterInfo describes some properties about the cluster where the installation is performed
+type clusterInfo struct {
```
alexgervais, line 35: I would move this struct definition towards the bottom of this file. It seems to be the style here as other structs are defined below.

cmd/edgectl/aes_install.go
```diff
@@ -437,166 +429,100 @@ func (i *Installer) Perform(kcontext string) error {
...
+		return fmt.Errorf(noCluster)
 	}
 
-	// Download AES manifests
-	crdManifests, err := getManifest(fmt.Sprintf("https://%s/yaml/aes-crds.yaml", manifestsDomain))
+	// Try to determine cluster type from node labels
+	ci := i.GetClusterInfo()
+	i.SetMetadatum("Cluster Info", "cluster_info", ci.name)
+
+	i.SetMetadatum("Cluster Info", "managed", "helm")
```
alexgervais, line 442: I think we only want to set the `managed` metadata if we run edgectl on top of an existing install, which is managed by `helm`. In the case of a first install, managed by edgectl, then this field should not be set.

cmd/edgectl/aes_install.go
```diff
@@ -437,166 +429,100 @@ func (i *Installer) Perform(kcontext string) error {
...
+
+	// create a new parsed checker for versions
+	chartVersion, err := helm.NewChartVersionRule("*")
 	if err != nil {
-		i.Report("fail_no_internet", ScoutMeta{"err", err.Error()})
-		return errors.Wrap(err, "download AES CRD manifests")
+		return err
```
alexgervais, line 447: We should also report an event to Metriton, using `i.Report()` whenever we handle a failure scenario. 

I saw a few places where this is not the case when handling Helm errors here and bellow.

cmd/edgectl/aes_install.go
```diff
@@ -437,166 +429,100 @@ func (i *Installer) Perform(kcontext string) error {
...
+		i.ShowWrapped(message, "\n")
+		i.Report("deploy")
 
-	// Grab Helm information if present
-	if managedDeployment, err := i.CaptureKubectl("get deployment labels", "", "get", "-n", "ambassador", "deployments", "ambassador", "-Lapp.kubernetes.io/managed-by"); err == nil {
-		if strings.Contains(managedDeployment, "Helm") {
-			i.SetMetadatum("Cluster Info", "managed", "helm")
-		}
+	} else {
+		// Third case: Ambassador is currently installed. Don't do anything.
```
alexgervais, line 510: What's the second case? 😂

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.13
 
 require (
```
alexgervais, line 5: Do all these dependencies need to be updated? Can we only add what's necessary for Helm?

cmd/edgectl/aes_install.go
```diff
@@ -437,166 +429,100 @@ func (i *Installer) Perform(kcontext string) error {
...
 	}
 
-	// Allow overriding the source domain (e.g., for smoke tests before release)
-	manifestsDomain := "www.getambassador.io"
-	domainOverrideVar := "AES_MANIFESTS_DOMAIN"
-	overrideMessage := "Downloading manifests from override domain %q instead of default %q because the environment variable %s is set."
-	if amd := os.Getenv(domainOverrideVar); amd != "" {
-		i.ShowWrapped(fmt.Sprintf(overrideMessage, amd, manifestsDomain, domainOverrideVar))
-		manifestsDomain = amd
+	namespace, err := i.kubeinfo.Namespace()
```
alexgervais, line 432: It would be safer to assume we are installing AES in the `ambassador` namespace (and create the namespace if necessary). 

Although I agree it would be great to allow customizing the namespace and allow installs anywhere, a lot of the installer code assumes we are in the `ambassador` namespace for now. I would feel more confident to stick with a hardcoded namespace for now and refactor in another PR.

inercia: Then shouldn't we set explictly the `ambassador` namespace in [`i.kubeinfo = k8s.NewKubeInfo(...)`](https://github.com/datawire/ambassador/blob/master/cmd/edgectl/aes_install.go#L379)?

cmd/edgectl/aes_install.go
inercia: :-) There was a second case for a while... Fixing

cmd/edgectl/aes_install.go
inercia: Ok, I think I have added `i.Report()`s in all the new failures...

cmd/edgectl/aes_install.go
inercia: Ok, I have defined

```go
	// defInstallNamespace is the default installation namespace
	defInstallNamespace = "ambassador"
```

and I will use `defInstallNamespace` instead of this `namespace`.

go.mod
inercia: You mean the `replace` block?

cmd/edgectl/aes_install.go
```diff
@@ -818,11 +741,17 @@ func (i *Installer) Quit() {
...
 
-func (i *Installer) ShowWrapped(text string) {
-	text = strings.Trim(text, "\n")                  // Drop leading and trailing newlines
-	for _, para := range strings.Split(text, "\n") { // Preserve newlines in the text
-		for _, line := range doWordWrap(para, "", 79) { // But wrap text too
-			i.show.Println(line)
+func (i *Installer) ShowWrapped(texts ...string) {
+	for _, text := range texts {
+		if text == "\n" {
+			i.show.Println()
```
ark3, line 747: You don't need this special case. Passing the empty string yields a blank line, and I find that to be clearer at the call site. Non-empty arguments don't need a trailing newline, so passing in a lone newline looks weird.

cmd/edgectl/aes_install.go
```diff
@@ -818,11 +741,17 @@ func (i *Installer) Quit() {
 	i.cancel()
 }
 
-func (i *Installer) ShowWrapped(text string) {
-	text = strings.Trim(text, "\n")                  // Drop leading and trailing newlines
-	for _, para := range strings.Split(text, "\n") { // Preserve newlines in the text
-		for _, line := range doWordWrap(para, "", 79) { // But wrap text too
-			i.show.Println(line)
+func (i *Installer) ShowWrapped(texts ...string) {
```
ark3, line 744:
```suggestion
// ShowWrapped displays to the user (via the show logger) the text items passed
// in with word wrapping applied. Leading and trailing newlines are dropped in
// each text item (to make it easier to use multiline constants), but newlines
// within each item are preserved. Use an empty string item to include a blank
// line in the output between other items.
func (i *Installer) ShowWrapped(texts ...string) {
```

cmd/edgectl/aes_install.go
```diff
@@ -437,166 +428,97 @@ func (i *Installer) Perform(kcontext string) error {
...
+	options := helm.HelmManagerOptions{
+		Version: chartVersion,
+		Logger:  i.log,
 	}
-	i.version = matches[1]
-	i.SetMetadatum("AES version being installed", "aes_version", i.version)
 
-	// Display version information
-	i.ShowWrapped(fmt.Sprintf("-> Installing the Ambassador Edge Stack %s.", i.version))
+	chartValues := map[string]string{} // TODO: set any Helm values we want to override,
```
ark3, line 450: Does the chart install three replicas? Can we change that to one replica (and would that be done here)? I suggest this for consistency with the old installer, but maybe that's not a good reason. Thoughts?

cmd/edgectl/aes_install.go
```diff
@@ -437,166 +428,97 @@ func (i *Installer) Perform(kcontext string) error {
...
+	if err := chartRel.Sync(i.ctx); err != nil {
+		i.log.Printf("Failed to sync release: %s", err)
+		i.Report("fail_download", ScoutMeta{"err", err.Error()})
+		return err
+	}
+
+	if !chartRel.IsInstalled() {
+		// Ambassador is not installed at all: perform the installation
+		i.SetMetadatum("AES version being installed", "aes_version", downVersion)
+		i.version = downVersion
```
ark3, line 486: I think we should perform these steps unconditionally, like the original code did. In this version, these won't be set if the chart is already installed.

cmd/edgectl/aes_install.go
```diff
@@ -437,166 +428,97 @@ func (i *Installer) Perform(kcontext string) error {
...
-		if strings.Contains(managedDeployment, "Helm") {
-			i.SetMetadatum("Cluster Info", "managed", "helm")
+		message := ""
+		if installedRelease.Info != nil {
+			message = installedRelease.Info.Notes
 		}
+		i.ShowWrapped(message, "\n")
+		i.Report("deploy")
+	} else {
+		// Ambassador is currently installed. Don't do anything.
```
ark3, line 506: What happens if the wrong version of Ambassador is installed? What if there is an incomplete installation? The old code handled those cases (poorly!), but this _appears to_ ignore them completely. I admit I don't know what the Helm library is doing under the covers, so maybe this is okay. Can you clarify?

cmd/edgectl/aes_install.go
```diff
@@ -437,166 +428,97 @@ func (i *Installer) Perform(kcontext string) error {
...
+		i.ShowWrapped(fmt.Sprintf("-> Found an existing installation of Ambassador Edge Stack %s.", downVersion))
+		i.ShowWrapped("-> ... will not try to install/update Ambassador")
+		i.Report("installed")
```
ark3, line 509: This is a new report. The old code reported either `deploy` or a failure. We should be careful about changing that. Whatever we do, `metq` will have to support the old and new reports for a long time.

cmd/edgectl/aes_install.go
ark3: My comment here renders weirdly in GitHub, so let me clarify. All I'm trying to do is suggest a doc comment for `ShowWrapped`. I wrote this comment in my branch, but I think it makes more sense in yours, since you added the variadic arguments (brilliant!).

cmd/edgectl/aes_install.go
inercia: According to the [defaults in the Chart](https://github.com/datawire/ambassador-chart/#configuration), replicas is 3. We could add a map with the default chart values, and use `replicas=1`.

richarddli: Let's keep at 3, please.

ark3: Okay, as long as we're aware that
- the old installer installed one replica because...
- the manual install YAML specifies one replica

cmd/edgectl/aes_install.go
alexgervais: I can see the confusion between `deploy` and `installed`. Maybe it would make it clearer to rename this one to `already_installed` or something that will better represent this new code path.

cmd/edgectl/aes_install.go
richarddli: I didn't realize we changed it. Then let's do 1.

cmd/edgectl/aes_install.go
ark3: This isn't really a new code path from the metrics perspective. If we want to distinguish between "installed this run" and "installed prior to this run" I suggest we use a metadata item. In other words, I say we always report `deploy` if a deployment exists (like the old installer) and add `already_installed: true` or `already_installed: false` metadata as is appropriate.

cmd/edgectl/aes_install.go
ark3: Yup, I'm not sure why the OSS YAML and the Helm chart say 3, but the AES YAML says 1. But that's the way it has been since at least 1.0.

cmd/edgectl/aes_install.go
inercia: Ok, I'll fix this.

cmd/edgectl/aes_install.go
inercia: If you mean to report the version of the Chart that is already installed, the API we are using does not provide that information. It provides the version for the Helm Chart that is currently installed though, but I'm not sure that would be useful...

cmd/edgectl/aes_install.go
inercia: What do you mean with a _wrong version of Ambassador_?

Regarding _incomplete installations_, this uninstalls everything if something goes wrong.

cmd/edgectl/aes_install.go
inercia: Ok, I can add that kind of metrics.

I think it would be convenient to document all these magic values for metrics. Something like:

```go
const (
    // reportDeployed is sent when the deployment exists
    reportDeployed = "deploy"

    // reportAlreadyInstalled is sent when a deployment was already installed in the system
    reportAlreadyInstalled = "already_installed"
)
```

cmd/edgectl/aes_install.go
inercia: Ok, I will add this comment.

cmd/edgectl/aes_install.go
inercia: I have modified `GetInstalledImageVersion` for getting the current version of Ambassador, even if it was installed with Helm.

cmd/edgectl/aes_install.go
```diff
@@ -527,149 +548,130 @@ func (i *Installer) Perform(kcontext string) error {
...
+			i.ShowWrapped(abortExisting, seeDocs)
+			i.Report("fail_existing_aes", ScoutMeta{"installing", i.version}, ScoutMeta{"found", installedVersion})
+			return errors.Errorf("existing AES %s found when installing AES %s", installedVersion, i.version)
+		default:
+			// continue with the rest of the setup
 		}
+	} else {
+		// Ambassador is definetively not installed: perform the installation
+		i.SetMetadatum("AES version being installed", "aes_version", i.version)
+		i.SetMetadatum("Cluster Info", "managed", "helm")
```
alexgervais, line 630: If we don't have a previous installation, the `managed` metadata should not be set.

cmd/edgectl/aes_install.go
```diff
@@ -535,124 +563,122 @@ func (i *Installer) Perform(kcontext string) Result {
...
+		alreadyInstalled = true
+	} else {
 		if err != nil {
-			i.show.Println("Failed to look for an existing installation:", err)
-			installedVersion = ""
-			// Things will likely fail when we try to apply manifests
+			i.log.Printf("when obtaining the chart manager: %s", err)
+			i.Report("fail_download", ScoutMeta{"err", err.Error()})
+			return UnhandledErrResult(err)
 		}
```
ark3, line 616: Shouldn't we do this higher up, where we set `installedVersion = ""`?

cmd/edgectl/aes_install.go
```diff
@@ -718,7 +742,6 @@ func (i *Installer) Perform(kcontext string) Result {
 		i.ShowWrapped(seeDocs)
 		return UnhandledErrResult(err)
 	}
-	i.Report("aes_listening")
```
ark3, line 745: Why delete this report?

cmd/edgectl/aes_install.go
inercia: Damn, this is a merge error. Fixing.

cmd/edgectl/aes_install.go
inercia: Good catch. Fixing.

cmd/edgectl/aes_install.go
```diff
@@ -694,6 +747,8 @@ func (i *Installer) Perform(kcontext string) Result {
 		return UnhandledErrResult(nil)
 	}
 
+	i.Report("aes_listening")
+
 	// Grab load balancer address
```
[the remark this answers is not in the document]
inercia: Damn. Fixing.

khussey:
@inercia it seems to me that we should be adding new "help" pages in conjunction with these changes, to cover the new Helm-specific failure cases. These cases should have new documentation URLs assigned to them (rather than repurposing existing ones), so that users who have downloaded the existing non-Helm installer can still find help for the obsolete failure cases.

We'll also need to think about the best way to deprecate the obsolete help pages that no longer apply (perhaps by adding a message at the top of them indicating that a newer version of the installer is available and that they should download it for an improved experience)...

cmd/edgectl/aes_install_errors.go
```diff
@@ -93,103 +100,30 @@ func (i *Installer) GetVersionsError(err error) Result {
 	}
 }
 
-// Unable to fetch the AES CRD manifests (aes-crds.yaml)
-func (i *Installer) AESCRDManifestsError(err error) Result {
+// Unable to download the latest Chart
+func (i *Installer) DownloadError(err error) Result {
 	i.Report("fail_no_internet", ScoutMeta{"err", err.Error()})
 
 	url := "https://www.getambassador.io/docs/latest/topics/install/help/aes-crd-manifests"
```
khussey, line 107: This needs to be a new help page (with a new URL).

cmd/edgectl/aes_install_errors.go
```diff
@@ -93,103 +100,30 @@ func (i *Installer) GetVersionsError(err error) Result {
...
-
-	i.log.Printf("matches is %+v", matches)
-
-	return Result{
-		Report:       "fail_bad_manifests",
-		ShortMessage: "The installer failed to parse AES manifests",
-		Message:      fmt.Sprintf("Visit %v for a more detailed explanation and step-by-step instructions about downloading AES manifests to continue installing Ambassador Edge Stack.", url),
-		URL:          url,
-		Err:          err,
+		Err:          errors.Wrap(err, "Failed to download AES CRD Chart"),
```
khussey, line 113: "AES CRD Chart" -> "AES Chart" (for consistency)?

cmd/edgectl/aes_install_errors.go
```diff
@@ -93,103 +100,30 @@ func (i *Installer) GetVersionsError(err error) Result {
...
-		ShortMessage: fmt.Sprintf("The installer found incompatible AES CRD versions: Existing AES %v found when installing AES %v", installedVersion, i.version),
-		Message:      message,
-		Err:          errors.Errorf("Ambassador Edge Stack %s already installed", installedVersion),
-	}
-}
-
-// Existing AES CRD's, unable to upgrade.
-func (i *Installer) ExistingCRDsError() Result {
+func (i *Installer) ExistingInstallationFoundError(installedVersion string) Result {
 	url := "https://www.getambassador.io/docs/latest/topics/install/help/existing-crds"
```
khussey, line 119: This needs to be a new help page (with a new URL).

cmd/edgectl/aes_install_errors.go
```diff
@@ -199,7 +133,7 @@ func (i *Installer) InstallAESError(err error) Result {
 
 	return Result{
 		Report:       "fail_install_aes",
-		ShortMessage: "An error occurred while applying Kubernetes manifests",
+		ShortMessage: "An error occurred while applying AES Chart",
 		Message:      fmt.Sprintf("Visit %v for a more detailed explanation and suggestions on how to continue installing Ambassador Edge Stack.", url),
 		URL:          url,
```
khussey, line 138: This needs to be a new help page (with a new URL).

cmd/edgectl/aes_install_errors.go
```diff
@@ -212,7 +146,7 @@ func (i *Installer) WaitForAESError(err error) Result {
 
 	return Result{
 		Report:       "fail_wait_aes",
-		ShortMessage: "An error occurred while applying Kubernetes manifests",
+		ShortMessage: "An error occurred while applying AES Chart",
 		Message:      fmt.Sprintf("Visit %v for a more detailed explanation and suggestions on how to continue installing Ambassador Edge Stack.", url),
 		URL:          url,
```
khussey, line 151: This needs to be a new help page (with a new URL).

docs/topics/install/help/index.md
```diff
@@ -5,21 +5,16 @@ We recommend edgectl for [installing Ambassador Edge Stack](../../../tutorials/g
 - [Could not find a Kubernetes cluster](no-cluster)
 - [Email request interrupted](email-request)
 - [Failed to create a Host resource in your cluster](host-resource-creation)
-- [Failed to download AES manifests](aes-manifests)
```
khussey, line 8: No, we can't remove or otherwise repurpose existing pages because they are already out there "in the wild". Rather, we need new pages for the (smaller number of) ways the Helm install can fail in these parts of the installation.

cmd/edgectl/aes_install_errors.go
khussey: Or perhaps event better "AES Chart" -> "Helm chart" (in both places here).

docs/topics/install/help/index.md
inercia: Ok, fixed.

cmd/edgectl/aes_install_errors.go
inercia: Ok, fixed.

cmd/edgectl/aes_install_errors.go
inercia: Ok, fixed

cmd/edgectl/aes_install_errors.go
inercia: I think we can reuse the error message by just saying `"Failed to download AES CRD"`

cmd/edgectl/aes_install_errors.go
inercia: Done

cmd/edgectl/aes_install_errors.go
inercia: Resolved by just saying `AES` instead of AES manifests or AES Charts.

PR #2493, opened 2020-03-26
Installer: Support multiple AES pods

ark3:
Support multiple AES pods with arbitrary labels. This should help address the Helm vs direct install incompatibility and generally make the installer more robust.

This changeset also starts collecting installer results in the Installer object rather than in local variables in the gigantic `Perform()` method. This helps to simplify some code already, but more importantly makes it practical to break up that giant function in the future. Also it will be a little easier to emit an installation summary at the end.

### DIFF
--- a/cmd/edgectl/aes_install.go
+++ b/cmd/edgectl/aes_install.go
@@ -200,14 +200,48 @@ func (i *Installer) loopUntil(what string, how func() error, lc *loopConfig) err
 // the Pod is Running (though not necessarily Ready). This should be good enough
 // to report the "deploy" status to metrics.
 func (i *Installer) GrabAESInstallID() error {
-	// Note: This is brittle. If there is more than one pod, this returns all
-	// the pod names squashed together. This should not occur in normal usage,
-	// i.e. when installing a single version of AES, maybe repeatedly.
-	pod, err := i.CaptureKubectl("get AES pod", "", "-n", "ambassador", "get", "pods", "-l", "service=ambassador", "-o", "go-template={{range .items}}{{.metadata.name}}{{end}}")
+	aesImage := "quay.io/datawire/aes:1.3.1" // FIXME
+	podName := ""
+	containerName := ""
+	podInterface := i.coreClient.Pods("ambassador") // namespace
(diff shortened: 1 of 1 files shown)


cmd/edgectl/aes_install.go
```diff
@@ -200,13 +200,24 @@ func (i *Installer) loopUntil(what string, how func() error, lc *loopConfig) err
...
+	aesImage := "quay.io/datawire/aes:" + i.version
+	pod := ""
+	podInterface := i.coreClient.Pods("ambassador") // namespace
+	pods, err := podInterface.List(k8sTypesMetaV1.ListOptions{})
 	if err != nil {
 		return err
 	}
+	for _, k8sPod := range pods.Items {
+		containers := k8sPod.Spec.Containers
+		if len(containers) != 1 || containers[0].Image != aesImage || containers[0].Command != nil {
```
alexgervais: I wonder if we could have a looser conditional here? I'd go with a simple `if containers[0].Image != aesImage` {}

In case we eventually package AES with a sidecar container or change the docker entrypoint to remove the command, it would introduce a breaking change in the installer.

ark3: Here are the missing comments. :slightly_frowning_face: 
- `len(containers) != 1` is to skip the Traffic Agent sidecar
- `containers[0].Command != nil` is to skip the Traffic Manager

ark3: Added.

alexgervais: Maybe iterating over the containers would be a better option then. Otherwise, if we run a single pod with the traffic agent, we'll be skipping over it and never will know the install_id!

ark3: Hmm? I'm confused. Can you explain what case we're messing up?

ark3: We synced elsewhere. I will update this PR to check for the Traffic Agent in a different way, rather than relying on `len(containers) == 1`, as that will break as soon as new Istio stuff lands, or if anything else injects a container alongside Ambassador.

PR #2495, opened 2020-03-27
Fast forward 1.3.2 to where 1.3 is...

ark3:
Because I foolishly landed commits on 1.3 that were not related to docs.


kflynn:
Merging before CI, since I know that this is literally a fast-forward.

PR #2499, opened 2020-03-29
Host crd: Fix insecure Route action

jfrabaute:
## Description
Host crd: Fix insecure Route action

For each listener, there will be one envoy filter per server_names
and also an envoy filter with empty filter_chain.

For http, the envoy filter_chain by server_names has no effect.
This is only for SNI and https.
So for the http listener, only the first envoy filter is used.

So, when the first vhost is redirecting, all the http traffic will be redirected.
When the first vhost is routing http traffic, all the http traffic will be routed, no redirection.

The fix is checking per route, within the vhost loop,
if there is a corresponding "Host" crd for the route having the hostname.
If so, it applies the action based on the Host CRD and not based on the vhost currently processed.

## Related Issues
https://github.com/datawire/ambassador/issues/2216

## Testing
manual tests

### DIFF
--- a/python/ambassador/envoy/v2/v2listener.py
+++ b/python/ambassador/envoy/v2/v2listener.py
@@ -1038,6 +1038,20 @@ class V2Listener(dict):
                                     action=None,
                                     insecure_action='Reject')
 
+        # Preprocess the list of "HOST" CRDs we have, asking for specific insecure action
+        hostsWithInsecureRoute = {}
+        for hname, hostconfig in config.ir.hosts.items():
+            insecure_config = hostconfig.get('requestPolicy', {}).get('insecure', {})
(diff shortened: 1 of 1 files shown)


jfrabaute:
Any chance to have this reviewed?

jfrabaute:
Not stale.

jfrabaute:
Not stale.

jfrabaute:
Not stale

kflynn:
After a number of issues in the area of how the various actions of various `Host`s interact with each other, we're in the middle of digging deep into this and cleaning it all up. I'm going to close this one since it's doomed to millions of confilcts, but @jfrabaute, many thanks for the effort, and we've definitely studied this PR and learned from it -- be looking for a shout out when we land the cleanup. 🙂

PR #2496, opened 2020-03-27
edgectl reports install ids when registering a domain name

alexgervais:
## Description
edgectl reports install ids when registering a domain name.

## Related Issues
https://github.com/datawire/apro/issues/1156

## Testing
Manual tests against the beta channel.

## Tasks That Must Be Done
- [ ] Did you update CHANGELOG.md?
- [ ] Did you add or update tests?
- [ ] Did you update documentation?
- [ ] Were there any special dev tricks you had to use to work on this code efficiently?
  + [ ] Did you add them to DEVELOPING.md?

PR #2505, opened 2020-03-30
Bruce/celebrate with user 1

brucehorn:
Some improvements in the output during edgectl install, working on the Edgectl Install Improvements story, "Celebrate with the User".  I made a first pass on the following (from the story):

-_change the edgectl install success message printed to the terminal to a happier, more celebratory success message: bold? Capital letters? Distinguished spacing? This seems silly, but it turns out that a little design goes a long way in helping people like the product enough to put more time into evaluating it._. I colorized and restyled the text a bit.

- _add to the terminal message, some information about how to log back into the UI in the future._  Text was added for this.

-_change the edgectl --help message about login to be more clear that this is how one opens the UI._ Very minor change in the Cobra command definition.

I would appreciate some feedback on the language here.

celebrate2

### DIFF
--- a/cmd/edgectl/aes_install.go
+++ b/cmd/edgectl/aes_install.go
@@ -83,6 +83,16 @@ func aesInstall(cmd *cobra.Command, args []string) error {
 	kcontext, _ := cmd.Flags().GetString("context")
 	i := NewInstaller(verbose)
 
+	// If Scout is disabled (environment variable set to non-null), inform the user.
+	// TODO: should be i.scout.Disabled() when Alex's PR #2496 is merged.
+	if os.Getenv("SCOUT_DISABLE") != "" {
+		i.show.Printf("[aesInstall] Scout is disabled; metrics will not be written to Metriton.")
+	}
+
+	// Both printed and logged when verbose.
+	i.log.Printf("[aesInstall] the install_id is: %s", i.scout.installID)
+	i.log.Printf("[aesInstall] the trace_id is:   %s", i.scout.metadata["trace_id"])
+
 	sup := supervisor.WithContext(i.ctx)
 	sup.Logger = i.log
 
@@ -415,6 +425,7 @@ func (i *Installer) Perform(kcontext string) error {
 
 	// Attempt to use kubectl
 	_, err := i.GetKubectlPath()
+	//err = errors.New("early error for testing")  // TODO: remove for production
 	if err != nil {
 		i.Report("fail_no_kubectl")
 		return fmt.Errorf(noKubectl)
@@ -1000,7 +1011,6 @@ const noTlsSuccess = "Congratulations! You've successfully installed the Ambassa
 
 const noKubectl = `
 The installer depends on the 'kubectl' executable. Make sure you have the latest release downloaded in your PATH, and that you have executable permissions.
-
 Visit https://kubernetes.io/docs/tasks/tools/install-kubectl/ for more information and instructions.`
 
 const noCluster = `
--- a/cmd/edgectl/aes_install.go
+++ b/cmd/edgectl/aes_install.go
@@ -761,7 +761,7 @@ func (i *Installer) Perform(kcontext string) error {
 	i.show.Println()
 	i.ShowWrapped(color.Info.Sprintf(fullSuccess,
 		chalk.Bold.TextStyle(i.hostname),
-		chalk.Bold.TextStyle("edgectl login --namespace=ambassador "+i.hostname)))
+		chalk.Bold.TextStyle("edgectl login "+i.hostname)))
 
 	i.show.Println()
 
@@ -1030,7 +1030,7 @@ const fullSuccess = `Congratulations! You've successfully installed the Ambassad
 
 In the future, to log back in to the Ambassador Edge Policy Console, run
 $ %s
-from the command line.` // hostname, "edgectl login --namespace=ambassador <hostname>"
+from the command line.` // hostname, "edgectl login <hostname>"
 
 const noTlsSuccess = "Congratulations! You've successfully installed the Ambassador Edge Stack in your Kubernetes cluster. However, we cannot connect to your cluster from the Internet, so we could not configure TLS automatically."
 
--- a/cmd/edgectl/aes_install.go
+++ b/cmd/edgectl/aes_install.go
@@ -757,7 +757,8 @@ func (i *Installer) Perform(kcontext string) error {
 	}
 
 	i.show.Println("AES Installation Complete!")
-	i.show.Println("================================\n")
+	i.show.Println("================================")
+	i.show.Println()
 
 	i.ShowWrapped(fmt.Sprintf(fullSuccess,
 		color.Bold.Sprintf(i.hostname),
--- a/cmd/edgectl/aes_install.go
+++ b/cmd/edgectl/aes_install.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/pkg/browser"
 	"io"
 	"io/ioutil"
 	"log"
@@ -76,7 +77,7 @@ func getEmailAddress(defaultEmail string, log *log.Logger) string {
 			return text
 		}
 
-		fmt.Printf("Sorry, %q does not match our email address filter.\n", text)
+		fmt.Printf("Sorry, %q does not appear to be a valid email address.  Please check it and try again.\n", text)
 	}
 }
 
@@ -88,12 +89,11 @@ func aesInstall(cmd *cobra.Command, args []string) error {
 	// If Scout is disabled (environment variable set to non-null), inform the user.
 	// TODO: should be i.scout.Disabled() when Alex's PR #2496 is merged.
 	if os.Getenv("SCOUT_DISABLE") != "" {
-		i.show.Printf(color.Notice.Sprintf("[aesInstall] Scout is disabled; metrics will not be written to Metriton."))
+		i.show.Printf(color.Info.Sprintf(phoneHomeDisabled))
 	}
 
-	// Both printed and logged when verbose.
-	i.log.Printf("[aesInstall] the install_id is: %s", i.scout.installID)
-	i.log.Printf("[aesInstall] the trace_id is:   %s", i.scout.metadata["trace_id"])
+	// Both printed and logged when verbose (Installer.log is responsible for --verbose)
+	i.log.Printf(color.Info.Sprintf(installAndTraceIDs, i.scout.installID, i.scout.metadata["trace_id"]))
 
 	sup := supervisor.WithContext(i.ctx)
 	sup.Logger = i.log
@@ -432,6 +432,7 @@ func (i *Installer) Perform(kcontext string) error {
 	//err = errors.New("early error for testing")  // TODO: remove for production
 	if err != nil {
 		i.Report("fail_no_kubectl")
+		err = browser.OpenURL(noKubectlURL)
 		return fmt.Errorf(noKubectl)
 	}
 
@@ -439,6 +440,7 @@ func (i *Installer) Perform(kcontext string) error {
 	i.kubeinfo = k8s.NewKubeInfo("", kcontext, "")
 	if err := i.ShowKubectl("cluster-info", "", "cluster-info"); err != nil {
 		i.Report("fail_no_cluster")
+		err = browser.OpenURL(noClusterURL)
 		return fmt.Errorf(noCluster)
 	}
 	i.restConfig, err = i.kubeinfo.GetRestConfig()
@@ -996,7 +998,7 @@ const tryAgain = "If this appears to be a transient failure, please try running 
 const abortExisting = `
 This tool does not support upgrades/downgrades at this time.
 
-Aborting the installer to avoid corrupting an existing installation of AES.
+The installer will now quit to avoid corrupting an existing installation of AES.
 `
 
 const abortCRDs = `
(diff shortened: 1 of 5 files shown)


brucehorn:
I noticed that I capitalized "To access" which is actually mid-sentence.  Will fix that and am looking forward to other feedback as well.

docs/docs/latest/tutorials/getting-started.md
```diff
@@ -25,15 +25,15 @@ Windows. For other options, such as Docker, read the [detailed instructions] (/d
...
-    sudo curl -fL https://metriton.datawire.io/downloads/darwin/edgectl -o /usr/local/bin/edgectl && sudo chmod a+X /usr/local/bin/edgectl
+    sudo curl -fL https://metriton.datawire.io/downloads/darwin/edgectl -o /usr/local/bin/edgectl && sudo chmod a+x /usr/local/bin/edgectl
     ```
 
     If you decide to download the file, you may encounter a security block. To change this:
     * Go to **System Preferences > Security & Privacy > General**.
     * Click the **Open Anyway** button.
     * On the new dialog, click the **Open** button.
 
-2. Run the installer with `./edgectl install`
```
bnfb: Why did you remove the ./ ? Many people's PATH does not include . and thus one has to use ./edgectl rather than just edgectl in order to execute it.

docs/docs/latest/tutorials/getting-started.md
```diff
@@ -63,9 +63,9 @@ Windows. For other options, such as Docker, read the [detailed instructions] (/d
 
 </summary>
 
-1. Download the `edgectl` file
+1. Download the `edgectl.exe` file
    [here](https://metriton.datawire.io/downloads/windows/edgectl.exe).
-2. Run the installer with `edgectl.exe install`
```
bnfb: This seems reasonable but I forget how Windows works: can you run a program without the exe on the end?

cmd/edgectl/aes_install.go
```diff
@@ -80,6 +85,16 @@ func aesInstall(cmd *cobra.Command, args []string) error {
 	kcontext, _ := cmd.Flags().GetString("context")
 	i := NewInstaller(verbose)
 
+	// If Scout is disabled (environment variable set to non-null), inform the user.
+	// TODO: should be i.scout.Disabled() when Alex's PR #2496 is merged.
+	if os.Getenv("SCOUT_DISABLE") != "" {
+		i.show.Printf(color.Notice.Sprintf("[aesInstall] Scout is disabled; metrics will not be written to Metriton."))
```
bnfb: This message is not going to be clear to the user (in the odd case that they use it) and thus it will cause them to be confused. Something like: "[aesInstall] Info: phone-home metrics have been disabled by environment variable." Why is this better? (a) it says "Info" so that users know that they can ignore it; (b) it explains a little bit about what is going on in language the user probably understands rather than Scout and Metriton".

cmd/edgectl/aes_install.go
```diff
@@ -80,6 +85,16 @@ func aesInstall(cmd *cobra.Command, args []string) error {
...
+	}
+
+	// Both printed and logged when verbose.
+	i.log.Printf("[aesInstall] the install_id is: %s", i.scout.installID)
```
bnfb: Same comment here: the language should be user-centric. Something like: "[aesInstall] Info: unique id for this install: %s, %s". And it should only print this if Scout is enabled. And it should be just one line (one line is less "it must be important" to the user than two lines).

cmd/edgectl/aes_install.go
brucehorn: thanks, that's helpful.  I was under the impression this was a "datawire only" message so it is clear that this is better.

cmd/edgectl/aes_install.go
brucehorn: Sounds good.

docs/docs/latest/tutorials/getting-started.md
brucehorn: this is some of the code that I somehow inherited--I didn't make those changes, so wanted to figure out how to separate out my changes from the others.

ark3:
This PR shows lots of unrelated commits because the "base" branch is master. Please edit the PR in the GitHub UI to set the base branch to release/v1.3.2.

cmd/edgectl/aes_install.go
```diff
@@ -612,7 +629,7 @@ func (i *Installer) Perform(kcontext string) error {
 		return err
 	}
 	i.Report("cluster_accessible")
-	i.show.Println("Your AES installation's address is", i.address)
+	i.show.Println("Your AES installation's address is", chalk.Bold.TextStyle(i.address))
```
ark3: Can we drop `chalk` and use `gookit/color`  for everything? These two dependencies overlap in functionality, and I prefer `gookit/color` because it supports Windows.

cmd/edgectl/aes_install.go
```diff
@@ -987,24 +1009,35 @@ Removing the CRDs will cause your existing Ambassador Mappings and other resourc
...
-Aborting the installer to avoid corrupting an existing (but undetected) installation.
+The installer will now quit to avoid corrupting an existing (but undetected) installation.
 `
 
-const seeDocs = "See https://www.getambassador.io/docs/latest/tutorials/getting-started/"
+const seeDocs    = "See https://www.getambassador.io/docs/latest/tutorials/getting-started/"
+const seeDocsURL = "https://www.getambassador.io/docs/latest/tutorials/getting-started/"
+
+const phoneHomeDisabled  = "[aesInstall] Info: phone-home is disabled by environment variable"
+const installAndTraceIDs = "[aesInstall] Info: install_id = %s; trace_id = %s"
```
ark3: Why the `[aesInstall]` prefix here? We don't use this anywhere else.

cmd/edgectl/main.go
```diff
@@ -183,7 +183,7 @@ func getRootCommand() *cobra.Command {
 	}
 	loginCmd := &cobra.Command{
 		Use:   "login [flags] HOSTNAME",
-		Short: "Access the Ambassador Edge Policy Console",
+		Short: "Log in and open the Ambassador Edge Policy Console web application",
```
ark3: The [marketing-approved documentation](https://www.getambassador.io/about/edge-policy-console/) says 

> After you install the Ambassador Edge Stack, you can log in to the Edge Policy Console (EPC) to manage your deployment.

Drawing on that phrasing, I suggest "Log in to the Ambassador Edge Policy Console"

cmd/edgectl/aes_install.go
```diff
@@ -83,6 +86,15 @@ func aesInstall(cmd *cobra.Command, args []string) error {
 	kcontext, _ := cmd.Flags().GetString("context")
 	i := NewInstaller(verbose)
 
+	// If Scout is disabled (environment variable set to non-null), inform the user.
+	// TODO: should be i.scout.Disabled() when Alex's PR #2496 is merged.
+	if os.Getenv("SCOUT_DISABLE") != "" {
+		i.show.Printf(color.Info.Sprintf(phoneHomeDisabled))
+	}
```
ark3: Now that #2496 has landed on the release branch, you can update this.

cmd/edgectl/aes_install.go
```diff
@@ -735,8 +752,13 @@ func (i *Installer) Perform(kcontext string) error {
...
 	}
 
+	// Add a little color
+
 	i.show.Println()
-	i.ShowWrapped(fmt.Sprintf(fullSuccess, i.hostname))
+	i.ShowWrapped(color.Info.Sprintf(fullSuccess,
+		chalk.Bold.TextStyle(i.hostname),
+		chalk.Bold.TextStyle("edgectl install "+i.hostname)))
+
```
ark3: I think this should be `login` not `install`.

ark3: Oh, and the namespace flag... You have it correct in the comment below.

cmd/edgectl/main.go
ark3: In particular, I don't think we should use the phrase "web application" to describe the console. Yes, I know that does describe it correctly, but our documentation consistently uses that phrase to describe a service that a user writes and then exposes using Ambassador.

docs/docs/latest/tutorials/getting-started.md
brucehorn: this is no longer the case now that the PR is correctly branched off of release/v1.3.2

docs/docs/latest/tutorials/getting-started.md
brucehorn: this is no longer code in this PR now that I've branched off of release/v1.3.2

cmd/edgectl/aes_install.go
brucehorn: Fixed in new push.

cmd/edgectl/aes_install.go
brucehorn: I use both because chalk allows different functionality (e.g. changing a word within a sentence to bold without terminating the style change at that point) and they interoperate well.  I will spend some more time to see if I can rewrite the code to properly work with just one or the other.

cmd/edgectl/aes_install.go
brucehorn: it was meant to be a parallel construct to the installer.SetMetadatum and Report,  which say:
"[Metrics]..."
but I can remove it of course.

cmd/edgectl/aes_install.go
```diff
@@ -987,24 +1009,35 @@ Removing the CRDs will cause your existing Ambassador Mappings and other resourc
 
 $ kubectl delete crd -l product=aes
 
-Aborting the installer to avoid corrupting an existing (but undetected) installation.
+The installer will now quit to avoid corrupting an existing (but undetected) installation.
 `
 
-const seeDocs = "See https://www.getambassador.io/docs/latest/tutorials/getting-started/"
+const seeDocs    = "See https://www.getambassador.io/docs/latest/tutorials/getting-started/"
```
khussey: Should seeDocs somehow reference or build on seeDocsURL (so the URL string doesn't exist in two places)?

cmd/edgectl/main.go
brucehorn: The point of the word "open" is that it describes the fact that it will open in the web browser; logging in is just the beginning of the process.

Removing the "web application" text sounds right given the need to stay consistent with the documentation.

cmd/edgectl/aes_install.go
brucehorn: will do

cmd/edgectl/aes_install.go
brucehorn: I was thinking about that too.  I'll see if Go will let me construct a const from string fragments, otherwise I'll have to pass it in as a parameter.

cmd/edgectl/aes_install.go
brucehorn: done

cmd/edgectl/aes_install.go
ark3: Windows support is a requirement. If `chalk` doesn't _break_ Windows (and just leaves things unstyled in that case), then I suppose that's okay. Also, it looks like `chalk` might be unmaintained.

cmd/edgectl/aes_install.go
brucehorn: I wasn't sure since reading elsewhere in the code says that `--namespace ambassador` is the default.  Do we want to always have our users specify the namespace?

cmd/edgectl/aes_install.go
ark3: The `[Metrics]` prefix et al are not seen by the user (except in verbose mode, which is really just "log to the screen" mode -- it could use a lot of refinement).

cmd/edgectl/aes_install.go
ark3: Ah yes, for `edgectl login` the default for the namespace argument works fine. You are right.

cmd/edgectl/aes_install.go
ark3: That feature exists in the code I'm working on. I think we can get Bruce's PR into 1.3.2, whereas my work will take longer. Let's leave in the duplication for now and improve things in the next release.

brucehorn:
Latest install looks like this:

Screen Shot 2020-03-30 at 12 28 51 PM

docs/docs/latest/tutorials/getting-started.md
khussey: @bnfb the commands to download and adjust permissions for the binary actually place it in /usr/local/bin so in fact attempting to run the command with a prefix of "./" will fail.

docs/docs/latest/tutorials/getting-started.md
khussey: @bnfb you can indeed run a program on Windows without specifying the .exe extension.

cmd/edgectl/main.go
khussey: @brucehorn did you decide to abandon the "and open" as well (which I thought was an improvement)?

brucehorn: I thought it was an improvement as well but deferred to @ark3

cmd/edgectl/aes_install.go
brucehorn: fixed already in this PR

brucehorn:
Still waiting on a pending reviewer before merging...

cmd/edgectl/aes_install.go
brucehorn: latest push resolves this--`chalk` is no longer a requirement.

cmd/edgectl/aes_install.go
brucehorn: resolved, fixed `login` and removed the namespace text.

cmd/edgectl/aes_install.go
brucehorn: Removed in this PR.

PR #2508, opened 2020-03-31
edgectl sends crash reports and logs to metriton

alexgervais:
## Description
- `--no-report` cmd flag to skip uploading report and logs on failure.
- Call metriton to get a crash-report id and url for upload.
- Provides edgectl metadata, including versions and install ids.
- Provides edgectl logs, ambassador logs and ambassador resource description outputs.

## Related Issues
https://github.com/datawire/apro/issues/1162

## Testing
Manual tests

## Tasks That Must Be Done

## Other

### DIFF
--- a/cmd/edgectl/aes_install.go
+++ b/cmd/edgectl/aes_install.go
@@ -256,7 +256,7 @@ PodsLoop:
 	}
 
 	// Retrieve the cluster ID
-	clusterID, err := i.CaptureKubectl("get cluster ID", true, "", "-n", "ambassador", "exec", podName, "-c", containerName, "python3", "kubewatch.py")
+	clusterID, err := i.CaptureKubectl("get cluster ID", "", "-n", "ambassador", "exec", podName, "-c", containerName, "python3", "kubewatch.py")
 	if err != nil {
 		return err
 	}
@@ -350,12 +350,12 @@ func (i *Installer) CheckHostnameFound() error {
 
 // CheckACMEIsDone queries the Host object and succeeds if its state is Ready.
 func (i *Installer) CheckACMEIsDone() error {
-	state, err := i.CaptureKubectl("get Host state", true, "", "get", "host", i.hostname, "-o", "go-template={{.status.state}}")
+	state, err := i.CaptureKubectl("get Host state", "", "get", "host", i.hostname, "-o", "go-template={{.status.state}}")
 	if err != nil {
 		return LoopFailedError(err.Error())
 	}
 	if state == "Error" {
-		reason, err := i.CaptureKubectl("get Host error", true, "", "get", "host", i.hostname, "-o", "go-template={{.status.errorReason}}")
+		reason, err := i.CaptureKubectl("get Host error", "", "get", "host", i.hostname, "-o", "go-template={{.status.errorReason}}")
 		if err != nil {
 			return LoopFailedError(err.Error())
 		}
@@ -383,7 +383,7 @@ func (i *Installer) CheckACMEIsDone() error {
 // GetAESCRDs returns the names of the AES CRDs available in the cluster (and
 // logs them as a side effect)
 func (i *Installer) GetAESCRDs() ([]string, error) {
-	crds, err := i.CaptureKubectl("get AES crds", true, "", "get", "crds", "-lproduct=aes", "-o", "name")
+	crds, err := i.CaptureKubectl("get AES crds", "", "get", "crds", "-lproduct=aes", "-o", "name")
 	if err != nil {
 		return nil, err
 	}
@@ -410,7 +410,7 @@ func (i *Installer) GetAESCRDs() ([]string, error) {
 // TODO: Also look for Ambassador OSS and do something intelligent.
 func (i *Installer) GetInstalledImageVersion() (string, error) {
 	aesVersionRE := regexp.MustCompile("quay[.]io/datawire/aes:([[:^space:]]+)")
-	deploys, err := i.CaptureKubectl("get AES deployment", true, "", "-nambassador", "get", "deploy", "-lproduct=aes", "-o", "go-template={{range .items}}{{range .spec.template.spec.containers}}{{.image}}\n{{end}}{{end}}")
+	deploys, err := i.CaptureKubectl("get AES deployment", "", "-nambassador", "get", "deploy", "-lproduct=aes", "-o", "go-template={{range .items}}{{range .spec.template.spec.containers}}{{.image}}\n{{end}}{{end}}")
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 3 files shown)


cmd/edgectl/aes_install.go
```diff
@@ -940,6 +947,69 @@ func (i *Installer) Report(eventName string, meta ...ScoutMeta) {
...
+	resp, err := http.Post(reportURL, "application/json", buf)
+	if err != nil {
+		i.log.Printf("failed to initiate anonymous crash report due to error: %v", err.Error())
+		return
+	}
+	defer resp.Body.Close()
+	content, err := ioutil.ReadAll(resp.Body)
+	if resp.StatusCode != 200 {
+		i.log.Print("skipping anonymous crash report and log submission for this failure")
+		return
```
ark3: Perhaps add
```
i.log.Printf("%q", string(content))
```

cmd/edgectl/main.go
```diff
@@ -123,6 +123,9 @@ func getRootCommand() *cobra.Command {
 		Long:         myHelp,
 		SilenceUsage: true, // https://github.com/spf13/cobra/issues/340
 	}
+	_ = rootCmd.PersistentFlags().Bool(
+		"no-report", false, "turn off anonymous crash reports and log submission on failure",
```
alexgervais: @ark3 I added this flag as a global option to edgectl (as opposed to an install cmd flag). This is something I feel we want top-level and applicable to other sub-commands in the future... Let me know what you think :)

```
[...]

Flags:
  -h, --help        help for edgectl
      --no-report   turn off anonymous crash reports and log submission on failure

Use "edgectl [command] --help" for more information about a command.
```

cmd/edgectl/aes_install.go
```diff
@@ -940,6 +962,101 @@ func (i *Installer) Report(eventName string, meta ...ScoutMeta) {
 	}
 }
 
+func (i *Installer) generateCrashReport(sourceError error) {
+	// TODO: Use the live endpoint
+	//reportURL := "https://metriton.datawire.io/crash-report"
```
alexgervais: TODO

cmd/edgectl/aes_install.go
```diff
@@ -453,6 +458,18 @@ func (i *Installer) Perform(kcontext string) error {
 		return err
 	}
 
+	versions, err := i.CaptureKubectl("get versions", false, "", "version", "-o", "json")
```
alexgervais: @ark3 I'm extracting kubectl and the remote cluster version here. We should reuse `i.k8sVersion` for Metriton. (another PR)

cmd/edgectl/main.go
ark3: Sensible. :+1:

cmd/edgectl/aes_install.go
```diff
@@ -453,6 +458,18 @@ func (i *Installer) Perform(kcontext string) error {
 		return err
 	}
 
+	versions, err := i.SilentCaptureKubectl("get versions", "", "version", "-o", "json")
```
ark3, line 461: I don't think this should be silent. The output is not that long and would be handy even for a user, e.g., to notice a surprising version number as a means of recognizing that they are using the wrong cluster.

cmd/edgectl/aes_install.go
```diff
@@ -453,6 +458,18 @@ func (i *Installer) Perform(kcontext string) error {
...
+	if err != nil {
+		i.Report("fail_no_cluster")
+		return err
+	}
+	k8sVersion := &kubernetesVersion{}
+	err = json.Unmarshal([]byte(versions), k8sVersion)
+	if err != nil {
+		i.log.Printf("failed to read Kubernetes client and server versions: %v", err.Error())
+	}
+	i.k8sVersion = k8sVersion
```
ark3, line 471: Is a zero-valued structure okay here in the error case? Would `nil` be better? A comment either way would be helpful. Otherwise I'll pause at this bit of code every time I'm looking through the file.

cmd/edgectl/aes_install.go
```diff
@@ -884,7 +902,7 @@ func (i *Installer) ShowKubectl(name string, input string, args ...string) error
 
 // CaptureKubectl calls kubectl and returns its stdout, dumping all the output
 // to the logger.
-func (i *Installer) CaptureKubectl(name, input string, args ...string) (res string, err error) {
+func (i *Installer) CaptureKubectl(name string, input string, args ...string) (res string, err error) {
```
ark3, line 905: This change can be removed now.

cmd/edgectl/aes_install.go
```diff
@@ -897,24 +915,45 @@ func (i *Installer) CaptureKubectl(name, input string, args ...string) (res stri
...
 	cmd := exec.Command(args[0], args[1:]...)
 	cmd.Stdin = strings.NewReader(input)
-	cmd.Stdout = io.MultiWriter(NewLoggingWriter(i.cmdOut), resAsBytes)
-	cmd.Stderr = NewLoggingWriter(i.cmdErr)
+	if logToStdout {
+		cmd.Stdout = io.MultiWriter(NewLoggingWriter(i.cmdOut), resAsBytes)
+		cmd.Stderr = NewLoggingWriter(i.cmdErr)
+	} else {
+		cmd.Stdout = io.MultiWriter(resAsBytes)
+	}
```
ark3, line 956: Why use an `io.MultiWriter` here? Just `resAsBytes` would suffice, wouldn't it? Also, what about `cmd.Stderr`?

cmd/edgectl/aes_install.go
ark3: Can you put this and all the other new functions in `install_crash.go` or something like that?

cmd/edgectl/aes_install.go
```diff
@@ -940,6 +979,102 @@ func (i *Installer) Report(eventName string, meta ...ScoutMeta) {
...
+	buffer.Write(fileContent)
+
+	buffer.WriteString("\n========== kubectl describe ==========\n")
+	describe, err := i.SilentCaptureKubectl("describe ambassador namespace", "", "-n", "ambassador", "describe", "all")
+	if err != nil {
+		i.log.Printf("failed to describe ambassador resources: %v", err.Error())
+	}
+	buffer.WriteString(describe)
+
+	buffer.WriteString("\n========== kubectl logs ==========\n")
```
ark3, line 1047: This is totally fine, but I prefer `fmt.Fprint(buffer, ...)` because that works for every `io.Writer` and is generally familiar, whereas `.WriteString(...)` only works for a few types and is somewhat unfamiliar. Please feel free to ignore me here.

PR #2520, opened 2020-04-01
Bruce/cleanup edgectl install

brucehorn:
Cleaned up error messages, spacing, and alignment of messages during the edgectl install process.  Also added initial messages before each long-running process so that there is no unexplained wait in the console output.

Screenshot below:

Screen Shot 2020-04-01 at 3 22 56 PM


cmd/edgectl/aes_install.go
```diff
@@ -605,79 +644,46 @@ func (i *Installer) Perform(kcontext string) error {
...
-		i.ShowWrapped(noTlsSuccess)
+		i.ShowWrapped(color.Bold.Sprintf(noTlsSuccess))
 		i.show.Println()
-		loginMsg := "Determine the IP address and port number of your Ambassador service.\n"
-		loginMsg += "(e.g., minikube service -n ambassador ambassador)\n"
-		loginMsg += fmt.Sprintf(loginViaIP, "IP_ADDRESS:PORT")
+		loginMsg := "Determine the IP address and port number of your Ambassador service, e.g.\n"
+		loginMsg += color.Bold.Sprintf("$ minikube service -n ambassador ambassador)\n\n")
+		loginMsg += fmt.Sprintf(loginViaIP)
+		loginMsg += color.Bold.Sprintf("$ edgectl login -n ambassador IP_ADDRESS:PORT")
```
ark3, line 652: There's a superfluous close-parenthesis in the `minikube` example.

cmd/edgectl/aes_install.go
```diff
@@ -709,20 +715,29 @@ func (i *Installer) Perform(kcontext string) error {
...
-	i.hostname = string(content)
-	i.show.Println("-> Acquiring DNS name", color.Bold.Sprintf(i.hostname))
-
 	// Wait for DNS to propagate. This tries to avoid waiting for a ten
 	// minute error backoff if the ACME registration races ahead of the DNS
 	// name appearing for LetsEncrypt.
+	i.hostname = string(content)
+	i.show.Println("-> Acquiring DNS name", color.Bold.Sprintf(i.hostname))
+
 	if err := i.loopUntil("DNS propagation to this host", i.CheckHostnameFound, lc2); err != nil {
```
ark3, line 741: I don't agree with this change. The comment applies to the `loopUntil` call, but you have moved it far away.

cmd/edgectl/aes_install.go
brucehorn: I'll add the TODO as requested and check the minikube example, thanks.

cmd/edgectl/aes_install.go
brucehorn: I'll move it to the loopUntil block, makes sense.

PR #2522, opened 2020-04-02
Add support to disable x-request-id override by ambassador

prakharjoshi:
## Description
Right now ambassador populates the `x-request-id` header for all the requests routing through it. This causes issues if the client is sending the `x-request-id` header it will be overridden with some random hash value by the ambassador.

Actually, envoy injects this header in every request but envoy now supports the option to disable this overriding by adding `preserve_external_request_id` as true to its connection manager described here https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/network/http_connection_manager/v2/http_connection_manager.proto.

Ambassador updates the envoy config and right now ambassador doesn't have the feature to pass this value to the envoy. This PR include the changes which will be allowed a configurable option to disable this header overriding by the envoy.

## Related Issues
https://github.com/datawire/ambassador/issues/2218

## Testing

## Tasks That Must Be Done

## Other


python/tests/t_request_header.py
```diff
@@ -0,0 +1,58 @@
...
+class XRequestIdHeaderDefaultTest(AmbassadorTest):
+    target: ServiceType
+
+    def init(self):
+        self.target = HTTP(name="target")
+
+    def config(self):
+        yield self.target, self.format("""
+---
+apiVersion: ambassador/v0
```
kflynn, line 41: I think that you should delete this `ambassador` module entirely from this test.

You should be able to run these tests locally by creating a cluster, setting `DEV_KUBECONFIG` and `DEV_REGISTRY` as described in `DEVELOPING.md`, then

```
make clean push pytest-only PYTEST_ARGS="-k XRequestId"
```

Does that pass for you?

ppeble: 😨 

I didn't run the whole test suite locally, to be honest, and I couldn't parse if the build failures in Travis were 'real' or not. I'll take a look at this and get back to you.

prakharjoshi: @kflynn test is breaking in my local too. Also if I revert the change and run this test it still breaks. Seems like this is not the correct place to right this test, thoughts?

prakharjoshi: as discussed @kflynn I have set the default test as `xfail`. It passes rest of the tests.

PR #2526, opened 2020-04-02
Refactor out from the build.mk the ingresstest target

inercia:
## Description

We should not run `ingresstest` as part of `test` in the `builder.mk`, as it leads to some other errors.


inercia:
Not necessary a

PR #2530, opened 2020-04-03
Move to new directory structure

concaf:


.travis.yml
```diff
@@ -76,10 +76,10 @@ after_script:
...
-    skip_cleanup: true
-    on:
-      branch: release/v1.3
+#deploy:
+#  # publish to getambassador.io
+#  - provider: script
+#    script: ./.ci/publish-website
+#    skip_cleanup: true
+#    on:
+#      branch: release/v1.3
```
concaf, line 85: There are breaking changes involved in this PR and in https://github.com/datawire/getambassador.io/pull/219. I'm _pausing_ the pushing for now till all CIs show green.

RELEASING.md
```diff
@@ -89,3 +89,139 @@ and the CI will be green.
...
+```js
+              <li>
+                <div className={classnames(styles.Dropdown, !isDocLink((location || {}).pathname) && styles.hidden)}>
+                  <button className={classnames(styles.DropdownButton, styles.DocsDropdownColor)}>{ docsVersion((location || {}).pathname) } ▾</button>
+                  <div className={styles.DropdownContent}>
+                    <Link to="/docs/latest/">Latest</Link>
+                    <Link to="/docs/1.4/">v1.4</Link>
+                    <Link to="/docs/1.3/">v1.3</Link>
+                    <Link to="/docs/1.2/">v1.2</Link>
+                    <Link to="/docs/1.2/">v1.1</Link>
```
khussey, line 203: Shouldn't this be docs/v1.1? We can probably just keep 1.3, 1.4, and latest here as an example.

Also, since the version numbers are shown without the "v" prefix everywhere else (in the URLs, in links, and besides DOCS once selected), perhaps we should drop it and stick with just 1.4, 1.3, etc..

RELEASING.md
```diff
@@ -89,3 +89,139 @@ and the CI will be green.
...
+
+##### Note:
+Now, the website must now display v1.4 docs under getambassador.io/docs/. Make sure everything looks right.
+
+##### In ambassador.git,
+
+Now that the latest release is `1.4`, we need to remove that tag from `1.3`.
+
+- In the branch `release/v1.3`,
+  - In `docs/js/doc-page.js`, update Sidebar's prefix to `/docs/1.3`
```
khussey, line 217: Is it not possible to compute the prefix dynamically based on the location of the doc-page.js file (i.e., which submodule it's under)?

RELEASING.md
khussey: Or, alternatively, include the "v" prefix consistently throughout.

d6e-automaton: Fixed

RELEASING.md
d6e-automaton: Yep, like discussed we decided to do this later after merging this one.

PR #2533, opened 2020-04-04
Intercept: Restore regex support

ark3:
Regular expression matching was unintentionally broken in the transition to using AES as the Agent. This PR restores that feature.

Fixes https://github.com/datawire/apro/issues/1185

```console
$ edgectl intercept add echo -m :path=.*ark3.* -t localhost:9000
Using deployment cept-1586022949 in namespace default
cept-1586022949: applying intercept mapping in namespace default
cept-1586022949: starting SSH tunnel
Added intercept "cept-1586022949"

$ curl -s echo | head -n 2
Request served by echo-6466548f88-lgzrn


$ curl -s echo/foo/ark3/bar | head -n 2
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN"
        "http://www.w3.org/TR/html4/strict.dtd">

$ edgectl intercept list
   1. cept-1586022949
      Intercepting requests to echo when
      - :path: .*ark3.*
      and redirecting them to localhost:9000

$ edgectl intercept add --help
Add a deployment intercept

Usage:
  edgectl intercept add DEPLOYMENT -t [HOST:]PORT -m HEADER=REGEX ... [flags]

Flags:
  -h, --help                   help for add
  -m, --match stringToString   match expression (HEADER=REGEX) (default [])
  -n, --name string            a name for this intercept
      --namespace string       Kubernetes namespace in which to create mapping for intercept
  -p, --prefix string          prefix to intercept (default /)
  -t, --target string          the [HOST:]PORT to forward to

Global Flags:
      --no-report   turn off anonymous crash reports and log submission on failure

$
```

### DIFF
--- a/cmd/edgectl/cli.go
+++ b/cmd/edgectl/cli.go
@@ -235,7 +235,7 @@ func (d *Daemon) getRootCommand(p *supervisor.Process, out *Emitter, data *Clien
 	})
 	intercept := InterceptInfo{}
 	interceptAddCmd := &cobra.Command{
-		Use:   "add DEPLOYMENT [--namespace NAMESPACE] [-p PREFIX] -t [HOST:]PORT -m HEADER=REGEX ...",
+		Use:   "add DEPLOYMENT -t [HOST:]PORT -m HEADER=REGEX ...",
 		Short: "Add a deployment intercept",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(_ *cobra.Command, args []string) error {
--- a/cmd/edgectl/cli.go
+++ b/cmd/edgectl/cli.go
@@ -87,7 +87,7 @@ func (d *Daemon) getRootCommand(p *supervisor.Process, out *Emitter, data *Clien
 			d.network = nil
 
 			out.Println("Network overrides paused.")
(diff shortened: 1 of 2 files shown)


cmd/edgectl/cli.go
```diff
@@ -235,7 +235,7 @@ func (d *Daemon) getRootCommand(p *supervisor.Process, out *Emitter, data *Clien
 	})
 	intercept := InterceptInfo{}
 	interceptAddCmd := &cobra.Command{
-		Use:   "add DEPLOYMENT [--namespace NAMESPACE] [-p PREFIX] -t [HOST:]PORT -m HEADER=REGEX ...",
+		Use:   "add DEPLOYMENT -t [HOST:]PORT -m HEADER=REGEX ...",
```
LukeShu, line 238: Why are namespace and prefix not worth mentioning here anymore?

ark3: They are optional. Since `edgectl intercept add` is a bit complicated, I chose to show what is _required_ in the usage section, leaving the optional stuff for the flags section. Those two options snuck in when we scrambled to fix namespace stuff after the Agent-_is_-AES rewrite, and I missed it in my review.

PR #2550, opened 2020-04-07
fix: proper case if check

puneetloya:
## Description
its a bug fix for my previous pull request

### DIFF
--- a/python/ambassador/envoy/v2/v2listener.py
+++ b/python/ambassador/envoy/v2/v2listener.py
@@ -811,8 +811,7 @@ class V2Listener(dict):
             if req_hdrs:
                 self.base_http_config["tracing"]["request_headers_for_tags"] = req_hdrs
 
-        proper_case = self.config.ir.ambassador_module.get('proper_case', False)
-        if proper_case:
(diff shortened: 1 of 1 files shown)


python/ambassador/envoy/v2/v2listener.py
```diff
@@ -811,8 +811,7 @@ def __init__(self, config: 'V2Config', service_port: int) -> None:
             if req_hdrs:
                 self.base_http_config["tracing"]["request_headers_for_tags"] = req_hdrs
 
-        proper_case = self.config.ir.ambassador_module.get('proper_case', False)
```
kflynn, line 814: I actually think that this is better left alone -- what're you trying to solve here?

puneetloya: the if check failed for me with the .get method, it was always evaluating as false(even after I set it to true in the config).

kflynn: Gotcha, thanks.

kflynn:
Superseded by #2555 since I needed to do hand testing -- thanks much!

PR #2555, opened 2020-04-08
proper_case fixes

kflynn:
Pull in @puneetloya's fix to the `proper_case` check (thanks!) from #2550.

Also remove `proper_case` from `Mapping`: it's a listener thing, and was a no-op in `Mapping`.

## Testing
Tested by hand:
- no Ambassador module
- Ambassador module with `proper_case: true`
- Ambassador module with `proper_case: false`
- Ambassador module that doesn't mention `proper_case` at all


kflynn:
Known broken link, merging.

puneetloya:
thanks a ton :)

PR #2560, opened 2020-04-08
Update doc links for v1.4

kflynn:
Update doc links for v1.4


kflynn:
BLC passed! no point in waiting for the rest of CI.

PR #2566, opened 2020-04-09
Bruce/edgectl install 14 fixes

brucehorn:
Added SimpleMessage, Message to all Results in edgectl install.  Checked all URL's to docs.


cmd/edgectl/aes_install_errors.go
```diff
@@ -30,9 +32,10 @@ func (i *Installer) EmailRequestError(err error) Result {
 // Unable to get a kubectl path.
 func (i *Installer) NoKubectlError(err error) Result {
 	url := "https://www.getambassador.io/docs/latest/topics/install/help/no-kubectl"
+
 	return Result{
 		Report:       "fail_no_kubectl",
-		ShortMessage: "kubectl not found in $PATH",
+		ShortMessage: "The installer was unable to find kubectl in your $PATH",
```
khussey, line 38: I'm not sure about prefixing messages like this one and others with "The installer was/could/etc." as opposed to just stating the error condition. I'm sure Bjorn will have a preference on the desired wording as part of future improvements but for now we need to ensure that every error path HAS a message...

cmd/edgectl/aes_install_errors.go
```diff
@@ -42,39 +45,51 @@ func (i *Installer) NoKubectlError(err error) Result {
...
 // Unable to get client configuration or namespace
 func (i *Installer) GetRestConfigError(err error) Result {
+	url := "https://www.getambassador.io/docs/latest/topics/install/help/get-rest-config"
+
 	return Result{
-		Report: "fail_no_cluster",
-		URL:    "https://www.getambassador.io/docs/latest/topics/install/help/get-rest-config",
-		Err:    err,
+		Report:       "fail_no_cluster",
+		ShortMessage: "The installer could not communicate with your Kubernetes cluster",
```
khussey, line 64: This short message is generic (like the page it references) but should ideally be a little more specific about the particular error case, IMO. In this case, the installer was unable to get the client configuration or namespace from the Kubernetes cluster...

khussey: Same/similar comment applies to other messages below where you've included a generic short message.

cmd/edgectl/aes_install_errors.go
```diff
@@ -42,39 +45,51 @@ func (i *Installer) NoKubectlError(err error) Result {
...
+	url := "https://www.getambassador.io/docs/latest/topics/install/help/new-for-config"
+
 	return Result{
-		Report: "fail_no_cluster",
-		URL:    "https://www.getambassador.io/docs/latest/topics/install/help/new-for-config",
-		Err:    err,
+		Report:       "fail_no_cluster",
+		ShortMessage: "The installer could not communicate with your Kubernetes cluster",
+		Message:      fmt.Sprintf("Visit %v for a more detailed explanation and step-by-step instructions about how to set up your Kubernetes environment to continue installing Ambassador Edge Stack.", url),
+		URL:          "https://www.getambassador.io/docs/latest/topics/install/help/new-for-config",
```
khussey, line 79: Should be `URL: url,` here.

cmd/edgectl/aes_install_errors.go
```diff
@@ -120,19 +137,19 @@ func (i *Installer) ManifestParsingError(err error, matches []string) Result {
...
-	url := "https://www.getambassador.io/docs/latest/topics/install/help/incompatible-crd-versions"
 	message := fmt.Sprintf("Visit %v for a more detailed explanation and step-by-step instructions about updating CRD versions to continue installing Ambassador Edge Stack.\n%s",
 		url, abortExisting)
 
-	i.Report("fail_existing_aes", ScoutMeta{"installing", i.version}, ScoutMeta{"found", installedVersion})
-
 	return Result{
 		URL:          url,
-		ShortMessage: fmt.Sprintf("Found incompatible AES CRD versions: Existing AES %s found when installing AES %s", installedVersion, i.version),
+		ShortMessage: fmt.Sprintf("The installe found incompatible AES CRD versions: Existing AES %s found when installing AES %s", installedVersion, i.version),
```
khussey, line 152: "installe" -> "installer"

cmd/edgectl/aes_install_errors.go
```diff
@@ -141,62 +158,85 @@ The installer will quit to avoid corrupting an existing installation of AES.`
...
 func (i *Installer) AESPodStartupError(err error) Result {
+	url := "https://www.getambassador.io/docs/latest/topics/install/help/aes-pod-startup"
+
 	return Result{
-		Report: "fail_pod_timeout",
-		URL:    "https://www.getambassador.io/docs/latest/topics/install/help/aes-pod-startup",
-		Err:    err,
+		Report:       "fail_pod_timeout",
+		ShortMessage: "The installer was unable to talk to your AES pod",
+		Message:      fmt.Sprintf("The installer failed to communicate with an AES Pod in your cluster to validate the installation.  Visit %v for a more detailed explanation and step-by-step instructions about resolving this problem.", url),
```
khussey, line 230: Message should start with "Visit ..." like all the others; move the specific message here to the short message (or remove it, since the short message seems OK already).

cmd/edgectl/aes_install_errors.go
```diff
@@ -256,29 +297,33 @@ func (i *Installer) AESACMEChallengeError(err error) Result {
...
+	url := "https://www.getambassador.io/docs/latest/topics/install/help/dns-name-body"
 	i.Report("dns_name_failure", ScoutMeta{"err", err.Error()})
 
-	url := "https://www.getambassador.io/docs/topics/install/help/dns-name-body"
 	return Result{
 		ShortMessage: "Failed to register DNS name for the current installation",
 		Message:      fmt.Sprintf("Visit %v for a more detailed explanation and step-by-step instructions about acquiring a DNS name to continue installing Ambassador Edge Stack.", url),
 		URL:          url,
-		Err:          errors.Wrap(err, "acquire DNS name (read body)"),
+		Err:          errors.Wrap(err, "Failed to acquire DNS name (read body)"),
```
khussey, line 320: It's unclear to me why some of these `errors.Wrap(...)` calls include a brief error phase and others do not... is that what's contributing to the redundancy Bjorn saw?

cmd/edgectl/aes_install_errors.go
brucehorn: This is what they look like now.

new AESPodStartupError

brucehorn: let me know if this is still something you want to change.

cmd/edgectl/aes_install_errors.go
brucehorn: good catch, will fix

cmd/edgectl/aes_install_errors.go
brucehorn: thanks :-)

cmd/edgectl/aes_install_errors.go
brucehorn: new AESPodStartupError

brucehorn: is this OK?  That's what it looks like now.

cmd/edgectl/aes_install_errors.go
brucehorn: probably, I will take a look.  I understood that we weren't going to worry about the redundancy this time around but if this is something that we want to address for 1.4.1 we can do that.

cmd/edgectl/aes_install_errors.go
khussey: We should remove the "The installer failed to communicate with an AES Pod in your cluster to validate the installation." sentence from the message and keep it to "Visit ..." just like the rest. (This was the only one that didn't follow the consistent pattern.)

cmd/edgectl/aes_install_errors.go
khussey: We don't need to address it for 1.4.1, was just thinking out loud here.

cmd/edgectl/aes_install_errors.go
khussey: It's fine, was just thinking we'd do the wordsmithing after we address the problems in 1.4.1.

khussey:
@brucehorn were you able to verify that the paths that were missing error messages are now emitting them?

khussey:
@ark3 please review ASAP and we'll get this merged and peel off a 1.4.1 release, thanks!

cmd/edgectl/aes_install_result.go
```diff
@@ -127,9 +128,6 @@ func (i *Installer) ShowResult(r Result) {
 		}
 
 		if r.Message != "" {
-			i.show.Println()
```
khussey, line 130: @brucehorn the one question I have about this change is whether the right messaging is still output for the `KnownLocalClusterResult` case. It would probably be worthwhile to review all of the "successful" exit cases (i.e., those without `err`s to make sure their results are handled properly (based on which fields have been populated in the struct) and the right messages are output. /cc @ark3

cmd/edgectl/aes_install_result.go
```diff
@@ -127,9 +128,6 @@ func (i *Installer) ShowResult(r Result) {
...
-			i.show.Println("AES Installation Complete!")
```
khussey, line 131: @brucehorn the one question I have about this change is whether the right messaging is still output for the `KnownLocalClusterResult` case. It would probably be worthwhile to review all of the "successful" exit cases (i.e., those without `err`s to make sure their results are handled properly (based on which fields have been populated in the struct) and the right messages are output. /cc @ark3

brucehorn:
**New flow for failed DNS:**
failed dns

**New flow for full success (existing install, but otherwise the same):**

full success

brucehorn:
**For a known local cluster:**

local cluster

cmd/edgectl/aes_install_result.go
brucehorn: fixed, see below

khussey: Thank you!

PR #2574, opened 2020-04-09
[WIP] Moving styles beside the page where they're used.

khussey:
Testing whether styles for docs pages can be referenced locally.


khussey:
Experiment successful!

PR #2577, opened 2020-04-10
edgectl adds some installation info as labels in the Deployment

inercia:
## Description

Add some installation info as labels in the `Deployment`. For example:

```yaml
apiVersion: apps/v1
kind: Deployment
  ...
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  template:
    metadata:
      labels:
        app.kubernetes.io/instance: ambassador
        app.kubernetes.io/managed-by: edgectl
        app.kubernetes.io/name: ambassador
        app.kubernetes.io/part-of: ambassador
        getambassador.io/install-cluster-type: K3D
        getambassador.io/install-date: 2020-04-10.23-09-21
        getambassador.io/installer-arch: amd64
        getambassador.io/installer-os: linux
        getambassador.io/installer-version: 1.4.0a-11-g4e4877715-dirty
        product: aes

```

### DIFF
--- a/cmd/edgectl/aes_install.go
+++ b/cmd/edgectl/aes_install.go
@@ -16,6 +16,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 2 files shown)


cmd/edgectl/aes_install.go
```diff
@@ -620,6 +624,25 @@ func (i *Installer) Perform(kcontext string) Result {
...
+			// replace all the forbidden chars in the value by a `-`
+			for _, forbidden := range []string{",", " ", "\\", "/", "(", ")", "\"", "*", "^", "$", "#", "@", "\n", "\t"} {
+				reportedVersion = strings.ReplaceAll(reportedVersion, forbidden, "-")
+			}
+		}
+
+		// Save some environment info as Helm values
+		currentTime := time.Now()
+		chartValues[defHelmValueRootEdgectlInfo] = map[string]interface{}{
+			"edgectl/install-cluster-type": i.clusterinfo.name,
```
alexgervais, line 639: I'd leave `install-cluster-type` aside... I don't think it's relevant to either Ambassador or the ops people behind the installation. Plus, what if we get it wrong? It's one thing to show it in egectl logs but not a definitive answer.

inercia: I think that reporting the environment where Ambassador has been installed could be valuable information from a statistical point of view. Knowing that users are using GKE or Minikube for trying Ambassador can give us an interesting insight, even if this information is not 100% accurate. Maybe @bnfb can comment on this.

alexgervais: We can already extract this information in metriton using a simple join on 2 fields if we want to build such a view. I think we should limit the number of labels to prevent bloating the pod template with things we can infer otherwise and make sure every instrumented label serves a purpose. Their usage and intent should also be documented.

Instrumenting the cluster-type could be dangerous as Ambassador needs to remain agnostic of the platform (and support them all without if==gke, if==aws)

bnfb: If we can report that information in metriton then great: let's have a metq report that does that and call it good. But the goal is to get the report.

inercia:
Closing in favor of some other metrics.

PR #2578, opened 2020-04-10
diagd: Move estats update to its own thread

jfrabaute:
The estat update thread was just posting a message to the
AmbassadorEventWatcher thread.
Then this AmbassadorEventWatcher thread was processing sequentially all
the message.
This thread is also updating the config. This message can take some
time, more than 20 seconds.
20 seconds is the delay used by ambassador to declare envoy dead when it
could not get an update status from envoy.
But when the config takes more than 20 seconds, this blocks the message
doing this envoy update status, and then ambassador alive endpoint is
starting to return 503.
Then, at the k8s level, the liveness probe is failing and the container
is restarted.

In order to prevent this lag in the envoy update state, the timer which
was posting the update message is now directly doing the update itself.

## Description

## Related Issues
List related issues.

## Testing

## Tasks That Must Be Done

## Other

### DIFF
--- a/python/ambassador_diag/diagd.py
+++ b/python/ambassador_diag/diagd.py
@@ -177,6 +177,13 @@ class DiagApp (Flask):
     def check_scout(self, what: str) -> None:
         self.watcher.post("SCOUT", (what, self.ir))
 
+    def update_estats(self) -> None:
+        try:
+            self.estats.update()
+        except Exception as e:
+            self.logger.error("could not update estats: %s" % e)
+            self.logger.exception(e)
+
 
 # get the "templates" directory, or raise "FileNotFoundError" if not found
 def get_templates_dir():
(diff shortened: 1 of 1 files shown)


jfrabaute:
Feedback: This patch has been running for two weeks now on our internal prod cluster and we don't see any restart at all now.

kflynn:
Glad to hear this is running smoothly for you! What kind of load is this running under?

kflynn:
After going through and reviewing this one... I think we're going to have to call this one unsafe. 🙁  

This change fundamentally relies on the atomicity of the `self.stats.update({...})` call in `EnvoyStats.update_envoy_stats()`. That`s the `dict.update()` call, which ends up calling `__next__` on its argument -- but `__next__` can do literally anything, with no guarantees about whether or not it's thread-safe. 

It's true that the current call is probably mostly safe, because of what we're passing to `self.stats.update()`, but it's not hard to imagine us changing that without much thought. It feels like a safer path forward is to move the configuration parsing off the main thread, so that it doesn't so completely block other events. Let's talk about that on Slack...

kflynn:
We can close this in favor of #2674, right?

jfrabaute:
Superseded by https://github.com/datawire/ambassador/pull/2674

PR #2580, opened 2020-04-13
Do not fail if a previous `edgectl install`ed installation has been found

inercia:
## Description

Do not fail if a previous `edgectl install`ed installation has been found. The installation will continue with the remaining steps after finding a previous `edgectl` installation.

On the first `edgectl install`:

edgectl-install-first

then a second `edgectl install` produces:

edgectl-install-second

## Testing

Install twice and check that the second time you get a nice, non-error message.


brucehorn:
**First install:**


First install

**Second install:**

Second install

brucehorn:
**Better flow now:**

Second install 2

PR #2594, opened 2020-04-15
Recognize other Operator installations

inercia:
## Description

Recognize other Operator installations.

### DIFF
--- a/cmd/edgectl/aes_install_info.go
+++ b/cmd/edgectl/aes_install_info.go
@@ -154,7 +154,7 @@ var defInstallationMethodsInfo = []installationMethodInfo{
 	},
 	{
 		Method:   instOperator,
-		Label:    "app.kubernetes.io/managed-by=amb-oper",
+		Label:    "getambassador.io/installer=operator",
 		Name:     "operator",
 		LongName: "the Ambassador Operator",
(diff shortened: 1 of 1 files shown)


khussey:
> What is this PR trying to achieve? I don't see any usage of the added elements, and I'm questioning why edgectl would need to know about those operator-related metadata blocks.

Perhaps the existing instOperator element can just be updated to match the various flavours of Operator install with pattern matching (amb-oper-<something>)?

inercia:
> What is this PR trying to achieve? I don't see any usage of the added elements, and I'm questioning why edgectl would need to know about those operator-related metadata blocks.

@alexgervais it is for recognizing previous installation by the operator, as the operator will set different `managed-by` depending on how it was originally installed: the operator will deploy AES with `app.kubernetes.io/managed-by` set:

* `amb-oper-azure` if installed from the Azure marketplace
* `amb-oper-helm` if installeed from a Helm chart
* `amb-oper-manifest` if installed with a manifest

alexgervais:
@inercia can we do something in the likes of what @khussey suggested? Knowing we'll probably add more values for this field, it would require a new release and user-update of `edgectl` every time.

PR #2602, opened 2020-04-16
Installer: Suggest login if an existing Host points to an existing AES

ark3:
## Goals

1. Don't behave any differently for users doing a fresh install
2. Precisely figure out that an installation already exists
3. Don't change anything in the cluster if an installation already exists

This PR addresses those goals for the full success case where the installer has (in a prior invocation) already set up DNS and created a Host object. 

This PR does not address those goals for the no-DNS success case. However, the installer already suggests login rather than doing the login automatically for that success case. We should think about whether we want to change things for that case, but that's outside the scope of this PR.

I don't think performance is a goal here, but with my laptop and GKE cluster, I can run the "already installed" scenario in under ten seconds.

## Review Questions

- Does this PR achieve goal 1 above? That's really important!
- Does the heuristic implemented in this PR (existing Host resource points to the LB address) correctly achieve goal 2?
- Does the existing code (not touched by the PR) achieve goal 3?

## Screenshots

Install a second time (new flow at the end):
image

Install a second time after nuking the existing Host object (unchanged from before, except that one line was added, "Looking for a Host resource..."):
image

Install for the first time is unchanged.

### DIFF
--- a/cmd/edgectl/aes_install.go
+++ b/cmd/edgectl/aes_install.go
@@ -752,6 +752,26 @@ func (i *Installer) Perform(kcontext string) Result {
 		return i.AESACMEChallengeError(err)
 	}
 	i.Report("aes_listening")
+
+	if installedVersion != "" {
+		i.show.Printf("-> Looking for a Host resource in the existing installation")
+		hostResource, err := i.FindMatchingHostResource()
+		if err != nil {
+			i.log.Printf("Failed to look for Hosts: %+v", err)
+			hostResource = nil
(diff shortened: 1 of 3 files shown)


bnfb:
@ark3 The user experience (the console output) looks good to me, thanks.

brucehorn:
+1 pending the change in my comment above re. alignment and phrasing

ark3:
Updated screenshot for "install a second time" with 80a6232:
image

PR #2605, opened 2020-04-16
Render a Welcome section on the AES Edge Policy Console after edgectl install

brucehorn:
After running edgectl install, add a new query string (`?first-install=true`) to indicate that the Edge Policy Console is being logged into for the first time after installation.  This PR works with https://github.com/datawire/apro/pull/1241.


cmd/edgectl/aes_login.go
```diff
@@ -32,17 +32,18 @@ func aesLogin(cmd *cobra.Command, args []string) error {
 	namespace, _ := cmd.Flags().GetString("namespace")
 	justShowURL, _ := cmd.Flags().GetBool("url")
 	showToken, _ := cmd.Flags().GetBool("token")
+	firstInstall, _ := cmd.Flags().GetBool("firstInstall")
```
alexgervais, line 35: Is this left over testing code or is there a specific reason why we would want this flag in `edgectl login`?

cmd/edgectl/aes_login.go
```diff
@@ -86,7 +87,7 @@ func do_login(kubeinfo *k8s.KubeInfo, context, namespace, hostname string, openI
 	}
 
 	// Output
-	url := fmt.Sprintf("https://%s/edge_stack/admin/#%s", hostname, tokenString)
+	url := fmt.Sprintf("https://%s/edge_stack/admin/?first-install=%v#%s", hostname, firstInstall, tokenString)
```
alexgervais, line 90: I'm thinking maybe we want the `?first-install=%v` string to be appended ONLY if it's a firstInstall, to prevent "hackers" from tempering with the URL as it'll show up in their browser as `?first-install=false`

brucehorn: interesting, I actually had the same thought.  But of course they can always type it in themselves after they saw it once.

PR #2607, opened 2020-04-17
Fixes a wrong match for installations by the operator

inercia:
## Description

Fix the matching for recognizing previous installations by the Operator.

Screenshot from 2020-04-17 16 37 57


khussey:
@inercia please include a screen capture showing the output when an operator install has (now) properly been detected.

PR #2614, opened 2020-04-17
Traffic Agent: Listen on a single port set by env or default to 9900

ark3:
## Description
- Listen on a single port rather than two
- Make that port 9900 by default, so it doesn't stomp on the application running in the same container
- Make that port configurable
- Accept either cleartext or TLS, not both

## Related Issues
Fixes datawire/apro#1239

## Testing
Manual testing with and without TLS

## Tasks That Must Be Done


richarddli:
Docs LGTM

PR #2616, opened 2020-04-17
Add ability to use query parameters in Mappings

ppeble:
## Description

Adds the ability to use [query parameters](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto.html?highlight=routematch#config-route-v3-routematch) in `Mapping` objects.

Please note that while I linked to the latest envoy docs the behavior of query parameters is the same as the datawire envoy fork. All tests were performed on the datawire envoy fork.

## Testing

I have added python tests to verify this functionality. I based them on the similar header tests since it behaves in a similar fashion.

## Tasks That Must Be Done
- [X] Did you add or update tests?

### DIFF
--- a/docs/topics/using/query_parameters.md
+++ b/docs/topics/using/query_parameters.md
@@ -54,7 +54,7 @@ will send requests that contain the `quote-mode` query parameter to the `quote-m
 
 ## `regex_query_parameters`
 
-The following mapping will route mobile requests from Android and iPhones to a mobile service:
+The following mapping will route requests with the `quote-mode` header that contain values that match the regex:
 
 ```yaml
 ---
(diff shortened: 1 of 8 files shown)


docs/topics/using/query_parameters.md
```diff
@@ -0,0 +1,70 @@
...
+---
+apiVersion: getambassador.io/v2
+kind:  Mapping
+metadata:
+  name:  quote-backend
+spec:
+  regex_query_parameters:
+    user-agent: "^(?=.*\\bAndroid\\b)(?=.*\\b(m|M)obile\\b).*|(?=.*\\biPhone\\b)(?=.*\\b(m|M)obile\\b).*$"
+  prefix: /backend/
+  service: quote
```
kflynn, line 69: Is there a different example that doesn't imply that every mobile app on Android and iPhone set the `user-agent` in the query parameter? 🙂  I realize that that sounds kind of silly, but `user-agent` is _just_ well-known enough that I can imagine search engines linking to this in a way that's wrong...

ppeble: Absolutely, I copy/pasted this from the `headers` page so I'm not attached to it. I'll throw something up later today. 🙇

ppeble: Updated!

kflynn:
Landing this because the failure is bogus.

PR #2622, opened 2020-04-20
[story] Route-to-Code

LukeShu:
This is the long-lived PR for the Route-to-Code story. Do not merge until the story is finished.


LukeShu:
I'm not sure what's going on, but GitHub's view of the diff us just wrong.  Look over this with local tooling.

python/tests/utils.py
```diff
@@ -105,58 +107,48 @@ def install_ambassador(namespace, envs=None):
...
+        ),
+    )))
+
+    for manifest in ambassador_yaml:
+        if manifest.get('kind', '') == 'Pod' and manifest.get('metadata', {}).get('name', '') == 'ambassador':
+
+            # Don't set AMBASSADOR_ID={self.k8s.path}
+            for envvar in manifest['spec']['containers'][0]['env']:
+                if envvar.get('name', '') == 'AMBASSADOR_ID':
+                    envvar['value'] = 'default'
```
kflynn, line 146: Wait what? 

The mechanics of what this is doing are clear, but... why?

LukeShu: Because the knative and watt_scaling tests don't expect `AMBASSADOR_ID` to be set, because `install_ambassador` wasn't setting it before.

kflynn: ...and IIRC `install_ambassador` is only used for them?

LukeShu: That is correct.

PR #2624, opened 2020-04-21
Minor: export the Char directory

inercia:
## Description

Minor change, for making public the directory where the Helm Chart is. This is required for using these libraries from the Operator.


alexgervais:
@inercia will this have an impact on `edgectl install` such as leaving a directory on the user's fs?

inercia:
> @inercia will this have an impact on `edgectl install` such as leaving a directory on the user's fs?

It does not have any impact. This is just some info that is currently held in a temporary variable and we will export it...

PR #2631, opened 2020-04-22
Revert (most of) “YAML for 1.4.2"

kflynn:
This reverts most of commit 8299e48a9145c9983cfa961037fd82e99f3450f4. We leave the ambassador ServiceAccount up towards the top of the file.


docs/yaml/aes.yaml
```diff
@@ -37,9 +37,6 @@ rules:
 #  3. It needs permission to "create" events.
 #  4. It needs permission to "get"/"create"/"update" Leases (or
 #     Endpoints, if on Kubernetes <1.12).
-#
-# Note: If you edit this, be sure to also edit the pytest
-# `manifests/rbac_*_scope.yaml` files.
```
LukeShu, line 42: This comment would be helpful to keep.

kflynn: Good point. Restored.

PR #2644, opened 2020-04-24
updated ubuntu to 18.04

ducksecops:
## Description
Update tests to use ubuntu 18.04 instead of 14.04

## Related Issues
Ubuntu 14.04 was end of life on 30th April 2019


kflynn:
Merging without CI since CI won't rebuild these anyway.

Issue #2589, opened 2020-04-14
./cmd/capabilties_wrapper doesn't compile on macOS


LukeShu:
**Describe the bug**

`go build ./cmd/capabilities_wrapper` doesn't work (so neither does `go build ./...`) on macOS, and that's a pain for all of the macOS-using developers we have.

ark3:
Fixed in #2648?

cindymullins-dw:
This issue has been addressed.

PR #2050, opened 2019-11-21
Launch envoy through capability wrapper

swalberg:
(if capabilities are present)

When capabilities are explicitly dropped
```
sean~/play/ambassador (use_capabilities_wrapper *%)$ kubectl logs ambassador-7cb95448b8-mjxhp | grep cap
2019-11-21 17:59:05 AMBASSADOR INFO cap_net_bind_service is not supported, launching Envoy directly
```

When capabilities are present
```
sean~/play/ambassador (use_capabilities_wrapper *%)$ kubectl logs ambassador-bdbc4675d-hmq82 | grep cap
2019-11-21 18:00:08 AMBASSADOR INFO cap_net_bind_service is supported, launching Envoy through a wrapper
2019/11/21 18:00:08 Starting Envoy with CAP_NET_BIND_SERVICE capability
2019/11/21 18:00:08 Succeeded in setting capabilities
```

Additionally, if the capabilities are present, you can bind Envoy to a
low port, e.g. 80 or 443:

```
bash-5.0$ netstat -anp | grep 0.0.0.0:80
tcp        0      0 0.0.0.0:80              0.0.0.0:*               LISTEN      -
bash-5.0$ ps -ef | grep [w]rapper
  167 8888      0:06 {envoy} wrapper -c /ambassador/bootstrap-ads.json
```

## Description
A few sentences describing the overall goals of the pull request's commits.

## Related Issues
List related issues.

## Testing
A few sentences describing what testing you've done, e.g., manual tests, automated tests, deployed in production, etc.

## Todos
- [ ] Tests
- [ ] Documentation

## Other
* If this is a documentation change, please open the pull request at https://github.com/datawire/ambassador-docs instead.
* Code changes should occur against `master`.

PR #2648, opened 2020-04-27
Fixup capabilities_wrapper

LukeShu:
## Description

- Stub out the linux-only stuff in `cmd/capabilities_wrapper`, so that it compiles on things other than `GOOS=linux`, and stops causing headaches for macOS-using developers. (#2589)
- Fix the filesystem capabilities on the wrapper binary in the final Docker image; they weren't getting set.  This was a regression in 1.3.0.

## Related Issues

#2589

## Testing

- `go build ./cmd/capabilities_wrapper` ran fine on on a macincloud
- `make images && docker run --rm -it --entrypoint=sh ambassador:latest -c 'apk add attr && getfattr -d -m - /usr/local/bin/wrapper'` verified that it now has the `security.capability` xattr set.

## Tasks That Must Be Done
- [x] Did you update CHANGELOG.md? - no applicable changes; #2050 hasn't landed yet
- [ ] Did you add or update tests? - no, should I?
- [x] Did you update documentation? - no applicable changes
- [x] Were there any special dev tricks you had to use to work on this code efficiently? - no
- [x] Is this a build change? - yes
  + [x] If so, did you test on both Mac and Linux? - yes

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -130,7 +130,7 @@ RUN setcap cap_net_bind_service=ei /usr/local/bin/envoy
 # Our Go binaries. See envoy section for setcap info.
 COPY --from=artifacts /opt/ambassador /opt/ambassador
 RUN ln -s /opt/ambassador/bin/* /usr/local/bin/
-RUN setcap cap_net_bind_service=p /usr/local/bin/wrapper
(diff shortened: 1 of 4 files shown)


kflynn:
Ignoring bad links because they all seem to be "we couldn't talk to haproxy.org" -- which is weird, but should be transient.

PR #2655, opened 2020-04-30
Add option to set envoy base-id in entrypoint script

ppeble:
## Description

Adds the ability to specify a new [Envoy base ID](https://www.envoyproxy.io/docs/envoy/latest/operations/cli.html?highlight=base%20id#cmdoption-base-id) via an environment variable.

If no variable is provided then there is no behavior change.

This is being added to allow for Ambassador to run side-by-side with an `istio` sidecar. Unfortunately with istio 1.5 secret handling behavior has made a breaking change. We can no longer rely on istio writing their SSL certificates to a kubernetes secret that we can consume. This means that it is now impossible to run Ambassador with a freshly installed istio 1.5 and still use strict mTLS.

One solution is to run Ambassador side-by-side with an istio sidecar like any other service. But because two envoy instances should not share a base id we need the ability to shift it for Ambassador.

Please note that there are other possibilities for solving this issue, including reading from istio directly for secrets, and this is just one possible path. This is the lowest cost solution at the moment. Other possibilities are discussed in the related issue below.

## Related Issues

Istio 1.5 issue: https://github.com/datawire/ambassador/issues/2587

## Testing

Deployed to testing k3d infrastructure locally to verify.

## Other

Please note that I am not updating documentation with this specific change. Unfortunately this breaking istio change has large implications for those using both Ambassador and istio when using strict mTLS. To add to this there are many possible configuration paths...it is apparently possible to upgrade to istio 1.5 _without_ including this breaking change. It seems that only fresh installs include the breaking change. So the documentation for this change will be complex and I would like more time to use it in my environments before putting something together.


ppeble:
> Thanks! How about go ahead and document it without specifically talking about Istio before we land this one?

Oh right! I forgot that there was documentation about the env vars as well. Let me at least do that, sure thing.

kflynn:
Yup, looks like `travis-cleanup.sh` is b0rked. Landing because the actual tests passed.

PR #2666, opened 2020-05-05
Move Ambassador CRD handling to resource processor

impl:
## Description
This change removes the `handle_k8s_crd` method of the resource fetcher, instead delegating that work to the `AmbassadorProcessor` class introduced in a previous refactor. The generated IR remains the same.

## Testing
I've run this through the Python test suite for validation.

## Tasks That Must Be Done
- [ ] Is this a build change?

## Other

### DIFF
--- a/python/ambassador/config/resourcefetcher.py
+++ b/python/ambassador/config/resourcefetcher.py
@@ -302,7 +302,7 @@ class ResourceFetcher:
                 # Nothing else to do.
                 return
 
-        handler_name = f'handle_k8s_{obj.gvk.kind.lower()}'
+        handler_name = f'handle_k8s_{obj.kind.lower()}'
         # self.logger.debug(f"looking for handler {handler_name} for K8s {kind} {name}")
         handler = getattr(self, handler_name, None)
 
@@ -310,7 +310,7 @@ class ResourceFetcher:
             self.logger.debug(f"{self.location}: skipping K8s {obj.gvk}")
             return
 
-        if not self.check_k8s_dup(obj.gvk.kind, obj.namespace, obj.name):
+        if not self.check_k8s_dup(obj.kind, obj.namespace, obj.name):
             return
 
         result = handler(raw_obj)
--- a/python/ambassador/config/resourcefetcher.py
+++ b/python/ambassador/config/resourcefetcher.py
@@ -234,6 +234,9 @@ class ResourceFetcher:
             # Then we add everything else to be processed.
             watt_k8s_keys += watt_k8s.keys()
 
+            # `dict.fromkeys(iterable)` is a convenient way to work around the
+            # lack of an ordered set collection type in Python. As Python 3.7,
+            # dicts are guaranteed to be insertion-ordered.
             for key in dict.fromkeys(watt_k8s_keys):
                 for obj in watt_k8s.get(key) or []:
                     # self.logger.debug(f"Handling Kubernetes {key}...")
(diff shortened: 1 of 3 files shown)


python/ambassador/config/resourcefetcher.py
```diff
@@ -233,16 +227,16 @@ def parse_watt(self, serialization: str, finalize: bool=True) -> None:
...
+            # These objects have to be processed first, in order, as they depend
+            # on each other.
+            watt_k8s_keys = ['service', 'endpoints', 'secret', 'ingressclasses', 'ingresses']
 
-            # ...then handle Ambassador CRDs.
-            for key in CRDTypes:
+            # Then we add everything else to be processed.
+            watt_k8s_keys += watt_k8s.keys()
+
+            for key in dict.fromkeys(watt_k8s_keys):
```
kflynn, line 237: Hmmmmm. So your comment on line 230 for `watt_k8s_keys` lists a sequence of keys that must be processed _in order_, then this line generates a dict (to deduplicate the keys, presumably)... so won't iterating over the dict shred the order you so carefully set up?

python/ambassador/config/resourcefetcher.py
```diff
@@ -308,25 +302,16 @@ def handle_k8s(self, raw_obj: dict) -> None:
...
-
-        if obj.gvk.kind in CRDTypes:
-            handler = self.handle_k8s_crd
-            # self.handle_k8s_crd will do its own dup checking.
-            check_dup = False
-        else:
-            handler_name = f'handle_k8s_{obj.gvk.kind.lower()}'
-            # self.logger.debug(f"looking for handler {handler_name} for K8s {kind} {name}")
-            handler = getattr(self, handler_name, None)
+        handler_name = f'handle_k8s_{obj.gvk.kind.lower()}'
```
kflynn, line 305: I kinda feel like there should be convenience accessors for `obj.gvk.kind` and `obj.gvk.kind.lower()`. 🙂  Probably a separate PR...

python/ambassador/config/resourcefetcher.py
impl: I will add a comment to clarify, but this is a basically a convenient way to get around the lack of a distinct `OrderedSet` type. As of [Python 3.7](https://mail.python.org/pipermail/python-dev/2017-December/151283.html), dictionaries are insertion-ordered:

```
Python 3.7.6 (default, Jan 19 2020, 22:34:52) 
[GCC 9.2.1 20200117] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> foo = ['bar', 'baz', 'quux', 'quux', 'baz']
>>> dict.fromkeys(foo)
{'bar': None, 'baz': None, 'quux': None}
>>> dict.fromkeys(foo)
{'bar': None, 'baz': None, 'quux': None}
>>> dict.fromkeys(foo)
{'bar': None, 'baz': None, 'quux': None}
>>> dict.fromkeys(foo)
{'bar': None, 'baz': None, 'quux': None}
```

python/ambassador/config/resourcefetcher.py
impl: This is actually a super small change so I'll throw it in here as well. Although there are actually almost no places where the `.lower()` is still used (other than generating the domain in the GVK itself), so maybe just `.kind`?

impl:
> To confirm: `AmbassadorProcessor` is picking up all the resource types for which nothing special needs to happen, correct? Definitely looks cleaner there.
> 
> I had one other question about the ordering of dictionary keys -- looks good other than that, thanks!!

Indeed, it's just directly shimming those Kubernetes objects across to Ambassador resources. That's why the processor's `process` method is just one line of code.

PR #2670, opened 2020-05-06
(WIP) Split fetcher/processor into its own module

impl:
## Description
This change takes the existing classes in `resourcefetcher.py` and `resourceprocessor.py` and splits them into separate files in their own module, `.fetch`. I started working on moving the service/endpoint handling code out of the fetcher and it became evident that having all of the processors in one file would make navigating through the classes difficult.

## Testing
We'll run this in our staging cluster for a bit before removing the PR from WIP status. We also ran it through the test suite, of course.

## Tasks That Must Be Done
- [ ] Is this a build change?


kflynn:
With the exception of the test additions, commit 40784e3 is just rearrangement, right?

impl:
It is, with a very minor change to prevent cyclic imports (`KubernetesObject.as_normalized_resource` -> `NormalizedResource.from_kubernetes_object`). Even the tests are just reorganized too.

PR #2673, opened 2020-05-07
Don't try to build the website if GH_TOKEN is unset.

kflynn:
This seems to the cause of the recent spate of community-PR build failures

## Tasks That Must Be Done
- [ ] Did you update CHANGELOG.md? no, because it's not part of a release
- [x] Is this a build change? yes. Tested by hand on Mac, CI will test on Linux. 🙂

### DIFF
--- a/.ci/pr-build-website-preview
+++ b/.ci/pr-build-website-preview
@@ -15,6 +15,14 @@
 # limitations under the License
 
 set -o errexit
+
+# Don't bother trying to build the website if GH_TOKEN is unset (as it will
+# be for a community PR).
+if [ -z "$GH_TOKEN" ]; then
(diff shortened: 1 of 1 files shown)


LukeShu:
But feedback on bash:

I notice that you put this above the `set -o nounset`.  If you want to do the same check with nounset; use `:-` to provide a "fallback" value for the unset case: `[ -z "${GH_TOKEN:-}" ]`

kflynn:
Personally I think it's just absurd that `nounset` fires an error when you're _testing to see if the variable is set_. Oh well. 🙂 

(More seriously, putting it there also made sense because I didn't want `xtrace` either.)

PR #2675, opened 2020-05-08
Updated docs on the operator

inercia:
## Description

Updated docs on the operator.

### DIFF
--- a/docs/topics/install/aes-operator.md
+++ b/docs/topics/install/aes-operator.md
@@ -120,11 +120,9 @@ helmValues:
     type: NodePort
 ```
 
-* Note that, `spec.helmValues.enableAES` should not conflict with the `spec.installOSS` field or else your
-  installation will error out.
-  To install OSS, only `installOSS: true` needs to be set. To migrate from OSS to AES, change `installOSS: true` to `installOSS: false`.
-  `helmValues.enableAES` does not need to be changed in any scenario.
-  However, if you end up with a configuration where both `installOSS` and `enableAES` are set and they conflict with each other (`installOSS: true` + `enableAES: true` or `installOSS: false` or `enableAES: false`), then AmbassadorInstallation and Ambassador Operator logs will throw related errors.
+* Note that the `spec.installOSS` parameter should be used instead of `spec.helmValues.enableAES` to control whether 
+  OSS or AES is installed. A configuration where both `installOSS` and `enableAES` are set to the same value will 
+  introduce a conclict and result in an error.
 
 ## Install via Helm Chart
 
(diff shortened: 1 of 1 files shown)


docs/topics/install/aes-operator.md
```diff
@@ -94,60 +94,34 @@ examples of `updateWindow` are:
 The Operator cannot guarantee minute time granularity, so specifying a minute in the crontab
 expression can lead to some updates happening sooner/later than expected.
 
+`installOSS` in an optional field which, if set to `true` install Ambassador API Gateway instead of
```
khussey, line 97: `true` install -> `true`, installs

docs/topics/install/aes-operator.md
```diff
@@ -94,60 +94,34 @@ examples of `updateWindow` are:
...
+    name: ambassador
+  service:
+    ports:
+      - name: http
+        port: 80
+        targetPort: 8080
+    type: NodePort
+```
+
+* Note that, `spec.helmValues.enableAES` should not conflict with the `spec.installOSS` field or else your
```
khussey, line 123: How should it not conflict? And why are there dots in this parameter name? Perhaps you are saying that if `installOSS` is set to true then `enableAES` must be set to false? I'm wondering if there's a better user experience here, e.g., where one of the parameters takes precedence over the other?

docs/topics/install/aes-operator.md
```diff
@@ -94,60 +94,34 @@ examples of `updateWindow` are:
...
+  installation will error out.
```
concaf, line 124:
```suggestion
  installation will error out.
  To install OSS, only `installOSS: true` needs to be set. To migrate from OSS to AES, change `installOSS: true` to `installOSS: false`.
  `helmValues.enableAES` does not need to be changed in any scenario.
  However, if you end up with a configuration where both `installOSS` and `enableAES` are set and they conflict with each other (`installOSS: true` + `enableAES: true` or `installOSS: false` or `enableAES: false`), then AmbassadorInstallation and Ambassador Operator logs will throw related errors.
```

docs/topics/install/aes-operator.md
inercia:
```suggestion
`installOSS` in an optional field which, if set to `true`, install Ambassador API Gateway instead of
```

docs/topics/install/aes-operator.md
```diff
@@ -94,60 +94,37 @@ examples of `updateWindow` are:
 The Operator cannot guarantee minute time granularity, so specifying a minute in the crontab
 expression can lead to some updates happening sooner/later than expected.
 
+`installOSS` in an optional field which, if set to `true`, install Ambassador API Gateway instead of
```
khussey, line 97: install -> installs

docs/topics/install/aes-operator.md
inercia: Resolved.

PR #2715, opened 2020-05-19
First chunk of changes off of quay.io back to DockerHub.

kflynn:
Lots of YAML changes...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,11 +33,11 @@ env:
     # - DBUILD_ARGS="--cache-from ${DEV_REGISTRY}/builder:latest"
 
     # Set from the repository settings:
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -33,11 +33,11 @@ env:
     # - DBUILD_ARGS="--cache-from ${DEV_REGISTRY}/builder:latest"
 
     # Set from the repository settings:
-    # - DOCKER_DEV_REGISTRY=quay.io/datawire-dev (or something different if not datawire/ambassador.git)
+    # - DOCKER_DEV_REGISTRY -- e.g. datawire-dev (or something different if not datawire/ambassador.git)
```
LukeShu, line 36: I like to use strings like `docker.io/datawire-dev`, just to visually root it; so you can understand what that string is.

kflynn: I like that idea -- will do after we get the live stuff updated.

PR #2720, opened 2020-05-20
Repatriate YAML changes for 1.4 back to master

kflynn:
For git consistency. This should be pretty much a no-op.

### DIFF
--- a/.ci/pr-build-website-preview
+++ b/.ci/pr-build-website-preview
@@ -15,6 +15,14 @@
 # limitations under the License
 
 set -o errexit
+
+# Don't bother trying to build the website if GH_TOKEN is unset (as it will
+# be for a community PR).
+if [ -z "$GH_TOKEN" ]; then
+	echo '[pr-build-website-preview] skipping: GH_TOKEN is not set'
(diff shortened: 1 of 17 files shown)


LukeShu:
`git diff $(git merge-base refs/heads/master refs/pull/2720/head) refs/pull/2720/head` is what the GitHub UI is showing, which is a stupid thing to show.

It should be showing `git diff refs/heads/master refs/pull/2720/merge`, which is an empty patch--this is a no-op PR (except for history, of course).

kflynn:
This PR ends up looking really weird, because GitHub is being kind of silly.

If you look at the diffs between the head of `master` and the synthesized merge for this PR (`refs/pull/2720/merge`), the diffs are empty, which is what we'd expect.

GitHub, however, is showing ten files changed, because _it_ is showing the diff between `git merge-base master release/v1.4` (which is 58f55a9f05e5e0f176eaab795549772ac114357e) and the head of `flynn/1.4-to-master`. So the diffs it's showing are correct for that, just... not necessarily very relevant. 

So. Merging.

PR #2721, opened 2020-05-20
Next tranche of DockerHub changes

kflynn:
This set of changes is what I did to test with `DEV_REGISTRY=docker.io/datawiredev`.
I'll flip DEV_REGISTRY in CI after this lands.


LukeShu:
So after this lands, and you flip it; then I can merge master in to all of my PRs, and they should start passing now?

kflynn:
That's the theory, yeah. Landing this without waiting for CI, then flipping.

PR #2722, opened 2020-05-21
Switch to using docker login with just the domain part of the registry

kflynn:


kflynn:
The auth step here is necessary. The build failed because we didn't clone the `ingress-controller-conformance` repo, so I'm doing that now.

kflynn:
(Given a review, I'm going to land this, now that I've copied the `ingress-controller-conformance` repo to DockerHub.)

PR #2728, opened 2020-05-24
Changes from Datawire

kflynn:
- Bind internal services to localhost only
- Fix label-selector matching on `Ingress` support
- Support `regex_rewrite` on `Mapping`s
- The usual `go mod` madness


kflynn:
Landing after @LukeShu and I discussed it yesterday.

PR #2732, opened 2020-05-26
Restore some changes lost in the shuffle

kflynn:
These are the code portion of 1.4.3a, which was never repatriated because it would've caused serious doc problems.


kflynn:
Ignoring "broken" links, since they work fine in the browser...

PR #2734, opened 2020-05-26
1.5 changes from Datawire

kflynn:
1.5 changes from Datawire -- mostly improvements to `edgectl install` and such.


kflynn:
Merging expecting that the conformance-test stuff is a flake. 🤞

PR #2740, opened 2020-05-28
Update for v1.5.0 GA

kflynn:
Ambassador 1.5.0 GA


kflynn:
Merging early since we have a good build on the tag.

PR #2745, opened 2020-05-29
Cakuros/cert manager docs

cakuros:
Cert manager documentation updates.  Changed documentation to reflect cert-manager v0.15's deprecation of old CRD's.  Both DNS-01 challenge and HTTP-01 challenge tested on Ambassador API Gateway and Edge Stack on GKE.

### DIFF
--- a/docs/howtos/cert-manager.md
+++ b/docs/howtos/cert-manager.md
@@ -55,7 +55,7 @@ The DNS-01 challenge verifies domain ownership by proving you have control over 
 
 2. Note the `accessKeyID` and create a `secret` named `prod-route53-credentials-secret` in the cert-manager namespace that has a key value: `secret-access-key` from your AWS IaM credentials.
 
-3. Create and apply a `ClusterIssuer`.  Make sure that the `Issuer` and the `secret` are in the same namespace.
+3. Create and apply a `ClusterIssuer`. 
 
     ```yaml
     ---
@@ -137,7 +137,7 @@ The HTTP-01 challenge verifies ownership of the domain by sending a request for 
           selector: {}
     ```
 
-2. Create and apply a `Certificate`.  Make sure the `Certificate` is in the same namespace as Ambassador.
+2. Create and apply a `Certificate`.
 
     ```yaml
     ---
--- a/docs/howtos/cert-manager.md
+++ b/docs/howtos/cert-manager.md
@@ -1,27 +1,25 @@
 # Cert-Manager and Ambassador Edge Stack
 
-**Note:** Cert-manager release v0.15 [removed legacy CRD support](https://cert-manager.io/docs/installation/upgrading/upgrading-0.14-0.15/).  This document has been updated to use CRD standards specified in v0.15.  For more details about upgrading existing installations of cert-manager, see the [Upgrading](https://cert-manager.io/docs/installation/upgrading/) page
-
-cert-manager is still required for DNS-01 challenges for wildcard domains and when using Ambassador OSS. 
+**Note:** This document assumes cert-manager  v0.15 or greater.   This document has been updated to use CRD standards specified in v0.15.  [Legacy CRD support](https://cert-manager.io/docs/installation/upgrading/upgrading-0.14-0.15/) was removed in cert-manager v0.15, see their [upgrading](https://cert-manager.io/docs/installation/upgrading/) document for more info.
 
 ---
 
 Creating and managing certificates in Kubernetes is made simple with Jetstack's [cert-manager](https://github.com/jetstack/cert-manager). Cert-manager will automatically create and renew TLS certificates and store them in Kubernetes secrets for easy use in a cluster. Ambassador will automatically watch for secret changes and reload certificates upon renewal.
 
-Note: Ambassador Edge Stack will automatically create and renew TLS certificates with the HTTP-01 challenge. You should use cert-manager if you need support for the DNS-01 challenge and/or wildcard certificates.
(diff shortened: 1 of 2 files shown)


docs/howtos/cert-manager.md
```diff
@@ -12,16 +12,19 @@ Note: Ambassador Edge Stack will automatically create and renew TLS certificates
...
-```
-kubectl create ns cert-manager
-kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v0.11.1/cert-manager-no-webhook.yaml
-```
+2. Install cert-manager
 
-**Note:** The resource validation webhook is not mandatory, but if implemented, requires additional configuration.
+    ```
+    kubectl create ns cert-manager
+    helm install cert-manager --namespace cert-manager --set installCRDs=true
```
nbkrause, line 26: I very much disagree with the way jetstack chose to support installing CRDs in the chart. This approach has no safeguards against `helm uninstall` completely removing all of the CRs in the cluster. 

We should use their recommended way of installing the crds by installing them first with a `kubectl apply` and then using the chart to install the rest of cert-manager

docs/howtos/cert-manager.md
```diff
@@ -47,7 +50,7 @@ The DNS-01 challenge verifies domain ownership by proving you have control over
 
 1. Create the IAM policy specified in the cert-manager [AWS Route53](https://cert-manager.readthedocs.io/en/latest/tasks/acme/configuring-dns01/route53.html) documentation.
 
-2. Note the `accessKeyID` and create a secret named `prod-route53-credentials-secret` holding the `secret-access-key`.
+2. Note the `accessKeyID` and create a secret named `cert-manager-route-53` holding the `secret-access-key`.
```
nbkrause, line 53: What is the reason for changing the name of the secret?

docs/howtos/cert-manager.md
```diff
@@ -187,9 +182,9 @@ The HTTP-01 challenge verifies ownership of the domain by sending a request for
 
 4. Create a Mapping for the `/.well-known/acme-challenge/` route.
 
-cert-manager uses an `Ingress` resource to issue the challenge to `/.well-known/acme-challenge/` but, since Ambassador is not an `Ingress`, we will need to create a `Mapping` so the cert-manager can reach the temporary pod.
+    cert-manager uses an `Ingress` resource to issue the challenge to `/.well-known/acme-challenge/` but, since Ambassador is not an `Ingress`, we will need to create a `Mapping` so the cert-manager can reach the temporary pod.
```
nbkrause, line 185: This is no longer true. Ambassador added support the `Ingress` since this was originally written.

It might actually be easier to use the `Ingress` cert-manager creates for this. You should take a look at it and see if you can get it to work.

docs/howtos/cert-manager.md
cakuros: My only concern with this is that it results in versioning.  The URL path that they use includes the version number, i.e. https://github.com/jetstack/cert-manager/releases/download/**v0.15.1**/cert-manager.crds.yaml (emphasis mine).  Additionally, applying it requires you use `--validate=false` which I'm not so sure about either.

docs/howtos/cert-manager.md
cakuros: It's just what I had in my cluster-issuer.yaml when I copied and pasted it, I'll change it back.

docs/howtos/cert-manager.md
nbkrause: You can use this: `https://github.com/jetstack/cert-manager/releases/latest/download/cert-manager.crds.yaml`

cakuros: Aha!  That's what I needed, I tried replacing the version number w/ latest and got 404'd. TYVM.

docs/howtos/cert-manager.md
cakuros: For the time being, I think, I know this works for a fact, I did spend some time last week trying to create an Ingress that worked similarly to the mapping and was struggling.  I also know is that the Ingress that is applied out of the box by `cert-manager` does not work.

docs/howtos/cert-manager.md
```diff
@@ -47,75 +55,74 @@ The DNS-01 challenge verifies domain ownership by proving you have control over
...
     ---
     apiVersion: cert-manager.io/v1alpha2
     kind: Certificate
     metadata:
       name: ambassador-certs
-      namespace: default
+      # cert-manager will put the resulting Secret in the same Kubernetes namespace
+      # as the Certificate. Therefore you should put this Certificate in the same namespace as Ambassador.
+      # eg. if you deploy ambassador to ambassador namespace, you need to change to namespace: ambassador
+      namespace: ambassador
```
nbkrause, line 99: I disagree with this note. The certificate should be created in whichever namespace they want to create a Host (i.e. whichever namespace they are using for the application.) Telling people it needs to be in the Ambassador namespace tells them that Ambassador cannot read tls configuration from other namespaces which is strictly not the case.

**Edit:** I see this was added to the HTTP-01 challenge yaml so you just made sure it was repeated here. Someone made a mistake merging that change in when it was so I am removing both.

docs/howtos/cert-manager.md
```diff
@@ -47,75 +55,74 @@ The DNS-01 challenge verifies domain ownership by proving you have control over
...
     spec:
+      # naming the secret name certificate ambassador-certs is important because
+      # ambassador just look for this particular name
```
nbkrause, line 102: This has not been true for a long time. Where are you getting this information?

**Edit:** I see this was added to the HTTP-01 challenge yaml so you just made sure it was repeated here. Someone made a mistake merging that change in when it was so I am removing both.

PR #2761, opened 2020-06-03
Move service and endpoints processing to a separate processor

impl:
## Description
This change moves service processing from the resource fetcher to a separate processor, `ServiceProcessor`. The functionality should be exactly the same.

## Testing
This passes the automated test suite.

## Tasks That Must Be Done
- [ ] Is this a build change?

### DIFF
--- a/python/ambassador/fetch/fetcher.py
+++ b/python/ambassador/fetch/fetcher.py
@@ -64,7 +64,6 @@ class ResourceFetcher:
 
         self.k8s_processor = DeduplicatingKubernetesProcessor(AggregateKubernetesProcessor([
             CountingKubernetesProcessor(self.aconf, KubernetesGVK.for_knative_networking('Ingress'), 'knative_ingress'),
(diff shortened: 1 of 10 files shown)


kflynn:
Also, yes, I know, we keep making more conflicts for you. Sorry. 😐 Let's talk (on Slack if it's easier) and get this landed. 😉

impl:
There's an explanation of why ClusterIngress was removed on [this commit](https://github.com/datawire/ambassador/pull/2761/commits/3f492e6a123562b6918d13ede1b0f007bbda6909), but gist of it is that ClusterIngress was removed from Knative in version 0.9 and the ingress controller annotation we look for was added in 0.10 so there's no way a compatible Knative installation could be using ClusterIngress even before this change.

kflynn:
...well. After conflicts are resolved. Again. 😂  Sorry about that.

kflynn (approved):
Looks good -- going to land this. Thanks, @impl!

PR #2774, opened 2020-06-09
Multicluster Linkerd docs

kflynn:
I'm sure this needs further work, but it's a start.

### DIFF
--- a/docs/howtos/linkerd2.md
+++ b/docs/howtos/linkerd2.md
@@ -283,7 +283,7 @@ $ kubectl -n $namespace patch svc $service -p='
 metadata:
     annotations:
         mirror.linkerd.io/gateway-name: ambassador
-        mirror.linkerd.io/gateway-ns: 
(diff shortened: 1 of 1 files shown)


docs/howtos/linkerd2.md
```diff
@@ -167,6 +167,146 @@ Congratulations! You're successfully routing traffic to the QOTM application, th
...
+
+4. Configure individual exported services. Adding the following annotations to a service will tell the service to use Ambassador as the gateway:
+
+```bash
+$ kubectl -n $namespace patch svc $service -p='
+metadata:
+    annotations:
+        mirror.linkerd.io/gateway-name: ambassador
+        mirror.linkerd.io/gateway-ns: 
+'
```
richarddli, line 287: shouldn't this also have `ambassador` for the `gateway-ns`?

kflynn: Yes indeed -- fixed, thanks!

PR #2788, opened 2020-06-11
Fix the BLC not actually running

LukeShu:
It was starting up before `python3 -m http.server` started up, and so immediately exited with an error. So block until it's ready.

Also save the `blc.txt` log as an artifact so we can check for if weird things happen, like this.


khussey:
Those are the 5 "broken" links I expected to see, w00t!

Issue #2193, opened 2020-01-14
TCP mapping does not support overriding circuit_breakers max_connections


PierreF:
**Describe the bug**

TCPMapping could not be configured with more than 1024 connections

**To Reproduce**

Create any TCPMapping and try to establish more than 1024 connections. Some connection will be refused.

Try to configure circuit_breakers:
* on TCPMapping: fail because "circuit_breakers" is not allowed in a TCPMapping
* globally: fail, it's not used :(

**Expected behavior**

Allow to override circuit_breakers per TCPMapping.

**Versions (please complete the following information):**
 - Ambassador: quay.io/datawire/aes:1.0.0-rc4
 - Kubernetes environment: minikube v1.6.1 - Kubernetes v1.17.0

**Additional context**

Full steps to reproduce (see https://gist.github.com/PierreF/5835c428a9e8699eb034711ffc6b026b for files)
```
minikube start

kubectl apply -f https://www.getambassador.io/early-access/yaml/aes-crds.yaml && \
kubectl wait --for condition=established --timeout=90s crd -lproduct=aes && \
kubectl apply -f https://www.getambassador.io/early-access/yaml/aes.yaml && \
kubectl -n ambassador wait --for condition=available --timeout=90s deploy -lproduct=aes
```

Create quote & mapping (TCP + HTTP):
```
kubectl apply -f quote.yaml  # This is quote from Quick start, only change is using nodeport
kubectl apply -f tcp-mapping.yaml
kubectl apply -f http-mapping.yaml
```

Get nodePort:
```
minikube service list
```

In my case:
* Minikube IP is 192.168.39.144
* Quote direct TCP port is 32354
* Quote TCP port though ambassador is 30281
* HTTP port of ambassador is 32508

Use provided Python script to open multiple persistent connection:
```
# Passing though ambassador, 1024 connection => all works
python3 open_quote.py --address 192.168.39.144 --port 30281 --count 1024

# Passing though ambassador, 1025 connection => one connection (the last one) fail
python3 open_quote.py --address 192.168.39.144 --port 30281 --count 1025
```

This is neither a limit of FDs on quote, the testing script or ambassador itself, because the following works:
```
# Direct access to quote: works so both quote and the testing script support 2000 open connections
python3 open_quote.py --address 192.168.39.144 --port 32354 --count 2000

# Using Ambassador HTTP
python3 open_quote.py --address 192.168.39.144 --path /backend/ --port 32508 --count 2000
```

PR #2789, opened 2020-06-11
Add circuit_breakers to TCP connections

PierreF:
## Description

Circuit breaker were not supported on TCP mapping, making impossible to have more than 1024 connections per ambassador/envoy. This PR add circuit_breakers to TCP mapping (it reuse exactly what is done in HTTP mapping).

More than 1024 TCP connections is important when the TCP service exposed is using persistent connection that are long lived and you expect more than 1024 client.

## Related Issues
It's the PR for #2193

## Testing

A testing step present in #2193 still apply.

## Tasks That Must Be Done
- [ ] Is this a build change?

### DIFF
--- a/python/ambassador/ir/irtcpmapping.py
+++ b/python/ambassador/ir/irtcpmapping.py
@@ -5,6 +5,7 @@ from ..config import Config
 
 from .irbasemapping import IRBaseMapping
 from .irbasemappinggroup import IRBaseMappingGroup
+from .irhttpmapping import IRHTTPMapping
 from .irtcpmappinggroup import IRTCPMappingGroup
 
 import hashlib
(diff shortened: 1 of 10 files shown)


kflynn:
Also, are you on our [Slack](https://d6e.co/slack)? That might be a good place to talk further about this. 🙂

PierreF:
Fixed your comment and I just joinned Slack.

python/ambassador/ir/irtcpmapping.py
```diff
@@ -5,6 +5,7 @@
 
 from .irbasemapping import IRBaseMapping
 from .irbasemappinggroup import IRBaseMappingGroup
+from .irhttpmapping import IRHTTPMapping
```
kflynn, line 8: This is unneeded now, right?

PierreF:
Added test (that works :)) for TCP mapping & circuit breaker.

kflynn:
Looks really good in general! Thanks in advance for helping clean stuff up for us. 🙂

PR #2791, opened 2020-06-11
Read Knative generation information from the right place

impl:
## Description
We were accidentally trying to get the generation of a Knative ingress from the spec instead of metadata, which caused the observed generation to never update. This change uses the standardized `obj.generation` attribute instead.

## Testing
In addition to not impacting tests, this change fixed the issue for us in our production environment.

## Tasks That Must Be Done
- [ ] Is this a build change?


kflynn:
@concaf Any thoughts before I land this? :wink:

PR #2795, opened 2020-06-12
split projects yaml into separate file

rhs:


kflynn:
(Yes, I know I approved after merge. 🙂 )

PR #2814, opened 2020-06-22
Removed the BLC from travis jobs on release/1.3

alexgervais:


alexgervais:
@khussey, It appears the environment variables holding the container registry credentials changed. I think it is safe for you to merge (with your super-powers) since we don't expect to ship any docker images on the 1.3 branch and keep it alive only for docs.

PR #2826, opened 2020-06-23
New doc Schema for 1.5

hcavalieri:
## Description
Adds Schema.org declarations to doc pages for richer search results.

## Testing
I've tested it manually by changing the file in my local copy of the website repo and ran the resulting schemas through [Google's Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool/u/0/).

### DIFF
--- a/docs/js/doc-page.js
+++ b/docs/js/doc-page.js
@@ -21,6 +21,104 @@ function removeDoubleSlashes(str) {
   return str.replace(/\/{2,}/g, '/');
 }
 
+function getPathFromSlug(str) {
+  return removeDoubleSlashes(`/${str}/`);
+}
+
+function relativeToAbsUrl(url) {
+  if (url.startsWith('http') || url.startsWith('mailto')) {
+    return url;
+  } else
+    return 'https://www.getambassador.io' + removeDoubleSlashes(`/${url}/`);
+}
+
+const getDocPageSchema = ({
+  slug,
+  // an array of { title, slug } relative to parent pages (only if applicable)
+  // This doesn't include the root docs page nor the page itself
+  breadcrumbs = [],
+  // the doc page title (required)
+  title,
+  // the current docs version (required)
+  version,
+  // the current sidebar section (only if applicable)
+  section,
+  isFAQ,
+  isLatest,
+}) => ({
+  '@context': 'http://schema.org',
+  '@type': isFAQ ? 'FAQPage' : 'WebPage',
+  '@id': relativeToAbsUrl(slug),
+  // Not every search engine has assemblyVersion figured out (see below),
+  // we want to separate this doc page from others of different versions 😉
+  version,
+  breadcrumb: {
+    '@type': 'BreadcrumbList',
+    itemListElement: [
+      // Every doc page will have this as the root
+      {
+        '@type': 'ListItem',
+        position: 1,
+        item: {
+          // be sure to update this /latest with the current version
+          '@id': isLatest
+            ? relativeToAbsUrl('/docs/latest/')
+            : relativeToAbsUrl(`/docs/${version}/`),
+          name: 'Ambassador Docs',
+        },
+      },
+      ...breadcrumbs.map((crumb, i) => ({
+        '@type': 'ListItem',
+        // start at position 2
+        position: i + 2,
+        item: {
+          '@id': relativeToAbsUrl(crumb.slug),
+          name: crumb.title,
+        },
+      })),
+      {
+        '@type': 'ListItem',
+        position: 1 + breadcrumbs.length + 1,
+        item: {
+          '@id': relativeToAbsUrl(slug),
+          name: title,
+        },
+      },
+    ],
+  },
+  mainEntity: {
+    '@type': 'APIReference',
+    headline: title,
+    inLanguage: 'en',
+    isPartOf: 'https://www.getambassador.io/#software',
+    // the current docs version
+    assemblyVersion: version,
+    // the sidebar section, if applicable,
+    articleSection: section,
+    // @TODO: check this license
+    license: 'Apache-2.0',
+    // @TODO: check this programmingModel
+    programmingModel: 'unmanaged',
+    // @TODO: adding keywords related to this doc page would be optimal
+    keywords:
+      'Kubernertes,API Gateway,Edge Stack,Envoy Proxy,Kubernetes Ingress,Load Balancer,Identity Aware Proxy,Developer Portal, microservices, open source',
+  },
+});
+
+// Used to get a flat array of *all* links with their corresponding parents
+function flatAllLinks(links, parent) {
+  return links.reduce((acc, cur) => {
+    let link = cur;
+    if (parent) {
+      link = { ...cur, parent };
+    }
+    if (!link.items) {
+      return [...acc, link];
+    }
+    return [...acc, link, ...flatAllLinks(link.items, link)];
+  }, []);
+}
+
 export default ({ data, location }) => {
   const page = data.mdx || {};
   const title =
@@ -50,6 +148,36 @@ export default ({ data, location }) => {
     `/${rest.join('/')}/`,
   )}`;
 
+  // For Schema purposes, we want to make the relationship between pages and their parents explicit
+  // So we flat all links (see flatAllLinks above)
+  const flatLinks = flatAllLinks(docLinks);
+  // Find the current expanded link
+  const expandedLink = flatLinks.find(
+    (l) => getPathFromSlug(l.link) === getPathFromSlug(slug),
+  );
+  let breadcrumbs = [];
+  let section;
+  // And check to see if it has any parent
+  if (expandedLink && expandedLink.parent) {
+    if (expandedLink.parent.link) {
+      // If it does, that's the Schema section and part of the breadcrumbs
+      section = expandedLink.parent.title;
+      breadcrumbs.push({
+        title: expandedLink.parent.title,
+        slug: expandedLink.parent.link,
+      });
+    }
+    // If the parent also has a parent, then that's also part of the breadcrumbs
+    if (expandedLink.parent.parent && expandedLink.parent.parent.link) {
+      breadcrumbs = [
+        {
+          title: expandedLink.parent.parent.title,
+          slug: expandedLink.parent.parent.link,
+        },
+        ...breadcrumbs,
+      ];
+    }
+  }
   return (
     <React.Fragment>
       <Helmet>
@@ -60,6 +188,19 @@ export default ({ data, location }) => {
         {metaDescription && (
           <meta name="description" content={metaDescription} />
         )}
+        <script type="application/ld+json">
+          {JSON.stringify(
+            getDocPageSchema({
+              isLatest: slug.includes('docs/latest'),
+              slug,
+              title,
+              version: versions.version,
+              isFAQ: slug.includes('about/faq/'),
+              breadcrumbs,
+              section,
+            }),
+          )}
+        </script>
       </Helmet>
       <Layout location={location}>
         <Sidebar location={location} prefix="" items={docLinks} />
--- a/docs/js/doc-page.js
+++ b/docs/js/doc-page.js
@@ -89,11 +89,7 @@ const getDocPageSchema = ({
     assemblyVersion: version,
     // the sidebar section / parent link, if applicable,
     articleSection: section,
-    // @TODO: check this license
     license: 'Apache-2.0',
-    // @TODO: check this programmingModel
-    programmingModel: 'unmanaged',
-    // @TODO: adding keywords related to this doc page would be optimal
     keywords:
       'Kubernertes,API Gateway,Edge Stack,Envoy Proxy,Kubernetes Ingress,Load Balancer,Identity Aware Proxy,Developer Portal, microservices, open source',
   },
--- a/docs/js/doc-page.js
+++ b/docs/js/doc-page.js
@@ -25,11 +25,8 @@ function getPathFromSlug(str) {
   return removeDoubleSlashes(`/${str}/`);
 }
 
-function relativeToAbsUrl(url) {
-  if (url.startsWith('http') || url.startsWith('mailto')) {
-    return url;
-  } else
-    return 'https://www.getambassador.io' + removeDoubleSlashes(`/${url}/`);
+function relativeToAbsUrl(slug) {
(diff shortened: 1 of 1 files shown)


docs/js/doc-page.js
```diff
@@ -21,6 +21,104 @@ function removeDoubleSlashes(str) {
   return str.replace(/\/{2,}/g, '/');
 }
 
+function getPathFromSlug(str) {
+  return removeDoubleSlashes(`/${str}/`);
+}
+
+function relativeToAbsUrl(url) {
+  if (url.startsWith('http') || url.startsWith('mailto')) {
```
khussey, line 29: How could the URL argument start with 'http' or 'mailto' if it's relative?

docs/js/doc-page.js
```diff
@@ -21,6 +21,104 @@ function removeDoubleSlashes(str) {
...
+  slug,
+  // an array of { title, slug } relative to parent pages (only if applicable)
+  // This doesn't include the root docs page nor the page itself
+  breadcrumbs = [],
+  // the doc page title (required)
+  title,
+  // the current docs version (required)
+  version,
+  // the current sidebar section (only if applicable)
+  section,
```
khussey, line 45: Do you mean "selection" here?

docs/js/doc-page.js
```diff
@@ -21,6 +21,104 @@ function removeDoubleSlashes(str) {
...
+    // @TODO: check this programmingModel
+    programmingModel: 'unmanaged',
+    // @TODO: adding keywords related to this doc page would be optimal
+    keywords:
+      'Kubernertes,API Gateway,Edge Stack,Envoy Proxy,Kubernetes Ingress,Load Balancer,Identity Aware Proxy,Developer Portal, microservices, open source',
+  },
+});
+
+// Used to get a flat array of *all* links with their corresponding parents
+function flatAllLinks(links, parent) {
```
khussey, line 109: Should this be called "flat**ten**AllLinks"?

docs/js/doc-page.js
```diff
@@ -50,6 +148,36 @@ export default ({ data, location }) => {
     `/${rest.join('/')}/`,
   )}`;
 
+  // For Schema purposes, we want to make the relationship between pages and their parents explicit
+  // So we flat all links (see flatAllLinks above)
```
khussey, line 152: Flat**ten** all links?

docs/js/doc-page.js
```diff
@@ -21,6 +21,104 @@ function removeDoubleSlashes(str) {
...
+}
+
+const getDocPageSchema = ({
+  slug,
+  // an array of { title, slug } relative to parent pages (only if applicable)
+  // This doesn't include the root docs page nor the page itself
+  breadcrumbs = [],
+  // the doc page title (required)
+  title,
+  // the current docs version (required)
```
khussey, line 42: Is this the currently **selected** docs version?

docs/js/doc-page.js
```diff
@@ -60,6 +188,19 @@ export default ({ data, location }) => {
         {metaDescription && (
           <meta name="description" content={metaDescription} />
         )}
+        <script type="application/ld+json">
+          {JSON.stringify(
+            getDocPageSchema({
+              isLatest: slug.includes('docs/latest'),
```
khussey, line 194: Why are we calling out whether the page is "latest" here? Does it matter that pages for the 1.5 version are actually "latest" (the content is identical) but their slugs don't include 'docs/latest' (they include 'docs.1.5' instead)?

docs/js/doc-page.js
hcavalieri: This is careless copy and pasting from the rest of the site where there's the possibility that a CMS-filled piece of content have both a relative or absolute URL. I'll remove it to keep it tidy, nice catch!

docs/js/doc-page.js
hcavalieri: Absolutely, sorry the misspell 😅

docs/js/doc-page.js
hcavalieri: Hm... maybe we've conflicting definitions? What I meant by "section" is where in the sidebar is this link located. If it's nested under a specific "section", such as "Core Concepts" under "Running and Using Ambassador", then its parent should be included in the schema `breadcrumbs`.

Does this make sense?

docs/js/doc-page.js
hcavalieri: This is merely for defining the `@id` of the first breadcrumb item in the Schema definition, which is the main page for the current docs. As the website has both `/docs/latest` and `/docs/1.5` live, it's important to make this distinction to prevent mixing both.

That said, we can get the main page's URL through the current page's slug, with something like:

`relativeToAbsUrl(slug.match(/\/docs\/[\d\.(latest)]*/)[0])`

docs/js/doc-page.js
khussey: Ah, OK. Yes that makes sense.

docs/js/doc-page.js
```diff
@@ -21,6 +21,94 @@ function removeDoubleSlashes(str) {
...
+  // the current sidebar section (only if applicable)
+  section,
+  isFAQ,
+}) => ({
+  '@context': 'http://schema.org',
+  '@type': isFAQ ? 'FAQPage' : 'WebPage',
+  '@id': relativeToAbsUrl(slug),
+  // Not every search engine has assemblyVersion figured out (see below),
+  // we want to separate this doc page from others of different versions 😉
+  version,
```
LukeShu, line 50: This will set the schema's `version` to the version of Ambassador, right? That wouldn't be quite right, would it? The version here should be the web-page's version, not the Ambassador release it describes, right?

docs/js/doc-page.js
```diff
@@ -50,6 +138,36 @@ export default ({ data, location }) => {
...
+  // So we flatten all links (see flattenLinks above)
+  const flatLinks = flattenLinks(docLinks);
+  // Find the current expanded link
+  const expandedLink = flatLinks.find(
+    (l) => getPathFromSlug(l.link) === getPathFromSlug(slug),
+  );
+  let breadcrumbs = [];
+  let section;
+  // And check to see if it has any parent
+  if (expandedLink && expandedLink.parent) {
```
LukeShu, line 151: Pages can be arbitrarily deep, right? Shouldn't this be a loop? Instead of just hard-coding 2 iterations?

docs/js/doc-page.js
```diff
@@ -21,6 +21,94 @@ function removeDoubleSlashes(str) {
...
+  version,
+  // the current sidebar section (only if applicable)
+  section,
+  isFAQ,
+}) => ({
+  '@context': 'http://schema.org',
+  '@type': isFAQ ? 'FAQPage' : 'WebPage',
+  '@id': relativeToAbsUrl(slug),
+  // Not every search engine has assemblyVersion figured out (see below),
+  // we want to separate this doc page from others of different versions 😉
```
LukeShu, line 49: Isn't it already separated by having a different `@id`?

docs/js/doc-page.js
hcavalieri: You mean `1.5.4` vs. `1.5`, for example? If so, absolutely agree!

How can we get the current web-page's version if the URL is `/latest`, though? With `/docs/1.4` and other numbered URLs it's pretty straightforward, but with `latest` I don't see a way around this.

Maybe we could get `versions.yml`'s `version` and get only the major and minor? In `1.5.4` we'd get only the `1.5` portion, for example. This would get complicated if there was ever any non-numbered version, though.

Thanks for your time, Luke!

docs/js/doc-page.js
hcavalieri: Indeed it's already separated, this is just an extra layer of cautiousness. If we didn't have this, search engines would understand the same page under different URLs (such as `/docs/latest/topics/concepts/` and `/docs/1.3/topics/concepts/`) as plain content duplication, which is treated quite severely.

With this property we're basically saying it's okay if most pages look mostly the same, they're part of the documentation for different versions of Ambassador. This can also potentially help search results for legacy users looking for "Service Preview Ambassador 1.3", for example. Whether or not Google and Bing will index the content this way is out of our control, though.

How do you suggest rephrasing this comment to make this distinction clearer, Luke? :)

docs/js/doc-page.js
hcavalieri: Done!

hcavalieri:
I've updated the code with @LukeShu 's suggestions and took the opportunity to abstract the complex Schema logic into a custom React hook (`useDocSEO`) to make organization easier. In the future I'd suggest moving this hook into its own file to make `doc-page.js` more approachable for contributors.

If you're both satisfied with the end result, let me know and I'll update the other branches for other releases 😉 

**PS:** This was my first ever thorough code review. As someone who rarely works with teams, your feedback has been immensely valuable and the experience was very enriching for me. Thank you so much for the patience and level of detail, I'm striving to meet your expectations!

docs/js/doc-page.js
hcavalieri: My final solution can be found here:
https://github.com/datawire/ambassador/blob/48ae642f45883fe4e68f84401ab6fe2190cf7701/docs/js/doc-page.js#L161-L163

What do you think??

docs/js/doc-page.js
LukeShu: Perhaps:

Search engines normally penalize sites for duplicating content; we want to communicate to the search engine "we're hosting multiple versions of the docs, please don't penalize us for having content that is duplicated in multiple versions". Principally, we do that with `assemblyVersion` (below), but some search engines don't have `assemblyVersion` (the version of the software being described) figured out, so, we set `version` (the version of the web page describing the software) too.

docs/js/doc-page.js
LukeShu: I meant the opposite of that.

I mean that
- The web page is describing 1.5.4
- But then someone improves the docs; not doing a release.
- That's a new "version" of the web page...
- ... which is what I understand `version` to mean at this place in the schema

That is: It is my understanding that the version here should be *more* granular than `versions.version`, not less.

(That understanding could very well be mistaken.)

Based on that understanding, I suspect the right thing to do would be to use `git describe --tags` as the doc version, but that's probably not worth the complexity at this point (maybe after we simplify how the versions get updated).

hcavalieri: Ohhh gotcha! We define the version in the schema in two different instances:

- the [`assemblyVersion`](https://schema.org/assemblyVersion) of the `APIReference`
  - according to Schema.org, `assemblyVersion` stands for _"Associated product/technology version. e.g., .NET Framework 4.5."_, so in this case it refers to the Ambassador version, not docs version
  - as we only have doc segments for minor releases (`1.5`) and not for patches ( `1.5.4`), I'm thinking only showing the minor version in `assemblyVersion` is the ideal, but this is open for debate :)
- the [`version`](https://schema.org/version) of the `WebPage` (or `FAQPage`)
  - according to Schema.org, this `version` represents _"The version of the CreativeWork embodied by a specified resource."_
  - so you're right, **it's the _version of the docs_, not the software**
  - that said, the goal here is mainly to show search engines there's a difference between the 1.5 and 1.4 docs version, but I don't think they're as granular in their understanding to make it worth the trouble of precisely explaining the docs version here
  - if we can't get to a solid answer, we could remove this property as the `@id` and the `mainEntity` of the pages are already different, let me know what you think

Thanks for spotting this nuance, @LukeShu !

docs/js/doc-page.js
hcavalieri: Sounds good! The discussion above on the `version` property will impact the final result of this comment, but I've copied it over locally in case we go through with it. Waiting on your answer above 😉

docs/js/doc-page.js
```diff
@@ -21,21 +21,90 @@ function removeDoubleSlashes(str) {
...
+function relativeToAbsUrl(slug) {
+  return 'https://www.getambassador.io' + getPathFromSlug(slug);
+}
 
-  const metaDescription = page.frontmatter
-    ? page.frontmatter.description
-    : null;
+function getMainDocsUrl(slug) {
+  return relativeToAbsUrl(slug.match(/\/docs\/[\d\.(latest)]*/)[0]);
+}
```
LukeShu, line 34: I guess this is the part that's erroring? It will throw a type error if the slug doesn't match that regex.

LukeShu: That regex doesn't make sense to me.

`return relativeToAbsUrl(getPathFromSlug(slug).match(/\/docs\/[^\/]+/)[0])` perhaps?

hcavalieri: There, I've made it clearer, commented it and fixed the build errors :D

hcavalieri:
@LukeShu build is fixed! Let me know if you need anything else to merge this :)

docs/js/doc-page.js
```diff
@@ -30,7 +30,11 @@ function relativeToAbsUrl(slug) {
 }
 
 function getMainDocsUrl(slug) {
-  return relativeToAbsUrl(slug.match(/\/docs\/[\d\.(latest)]*/)[0]);
+  // All doc pages slugs follow the pattern /docs/{VERSION}/{PATH}
+  // This regex extracts the /docs/{VERSION}/ part to find the current page's main docs slug
+  const mainDocsSlug = getPathFromSlug(slug).match(/\/docs\/[\d\.(latest)]*/)
```
LukeShu, line 35: Let me rephrase: I don't think that regex does what you think it does.

```console
$ node
Welcome to Node.js v14.4.0.
Type ".help" for more information.
> "/docs/()tassel()/foo".match(/\/docs\/[\d\.(latest)]*/)
[
  '/docs/()tassel()',
  index: 0,
  input: '/docs/()tassel()/foo',
  groups: undefined
]
```

hcavalieri: This is exactly the expected behavior: `()tassel()` is the docs version in the `/docs/()tassel()/foo` URL, so the main docs path would be `getPathFromSlug('/docs/()tassel()')`, which is the first element of the resulting `match` operation (`match(...)[0]`) 😄

Does this make sense? I've tested it manually with all docs versions, but maybe we can create a new PR adding unit tests?

LukeShu: So that's not what a glance at the regex suggests.  If we don't care specifically about the string `latest`, it probably shouldn't be confusingly in the regex.  Perhaps the simpler `/\/docs\/[^/]*/`?

PR #2837, opened 2020-06-30
New doc Schema (v1) for master

hcavalieri:
## Description
Adds Schema.org declarations to doc pages for richer search results.

## Testing
I've tested it manually by changing the file in my local copy of the website repo and ran the resulting schemas through [Google's Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool/u/0/).

## This is a provisional PR

This is related to #2826 and although it's highly positive for the docs SEO right off the batch, it's not a final solution. Eventually we'll want to go back to polish the code based on the following questions:

- [ ] Define the license in the Schema. Right now I'm using Apache-2.0 following this repo's license, can we go through with it or is the subject more nuanced?
- [ ] Can we use "unmanaged" as the ["programmingModel"](https://schema.org/programmingModel) in  the Schema?
- [ ] Can we move parts of the code to another file to make it neater? `docs/js/doc-page.js` is currently 261 lines long.

### DIFF
--- a/docs/js/doc-page.js
+++ b/docs/js/doc-page.js
@@ -21,6 +21,104 @@ function removeDoubleSlashes(str) {
   return str.replace(/\/{2,}/g, '/');
 }
 
+function getPathFromSlug(str) {
+  return removeDoubleSlashes(`/${str}/`);
+}
+
+function relativeToAbsUrl(url) {
+  if (url.startsWith('http') || url.startsWith('mailto')) {
+    return url;
+  } else
+    return 'https://www.getambassador.io' + removeDoubleSlashes(`/${url}/`);
+}
+
+const getDocPageSchema = ({
+  slug,
+  // an array of { title, slug } relative to parent pages (only if applicable)
+  // This doesn't include the root docs page nor the page itself
+  breadcrumbs = [],
+  // the doc page title (required)
+  title,
+  // the current docs version (required)
+  version,
+  // the current sidebar section (only if applicable)
+  section,
+  isFAQ,
+  isLatest,
+}) => ({
+  '@context': 'http://schema.org',
+  '@type': isFAQ ? 'FAQPage' : 'WebPage',
+  '@id': relativeToAbsUrl(slug),
+  // Not every search engine has assemblyVersion figured out (see below),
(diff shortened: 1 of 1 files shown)


docs/js/doc-page.js
```diff
@@ -60,6 +188,19 @@ export default ({ data, location }) => {
         {metaDescription && (
           <meta name="description" content={metaDescription} />
         )}
+        <script type="application/ld+json">
+          {JSON.stringify(
+            getDocPageSchema({
+              isLatest: slug.includes('docs/latest'),
+              slug,
+              title,
+              version: versions.version,
```
LukeShu, line 197: This will set the schema's `version` to the version of Ambassador, right? That wouldn't be quite right, would it; the version there should be the web-page's version, not the Ambassador release it describes, right?

docs/js/doc-page.js
```diff
@@ -50,6 +148,36 @@ export default ({ data, location }) => {
...
+    if (expandedLink.parent.link) {
+      // If it does, that's the Schema section and part of the breadcrumbs
+      section = expandedLink.parent.title;
+      breadcrumbs.push({
+        title: expandedLink.parent.title,
+        slug: expandedLink.parent.link,
+      });
+    }
+    // If the parent also has a parent, then that's also part of the breadcrumbs
+    if (expandedLink.parent.parent && expandedLink.parent.parent.link) {
```
LukeShu, line 171: Pages can be arbitrarily deep, right? Shouldn't this be a loop? To check for great^n-grandparents and not just grandparents?

docs/js/doc-page.js
```diff
@@ -21,6 +21,104 @@ function removeDoubleSlashes(str) {
...
+  // the current sidebar section (only if applicable)
+  section,
+  isFAQ,
+  isLatest,
+}) => ({
+  '@context': 'http://schema.org',
+  '@type': isFAQ ? 'FAQPage' : 'WebPage',
+  '@id': relativeToAbsUrl(slug),
+  // Not every search engine has assemblyVersion figured out (see below),
+  // we want to separate this doc page from others of different versions 😉
```
LukeShu, line 53: Isn't it already separated by having a different `@id`?

PR #2865, opened 2020-07-17
Correct prometheus operator install information

nbkrause:
From a conversation in Slack, it looks like #2863 was not correct. 

1. `apiextensions.k8s.io/v1` is added in 1.16

2. The sed command was fat fingered and left out another needed string replacement.

Sorry for the confusion this caused. I am glad it was caught in the `pre-release` docs.


richarddli:
This should be cherry-picked onto `release/v1.5`.

PR #2904, opened 2020-07-31
Service Preview org

richarddli:
- make the introduction the first page that people see when they get to the SP docs
- delete the ref material from the SP docs
- try to clarify the use case of SP

### DIFF
--- a/docs/topics/using/edgectl/index.md
+++ b/docs/topics/using/edgectl/index.md
@@ -2,31 +2,13 @@
 
 One of the challenges in adopting Kubernetes and microservices is the development and testing workflow. Creating and maintaining a full development environment with many microservices and their dependencies is complex and hard.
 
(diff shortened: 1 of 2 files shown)


docs/topics/using/edgectl/index.md
```diff
@@ -2,31 +2,13 @@
 
 One of the challenges in adopting Kubernetes and microservices is the development and testing workflow. Creating and maintaining a full development environment with many microservices and their dependencies is complex and hard.
 
-Service Preview addresses this challenge by connecting your CI system or local development infrastructure to the Kubernetes cluster, and dynamically routing specific requests to your local environment.
+Service Preview, based on [Telepresence](https://www.telepresence.io), enables different developers to access different virtual versions of the same microservice. These virtual versions are deployed on your CI system or local development infrastructure, enabling fast development and testing workflows.
```
khussey, line 5: Would "run" make more sense here than "access"?

Issue #2889, opened 2020-07-27
MicroCD/Projects does not work OOTB with `edgectl install`.


cakuros:
## Background/Summary

When following the `edgectl install` process and adding `projects.yaml`, navigating to the projects tab in the EPC shows a service-accounts permissions error.  It is likely that the RBAC portions for Projects are not being correctly created during this process.

## Issue/Request

Running `kubectl apply -f https://getambassador.io/yaml/projects.yaml` should work without errors after running edgectl install.

## Replicator

A step-by-step replicator for the issue. Start with how and where to install Ambassador and end with the manifestation of the bug.

1. Kubernetes cluster info
   * 1.16
   * GKE

2. Ambassador install info
   * Ambassador version 1.6.0 via `edgectl install`.
   * Edgectl version 1.6.0

3. Ambassador configuration
   * `kubectl apply -f https://getambassador.io/yaml/projects.yaml`
   * Single AES deployment

4. Example request that shows the replication
   1. Run `edgectl install` on fresh cluster
   2. Run `kubectl apply -f https://getambassador.io/yaml/projects.yaml`
   3. Run `edgectl login -n ambassador {{AMBASSADOR_HOST}}

![Screenshot from 2020-07-22 12-28-15](https://user-images.githubusercontent.com/63562580/88557849-c14b5480-cff8-11ea-9e4d-17c685fe9bc7.png)

## Notes

Need to check and see if this can be resolved by applying the correct `ClusterRole` and `ClusterRoleBinding` to the service-account.  Will report back after checking.

gz#457

cakuros:
Need to add rbac and restart the Ambassador pod.
```yaml
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRole
metadata:
  name: ambassador-projects
  labels:
    app.kubernetes.io/name: ambassador
    helm.sh/chart: ambassador-6.5.1
    app.kubernetes.io/instance: ambassador
    app.kubernetes.io/managed-by: Helm
    product: aes
rules:
- apiGroups: [""]
  resources: [ "secrets", "services" ]
  verbs: [ "get", "list", "create", "patch", "delete", "watch" ]
- apiGroups: ["apps"]
  resources: [ "deployments" ]
  verbs: [ "get", "list", "create", "patch", "delete", "watch" ]
- apiGroups: ["batch"]
  resources: [ "jobs" ]
  verbs: [ "get", "list", "create", "patch", "delete", "watch" ]
[...]

PR #2910, opened 2020-08-04
Project quickstart, added RBAC manifest for manual install.

cakuros:
It's not immediately clear that after applying `kubectl apply -f https://getambassador.io/yaml/projects.yaml` that you also have to update the RBAC and restart the pods for it to take effect.  In the most apparant scenario, running `edgectl install` and then `kubectl apply -f https://getambassador.io/yaml/projects.yaml` results in the `ambassador` service account not getting the proper RBAC updates.  This would also likely occur if someone ran `helm install ambassador -n ambassador datawire/ambassador` without setting the registry flag and then applying the `projects` yaml.

Acts as a temporary fix for #2889

### DIFF
--- a/docs/tutorials/projects.md
+++ b/docs/tutorials/projects.md
@@ -29,7 +29,7 @@ the project controller by running the following command:
 kubectl apply -f https://www.getambassador.io/yaml/projects.yaml
 ```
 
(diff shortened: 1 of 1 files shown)


cakuros:
Tested on edgectl install, manual yaml, and helm.  Doesn't really do anything on the manual yaml install, but it also doesn't hurt it.

PR #2961, opened 2020-09-04
Link to the new best list of aws lb annotations

nbkrause:
It looks like Kubernetes recently removed third-party documentation from their docs without pointing to any other source for the information on that page. https://github.com/kubernetes/website/pull/23517

This is the next best source for that info I could find.


khussey (approved):
LGTM, thanks

PR #2970, opened 2020-09-09
Minor addition to Timeouts + Module page Formatting

cakuros:
This started out as a minor insert of an example into Modules + timeouts page and turned into a formatting update for the Modules page.

Modules is now split into 2 categories:
1) Global mapping config examples
2) Config table enumeration/examples

Part 2 is ordered by appearance in the table, and each header follows the same format of "Title (`table_entry`)".

I added `diagnostics` to the table and moved it into the config table example since I thought it fit there better than where it was.  I also condensed the Liveness and Readiness probes into 1 entry, instead of having them split apart across the document.

### DIFF
--- a/docs/topics/running/ambassador.md
+++ b/docs/topics/running/ambassador.md
@@ -196,7 +196,7 @@ readiness_probe:
   rewrite: /backend/health
 ```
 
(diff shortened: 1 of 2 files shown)


richarddli (approved):
LGTM.

PR #2976, opened 2020-09-14
Fix ugly link in sp install

nbkrause:
Fixes ugly link in docs/topics/using/edgectl/service-preview-install/


khussey (approved):
LGTM, thanks!

PR #2987, opened 2020-09-24
Add documentation for running SP with Istio

nbkrause:
Service preview can intercept request to services in an Istio mesh. 

There are some configuration changes that need to be made to support this. 

1. The `istio-proxy` **must not** run alongside service preview components

2. The `traffic-agent` **must** set `AMBASSADOR_ENVOY_BASE_ID: "1"` so it can
run in the same pod as the `istio-proxy`.

### DIFF
--- a/docs/topics/using/edgectl/service-preview-reference.md
+++ b/docs/topics/using/edgectl/service-preview-reference.md
@@ -73,6 +73,9 @@ spec:
       app: telepresence-proxy
   template:
     metadata:
+      annotations:
+        consul.hashicorp.com/connect-inject: 'false'
+        sidecar.istio.io/inject: 'false'
       labels:
         app: telepresence-proxy
     spec:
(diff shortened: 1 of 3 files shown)


nbkrause:
Shoot I need to rebase this 🤦

docs/topics/using/edgectl/service-preview-reference.md
```diff
@@ -669,6 +683,39 @@ spec:
 
 **Note**: If you already had an active Edge Control Daemon connection to the cluster, you must reconnect to the cluster for the Edge Control Daemon to detect the change to the Host resource. This limitation will be removed in the future.
 
+## Service Mesh Support
+
+Much like a service mesh, the Service Preview `traffic-agent` runs as a sidecar to the container running your application. Luckily, due to the magic of service mesh Pod networking, Service Preview will run alongside a service mesh without any changes needed to be made on your part.
```
richarddli, line 688: Delete "Much like a service mesh" because a service mesh isn't just the sidecar, it's also other stuff. So we can avoid the comparison.

richarddli (changes requested):

richarddli (approved):

PR #2990, opened 2020-09-28
Add HOST_IP env var to Ambassador deployments

nbkrause:
Add HOST_IP env var to Ambassador deployments

    HOST_IP is referenced in a couple locations in our documentation for as
    a resolvable env var that allows Ambassador to dynamically target a
    service running as a `DaemonSet` (like Consul server or DataDog collector).

    This env var is available by default with the YAML deploys but not other
    methods. This commit adds it to the other AES deployments

### DIFF
--- a/docs/yaml/aes.yaml
+++ b/docs/yaml/aes.yaml
@@ -316,6 +316,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: HOST_IP
+          valueFrom:
+            fieldRef:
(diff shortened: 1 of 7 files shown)


kflynn (approved):
Hit it!

LukeShu:
I glanced at this PR and didn't catch it, but CI caught it merging this in to apro.git:

Most of these files have a comment at the top saying "# GENERATED FILE: DO NOT EDIT."--the proper way to have edited them is to edit `k8s-aes-src/*.yaml` and then run `make update-yaml`.

PR #2994, opened 2020-09-29
Repatriate 1.7.3 in to master

LukeShu:
1. Unlike most of the repatriation PRs for .Z releases, this also includes apro.git master
2. There's actually some complex merges going on here, so probably give it some more scrutiny than usual for repatriation PRs.

I think it makes sense to review+merge https://github.com/datawire/ambassador/pull/2995 / https://github.com/datawire/apro/pull/1902 first, then flush the GitHub cache, then review this, to keep separate things separate.  Or you can just merge this, and that will automatically merge #2995, since this is on top of #2995.


kflynn (approved):
And this also looks OK, I think.

PR #2995, opened 2020-09-29
Sync from apro.git master

LukeShu:
Normally we only do this when doing an X.Y.0 release, but it's just easier for conflicts if we do it now.  I could have rolled this in to #2994, but let's separate it so it's more feasible to review.


kflynn (approved):
Hit it.

PR #2998, opened 2020-10-01
gold file generation: emit kubectl stderr, so it's easier to debug

jwm:
## Description
I was regenerating gold files as part of another PR, but all the gold file copies failed. This stops the gold file generator from hiding stderr, so it's easier to troubleshoot.

In my case, the problem is that gold file generation doesn't honor `DEV_KUBECONFIG`, and auth was failing against the (different) cluster described in my normal `KUBECONFIG`. It feels like fixing that is as easy as adding `KUBECONFIG=$DEV_KUBECONFIG` to the `kubectl` invocation here, but I wasn't sure if there were any other workflows that depend on the existing behavior.

### DIFF
--- a/builder/copy-gold.sh
+++ b/builder/copy-gold.sh
@@ -9,7 +9,7 @@ copy_gold () {
 	local pod="$1"
 	local namespace="${2:-default}"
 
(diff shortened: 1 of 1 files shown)


kflynn (approved):
Let's do it. 🙂

PR #3010, opened 2020-10-08
1.7 is no longer latest.

kflynn:
Update doc links.


LukeShu (changes requested):
Doesn't `aes-pages.yml` also need updated?

LukeShu (approved):

PR #3022, opened 2020-10-15
Fix outside contributor CI

LukeShu:
## Description

This should fix the failing CI that all outside contributors are currently seeing.

As usual for me, I suggest reviewing this commit-by-commit.  This includes the things that are currently on `apro.git` `release/v1.8` but haven't yet made their way to `ambassador.git` `master`, because some of those changes are relevant to this PR. Only the last 3 commits are new.  That last commit isn't actually related to OSS CI, but it's something that came up when Flynn was validating that this didn't break macOS.

## Tasks That Must Be Done
- [x] Did you update CHANGELOG.md? - no applicable changes
- [x] Did you add or update tests? - not nescessary
- [x] Did you update documentation? - yes, `make help` got updated
- [x] Were there any special dev tricks you had to use to work on this code efficiently? - no
- [x] Is this a build change? - yes
  + [ ] If so, did you test on both Mac and Linux? - no, not macOS (cc @kflynn)

### DIFF
--- a/.ci/kubernaut-claim
+++ b/.ci/kubernaut-claim
@@ -33,8 +33,8 @@ if [ -n "${KUBECEPTION_TOKEN}" ]; then
     echo ${CLAIM_NAME} > ~/kubernaut-claim.txt
     mkdir -p ~/.kube
     export DEV_KUBECONFIG=~/.kube/${CLAIM_NAME}.yaml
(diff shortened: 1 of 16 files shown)


kflynn (approved):
I think this looks OK, and it seems good on my Mac with and without `BASE_REGISTRY` set.

PR #3039, opened 2020-10-28
Guard /metrics from uninitialized ir

markusjevringsesame:
## Description
A missed check causes 1.8.1 to throw an exception. This value is None, for unknown reasons. Other endpoints have this check, so it felt easy to add it here as well.

## Related Issues
https://github.com/datawire/ambassador/issues/3037

## Testing
I built this locally using `make images`, but I did *not* run the tests, as I don't have a cluster that the build can make indiscriminant changes to at the moment. 
Simple testing consisted of deploying the images to our environment and verifying that the exception was not present anymore, which it was not.
This solves *one* of the problems in the linked issue, but it does not have any impact on the other problem, the cluster reconfiguration, which remains.

## Tasks That Must Be Done
- [X] Did you update documentation?
- [X] Were there any special dev tricks you had to use to work on this code efficiently?
  + [X] Did you add them to DEVELOPING.md?
- [ ] Is this a build change?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ Note that Ambassador Edge Stack `External` Filters already unconditionally use t
 
 ## Next Release
 
+- Bugfix: The /metrics endpoint will no longer break if invoked before configuration is complete
 - Bugfix: Ambassador will no longer mistakenly post notices regarding `regex_rewrite` and `rewrite` directive conflicts in `Mapping`s due to the latter's implicit default value (`/`).
 - Feature: Support configuring the gRPC Statistics Envoy filter to enable telemetry of gRPC calls (see the `grpc_stats` configuration flag)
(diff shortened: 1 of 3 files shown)


markusjevringsesame:
As this isn't a release, I wasn't sure if I should update the changelog or if that's something that's only done on release.
Also I don't have linux or mac, so I can't test there.
And finally, no tests added or updated.
 
That's why those check-boxes remain unchecked.

kflynn (approved):
Taking this, thanks! And if that uppercase-hostname thing is a bug in our build scripts, please file an issue for that, too?

markusjevringsesame:
@kflynn Thanks! It's hard for me to verify if it's a WSL issue or a general issue, since I can only test with WSL, but I'll create the issue and someone else can verify it.

PR #3048, opened 2020-11-05
Add external filters HOWTO

mattmcclure-dw:
As discussed:

- Created external filter HOWTO by copying Auth HOWTO and swapping AuthServer YAML for Filter YAML
- Changed references from AuthService to Filter
- Added links

### DIFF
--- a/docs/howtos/ext-filters.md
+++ b/docs/howtos/ext-filters.md
@@ -87,7 +87,7 @@ Note that the `READY` field says `1/1` which means the pod is up and running.
 
 ## 2. Configure Ambassador Authentication
 
-Once the auth service is running, we need to tell Ambassador about it. The easiest way to do that is to map the `example-auth` service with the following:
+Once the auth service is running, we need to tell Ambassador about it. The easiest way to do that is to first map the `example-auth` service with the following `Filter`:
 
 ```yaml
 ---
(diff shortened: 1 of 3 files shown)


docs/howtos/ext-filters.md
```diff
@@ -0,0 +1,170 @@
...
+Wait for the pod to be running before continuing. The output of `kubectl get pods` should look something like
+
+```console
+$ kubectl get pods
+NAME                            READY     STATUS    RESTARTS   AGE
+example-auth-6c5855b98d-24clp   1/1       Running   0          4m
+```
+Note that the `READY` field says `1/1` which means the pod is up and running.
+
+## 2. Configure Ambassador Authentication
```
nbkrause, line 88: @mattmcclure-dw You need a `FilterPolicy` as well to actually configure Ambassador to use this `Filter`.

Something like this:

```yaml
---
kind: FilterPolicy
metadata:
  name: authentication
spec:
  rules:
  - host: "*"
    path: /*
    filters:
    - authentication
```
will enable that `Filter` to run on all requests to all hosts.

docs/howtos/ext-filters.md
```diff
@@ -0,0 +1,170 @@
...
+This configuration tells Ambassador about the filter, notably that it needs the `/extauth` prefix, and that it's OK for it to pass back the `x-qotm-session` header. Note that `path_prefix` and `allowed_headers` are optional.
+
+If the auth service uses a framework like [Gorilla Toolkit](http://www.gorillatoolkit.org) which enforces strict slashes as HTTP path separators, it is possible to end up with an infinite redirect where the filter's framework redirects any request with non-conformant slashing. This would arise if the above example had ```path_prefix: "/extauth/"```, the filter would see a request for ```/extauth//backend/get-quote/``` which would then be redirected to ```/extauth/backend/get-quote/``` rather than actually be handled by the authentication handler. For this reason, remember that the full path of the incoming request including the leading slash, will be appended to ```path_prefix``` regardless of non-conformant slashing.
+
+## 3. Test Authentication
+
+If we `curl` to a protected URL:
+
+```shell
+$ curl -Lv $AMBASSADORURL/backend/get-quote/
```
nbkrause, line 117: Re: My comment about `FilterPolicy` above.

Alternatively, you could have the docs show how cool™ it is that you can select which requests get authenticated.

You could have the `FilterPolicy` only run the `Filter` on requests to `path: /backend/get-quote/` to show how you can have authentication on different routes.

```yaml
---
kind: FilterPolicy
metadata:
  name: authentication
spec:
  rules:
  - host: "*"
    path: /backend/get-quote/
    filters:
    - authentication
```

docs/js/doc-links.yml
```diff
@@ -226,6 +226,8 @@
           link: /docs/pre-release/howtos/tracing-datadog
         - title: Zipkin
           link: /docs/pre-release/howtos/tracing-zipkin
+    - title: External Filters
+      link: /docs/pre-release/howtos/ext-filters
```
nbkrause, line 230: The goal of this documentation change was to keep people clicking on "Custom Filters" from being pointed at the `Plugin` `Filter`. I think we should arrange the ToC so that this is under a drop down called "Custom Filters" and have this one and the current `Plugin` `Filter` doc as in that drop down. 

This one should be first. We should also call out why you would need custom filters and why the `External` `Filter` is prefered to the `Plugin` `Filter`.

nbkrause (changes requested):
Great work! I have some comments about the content that needs to be fixed and the strategic placement in the ToC.

Thanks!

docs/howtos/ext-filters.md
mattmcclure-dw: see below

docs/howtos/ext-filters.md
mattmcclure-dw: I added both `FilterPolicy` examples, first with all hosts/path, then with the how cool™ example, hows's that look?

docs/js/doc-links.yml
mattmcclure-dw: I rearranged the ToC, lines 219-224, I think that matches what you described.
For calling out using `External` over `Plugin` `Filter`s, are you thinking I add a line to the top of the `Plugin` `Filter` doc saying something like "Yeah you can do this but it's a pain in the butt to build and maintain, what you really want is an `External` `Filter`, trust us"?  Would we even provide support if someone was having a problem writing their own `Plugin` `Filter`?

PR #3053, opened 2020-11-08
Don't allow mixing host_redirect and non-host_redirect Mappings in the same group

kflynn:
If an `IRHTTPMappingGroup` has some `Mapping`s with `host_redirect` set, and some without, Bad Things Happen™: notably, Ambassador could crash with a Python exception on a missing `IRCluster` if the `host_redirect` `Mapping` is processed first. This change disallows that situation.

(In a future release, we should probably support doing canaries that span `host_redirect` and non-`host_redirect` `Mapping`s.)

This was tested using a `snapshot.yaml` that exhibited this bug. Unit test coming soon.


esmet (approved):
Looks good

acookin (approved):

LukeShu (approved):
Seems sane.

PR #3054, opened 2020-11-09
host_redirect fix improvements

kflynn:
- Make the `host_redirect` fix work irrespective of which `Mapping` comes "first"
- Woo unit test! 🙂

### DIFF
--- a/python/tests/test_host_redirect_errors.py
+++ b/python/tests/test_host_redirect_errors.py
@@ -1,0 +1,185 @@
+from typing import Any, Dict, List, Tuple
+
+import difflib
+import json
+import logging
+import os
+import random
+import sys
+import yaml
+
+import pytest
+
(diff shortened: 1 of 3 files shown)


acookin (approved):

acookin (approved):

python/tests/test_host_redirect_errors.py
```diff
@@ -0,0 +1,179 @@
+from typing import Any, Dict, List, Tuple
```
LukeShu, line 1: > Y'know, it's OK for tests to have comments, too. :slightly_smiling_face:

kflynn: ..._something something_ hoist _something_ petard _something something_...

Yeah. You're right. I'll fix it soon. 🙂

PR #3055, opened 2020-11-09
Fix broken ambassador_id link

mattmcclure-dw:
Fixed broken link to info about ambassador_id.  Also removed confusing link text in YAML example and replaced with "see below" since I can't find a way to make a link within a comment within a code block.

### DIFF
--- a/docs/topics/running/ambassador.md
+++ b/docs/topics/running/ambassador.md
@@ -13,7 +13,7 @@ metadata:
   name:  ambassador
 spec:
 # Use ambassador_id only if you are using multiple ambassadors in the same cluster.
-# For more information: ../../running#ambassador_id.
+# See below for more information.
   # ambassador_id: "<ambassador_id>"
   config:
 # Use the following table for config fields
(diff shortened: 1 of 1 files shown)


docs/topics/running/ambassador.md
```diff
@@ -23,7 +23,7 @@ spec:
 | :----- | :----- | :-- |
 | `add_linkerd_headers` | Should we automatically add Linkerd `l5d-dst-override` headers? | `add_linkerd_headers: false` |
 | `admin_port` | The port where Ambassador's Envoy will listen for low-level admin requests. You should almost never need to change this. | `admin_port: 8001` |
-| `ambassador_id` | Use only if you are using multiple ambassadors in the same cluster. [Learn more](#ambassador_id). | `ambassador_id: "<ambassador_id>"` |
+| `ambassador_id` | Use only if you are using multiple ambassadors in the same cluster. See [this page](../running/#ambassador_id) for more information. | `ambassador_id: "<ambassador_id>"` |
```
LukeShu, line 26: That's not the right link either, is it? There's no "ambassador_id" section at https://www.getambassador.io/docs/pre-release/topics/running/#ambassador_id

mattmcclure-dw: The intended absolute link is https://www.getambassador.io/docs/pre-release/topics/running/running/#ambassador_id.
To test I can run the site locally with Gatsby and hover over the link and it does render correctly:
Screen Shot 2020-11-12 at 1 48 38 PM
I agree that the logic is weird in how it treats paths and it looks like it's shouldn't work.

kflynn: I can confirm that clicking the link above does work, so I'm'a approve these. Separately, though, when I follow the link, the header `AMBASSADOR_ID` is invisible off the top of the page -- is that Chrome being stupid, or something we should fix?

kflynn (approved):
Hit it.

docs/topics/running/ambassador.md
mattmcclure-dw: I think it's a bigger problem around how the site is built and formatted, the browser goes to the right place on the page but the navbar at the top is overlayed over that page such that it covers the top of the page (the header in this case). I *think*. I've seen this behavior consistently on anchors at least. If we end up doing some major design overhaul of the entire docs site hopefully this will go away, otherwise I'll spend some time trying to figure out how to fix it.

PR #3062, opened 2020-11-12
Updated error response doc (RELEASE BRANCH)

mattmcclure-dw:
Fixed a broken link, confirmed it's good now by testing with Gatsby locally. I'm getting the hang of a workflow that should let me better test in the future.
Also changed the title in the ToC to match the title on the page itself and what was published on the release announcement blog. I renamed the file as well to match, I thought about not doing this in case it broke links anywhere but I couldn't find any published links in the blog or elsewhere and OCD wouldn't allow leaving the old file name.

### DIFF
--- a/docs/js/doc-links.yml
+++ b/docs/js/doc-links.yml
@@ -133,6 +133,8 @@
       items:
         - title: The Ambassador Module
           link: /docs/latest/topics/running/ambassador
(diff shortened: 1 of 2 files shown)


esmet (approved):
looks good, thanks!

PR #3063, opened 2020-11-13
Fix a bug in kates where it thinks it is bootstrapped prematurely.

rhs:
## Description

Fix a bug in kates where it thinks it is bootstrapped before it actually is. This change also includes some changes to testing related utilities, namely:

1. Split k3s functionality out of the dtest package into the dtest_k3s package while keeping aliases for it in the dtest package. This is just to avoid circular imports. I would like to fix this in a better way later, so kates is the only thing that should ever use dtest_k3s.

2. Update the k3s functionality to use dexec rather than supervisor.

3. Add an option to dexec.Cmd to optionally disable logging.

## Testing

## Tasks That Must Be Done
  + [ ] Is this a new feature?
      + memory limits?
      + cpu limits?
      + replica counts?
- [ ] Is your change adequately tested?
  + [ ] Is the bulk of your code covered by unit tests?
  + [ ] Does at least one end-to-end test cover the integration points your change depends on?
- [ ] Is this a build change?

## Other


pkg/dexec/borrowed_exec.go
```diff
@@ -125,6 +127,47 @@ func (w *prefixSuffixSaver) Write(p []byte) (n int, err error) {
...
+}
+
+// CaptureErr runs a command with the supplied input and captures
+// stdout and stderr as a string.
+func (c *Cmd) CaptureErr(stdin io.Reader) (output string, err error) {
+	c.Stdin = stdin
+	out := strings.Builder{}
+	c.Stdout = &out
+	c.Stderr = &out
+	err = c.Run()
```
esmet, line 157: is it safe to give the same string builder pointer for stdout and stderr? thinking about races and weird interleaving

rhs: This is a quote from the exec packages godoc:

    // If Stdout and Stderr are the same writer, and have a type that can
    // be compared with ==, at most one goroutine at a time will call Write.

esmet: cool. that's convenient.

esmet (approved):

pkg/kates/client.go
```diff
@@ -357,26 +358,71 @@ func isExpiredError(err error) bool {
 }
 
 type lw struct {
```
kflynn, line 360: I know this isn't a change in this PR, but I just realized how much happier I'd've been to have this be `type listWatcher struct { ... }`.

pkg/kates/client.go
```diff
@@ -357,26 +358,71 @@ func isExpiredError(err error) bool {
...
+// This computes whether we have synced a given watch. We used to use SharedInformer.HasSynced() for
+// this, but that seems to be a blatant lie that always return true. My best guess as to why it lies
+// is that it is actually reporting the synced state of an internal queue, but because the
+// SharedInformer mechanism adds another layer of dispatch on top of that internal queue, the
+// syncedness of that internal queue is irrelevant to whether enough layered events have been
+// dispatched to consider things synced at the dispatch layer.
+//
+// So to track syncedness properly for our users, when we do our first List() we remember how many
+// resourcees there are and we do not consider ourselves synced until we have dispatched at least as
+// many Add events as there are resources.
```
kflynn, line 401: Is this deterministic, or is there a way that it could conceivably still fail? (I _think_ it sounds deterministic, but e.g. suppose a given resource is deleted right as we're coming up -- are we certain to get an add and then a delete, or might we just never see anything for that resource?)

kflynn (approved):
One note, one question, but I'm good landing this.

pkg/kates/client.go
rhs: I believe it is deterministic, because client-go computes the Add events based on the result of the List() method. It's client-go of course so my understanding could prove to be wrong, but that's the theory. ;-)

kflynn: Sounds sane, anyway. 🙂

PR #3068, opened 2020-11-16
(1.9 docs) Fix doc-page link parsing

hdoro:
Starting with 1.9, some `docLinks` from `doc-links.yml` are breaking a function inside `doc-page.js`. This small change should fix it. Will also implement to `master`, but I don't see why apply it to 1.8 and lower.

### DIFF
--- a/docs/js/doc-page.js
+++ b/docs/js/doc-page.js
@@ -38,7 +38,10 @@ function getMainDocsUrl(slug) {
 }
 
 // Used to get a flat array of *all* links with their corresponding parents
-function flattenLinks(links, parent) {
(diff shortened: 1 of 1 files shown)


khussey:
Thanks @hdoro . I fixed what I believe was the root cause in https://github.com/datawire/ambassador/pull/3071. I think we need to better understand how the malformed file got merged in the first place before changing the script that consumes it (in case doing so would mask similar problems in the future).

kflynn:
Overall I think that it's better to let the build fail if people get things completely wrong, so I'm closing this one. Thanks @hdoro!

PR #3079, opened 2020-11-17
[v1.9] fix links

mattmcclure-dw:
Fixing links around error response doc
Also added this: https://github.com/datawire/ambassador/commit/e0425bd7955384d1a8fa865a17d4b1bc45ae1bbf since it was done on master but not the 1.9 branch and had shown up on #blc as well


khussey (approved):
LGTM, thanks!

PR #3092, opened 2020-11-24
Update gRPC example for insecure host

kflynn:
(cherry picked from commit b602adf80a52ca85541ca4050dd58a7a5786ad4e)

### DIFF
--- a/docs/howtos/grpc.md
+++ b/docs/howtos/grpc.md
@@ -153,7 +153,7 @@ spec:
       restartPolicy: Always
 ```
 
-The Host is declared here because we are using gRPC without TLS.  Since Ambassador terminates TLS by default, in the Host we add a `requestPolicy` which allows insecure connections . After adding the Ambassador Edge Stack mapping to the service, the rest of the Kubernetes deployment YAML file is pretty straightforward. We need to identify the container image to use, expose the `containerPort` to listen on the same port the Docker container is listening on, and map the service port (80) to the container port (50051).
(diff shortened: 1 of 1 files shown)


docs/howtos/grpc.md
```diff
@@ -141,7 +153,7 @@ spec:
       restartPolicy: Always
 ```
 
-After adding the Ambassador Edge Stack mapping to the service, the rest of the Kubernetes deployment YAML file is pretty straightforward. We need to identify the container image to use, expose the `containerPort` to listen on the same port the Docker container is listening on, and map the service port (80) to the container port (50051).
+The Host is declared here because we are using gRPC without TLS.  Since Ambassador terminates TLS by default, in the Host we add a `requestPolicy` which allows insecure connections . After adding the Ambassador Edge Stack mapping to the service, the rest of the Kubernetes deployment YAML file is pretty straightforward. We need to identify the container image to use, expose the `containerPort` to listen on the same port the Docker container is listening on, and map the service port (80) to the container port (50051).
```
khussey, line 156: Extraneous space before period here...

khussey (changes requested):
Minor nit.

docs/howtos/grpc.md
kflynn: Oops. Thanks! (Sorry, I missed this comment for awhile. 🙁 )

khussey (approved):
+1

PR #2053, opened 2019-11-22
Add X-Forwarded-Client-Certificate (XFCC) support

jwm:
## Description
Adds a configurable SNI hostname on originated TLS connections.

## Related Issues
#1543

## Testing
We're running this in production, but with Ambassador 0.75. I ported it to the current `master`, but hit a wall running the test suite (see #2052). IIRC, the test was working at one point, but broke when I rebased onto an Ambassador release with Python type checking enabled. It looked like there was a legit type problem at the time, but I haven't touched the test recently.

## Todos
- [ ] Tests
- [X] Documentation

PR #3121, opened 2020-12-21
add XFCC-related config tests for invalid values

jwm:
## Description
Following up with XFCC-related tests to ensure invalid config values result in errors.

## Related Issues
#2053

### DIFF
--- a/python/tests/test_ambassador_module_validation.py
+++ b/python/tests/test_ambassador_module_validation.py
@@ -1,0 +1,150 @@
+from typing import List, Tuple
+
+import logging
+
+import pytest
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s test %(levelname)s: %(message)s",
+    datefmt='%Y-%m-%d %H:%M:%S'
+)
+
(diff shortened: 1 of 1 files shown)


jwm:
One interesting thing I noticed when plumbing the test code: `ResourceFetcher.handle_k8s()` silently ignores blocks that `KubernetesObject` won't parse. I forgot to place the `name` attr in a `metadata` section in the config hunks in this PR, which led to some tail chasing.

I don't have a ton of context here, might make sense to log in this case? Looks like it's got access to the `logger` instance.

```
        try:
            obj = KubernetesObject(raw_obj)
        except ValueError:
            # The object doesn't contain a kind, API version, or name, so we
            # can't process it.
            return
```

kflynn:
The challenge with logging in that case is... what would we log? "Hey, you have an illegally-formed object to which we can't refer.... somewhere"? 🙂 

(We could always spit out its serialization, but it could be very large, so we'd need to be careful to only do it once... hmmm...)

kflynn (approved):
Oh, this looks wonderful -- thanks!!

PR #3135, opened 2021-01-06
update the output image name in the developer docs

jwm:
## Description
When rebuilding recently, I noticed the build process' output image name changed.


kflynn (approved):
Ouch. Many thanks!!

Issue #3142, opened 2021-01-08
Regression: Ambassador fails to start if there is an IngressClass resource in the cluster


vkamra:
**Describe the bug**

Ambassador fails to start in a cluster that has an `IngressClass` resource (see [1] for log output).

The error appears to be related to the logic [here](https://github.com/datawire/ambassador/blob/cdda91fbebd878934966c16f402334fc55b39863/python/ambassador/fetch/fetcher.py#L348) when working with IngressClasses - where the code calls `obj.namespace` which fails for `IngressClass` because it is a cluster-scoped resource.

This does not fail with Ambassador version 1.8.1 - I believe because the validation in `def namespace (self)` was only added in [this](https://github.com/datawire/ambassador/commit/233264aca06809b50aead1195f4fa4c91500d1e7#diff-d9597ed3557cc8d4858aec4845cee74fc78ba45665848ac6ec2ead7a5a90e43dR119) commit.

**To Reproduce**
1. Create an IngressClass resource e.g. 
```
apiVersion: networking.k8s.io/v1beta1
kind: IngressClass
metadata:
  creationTimestamp: "2021-01-05T22:37:32Z"
  generation: 1
  managedFields:
  - apiVersion: networking.k8s.io/v1beta1
    fieldsType: FieldsV1
    fieldsV1:
      f:spec:
        f:controller: {}
    manager: oc
    operation: Update
    time: "2021-01-05T22:37:32Z"
  name: nginx
  resourceVersion: "5440401"
  selfLink: /apis/networking.k8s.io/v1beta1/ingressclasses/nginx
  uid: 9f33c857-c6fa-4adb-90cc-a2954b9c962b
spec:
  controller: nginx.org/ingress-controller
```
2. Start Ambassador - the pod will fail to start with the error in [1]

**Expected behavior**
Ambassador startup is not impacted by `IngressClass` resources

**Versions (please complete the following information):**
 - Ambassador: 1.9.1, 1.10.0
 - Kubernetes environment [e.g. Minikube, bare metal, Google Kubernetes Engine]
 - Version [e.g. 1.8.1]

**Additional context**
Note - we are using `annotations` for Ambassador configuration instead of CRDs.

Startup logs:
[1]
```
2021-01-05 23:43:01 diagd 1.9.1 [P61TAEW] ERROR: could not reconfigure: KubernetesObject ingressclass.networking.k8s.io nginx has no namespace (it is cluster-scoped)
[...]

PR #3143, opened 2021-01-09
Use a resource processor for ingresses and ingress classes

impl:
## Description
This change moves the handlers for the ingress classes and ingresses in the fetcher to their own resource processors. Because of the strange interdependence of the Ambassador service, which must be discovered, ingress classes, and ingresses, we also introduce a dependency handling mechanism that (a) makes it clear where that information is coming from and (b) automatically generates the desired handling order for data coming in from watt.

## Related Issues
* #3142 

## Testing
New tests were added, and this change was tested against the current integration tests in K3d as well.

## Tasks That Must Be Done
  + [ ] Is this a new feature?
      + memory limits?
      + cpu limits?
      + replica counts?
- [x] Is this a build change?


impl:
Closing in favor of #3155, #3156, and #3157.

PR #3144, opened 2021-01-11
Fix link problem. Mea culpa.

kflynn:
There's no 'v' in the links.


khussey (approved):
Oops, I missed this too. :(

PR #3156, opened 2021-01-19
Add processor dependency management

impl:
## Description

This is a subset of #3143.

Frequently, one or more processors will depend on something discovered by another processor to work correctly. This had been implemented by manually keeping track of these dependencies and putting them into a list in the fetcher so that the order would be known in advance. However, this approach is brittle and it's hard to find a good place to store the dependencies that makes sense to every dependent processor.

This change adds support for dependency management, where a processor may provide or want a specific set of data associated with the processing of a given type (expressed as a Watt key). The ordering is then automatically computed by traversing a graph of the dependencies.

Since we've yet to migrate IngressClasses and Ingresses to new processors, there's a bit of rough behavior around them, but the overall approach works as intended.

## Testing

The automated test suite passes with these changes. No changes to behavior are part of this PR.

## Tasks That Must Be Done
  + [ ] Is this a new feature?
      + memory limits?
      + cpu limits?
      + replica counts?
  + [x] Was their sufficient test coverage for the area changed?
  + [x] Do the existing tests capture the requirements for the area changed?
- [ ] Is this a build change?

### DIFF
--- a/python/ambassador/fetch/dependency.py
+++ b/python/ambassador/fetch/dependency.py
@@ -1,0 +1,168 @@
+from typing import Any, Collection, Iterator, Mapping, MutableSet, Optional, Protocol, Sequence, Type, TypeVar
+
+from collections import defaultdict
+import dataclasses
+
+from .k8sobject import KubernetesObject
+
+
+class Dependency (Protocol):
+    """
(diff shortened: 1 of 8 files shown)


kflynn (approved):
Taking this one. Thanks! 🙂

PR #3157, opened 2021-01-19
Move ingress handling to a resource processor

impl:
## Description
This change moves the handling of IngressClasses and Ingresses to their own resource processors. Because this was the last Kubernetes type being handled by the fetcher, we also remove the legacy code for handling different types. Going forward, all new Kubernetes object handling will go into a separate processor type.

This change depends on #3156.

## Testing
This change passes the automated test suite. No behavior is changing.

## Tasks That Must Be Done
  + [ ] Is this a new feature?
      + memory limits?
      + cpu limits?
      + replica counts?
  + [x] Was their sufficient test coverage for the area changed?
  + [x] Do the existing tests capture the requirements for the area changed?
- [ ] Is this a build change?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,10 @@ Please see the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest
 
 ## Next Release
 
+### Ambasssador API Gateway + Ambassador Edge Stack
+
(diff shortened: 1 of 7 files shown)


kflynn (approved):
Given that you ran the `Ingress` tests offline and they passed, I'm'a take this. Thanks!! 🙂

PR #3158, opened 2021-01-20
[master] Update doc on requestPolicy behavior

mattmcclure-dw:
updated doc to warn about global behavior

### DIFF
--- a/docs/topics/running/host-crd.md
+++ b/docs/topics/running/host-crd.md
@@ -126,7 +126,7 @@ requestPolicy:
     additionalPort: insecure-port
 ```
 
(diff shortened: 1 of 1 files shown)


docs/topics/running/host-crd.md
```diff
@@ -126,6 +126,8 @@ requestPolicy:
     additionalPort: insecure-port
 ```
 
+> **WARNING:** `requestPolicy` is applied globally, even if it is applied to only one `Host` and you have multiple `Host`s configured. Different `requestPolicy` behaviors cannot be applied to different `Host`s.   A `requestPolicy` defined on one `Host` will override the default behavior on another `Host` with no `requestPolicy` defined.
```
cakuros, line 129: I think there should be guidance in telling people that if they have `requestPolicy` that they want set, to copy it on to ALL Host resources for that Ambassador installation rather than having just one dictate and overwrite the policy of others (since this behavior is not always obvious)

cakuros (changes requested):

cakuros (approved):
LGTM!

PR #3159, opened 2021-01-20
[1.10] Update doc on requestPolicy behavior

mattmcclure-dw:
updated doc to warn about global behavior


cakuros (changes requested):
(Copied from Master): I think there should be guidance in telling people that if they have requestPolicy that they want set, to copy it on to ALL Host resources for that Ambassador installation rather than having just one dictate and overwrite the policy of others (since this behavior is not always obvious)

cakuros (approved):
LGTM!

PR #3197, opened 2021-02-03
[master] Code block titles

mattmcclure-dw:
Removing the language type from code block markdown wherever the code is a terminal command(s) but something else is specified ("bash," "shell," etc.) since the new code block styling will put that into the title. We want them to consistently be "Terminal" (which it uses if nothing is specified).


brianfleming (approved):
lgtm

PR #3227, opened 2021-02-17
Adding Host warning message where host is used in examples

AliceProxy:
Adding the warning about Host behavior wherever host is used, and adding further clarification about what changes are made to the cluster.

I encountered a user that was using `Host.requestPolicy` which modified the `requestPolicy` for the entire cluster for a user. They did not understand that this change was being applied outside of the `Host` in which they applied the configuration.
The source of the error with `requestPolicy` behavior was not immediately apparent until support asked the user if they had
applied other hosts with a different `requestPolicy`.

The docs state in the warning from this PR that only one `requestPolicy` can be applied to an Ambassador instance, and that all `Host`s should share the same `requestPolicy` configuration.

This is expected behavior; however, we felt that it was a difficult issue for a user to figure out on their own without the assistance of support.  The purpose of this PR is to add clarity to the warning statement, describe in detail what changes are being made to the user's configuration, and copy the warning to other pages that make use of the `Host` `CRD` so that users can be aware of changes that they are making to the configuration of their cluster.

There [is an issue ](https://github.com/datawire/apro/issues/2328) that further addresses the behavior described above.

### DIFF
--- a/docs/howtos/grpc.md
+++ b/docs/howtos/grpc.md
@@ -155,6 +155,21 @@ spec:
 
 The Host is declared here because we are using gRPC without TLS.  Since Ambassador terminates TLS by default, in the Host we add a `requestPolicy` which allows insecure connections. After adding the Ambassador Edge Stack mapping to the service, the rest of the Kubernetes deployment YAML file is pretty straightforward. We need to identify the container image to use, expose the `containerPort` to listen on the same port the Docker container is listening on, and map the service port (80) to the container port (50051).
 
+> **WARNING - Host Configuration:** The `requestPolicy` property of the `Host` `CRD` is applied globally within an Ambassador instance, even if it is applied to only one `Host` when multiple `Host`s are configured. Different `requestPolicy` behaviors cannot be applied to different `Host`s. It is recommended to apply an identical `requestPolicy` to all `Host`s instead of assuming the behavior, to create a more human readable config. 
+> 
+> If you intend to use more than one type of `requestPolicy`, you will need a separate Ambassador instance for each separate type.
+> 
+> If multiple `Host`s are applied, the `requestPolicy` from the `Host` with the first alphabetical `metadata.name` is always the one that is applied. Order does not matter.
+> If a `requestPolicy` is not defined for a `Host`, it's assumed to be `Redirect`, and so even if a host named `a` does not specify it, the default `requestPolicy` of `Redirect` will be applied to all `Host`s in that Ambassador instance.
+>
+> The `insecure-action` can be one of:
+>
+> * `Redirect` (the default): redirect to HTTPS
+> * `Route`: go ahead and route as normal; this will allow handling HTTP requests normally
+> * `Reject`: reject the request with a 400 response
+>
+> For more information, please refer to the [`Host` documentation](../topics/running/host-crd#secure-and-insecure-requests).
+
 Once you have the YAML file and the correct Docker registry, deploy it to your cluster with `kubectl`.
 
 ```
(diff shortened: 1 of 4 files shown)


docs/howtos/grpc.md
```diff
@@ -155,6 +155,21 @@ spec:
...
 
+> **WARNING - Host Configuration:** The `requestPolicy` property of the `Host` `CRD` is applied globally within an Ambassador instance, even if it is applied to only one `Host` when multiple `Host`s are configured. Different `requestPolicy` behaviors cannot be applied to different `Host`s. It is recommended to apply an identical `requestPolicy` to all `Host`s instead of assuming the behavior, to create a more human readable config. 
+> 
+> If you intend to use more than one type of `requestPolicy`, you will need a separate Ambassador instance for each separate type.
+> 
+> If multiple `Host`s are applied, the `requestPolicy` from the `Host` with the first alphabetical `metadata.name` is always the one that is applied. Order does not matter.
+> If a `requestPolicy` is not defined for a `Host`, it's assumed to be `Redirect`, and so even if a host named `a` does not specify it, the default `requestPolicy` of `Redirect` will be applied to all `Host`s in that Ambassador instance.
```
cakuros, line 163: Again, I think leaving out the bit about alphabetical here.

"If a `requestPolicy` is not defined for a `Host`, it's assumed to be `Redirect`.  If the behavior expected out of Edge Stack is anything other than `Redirect`, it must be explicitly enumerated in all `Host` resources."

docs/howtos/grpc.md
```diff
@@ -155,6 +155,21 @@ spec:
 
 The Host is declared here because we are using gRPC without TLS.  Since Ambassador terminates TLS by default, in the Host we add a `requestPolicy` which allows insecure connections. After adding the Ambassador Edge Stack mapping to the service, the rest of the Kubernetes deployment YAML file is pretty straightforward. We need to identify the container image to use, expose the `containerPort` to listen on the same port the Docker container is listening on, and map the service port (80) to the container port (50051).
 
+> **WARNING - Host Configuration:** The `requestPolicy` property of the `Host` `CRD` is applied globally within an Ambassador instance, even if it is applied to only one `Host` when multiple `Host`s are configured. Different `requestPolicy` behaviors cannot be applied to different `Host`s. It is recommended to apply an identical `requestPolicy` to all `Host`s instead of assuming the behavior, to create a more human readable config.
```
cakuros, line 158: I *think* we want to be renaming all instances of `Ambassador` as `Edge Stack`.  Check with Matt McClure on this.

docs/howtos/grpc.md
```diff
@@ -155,6 +155,21 @@ spec:
 
 The Host is declared here because we are using gRPC without TLS.  Since Ambassador terminates TLS by default, in the Host we add a `requestPolicy` which allows insecure connections. After adding the Ambassador Edge Stack mapping to the service, the rest of the Kubernetes deployment YAML file is pretty straightforward. We need to identify the container image to use, expose the `containerPort` to listen on the same port the Docker container is listening on, and map the service port (80) to the container port (50051).
 
+> **WARNING - Host Configuration:** The `requestPolicy` property of the `Host` `CRD` is applied globally within an Ambassador instance, even if it is applied to only one `Host` when multiple `Host`s are configured. Different `requestPolicy` behaviors cannot be applied to different `Host`s. It is recommended to apply an identical `requestPolicy` to all `Host`s instead of assuming the behavior, to create a more human readable config. 
+> 
+> If you intend to use more than one type of `requestPolicy`, you will need a separate Ambassador instance for each separate type.
```
cakuros, line 160: Ambassador -> Edge Stack

cakuros (changes requested):
Some changes required.

docs/howtos/grpc.md
```diff
@@ -155,6 +155,21 @@ spec:
...
+> 
+> If multiple `Host`s are applied, the `requestPolicy` from the `Host` with the first alphabetical `metadata.name` is always the one that is applied. Order does not matter.
```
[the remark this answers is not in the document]
AliceProxy: Agreed, I was not considering that describing this quirk of the behavior might cause users to plan for its existence when configuring `Host`s. Updated warnings to advise against using this behavior rather than to expect it as you suggested.

AliceProxy:
Pushed a new commit to this branch to remove descriptions of the alphabetical behavior as well as change `Ambassador` to `Edge Stack`.

PR #3231, opened 2021-02-17
[v1.4] Removing documentation source files (not YAML)

khussey:
This pull request removes documentation source files (but not YAML) now that they live in ambassador-docs.git.


kflynn (approved):
Hit it!

Issue #3224, opened 2021-02-16
Ambassador should route any knative service endpoint


maticue:
**Describe the bug**

Per default knative services with endpoints different than `/` are not routed by ambassador. 

Let's suppose the following curl:

````
curl -v "http://event-display.default.svc.cluster.local/hello" -X POST -H "Content-Type: application/json" -d '{"foo":"bar"}'
````

Ambassador logs show `No Route` value:

````
ACCESS [2021-02-12T23:18:02.936Z] "POST /hello HTTP/1.1" 404 NR 0 0 0 - "10.233.88.35" "curl/7.75.0" "ad25b510-5b9e-48c2-a3de-db876fc414b5" "event-display.default.svc.cluster.local" "-"
````
This is because [Ambassador default mapping for knative routing has](https://github.com/datawire/ambassador/blob/master/python/ambassador/fetch/knative.py#L79):

````
prefix: /
prefix_regex: true
````

which, based on [ambassador documentation](https://github.com/datawire/ambassador/blob/master/python/ambassador/fetch/knative.py#L79):

> When the prefix_regex attribute is set to true, Ambassador Edge Stack configures a regex route instead of a prefix route in Envoy. This means the entire path must match the regex specified, not only the prefix.

the scope is limited to only `/`.

On the other hand, knative service definition does not allow to set up `path` attribute which is expected by `knative.py`. Knative also creates a knative route + knative ingress when a knative service is created.

We can workaround this creating another knative ingress manifest different from the default one created automatically by knative.

**To Reproduce**
1. Install ambassador and knative. [Configure them to work together.](https://knative.dev/docs/install/any-kubernetes-cluster/#serving_networking-0)
2. Create any knative service like the following

````
apiVersion: serving.knative.dev/v1 # Current version of Knative
kind: Service
metadata:
  name: helloworld-go # The name of the app
spec:
  template:
    spec:
      containers:
        - image: gcr.io/knative-samples/helloworld-go # Reference to the image of the app
          env:
            - name: TARGET # The environment variable printed out by the sample app
              value: "Go Sample v1"
````

2. Wait until ambassador mapping is created
3. Try to access to GET any endpoint for the designated service different than `/`. Like `/hello`
4. Get a 404 and see the error on ambassador logs

**Expected behavior**
If the backend service has a service running on `/hello`, I should see a 2xx HTTP code.

**Versions (please complete the following information):**
 - Ambassador: 1.11.1
 - Kubernetes environment: OKE
[...]

PR #3232, opened 2021-02-19
Convert Knative Ingress paths from prefix match to path match

impl:
## Description
This change makes the paths specified in each Knative Ingress rule prefix-matching. The previous behavior required the path pattern to match the entire actual path. With a default of `/`, this in particular meant that any path other than the root would not match, which explicitly deviates from the "catch all" requirement in the Knative spec.

## Related Issues
* #3224 

## Testing
This change passes existing automated tests, although it could probably use a test of its own as well.

## Tasks That Must Be Done
  + [ ] Is this a new feature?
      + memory limits?
      + cpu limits?
      + replica counts?
- [ ] Is your change adequately tested?
- [ ] Is this a build change?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,11 @@ Please see the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest
 
 ## Next Release
 
-(no changes yet)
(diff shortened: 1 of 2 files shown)


kflynn (approved):
Taking this one. 🙂 Thanks!

PR #3235, opened 2021-02-23
[master] Quick start YAML

mattmcclure-dw:
Adding Edge Stack quick start YAML file

### DIFF
--- a/docs/yaml/quickstart/qotm.yaml
+++ b/docs/yaml/quickstart/qotm.yaml
@@ -1,0 +1,48 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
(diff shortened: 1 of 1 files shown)


khussey (approved):
Looks OK to me but I defer to @kflynn on how well this fits with our other YAML (and upcoming changes to how we manage it).

mattmcclure-dw:
@khussey sounds good, I could put it in /docs/latest/... if needed instead but was trying to avoid using that super long URL in the new quick start

kflynn (approved):
I'm OK with adding YAML to the website here. Upcoming work might move it in the future, but we'll be in touch on that before it happens. 🙂

PR #3243, opened 2021-02-26
Adding a versions.yml file alongside YAML

khussey:
## Description
This pull request adds a versions.yml file alongside the YAML files so that the versions used during variable substitution can be maintained in the same place where they are used.


kflynn (approved):
Hit it!

PR #3391, opened 2021-05-05
Prep work for making host actions fully independent

kflynn:
This is prep for making `Host`s have independent insecure routing actions -- it's mostly a combination of splitting out the parts I think will be important to keep around in the `V2Listener` from the parts that won't, reorganizing a few things, and scrapping some old dead code. It should handle both Envoy V2 and Envoy V3.

Testing is all KAT-based right now.

 - [x] I didn't need to update `CHANGELOG.md` -- this is all internals.
 - [x] This is unlikely to impact how Ambassador performs at scale, because it's not changing what we're actually doing for the most part.


acookin (approved):

kflynn:
Closing and reopening due to `master` rewrite.

PR #3397, opened 2021-05-07
emissary k8s manifests

acookin:
## Description
Not yet plugged into CI but i can run the script locally and it works.

## Related Issues
List related issues.

## Testing

## Checklist

    + Any new features
    + Any deprecations
 

kflynn (approved):
I think this looks OK too, thanks!

PR #3403, opened 2021-05-12
Repatriate 1.13.4

kflynn:
Repatriate 1.13.4


python/ambassador/ir/irambassador.py
```diff
@@ -131,7 +132,7 @@ def __init__(self, ir: 'IR', aconf: Config,
             use_proxy_proto=False,
             enable_http10=False,
             proper_case=False,
-            prune_unreachable_routes=True,          # default True; can be updated in finalize()
+            prune_unreachable_routes=False,
```
acookin, line 135: this default should remain True?

acookin (changes requested):
i think the `prune_unreachable_routes` change may have been undone here 😬

python/ambassador/ir/irambassador.py
kflynn: Great catch, thanks!

acookin (approved):

esmet (approved):

PR #3413, opened 2021-05-12
emissary/ambassador/aes manifests/charts to s3

acookin:
## Description

Push all the things to s3.

The result of the scripts in this PR is that:
* all the manifests in ./manifests/ directory get published to s3. there is a separate key for each product in s3, and those things are accessible via app.getambassador.io/yaml/$product/$version/*.yaml, where $version is taken from docs/yaml/version.yml
** e.g. ambassador.git/manifests/edge-stack/aes.yaml is accessible at `https://app.getambassador.io/yaml/edge-stack/1.13.3/aes.yaml` (on this branch version.yml is 1.13.3)
* emissary-ingress and ambassador charts are published to the same helm repo in s3. if you run `helm repo add datawire helm repo add datawire https://app.getambassador.io` then run `helm search repo datawire`, you should see `ambassador`, `emissary-ingress` and `ambassador-operator`

Notable things for the manifests:
* i only published the things to s3 that we generate from the helm chart and are referenced in documentation
** a bunch of stuff like the emojivoto, quote, etc should actually be moved to the docs repo probably? or to another repo that has some CD that publishes those manifests. should sync with support on this
** i didn't pull in traffic agent, traffic manager, project stuff because that's going to be deprecated and i didn't want to pull in any cruft
** ambassador-agent exists in all the manifests/charts by default, and is not referenced in docs anywhere.
* all the other things don't seem to have docs or any reference on how to keep them up to date, so i would like that to happen before i just blindly move them over
** this includes things like ambassador-prometheus-rbac.yaml, ambassador-istio-configmap.yaml, etc.
* I changed the manifest generation so that the image tag is no longer `$$version$$`, and made it the version that's in the helm chart (since $$version$$ won't be a thing that works when we're reading things from s3)

If you think I've missed something, i probably have so lmk :).

I also haven't hooked this into the release process yet. Mostly because I'm not sure if we should publish _just_ the emissary chart, _just_ the ambassador chart, or both (in which case it would be easier to version them together at this point). At any rate, i think this PR does enough good on it's own and we can address the release bits in a different PR.
## Related Issues
List related issues.

## Testing

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 

kflynn (approved):
This looks much more graceful than the one for `master`. 😂

PR #3418, opened 2021-05-13
publish yaml in ./docs to s3

acookin:
## Description
Publish everything in ./docs/yaml to s3.
Probably kinda hacky, but I'm pretty sure it works.

We need to "generate" the yaml that we publish because we need to replace the variables in versions.yml in each yaml file, so yeah that's kinda gross.

## Related Issues
List related issues.

## Testing

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 
### DIFF
--- a/docs/publish_yaml_s3.sh
+++ b/docs/publish_yaml_s3.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+
+set -e
+
+CURR_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+[ -d "$CURR_DIR" ] || { echo "FATAL: no current dir";  exit 1; }
+basedir=$1
+shift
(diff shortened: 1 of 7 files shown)


docs/publish_yaml_s3.sh
```diff
@@ -0,0 +1,72 @@
...
+fi
+
+echo ${version} > stable.txt
+if [ -z "$AWS_BUCKET" ] ; then
+    AWS_BUCKET=datawire-static-files
+fi
+
+# make this something different than ambassador, emissary, or edge-stack
+# so we don't conflict with the new hotness we're doing for 2.0
+unversioned_base_s3_key=yaml/ambassador-docs/
```
acookin, line 42: yeah this is a terrible path for this. other options had something to do with `legacy`... any suggestions welcome lol.

kflynn (approved):
Hit it!

PR #3421, opened 2021-05-14
Set up a nightly job to push 2.0.0-nightly artifacts

kflynn:
This'll be gaining intelligence as we go, but, here're the highlights:

- Introduce a nightly job for `shared/emissary` to run the new `push-nightly` job. 
- The `push-nightly` job currently 
   - takes the existing image that we already built,
   - retags and pushes it as `2.0.0-nightly.YYYYMMDDTHHMMSS`, and also
   - retags and pushes it as `2.0.0-nightly.YYYYMMDD`.
   - In addition to `ambassador` images, it also pushes `ambassador-ea` images (see below).
- It does this _before testing_ so that we always have that image.
- **NOTE WELL** that this is _not_ a real, tested image yet! There's still a lot of work to do. (If you're reading this from outside Ambassador Labs, for the love of all that is holy please _do not run this image_ before talking to us!)
- The `ambassador-ea` images are the same code as the `ambassador` images, but all the interesting knobs that exist are turned on by default.
   - At present, that means setting `AMBASSADOR_FAST_RECONFIGURE=true` and `AMBASSADOR_ENVOY_API_VERSION=V3`.
   - Soon enough it might also include e.g. making the endpoint resolver standard or something. Who knows.
- Next up is teaching `push-nightly` to also push the Helm chart, YAML, etc.

Some underlying changes:

- Version information always reads `docs/yaml/versions.yml` now, which implies that
- We'll be updating `docs/yaml/versions.yml` to reflect the next upcoming release very early in our development cycles (you'll note that in this PR, it's set to `2.0.0`).

Note that this PR currently builds the EA images from `Dockerfile-ea`, which I find a little unsatisfying. Either we should spilt the KAT images out of `Dockerfile` (which I would support), or we should include the EA images in `Dockerfile` (a quick cut at that turned up hassles with needless build args, so I did this the simple way instead).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,6 +118,14 @@ commands:
     - run: |
         [[ "$DEV_REGISTRY" == 127.0.0.1:31000 ]] || make push
     - when:
+        condition:
+          not: << parameters.release >>
+        steps:
+        - run:
+            name: "Push dev images"
+            command: |
+              [[ "$DEV_REGISTRY" == 127.0.0.1:31000 ]] || make push-dev
+    - when:
         condition: << parameters.release >>
         steps:
         - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -143,6 +143,21 @@ commands:
     # teardown
     - dirty-check
     - amb-save-workspace
+  "job-push-ci":
+    steps:
+    # setup
+    - amb-linux-install
+    - amb-checkout
+    - amb-skip-if-no-code-changes
+    - amb-config-registry
+    - # main
+      run: make images
+    - run: |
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -703,11 +736,19 @@ jobs:
       release:
         type: boolean
         default: false
+      push_nightly:
+        type: boolean
+        default: false
     executor: oss-linux
-    resource_class: xlarge
+    resource_class: large
```
esmet, line 743: this does slow down the build considerably but it might be OK since the build is fast-ish now.

esmet (approved):

acookin (approved):

PR #3422, opened 2021-05-14
fix: set diagd gunicorn keepalive to 300s

neufeldtech:
### The Problem
Intermittent 503's from the `/ambassador/v0/check_alive` endpoint even though gunicorn/diagd is healthy.

This is to help avoid 503's on the /ambassador/v0/check_alive and similar endpoints. Currently, use-cases that probe this health check are subject to race-conditions where gunicorn closes idle TCP connections after 2s and does not send a FIN or RST to envoy. Envoy, believing the tcp connection is still active, attempts to route the check_alive request to gunicorn, but receives a RST as the TCP connection is no longer valid according to gunicorn. This results in a failing health check even though gunicorn is 'up'. This false-positive failure mode is problematic especially in situations where one is leveraging this health check to detect ambassador health from a CDN or other downstream load balancer.

See Wireshark image of the default 2s tcp idle timeout from gunicorn, and the implications of this low default. Our CDN is probing the backend between once every 500ms and 2.5s. When the HTTP probes do this pod are more than 2s apart, the above race condition is triggered and Envoy reports receiving a 503 from gunicorn, indicating a 'failed' health check.

image

Setting the keepalive timeout to 300s should be a safe default as typical implementations will probably be polling this health endpoint more than once every 5 minutes.

## Related Issues
N/A

## Testing
I've manually tested & captured this in Production some time ago, but since have upgraded back to mainline ambassador and am no longer running this fork. Would love to get this change merged into Ambassador to fix my 503 problem 😬 .

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 

tony-ong (dismissed):

neufeldtech:
@esmet @kflynn @acookin @rhs I'd appreciate your review and feedback on this. If further clarification is needed on how this solves my issue let me know what further detail you'd like. Let me know if you have any questions.

Thanks

bullywhippet:
This keepalive would be super helpful, any chance of getting this feature merged in?

kflynn:
Superseded by #4008 -- many thanks for the original, @neufeldtech!

PR #3429, opened 2021-05-15
Have a way to promote Emissary builds all the way to "GA"

kflynn:
Review commit-by-commit. **NOTE WELL**: several paths are deliberately tweaked to be _certain_ that we won't collide with Ambassador. Also, there's a _lot_ of opportunity to factor out common code in here.

Flow:
- Develop as usual.
- PRs build and push dev-versioned images, Helm charts, and YAML.
   - These builds are also recorded in S3 so you can find the dev build corresponding to a given commit.
- To promote to GA, tag with a `vX.Y.Z` tag and, for now, run `make release/promote-oss/to-ga`
   - This will look up the dev build in S3, retag it as "GA" and push, then push Helm and manifests
      - Right now, the "GA" tag looks like `vX.Y.Z-wip` to avoid collisions with real builds.
 
Major caveats:
- The EA images are pushed, but the EA Helm charts are not (yet).
- The paths in S3 and such are hella disorganized, and we need to clean them up.
- CI isn't hooked up on the GA tag, but that's deliberate.
- The whole record-the-dev-version in S3 bugs me; if we keep it we'll need a pruning job.
   - OTOH we can probably ditch the whole `-ci` version thing by keeping the S3 thing.

### DIFF
--- a/builder/builder.mk
+++ b/builder/builder.mk
@@ -434,9 +434,11 @@ push-nightly: docker/$(LCNAME).docker.tag.local docker/$(LCNAME)-ea.docker.tag.l
 				docker push $$tag ;\
 			done ;\
 		done ;\
+		CHART_VERSION_SUFFIX=-nightly.$$today ;\
+		IMAGE_TAG=$${base_version}$${CHART_VERSION_SUFFIX} ;\
 		$(MAKE) \
-			CHART_VERSION_SUFFIX=-nightly.$$today \
-			IMAGE_TAG=$${base_version}-nightly.$${suffix} \
+			CHART_VERSION_SUFFIX="$${CHART_VERSION_SUFFIX}" \
+			IMAGE_TAG="$${IMAGE_TAG}" \
 			IMAGE_REPO="$(DEV_REGISTRY)/$(LCNAME)" \
 			chart-push-ci ; \
 		$(MAKE) update-yaml --always-make; \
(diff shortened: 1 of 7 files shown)


charts/charts.mk
```diff
@@ -26,3 +26,9 @@ chart-push-ci:
 		$(call _set_repo,$$chart,${IMAGE_REPO}) ; \
 		$(call _push_chart,$$chart) ; \
 	done ;
+
+# This is pretty Draconian. Use with care.
```
acookin, line 43: meh yeah i really wanted to avoid doing something like this bc it would blow away your local changes, but it's still probably. better than not cleaning up.

also i think you want to use $(OSS_HOME) to prefix the paths

builder/builder.mk
```diff
@@ -390,6 +390,9 @@ push-dev: docker/$(LCNAME).docker.tag.local docker/$(LCNAME)-ea.docker.tag.local
 			docker tag $$(cat docker/$$image.docker) $$tag && \
 			docker push $$tag ;\
 		done ;\
+		commit=$$(git rev-parse HEAD) ;\
+		printf "$(CYN)==> $(GRN)recording $(BLU)$$commit$(GRN) => $(BLU)$$suffix$(GRN) in S3...$(END)\n" ;\
+		echo "$$suffix" | aws s3 cp - s3://datawire-static-files/dev-builds/$$commit ;\
```
acookin, line 395: if this is going to get used to decide if a commit is good for GA should there be a dirty check somewhere?

kflynn: Oooh, nice catch.

acookin (approved):

PR #3436, opened 2021-05-18
move/massage some toybox scripts, tie in manifests/helm to GA release

acookin:
## Description
TODO: i need to add a bunch more comments

This PR changes the release workflow to look something like this (probably best to review with each step):

1. At some point (beginning of cycle or before a patch release), run `make release/start`, which:
    1. Runs `releng/start-sanity-check`
        1. makes sure you're in `git@github.com:datawire/ambassador` and that you're in a clean checkout
        2. either checks out or creates the release/vX.Y branch and makes sure you're up to date with that branch
    2. Runs `releng/start-update-version X.Y.Z`
        1. updates `docs/yaml/versions.yml` to have version: X.Y.Z
        2. puts `## [X.Y.Z] (TBD)` with (no changes) in the changelog
        3. commits changes. 
2. "At some point" there will be a ref that has been tested/smoked and is ready to be promoted to GA. This ref should be checked out and then you can run `make release/go` which:
    1. runs `releng/start-sanity-check` again (see above for details)
    2. creates+pushes the tag. this will trigger `release/promote-oss/to-ga`, probably in CI (see next step)
    3. runs `releng/release-go-changelog-update`
        1. updates the date in the changelog (there is already a `## [X.Y.Z] (TBD)` entry, this just changes it to `## [X.Y.Z] May 25, 2021` e.g.
        2. creates a branch off of release/vX.Y called `rel/X.Y.Z/changelog-date-update` (TODO: actually ensure this branches from the right place)
        3. commits+pushes changelog changes to changelog branch
        4. creates a PR for these changes
3: Run `make RELEASE_VERSION=X.Y.Z release/promote-oss/to-ga`, where X.Y.Z is the tag created in the step above (TODO: CI does this)
    1. takes the dev version from `s3://datawire-static-files/dev-builds/$commit`, pulls the known image from dev version, re-tags it with GA and pushes the new GA image (i didn't change this bit)
4. Once `make release/promote-oss/to-ga` is run, we can now update the manifests with the GA lookin tag by running `make VERSION=X.Y.Z release/manifests`. (Note: At this point, we've already tested everything with the dev versions)
    1. `git checkout vX.Y.Z -b rel/vX.Y.Z/chart-img-updates`
    1. update all the image tag references in the chart (chart.yaml, values.yaml, README, CHANGELOG entry)
    2. bumps the chart versions (either bumps patch or minor depending on the VERSION).
    3. runs `make update-yaml` to generate the yaml from the chart, so the manifests all have the updated version.
    4. commits+pushes to branch
    5. create a PR back to release/vX.Y
    6. Tags the chart off of the HEAD of rel/vX.Y.Z/chart-img-updates. This tag will trigger CI to publish the chart and manifests. (Note: triggering manifest publishing here because we can't trigger jobs off of release/vX.Y branches without needing to touch almost every PR circle config)

## Related Issues
List related issues.

## Testing

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 
### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1040,6 +1040,23 @@ workflows:
             only: /^chart\-v[0-9]+\.[0-9]+\.[0-9]+$/
           branches:
             ignore: /.*/
+  'OSS: Manifests Release':
+    when:
+      or:
+      - equal:
+        - "https://github.com/datawire/ambassador"
+        - << pipeline.project.git_url >>
+      - equal:
+        - "https://github.com/datawire/ambassador-private"
+        - << pipeline.project.git_url >>
+    jobs:
+    - oss-manifest-publish:
+        name: "oss-rc-manifests-publish"
+        filters:
+          tags:
+            only: /^chart\-v[0-9]+\.[0-9]+\.[0-9]+$/
+          branches:
+            ignore: /.*/
   'OSS: Release':
     when:
       or:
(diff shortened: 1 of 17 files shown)


kflynn (approved):
I think this looks OK, thanks!

esmet (approved):

PR #3437, opened 2021-05-18
Add support for AMBASSADOR_JSON_LOGGING

esmet:
## Description
Support AMBASSADOR_JSON_LOGGING to get more of the control plane components logging in json. This should compliment our Envoy json logging options. There are still known issues with gunicorn workers not printing in json as well as the k8 client-go library still not printing in json.

## Testing
- Still needs tests

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 

Makefile
```diff
@@ -39,3 +40,21 @@ SHELL = bash
...
 .DELETE_ON_ERROR:
 .PHONY: FORCE
+
+deploy: push preflight-cluster
+	$(MAKE) deploy-only
+.PHONY: deploy
+
+deploy-only:
+	kubectl --kubeconfig $(DEV_KUBECONFIG) create namespace ambassador 2>/dev/null || true && \
+		kubectl apply -f https://getambassador.io/yaml/aes-crds.yaml && \
```
acookin, line 50: kubectl apply -f docs/yaml/ambassador/ambassador-crds.yaml ?

cmd/entrypoint/notify.go
```diff
@@ -109,7 +111,7 @@ func notifyWebhookUrl(ctx context.Context, name, xurl string) bool {
 		if errors.Is(err, syscall.ECONNREFUSED) {
 			// We couldn't succesfully connect to the sidecar, probably because it hasn't
 			// started up yet, so we log the error and return false to signal retry.
-			log.Println(err)
+			dlog.Errorf(context.TODO(), err)
```
acookin, line 114: what's wrong with the context passed to this function?

acookin (changes requested):
just some smol nits

cmd/entrypoint/notify.go
esmet: good call

Makefile
```diff
@@ -20,6 +20,7 @@ unexport SSH_CLIENT
 unexport SSH2_CLIENT
 
 NAME ?= ambassador
+DEV_KUBECONFIG ?= ~/.kube/config
```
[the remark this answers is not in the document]
kflynn: I think so too -- if `$DEV_KUBECONFIG` isn't set when you try to talk to a cluster, it should fail.

kflynn (changes requested):
There are like three PRs in here: `make deploy`, clean up `dlog` usage, and JSON logging. As much as I like these things, I think we shouldn't pile them all into one PR... 🙂

esmet:
> There are like three PRs in here: `make deploy`, clean up `dlog` usage, and JSON logging. As much as I like these things, I think we shouldn't pile them all into one PR... 🙂

Sounds good, I created https://github.com/datawire/ambassador/pull/3441 and https://github.com/datawire/ambassador/pull/3442

kflynn (approved):
I'm curious why the test is in `pkg/busy/busy.go` but hit it!

esmet:
> I'm curious why the test is in `pkg/busy/busy.go` but hit it!

The `busyambassador` change to use JSON logging is in the busy package, so I added a unit test for the new functionality in that package.

PR #3441, opened 2021-05-19
Add make deploy

esmet:
Make local development easier by adding a make deploy target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,3 +39,29 @@ SHELL = bash
 .SECONDARY:
 .DELETE_ON_ERROR:
 .PHONY: FORCE
+
+preflight-dev-kubeconfig:
+	@if [ -z "$(DEV_KUBECONFIG)" ] ; then \
+		echo "DEV_KUBECONFIG must be set"; \
+		exit 1; \
+	fi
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -20,6 +20,7 @@ unexport SSH_CLIENT
 unexport SSH2_CLIENT
 
 NAME ?= ambassador
+DEV_KUBECONFIG ?= ~/.kube/config
```
acookin, line 23: same feedback, this should probaly just fail and tell the user to export a DEV_KUBECONFIG

acookin (changes requested):

Makefile
kflynn: I see the `preflight-dev-kubeconfig` target, which is great! need to drop this assignment too. 😉

kflynn (changes requested):
One last tweak. 🙂 Thanks!

Makefile
esmet: Done

acookin (approved):

kflynn (approved):
Hit it! Thanks. 🙂

PR #3447, opened 2021-05-21
Disable logging in invokeHook

esmet:
This will reduce log noise when using AMBASSADOR_LEGACY_MODE=true


kflynn (approved):
Hit it!

khussey:
@esmet seems like this may be a candidate for merging to `release/v1.13` as well?

esmet:
Declining this in favor of https://github.com/datawire/ambassador/pull/3448

PR #3449, opened 2021-05-24
Docker build without the build container

esmet:
This will help with layer caching and make the build look and feel more typical.

This also:
- Updates the CI build to an xlarge builder now that we have more CPU-hungry tasks on the critical path to pushing images.
- Vendors go deps, which helps the docker build.
- Combines the images + push CI steps so that we do not need to run through the make machinery twice.


builder/builder.sh
```diff
@@ -172,7 +172,7 @@ print("stage2_tag=%s" % stage2)
     local name2="${BASE_REGISTRY}/builder-base:stage2-${stage2_tag}"
 
     msg2 "Using stage-1 base ${BLU}${name1}${GRN}"
-    if ! docker run --rm --entrypoint=true "$name1"; then # skip building if the "$name1" already exists
+    if ! docker image inspect "$name1" &>/dev/null; then # skip building if the "$name1" already exists
```
esmet, line 175: We use this pattern in builder.mk so I decided that it was sufficient. It is also faster than docker run.

builder/builder.mk
```diff
@@ -281,79 +273,54 @@ docker/base-envoy.docker.stamp: FORCE
 	    docker image inspect $(ENVOY_DOCKER_TAG) --format='{{ .Id }}' >$@; \
 	  fi; \
 	}
-docker/$(LCNAME).docker.stamp: %/$(LCNAME).docker.stamp: %/snapshot.docker.tag.local %/base-envoy.docker.tag.local %/builder-base.docker $(BUILDER_HOME)/Dockerfile FORCE
+docker/$(LCNAME).docker.stamp: %/$(LCNAME).docker.stamp: %/base-envoy.docker.tag.local %/builder-base.docker $(BUILDER_HOME)/Dockerfile FORCE
```
esmet, line 276: Removing `snapshot.docker.tag.local` etc is key. It means that we no longer let make track whether a Docker build needs to happen or not. We just let Docker decide how to build layers and whether to use a cached version or not.

builder/builder.mk
```diff
@@ -231,6 +231,10 @@ raw-version:
 	@$(BUILDER) raw-version
 .PHONY: raw-version
 
+python/ambassador.version:
+	$(BUILDER) raw-version > python/ambassador.version
```
esmet, line 235: This is a hack but it gets the job done for now.

builder/Dockerfile
```diff
@@ -148,8 +194,8 @@ FROM ${base} as kat-server
 COPY --from=artifacts /buildroot/bin/kat-server /usr/bin/kat-server
 
 WORKDIR /work
-COPY server.crt server.crt
-COPY server.key server.key
+COPY builder/server.crt server.crt
+COPY builder/server.key server.key
```
esmet, line 198: We run docker build from the top level so we need to reference these files inside the builder/ directory.

builder/Dockerfile
```diff
@@ -65,12 +60,63 @@ USER dw
...
+# These will be extracted into the optimized image later
+ADD demo demo
+ADD docs/yaml/ambassador/ambassador-crds.yaml docs/yaml/ambassador/ambassador-crds.yaml
 
 ########################################
 # The optimized images
 ########################################
 
-FROM ${artifacts} as artifacts
+FROM artifact-stage as artifacts
```
esmet, line 119: This is key. We now start from the artifact-stage, defined above, instead of $artifacts, which would have been a build container snapshot.

builder/Dockerfile
```diff
@@ -65,12 +60,63 @@ USER dw
 
 ENTRYPOINT [ "/bin/bash" ]
 
+########################################
+# The golang build stage
```
esmet, line 64: Separating the golang build stage and the artifact build stage allows us to build go code separately from the python code

kflynn (approved):
Hit it!

PR #3453, opened 2021-05-25
Fix the telepresence download link to be OS and ARCH aware

esmet:


kflynn (approved):
Looks good, thanks! 🙂

PR #3460, opened 2021-05-27
dtest_k3s: Work with newer cgroups

LukeShu:
## Description

If something's going wrong, don't silently try to retry repeatedly until `go test` finally times out and panics. Detect errors! Log things!

And that error detecting and logging gets us to the real fix: Work with Linux kernels with newer cgroups.

v2:
- Rebased to be closer to master

## Testing

When I stick a `replace` line in the Telepresence `go.mod` to point to this, the tests once again work on my laptop.
## Checklist

    + Any new features
    + Any deprecations
 
 - [x] This is unlikely to impact how Ambassador performs at scale. - should only affect the tests
 
 - [x] My change is adequately tested. - no, but that's an existing problem with `dtest` being under-tested
 
      (This will be fixed soon.)
 
### DIFF
--- a/pkg/dtest_k3s/k3s.go
+++ b/pkg/dtest_k3s/k3s.go
@@ -287,7 +287,7 @@ func DockerRegistry(ctx context.Context) string {
 
 const dtestKubeconfig = "DTEST_KUBECONFIG"
 const k3sPort = "6443"
-const k3sImage = "rancher/k3s:v1.17.3-k3s1"
+const k3sImage = "rancher/k3s:v1.20.4-k3s1"
 
 const k3sMsg = `
 kubeconfig does not exist: %s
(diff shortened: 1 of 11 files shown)


pkg/dtest_k3s/k3s.go
```diff
@@ -93,13 +100,18 @@ func isKubeconfigReady(ctx context.Context) bool {
...
 		panic(err)
 	}
-	_ = cmd.Wait()
-	return cmd.ProcessState.ExitCode() == 0
+	outString := strings.TrimSpace(string(outBytes))
+	ret, err := strconv.ParseBool(outString)
+	if err != nil {
+		panic(fmt.Errorf("ParseBool: %q: %w", outString, err))
+	}
+	return ret
```
LukeShu, line 114: This is a key part of "better error handling"; it distinguishes between "`docker exec` failed" and "`docker exec` succeeded, and we're just not ready yet".

pkg/dtest_k3s/k3s.go
```diff
@@ -298,13 +310,15 @@ func Kubeconfig(ctx context.Context) string {
 
 	K3sUp(ctx)
 
-	for {
+	dlog.Printf(ctx, "Polling for k3s to be ready...")
+	for ctx.Err() == nil {
```
LukeShu, line 314: This isn't quiet error handling, but it means that this doesn't block forever if the ctx dies.

LukeShu:
Any idea what's going on with the CodeQL-Build CI?

LukeShu:
Now lives separately at https://github.com/datawire/dtest

PR #3465, opened 2021-05-28
[1.13] release machinery updates 

acookin:
This should look familiar if you've done the 2.0 PR.

This should be self-documenting with https://www.notion.so/datawire/X-Y-Z-Emissary-Checklist-eef145314ce4496c8424d15f35595a18
and if it's not then plz leave comments 4 comments :)


esmet (approved):

kflynn (approved):
I'm assuming that the bits that look like copies really are. :wink:

Also, this says we'll run a nightly on `master`. We should probably run a nightly on `master` _and_ on `release/v1.13`, right? What branch should that go on?

esmet (approved):

PR #3473, opened 2021-05-31
(from AES) Implement Deployment and ConfigMap report in Agent

leoluz:
## Description
This PR adds ConfigMaps and Deployments to the report sent to Ambassador Cloud.

## Testing
Added unit-tests to the new core-store and manually ran e2e test deploying in my personal kubeception cluster and analyzing the report sent to Ambassador Agent

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 

pkg/agent/agent.go
```diff
@@ -492,9 +500,19 @@ func (a *Agent) ProcessSnapshot(ctx context.Context, snapshot *snapshotTypes.Sna
 	}
 
 	if snapshot.Kubernetes != nil {
-		if a.podStore != nil {
-			snapshot.Kubernetes.Pods = a.podStore.GetPodsForServices(snapshot.Kubernetes.Services)
-			dlog.Debugf(ctx, "Found %d pods and %d services", len(snapshot.Kubernetes.Pods), len(snapshot.Kubernetes.Services))
+		if a.coreStore != nil {
+			if a.coreStore.podStore != nil {
+				snapshot.Kubernetes.Pods = a.coreStore.podStore.StateOfWorld()
+				dlog.Infof(ctx, "Found %d pods", len(snapshot.Kubernetes.Pods))
```
acookin, line 506: maybe add this as a debug line? cuz that'll be a lotta logs. same for the below log lines

pkg/agent/corestore.go
```diff
@@ -0,0 +1,127 @@
...
+	for _, v := range store.sotw {
+		deployments = append(deployments, v)
+	}
+	return deployments
+}
+
+// allowedNamespace will check if resources from the given namespace
+// should be reported to Ambassador Cloud.
+func allowedNamespace(namespace string) bool {
+	if namespace != NamespaceKubeSystem {
```
acookin, line 123: but what if the user is running services in kube system?

pkg/agent/corestore.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+}
+
+// NewPodStore will create a new podStore filtering out undesired resources.
+func NewPodStore(pods []*kates.Pod) *podStore {
+	sotw := make(map[string]*kates.Pod)
+	store := &podStore{sotw: sotw}
+
+	for _, pod := range pods {
+		if allowedNamespace(pod.GetNamespace()) && pod.Status.Phase != v1.PodSucceeded {
```
acookin, line 57: does it also make sense to filter out failed here?

acookin (changes requested):
overall just looks good. definitely want to fix that log line, but i'm willing to budge on the other feedback

pkg/agent/agent.go
leoluz: Definitely. 
Done.

pkg/agent/corestore.go
leoluz: I had a discussion with Alex and Max about the rules to apply while reporting to Ambassador Cloud. No one should be deploying their services in the `kube-system` namespace. I believe that the amount of people that will red flag this is higher than the ones the actually deploy services in `kube-system`

pkg/agent/corestore.go
leoluz: We also spoke about this and we found useful to send pods that are in failed state so we could inspect them and provide useful info to our users on how to address those issues.

acookin (approved):

PR #3487, opened 2021-06-02
Independent Host Actions

kflynn:
Best reviewed commit by commit. **Ask questions** if things don't make sense.

This isn't quite the final product (note that some tests are still `XFAIL`ed) but it's definitely close enough for folks to start messing with.

## New things:

- Setting `requestPolicy.insecure.action` actually works. 😂 
- We now have a `Listener` CRD that defines ports on which Ambassador will listen, and what the security model is for connections arriving on that port.
   - If you have no `Listener`s, the default is to listen on port 8080 for HTTP only...
   - ...**unless** you have already configured TLS termination, in which case we'll listen on 8080 and 8443 and redirect HTTP to HTTPS.
- We now _require_ a `Host` CRD to be present to terminate TLS.
   - Just defining a `TLSContext` alone isn't enough.
   - The minimal "just get TLS going" setup is now a `Host` that sets `tlsSecret`.
   - If you have no `Host`s at all, the default is to create an HTTP-only `Host`.
     - There's one exception to this: if you have no `Host`s, but you _do_ have a TLS secret named `fallback-self-signed-cert`, the default `Host` will use that fallback cert and redirect HTTP to HTTPS.
     - This is prep for allowing Emissary to create the fallback cert.
- `redirect_cleartext_from` in a `TLSContext` is no longer supported -- use an extra `Listener` instead!

## Separate PRs coming soon:

- Official release notes.
- **More tests** and fixes for the `XFAIL`ed tests. The configuration changes above mean a certain amount of thinking about whether some of these tests are worth keeping.
- Label selectors for `Listener`s, `Host`s, and `Mapping`s, to limit the associations that get set up.
- It is very likely to become mandatory to use `Mapping.spec.host`, or the upcoming label selector, so that the default behavior for a `Mapping` isn't to be associated with the entire world.
- `requestPolicy.insecure.additionalPort` in the `Host` will be removed -- use an extra `Listener` instead! 
- The `tls` `Module` will be desupported -- use `Host`/`TLSContext` instead!
- The `tls` element of the `ambassador` `Module` will be desupported -- use `Host`/`TLSContext` instead!
 
## Examples

Consider two `Host`s and a `Mapping` (secrets and services are not shown, but are present):

```yaml
---
kind: Host
metadata:
  name: h1-host
spec:
  hostname: h1.example.com
  tlsSecret:
    name: h1-secret
  requestPolicy:
    insecure:
      action: Route
---
kind: Host
metadata:
  name: h2-host
spec:
  hostname: h2.example.com
  tlsSecret:
    name: h2-secret
  requestPolicy:
    insecure:
      action: Redirect
---
kind: Mapping
metadata:
  name: qotm-mapping
spec:
  prefix: /quote/
  service: quote.ambassador
```

(Note that these `Host`s would've tried to do ACME in the old world, but in the new world, they won't.) The old world would generate something this (with the `/ambassador/v0` routes not shown):

```
LISTENER ambassador-listener-8443 on port 8443 (chains 2)
  CHAIN * (tls cert h1-secret, A7AA687AF455235146377741ED5E6C86D951D48D)
    VHost ambassador-listener-8443-h1.example.com-forced-star
      https://*/quote/ ROUTE cluster_quote_ambassador_default
      implicit-http://*/quote/ ROUTE cluster_quote_ambassador_default
  CHAIN h2.example.com (tls cert h2-secret, A7AA687AF455235146377741ED5E6C86D951D48D)
    VHost ambassador-listener-8443-h2.example.com
      https://h2.example.com/quote/ ROUTE cluster_quote_ambassador_default
      implicit-http://h2.example.com/quote/ REDIRECT HTTPS

LISTENER ambassador-listener-8080 on port 8080 (chains 2)
  CHAIN * (any)
    VHost ambassador-listener-8080-h1.example.com-forced-star
      implicit-http://*/quote/ ROUTE cluster_quote_ambassador_default
  CHAIN h2.example.com (any)
    VHost ambassador-listener-8080-h2.example.com
      implicit-http://h2.example.com/quote/ REDIRECT HTTPS```
```

Some help to read this output:
- `CHAIN $domain (tls cert $certname, $certfingerprint)` is a filter chain that
   - does a server name match for `$domain`
   - requires a TLS connection
   - will offer cert `$certname` on termination (and `$certname` has MD5 fingerprint `$certfingerprint`)

- `CHAIN $domain (any)` is a filter chain that
   - does a server name match for `$domain`
   - does _not_ require TLS (and, in fact, won't allow it -- the TLS inspector is not active here in the old world)
   - As it happens, these will only function if `$domain` is `*` -- you can't do server name matching without TLS.

Both chains here have a `VHost` with a name ending in `-forced-star`, with a wildcard domain: this is really the `h1-host` chain, but overwritten with the wildcard. Note also that in the 8080 listener, the `CHAIN h2.example.com (any)` can't actually match anything, as noted above. That means that the only functional chain on port 8080 is the wildcard chain, which has the `ROUTE` action from `h1-host`. That's the original bug.

In the new world, we don't do the `-forced-star` thing (since `Host`s and `Listener`s can tell us where we're listening, and for what), and we _do_ split apart the insecure actions into separate chains:
 
```
LISTENER ambassador-listener-8080 on port 8080 (chains 3)
  CHAIN h1.example.com (tls)
    VHost ambassador-listener-8080-h1.example.com
      https://h1.example.com/quote/ ROUTE cluster_quote_ambassador_default
      implicit-http://h1.example.com/quote/ ROUTE cluster_quote_ambassador_default
  CHAIN h2.example.com (tls)
    VHost ambassador-listener-8080-h2.example.com
      https://h2.example.com/quote/ ROUTE cluster_quote_ambassador_default
      implicit-http://h2.example.com/quote/ REDIRECT HTTPS
  CHAIN * (any)
    VHost ambassador-listener-8080-h1.example.com
      https://h1.example.com/quote/ ROUTE cluster_quote_ambassador_default
      implicit-http://h1.example.com/quote/ ROUTE cluster_quote_ambassador_default
    VHost ambassador-listener-8080-h2.example.com
      https://h2.example.com/quote/ ROUTE cluster_quote_ambassador_default
      implicit-http://h2.example.com/quote/ REDIRECT HTTPS

LISTENER ambassador-listener-8443 on port 8443 (chains 3)
  CHAIN h1.example.com (tls)
    VHost ambassador-listener-8443-h1.example.com
      https://h1.example.com/quote/ ROUTE cluster_quote_ambassador_default
      implicit-http://h1.example.com/quote/ ROUTE cluster_quote_ambassador_default
  CHAIN h2.example.com (tls)
    VHost ambassador-listener-8443-h2.example.com
      https://h2.example.com/quote/ ROUTE cluster_quote_ambassador_default
      implicit-http://h2.example.com/quote/ REDIRECT HTTPS
  CHAIN * (any)
    VHost ambassador-listener-8443-h1.example.com
      https://h1.example.com/quote/ ROUTE cluster_quote_ambassador_default
      implicit-http://h1.example.com/quote/ ROUTE cluster_quote_ambassador_default
    VHost ambassador-listener-8443-h2.example.com
      https://h2.example.com/quote/ ROUTE cluster_quote_ambassador_default
      implicit-http://h2.example.com/quote/ REDIRECT HTTPS
```

Here, note that we have chains separately naming `h1.example.com`, `h2.example.com`, and `*`. The `*` chain is only for HTTP, so it has routes separated out by `VHost`s. Each `VHost` uses a `domain` to pull out only the relevant hostnames.

One effect of no longer having the `-forced-star` entry is that you really do need to supply the correct `:authority` even for HTTP! However, since it's easy to add a `Host` with a wildcard `hostname`, this shouldn't be a problem if the wildcard behavior is actually desired.


python/ambassador/envoy/v2/v2route.py
```diff
@@ -419,62 +579,31 @@ def __init__(self, config: 'V2Config', group: IRHTTPMappingGroup, mapping: IRBas
...
+        route_hosts = self["_host_constraints"]
 
-                    else:
-                        # The exact_match that this route is looking for isn't matched by any of the
-                        # hostglobs; so this route is unreachable.
-                        hostglobs = set()
-                        break # hostglobs is empty, no point in doing more work
+        self.logger.debug(f"    - matches_domains: route_hosts {', '.join(sorted(route_hosts))}")
+        self.logger.debug(f"    - matches_domains: domains {', '.join(sorted(domains))}")
+
```
kflynn, line 592: All these `debug()`s can go.

acookin (approved):
yay!
the sync review was really helpful for this, then that example in the description was great when coming back to this the next day, thanks for that.

PR #3497, opened 2021-06-03
ready: Add option to enable envoy readiness endpoint from worker

jfrabaute:
## Description
default).

pool.

admin port.

This listener is disabled by default. the config "read_port" can be used

## Related Issues

## Testing
Manual tests


jfrabaute:
Hi,

I'm thinking about abandoning this PR in favor of the one for the master/2.x branch.
Is that ok?

PR #3500, opened 2021-06-03
[v1.13.7] Chart and manifest image updates

esmet:
Update charts and manifests to use the newly release images for 1.13.7


acookin (approved):
yes much more right 😬

PR #3528, opened 2021-06-09
Deprecate the tls field on the Ambassador module

esmet:
## Description

## Related Issues
List related issues.

## Testing

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 
### DIFF
--- a/python/ambassador/ir/irtls.py
+++ b/python/ambassador/ir/irtls.py
@@ -94,24 +94,9 @@ class TLSModuleFactory:
 
             amod_tls = amod.get('tls', None)
 
+            # Check for an Ambassador module tls field so that we can warn the user that this field is deprecated!
             if amod_tls:
-                # XXX What a hack. IRAmbassadorTLS.from_resource() should be able to make
-                # this painless.
-                new_args = dict(amod_tls)
-                new_rkey = new_args.pop('rkey', amod.rkey)
-                new_kind = new_args.pop('kind', 'Module')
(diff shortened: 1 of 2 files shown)


python/ambassador/ir/irtls.py
```diff
@@ -94,24 +94,9 @@ def load_all(cls, ir: 'IR', aconf: Config) -> None:
...
-                                                kind=new_kind,
-                                                name=new_name,
-                                                location=new_location,
-                                                **new_args)
-
-                ir.logger.debug("TLSModuleFactory saving TLS module from Ambassador module: %s" % ir.tls_module.as_json())
+                ir.post_error("The 'tls' field on the Ambassador module is deprecated! Please use a TLSContext instead https://www.getambassador.io/docs/edge-stack/latest/topics/running/tls/#tlscontext")
 
         # Finally, if we have a TLS Module, turn it into a TLSContext.
         if ir.tls_module:
```
kflynn, line 102: If `tls` `Module`s are also on the list (I think they are) you can search for `tls_module` to turn that off, too. Or just `ir.post_error` here like you did above.

kflynn (approved):
Hit it! Also note the comment about the `tls` `Module`.

python/ambassador/ir/irtls.py
esmet: Ah! OK I will submit a new PR for this

PR #3534, opened 2021-06-10
More IHA improvements

kflynn:
- Make domain-handling more straightforward, especially for the HTTP chain
- Introduce `Listener.spec.hostBinding`, as a mandatory field, and use it.
- Actually use `Mapping.spec.selector`. For a `Mapping` and a `Host` to associate, at least one of the following must be true:
    - The `Host` specifies a `selector`, and the `Mapping` has matching `labels`.
    - The `Mapping` specifies a `host` which matches the `Host'.
- Many test tweaks to line up with the new world.

There's always more test work to do; those will follow in separate PRs.


python/ambassador/envoy/v2/v2listener.py
```diff
@@ -551,7 +553,7 @@ def finalize_tcp(self) -> None:
                 tcp_filter = {
                     'name': 'envoy.filters.network.tcp_proxy',
                     'typed_config': {
-                        '@type': 'type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy',
+                        '@type': 'type.googleapis.com/envoy.config.filter.network.tcp_proxy.v2.TcpProxy',
```
kflynn, line 566: This really should've been a separate commit. Sorry about that!

python/ambassador/envoy/v3/v3listener.py
```diff
@@ -582,17 +584,9 @@ def finalize_tcp(self) -> None:
...
                     filter_chain_match["transport_protocol"] = "tls"
 
-                    # ...and match server names.
-                    #
-                    # (The if here is because if chain_hosts is empty, it means '*'. If there are
-                    # no specific names in the list, or if one of the names is actually '*', don't 
-                    # bother matching: just accept everything.)
-                    if (len(chain_hosts) > 0) and ('*' not in chain_hosts):
-                        filter_chain_match['server_names'] = chain_hosts
-
```
kflynn, line 588: Don't panic: this is a move, not a deletion. We handle `server_names` below now.

python/ambassador/envoy/v2/v2listener.py
```diff
@@ -249,6 +249,15 @@ def add_tcp_group(self, irgroup: IRTCPMappingGroup) -> None:
...
             for host in sorted(self.config.ir.get_hosts(), key=lambda h: h.hostname):
+                # They're asking for a hostname match here, which _cannot happen_ without
+                # SNI -- so don't take any hosts that don't have a TLSContext.
+
+                if not host.context:
+                    if self._log_debug:
+                        self.config.ir.logger.debug("V2Listener %s @ %s TCP %s: skip %s", 
+                                                    self.name, self.bind_to, group_host, host)
+                    continue
+
```
kflynn, line 261: If you're not going to set up SNI, it's fundamentally nonsensical to try a host-based TCP match.

python/ambassador/envoy/v2/v2listener.py
```diff
@@ -746,6 +750,36 @@ def compute_routes(self) -> None:
                                 logger.debug("      %s - %s: drop from %s %s%s",
                                              matcher, action, self.name, hostname, extra_info)
 
+            # If we're on Edge Stack and we don't already have an ACME route, add one.
+            if self.config.ir.edge_stack_allowed and not found_acme:
```
kflynn, line 760: This needs to go away -- Edge Stack should inject this, rather than having us have magic Edge Stack functionality in Emissary.

python/ambassador/envoy/v2/v2listener.py
```diff
@@ -868,13 +868,11 @@ def finalize_http(self) -> None:
             # OK, we have the filter_chain variable set -- build the Envoy virtual_hosts for it.
 
             for host in chain.hosts.values():
-                # This bit of rank paranoia is probably unnecessary: it boils down to 
-                # double-checking host constraints, and making sure that no internal keys
```
kflynn, line 778: ...and double-checking the host constraints borked the ACME hole-puncher, because we don't synthesize the hole-puncher with any host right now. It's not necessary any more -- after other IHA changes, you can't get a route in a chain anymore without a hostname match.

pkg/api/getambassador.io/v2/host_types.go
```diff
@@ -94,7 +94,7 @@ type HostSpec struct {
 	// Hostname by which the Ambassador can be reached.
 	Hostname string `json:"hostname,omitempty"`
 
-	// Selector by which we can find further configuration. Defaults to hostname=$hostname
```
kflynn, line 97: Nope, it defaults to empty. The Edge Stack ACME client does the wrong thing currently, but it's about to get fixed.

python/ambassador/ir/irambassador.py
```diff
@@ -410,7 +410,7 @@ def add_mappings(self, ir: 'IR', aconf: Config):
                 name = "internal_%s_probe_mapping" % name
 
                 mapping = IRHTTPMapping(ir, aconf, rkey=self.rkey, name=name, location=self.location,
-                                        timeout_ms=10000, **cur)
+                                        timeout_ms=10000, host="*", **cur)
```
kflynn, line 413: Same deal. Some bewildering errors if we don't do this -- also note that a user who doesn't want `host: *` for these `Mapping`s for the diagnostics, `check_ready`, etc., can set those things as disabled and add `Mapping`s by hand.

python/ambassador/envoy/v2/v2route.py
```diff
@@ -697,7 +697,10 @@ def generate_headers(config: 'V2Config', mapping_group: IRHTTPMappingGroup) -> L
                     # (This works without the user marking it as such because '*' isn't
                     # valid in DNS names, so we know that treating a name with a '*' as
                     # as exact match will always fail.)
-                    if header_value.startswith('*'):
+                    if header_value == "*":
+                        # This is actually a noop, so just don't include this header.
+                        continue
```
kflynn, line 702: Without this, `host: "*"` ends up rendering as `suffix_match: ""`, which... ew.

acookin (approved):

acookin (approved):

PR #3545, opened 2021-06-16
[WIP] Migrate canary helm chart updates to Emissary

AliceProxy:
## Description

Updates to the helm chart that will allow for the optional creation of a new deployment for canary pods that can have a separate image, and additional environment variables from the main deployment while using all of its other resources.

### New values:

new `canary` section to hold all of the following values.

- `canary.enabled`: When true creates a canary deployment that creates canary pods and a canary Ambassador service that will only route traffic to pods from the canary deployment. Everything excluding the following new values are shared with the main deployment.

- `canary.mixPods`: When true allows for requests to the main Ambassador service to hit **EITHER** the pods from the main deployment or pods from the canary deployment.

- `canary.replicaCount`: Determines the number of Canary pods to be created.

- `canary.image.repository`: Allows for setting the repository for the canary image.

- `canary.image.tag` When set allows for the canary deployment to use a different tag for the image. 

- `canary.envRaw`: Takes the place of `envRaw` for the canary deployment. This ensures that the main `env` will be copied over to the Canary deployment, but gives a way to provide additional environment variables for only the canary deployment. 
> NOTE:  do not put a duplicate environment variable in `canar.envRaw` if it exists in `env`

`_helpers.tpl`:
New `ambassador.canaryImage` vaue that  will set the imgae for the canary deployment. Gives you the ability to overwrite the `image.tag` and/or the `image.repository` for the canary deployment.

## Testing
Manual testing is all that has been done at the moment until a reviewer can look over this PR.

## Checklist

 - [ ] I made sure to update `./ambassador/CHANGELOG.md`.
   
    + Any new features
    + Whenever the Envoy version changes
    + Any deprecations
 
    + Any change to how we talk to Redis
 

AliceProxy:
Squashed into different branch targeting 2.0

PR #3604, opened 2021-07-14
Visibility into what ambex is doing with xDS

kflynn:
Provide some visibility into the snapshots that `ambex` is using to configure Envoy. In particular, this lets us see exactly what endpoints are actually being sent to Envoy.

This change works much like the snapshots logged by `diagd`: snapshots named `ambex-#.json` get written into `/ambassador/snapshots`. The current snapshot is `ambex-1.json`, and the snapshots get older as the number gets larger. Every time a new snapshot gets written, the older snapshots get renamed to the next higher number.

`AMBASSADOR_AMBEX_SNAPSHOT_COUNT` controls how many snapshots are saved. The default is currently 30. If you set `AMBASSADOR_AMBEX_SNAPSHOT_COUNT` to 0, no snapshots are saved at all.

This change also re-enables `make deploy` for 1.13.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,7 +68,8 @@ Please see the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest
 
 ### Emissary Ingress and Ambassador Edge Stack
 
-(no changes yet)
+- Change: Envoy-configuration snapshots get saved (as ambex-#.json) in /ambassador/snapshots. The number of snapshots is 
+  controlled by the `AMBASSADOR_AMBEX_SNAPSHOT_COUNT` environment variable; set it to 0 to disable. The default is 30.
 
 ## [1.13.9] June 30, 2021
 [1.13.9]: https://github.com/emissary-ingress/emissary/compare/v1.13.8...v1.13.9
--- a/cmd/ambex/main.go
+++ b/cmd/ambex/main.go
@@ -47,8 +47,10 @@ import (
 	"net"
 	"os"
 	"os/signal"
+	"path"
 	"path/filepath"
 	"reflect"
+	"strconv"
 	"strings"
 	"syscall"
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -63,6 +63,13 @@ Please see the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest
 
 ## RELEASE NOTES
 
+## Next Release
```
acookin, line 66: `make VERSION=1.13.10 release/hotfix/start` or the release process will be all jammed up for the next person :)

cmd/ambex/main.go
```diff
@@ -506,12 +613,13 @@ func update(ctx context.Context, config v2cache.SnapshotCache, configv3 v3cache.
...
 	// the ratelimiting logic decides.
-	//
-	// We also need to pay attention to contexts here so we can shutdown properly. If we didn't
-	// have the context portion, the ratelimit goroutine could shutdown first and we could end
-	// up blocking here and never shutting down.
-	select {
-	case updates <- Update{version, func() error {
+
+	log.Debugf("Created snapshot %s", version)
+	csDump(snapdirPath, numsnaps, curgen, &snapshot, &snapshotv3)
```
acookin, line 618: probably fine for hotfix but should probably not call this function for reals unless in some form of debug mode.

acookin (changes requested):
yay nice, just one nit change so the release process doesn't freak out haha

CHANGELOG.md
kflynn: Thanks for the catch!

cmd/ambex/main.go
kflynn: As discussed: we want it to run all the time. If needed you can set `AMBASSADOR_AMBEX_SNAPSHOT_COUNT=0` to disable it. 🙂

PR #3620, opened 2021-07-16
Allow overriding the stats_prefix for an AmbassadorListener

kflynn:
`AmbassadorListener.spec.statsPrefix` determines the stats prefix for a	given AmbassadorListener.
If not present,	the default depends on the protocol configured for a given AmbassadorListener:

- If HTTPS is configured, use `ingress_https`;
- Otherwise, if	HTTP is configured, use	`ingress_http`;
- Otherwise, if	TLS is configured, use `ingress_tls_$port`;
- Otherwise, use `ingress_tcp_$port`.


aidanhahn (approved):
Looks good to me!

PR #3626, opened 2021-07-22
ready: Add option to enable envoy readiness endpoint from worker

jfrabaute:
default).

pool.

admin port.

This listener is disabled by default. the config "ready_port" can be used

## Related Issues

## Testing
Manual tests


python/ambassador/envoy/v2/v2ready.py
```diff
@@ -0,0 +1,86 @@
...
+                        {
+                            'name': 'envoy.filters.network.http_connection_manager',
+                            'typed_config': {
+                                '@type': 'type.googleapis.com/envoy.config.filter.network.http_connection_manager.v2.HttpConnectionManager',
+                                "access_log": [
+                                    {
+                                        "name": "envoy.access_loggers.file",
+                                        "typed_config": {
+                                            "@type": "type.googleapis.com/envoy.config.accesslog.v2.FileAccessLog",
+                                            "format": "ACCESS [%START_TIME%] \"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%\" %RESPONSE_CODE% %RESPONSE_FLAGS% %BYTES_RECEIVED% %BYTES_SENT% %DURATION% %RESP(X-ENVOY-UPSTREAM-SERVICE-TIME)% \"%REQ(X-FORWARDED-FOR)%\" \"%REQ(USER-AGENT)%\" \"%REQ(X-REQUEST-ID)%\" \"%REQ(:AUTHORITY)%\" \"%UPSTREAM_HOST%\"\n",
```
kflynn, line 53: I'm kind of torn about logging for this, honestly. 🙂  If it's present, though, I think it should honor the existing logging configuration stuff, e.g. JSON logging if that's enabled, custom formats, etc...

kflynn (changes requested):
I have one thing I'd like to see for logging, but overall, this is really cool. 🙂  Thanks!!

python/ambassador/envoy/v2/v2ready.py
jfrabaute: Thanks for the review.

You are totally right, I need to honor the logging config.

I want to add one more config info as well, regarding the ip binding.
Let me give more details on the 3 changes I'd like to make:

1/ Make the logging honor the existing logging config (default)
2/ Add an option to disable logging completely for this listener.
3/ Add an option for which ip is binded by the listener. By default I'll make it 127.0.0.1.

For this config, I'd like to make the config an object in the module config with "ready_listener" as the key and the following keys for this object:

```
ready_listener:
  port: -1 #-1 disables this listener
  nolog: false # disable logging entirely if "true"
  ip: 127.0.0.1
```

Reason for 3/: 
This listener should not need to be exposed, so 127.0.0.1 makes sense.
But as it is not used yet by diagd and the go code, exposing this listener to the k8s readiness means it must bind the external ip of the pod. That's why the option is present o change it (to 0.0.0.0 for instance) as a first step.

I'll try to update the review with those changes, unless you have some feedback.

Thanks.

jfrabaute: As in the long term, I see this enabled and used all the time, I'm wondering if an object makes sense.
Maybe just:
```
 ready_port:
 ready_log: true # default true
```
could be enough?

jfrabaute:
Hi @kflynn 

Sorry for the delay, I finally could find some time to do a few fixes and send an updated PR.

I've added code to match the current logging config on this listener.
I've added 3 configs to enable and tune this new feature:

ready_port: -1 # (default to -1 means feature disabled by default) port used by the listener to expose the /ready endpoint
ready_ip: 127.0.0.1 # (default to 127.0.0.1) ip to bind
ready_log: true/false # (default to true) logs the hit or not.

jfrabaute:
Hi @kflynn 

I just rebased the PR on top of ambassador 2.1.0
I tested and it works fine.

jfrabaute:
Sorry, I think I have a small bug due to the latest changes for the python lint.
Working on it.

jfrabaute:
I found the problem, I'll update the review soon.

jfrabaute:
PR updated/fixed.
Thanks.

jfrabaute:
Superseded by: https://github.com/emissary-ingress/emissary/pull/4253

PR #3627, opened 2021-07-22
Create a timeout field for docs endpoint

aidanhahn:
## Description
Adds a new field to the mapping CRD allowing the user to configure a timeout for a service's documentation endpoints.
This PR contains the additional schema field as well as a config generation unit test.

## Related Issues
This PR correlates to upcoming changes to apro to fix issue #2722

## Testing
- The complete change across emissary and apro was manually tested with a [slow to respond service](https://github.com/aidanhahn/slowsvc)
- Unit tests were added in this PR for config generation

## Checklist

   (This change does not seem to warrant a CHANGELOG.md update)
    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 
### DIFF
--- a/pkg/api/getambassador.io/v2/devportal_types.go
+++ b/pkg/api/getambassador.io/v2/devportal_types.go
@@ -54,6 +54,10 @@ type DevPortalDocsSpec struct {
 
 	// URL is the URL used for obtaining docs
 	URL string `json:"url,omitempty"`
+
+	// Timeout speifies the amount of time devportal will wait
+	// for the downstream service to report an openapi spec back
+	Timeout int `json:"timeout_ms,omitempty"`
 }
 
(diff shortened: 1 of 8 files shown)


pkg/api/getambassador.io/v2/devportal_types.go
```diff
@@ -54,6 +54,10 @@ type DevPortalDocsSpec struct {
 
 	// URL is the URL used for obtaining docs
 	URL string `json:"url,omitempty"`
+
+	// Timeout speifies the amount of time devportal will wait
```
kflynn, line 58: Nit: `s/speifies/specifies/` 🙂

kflynn (approved):
One spelling nit. 🙂 

(I could bikeshed the test loop but no point. 😉 )

kflynn:
Forced DCO since Aidan is an Ambassador Labs employee. 🙂

PR #3639, opened 2021-07-27
Add support for initial_metadata to AuthService

rnburn:
## Description
Support specifying initial_metadata in AuthService.

When specified, envoy will add the initial_metadata to its ext auth request.

From Envoy's [docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/grpc_service.proto.html#config-core-v3-grpcservice):
> initial_metadata
(repeated config.core.v3.HeaderValue) Additional metadata to include in streams initiated to the GrpcService. This can be    used for scenarios in which additional ad hoc authorization headers (e.g. x-foo-bar: baz-key) are to be injected. For more    information, including details on header value syntax, see the documentation on custom request headers.

## Testing
Added a kat test case to verify the auth service receives the metadata.

## Checklist

 - [ x] I made sure to update `CHANGELOG.md`.
   
    + Any new features
    + Any deprecations
 
 - [ x] This is unlikely to impact how Ambassador performs at scale.
 
      (This will be fixed soon.)
 

pkg/api/getambassador.io/v2/authservice_types.go
```diff
@@ -51,6 +51,7 @@ type AuthServiceSpec struct {
 	AllowedRequestHeaders       []string                  `json:"allowed_request_headers,omitempty"`
 	AllowedAuthorizationHeaders []string                  `json:"allowed_authorization_headers,omitempty"`
 	AddAuthHeaders              map[string]BoolOrString   `json:"add_auth_headers,omitempty"`
+	InitialMetadata             map[string]BoolOrString   `json:"initial_metadata,omitempty"`
```
kflynn, line 54: Would it make sense to just make this a `map[string]string`? `AddAuthHeaders` uses the `bool` option in a very specific way, and we generally would rather not add more of those...

kflynn (changes requested):
One quick change, if you could -- thanks very much!! 🙂

pkg/api/getambassador.io/v2/authservice_types.go
rnburn: done

cortopy:
I also need to add `initial_metadata`. Any chance of getting this merged?

the-wondersmith:
🍃 Housekeeping 🧹

PR #3644, opened 2021-07-29
enable pytest-kat/envoy/envoy-v2 in github-actions

anodelman:
- split up kat tests to be able to run on the provided github-actions
  hosts
  * alphabetically (a-h, i-p, q-z)
  * support a new command line option for pytest (--letter-range) to
    indicate which set of tests to execute

Signed-off-by: Alice Nodelman <anode@datawire.io>

## Description
Enable kat/envoy/envoy-v2 tests in github-actions.

## Related Issues
List related issues.

## Testing
Enabled push and executed in github-actions.  Ran green.

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 
### DIFF
--- a/python/tests/kat/t_mappingtests.py
+++ b/python/tests/kat/t_mappingtests.py
@@ -305,7 +305,7 @@ name:  {self.name}
 hostname: "*"
 prefix: /{self.name}/status/
 rewrite: /status/
-service: httpbin.default
+service: httpbin.plain-namespace
 """)
 
     def queries(self):
@@ -341,7 +341,7 @@ name:  {self.name}
 hostname: "*"
 prefix: /{self.name}/status/
 rewrite: /status/
-service: httpbin.default
+service: httpbin.plain-namespace
 """)
 
     def queries(self):
@@ -370,7 +370,7 @@ name:  {self.name}
 hostname: "*"
 prefix: /{self.name}/status/
 rewrite: /status/
-service: httpbin
+service: httpbin.plain-namespace
 """)
 
     def queries(self):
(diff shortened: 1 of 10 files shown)


python/tests/kat/t_mappingtests.py
```diff
@@ -305,7 +305,7 @@ def config(self):
 hostname: "*"
 prefix: /{self.name}/status/
 rewrite: /status/
-service: httpbin.default
+service: httpbin.plain-namespace
```
kflynn, line 308: Huh. Why is this necessary?

python/tests/kat/test_ambassador.py
```diff
@@ -1,47 +1,57 @@
...
+	import t_ip_allow_deny
+	import t_listeneridletimeout
+	import t_loadbalancer
+	import t_logservice
+	import t_lua_scripts
+	import t_mappingtests
+	import t_max_req_header_kb
+	import t_no_ui
+	import t_optiontests
+	import t_plain
```
kflynn, line 39: The `plain` thing still worries me. How about comment in here that `plain`, `mappingtests`, and `optiontests` must go together... or, even better, try moving 

```
	import t_mappingtests
	import t_optiontests
```

into `t_plain.py` and let's see what blows up.

kflynn (changes requested):
One major question and one thought about the `plain` thing -- pretty cool though! I didn't know how to add options to `pytest`... 😂

python/tests/kat/t_mappingtests.py
anodelman: There's a question for @acookin

python/tests/kat/test_ambassador.py
anodelman: Will do!

python/tests/kat/t_mappingtests.py
acookin: because that's the naemsapce that the actual service gets deployed to. default was only working because another test was creating it

acookin (approved):

acookin (approved):
:shipit:

PR #3655, opened 2021-08-02
make target to promote hotfix build

acookin:
Signed-off-by: Alix Cook <alixcook@datawire.io>

## Description

Make target to make hotfixes easier

## Related Issues
List related issues.

## Testing
Can't really test it, we'll test it when we do the next hotfix :)

## Checklist

    + Any new features
    + Any deprecations

      (This will be fixed soon.)

### DIFF
--- a/builder/builder.mk
+++ b/builder/builder.mk
@@ -858,7 +858,7 @@ release/promote-oss/pr-to-passed-ci:
 		printf "$(CYN)==> $(GRN)Promoting $(BLU)$$commit$(GRN) => $(BLU)$$dev_version$(GRN) in S3...$(END)\n" ;\
 		echo "$$dev_version" | aws s3 cp - s3://datawire-static-files/passed-pr/$$commit ;\
 	}
-.PHONY: release/promote-oss/dev-to-passed-ci
+.PHONY: release/promote-oss/pr-to-passed-ci
 
 release/promote-oss/to-hotfix:
 	@test -n "$(RELEASE_REGISTRY)" || (printf "$${RELEASE_REGISTRY_ERR}\n"; exit 1)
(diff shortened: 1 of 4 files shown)


builder/builder.mk
```diff
@@ -845,6 +845,59 @@ release/promote-oss/dev-to-passed-ci:
...
+		commit=$$(git rev-parse HEAD) ;\
+		dev_version=$$(aws s3 cp s3://datawire-static-files/dev-builds/$$commit -) ;\
+		if [ -z "$$dev_version" ]; then \
+			printf "$(RED)==> found no dev version for $$commit in S3...$(END)\n" ;\
+			exit 1 ;\
+		fi ;\
+		printf "$(CYN)==> $(GRN)Promoting $(BLU)$$commit$(GRN) => $(BLU)$$dev_version$(GRN) in S3...$(END)\n" ;\
+		echo "$$dev_version" | aws s3 cp - s3://datawire-static-files/passed-pr/$$commit ;\
+	}
+.PHONY: release/promote-oss/dev-to-passed-ci
```
kflynn, line 861: This should be `pr-to-passed-ci`, right? 

Also, the Circle config looks to me like this'll only run on PRs to `master` or to `rel/vX.Y.Z`, right?

kflynn (changes requested):
Two kinda nitpicky questions. 

Overall, though, looks good, thanks!! The idea is that you run `make release/promote-oss/to-hotfix` manually, right?

builder/builder.mk
acookin: whoops and whoops. fixed both things

kflynn (approved):
Hit it! and thanks. 🙂

kflynn:
(@acookin confirmed that, yes, the idea is that you run make release/promote-oss/to-hotfix manually.)

PR #3665, opened 2021-08-05
Fix letter_range all

kflynn:
Make `--letter_range=all` be all pytests, rather than actually being just the A-H pytests.

Also add some logging about what's actually going on.

- [x] I don't need to update `CHANGELOG.md`.
- [x] I don't need to update `DEVELOPING.md`.


anodelman (changes requested):
Please update the fall through state to something like:

if letter_range not in ["all", "ah", "ip", "qz"]:
    print("Unknown test file name letter range: %s!" % letter_range)
    sys.exit(1)

anodelman (dismissed):

acookin (approved):

PR #3679, opened 2021-08-09
[2.0.1] Support alt stats name.

kflynn:
Our cluster naming is very hard to explain, and people would rather have control over
what shows up in the stats. Support `stats_name` in all the types where we create
clusters. Change the default stats name to be just the `service` with any
non-alpha-or-digit characters translated to underscore.

This is a breaking change.

- Support stats_name field in all the cluster-generating types.
- Link up stats_name when actually generating clusters.
- Update - and re-enable(!) - stats tests
- Quiet unneeded log.
- Update gold files

Still to come: more test coverage. Since the `statsd` and `dogstatsd` tests are updated
and functioning again, I think that this in OK for 2.0.1.

### DIFF
--- a/pkg/api/getambassador.io/v3alpha1/roundtrip_test.go
+++ b/pkg/api/getambassador.io/v3alpha1/roundtrip_test.go
@@ -1,0 +1,96 @@
+package v3alpha1
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"path"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// func TestAuthSvcRoundTrip(t *testing.T) {
+// 	var a []AuthService
+// 	checkRoundtrip(t, "authsvc.json", &a)
+// }
+
+// func TestDevPortalRoundTrip(t *testing.T) {
+// 	var d []DevPortal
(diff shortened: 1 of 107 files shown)


pkg/api/getambassador.io/v3alpha1/roundtrip_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"encoding/json"
+	"io/ioutil"
+	"path"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// func TestAuthSvcRoundTrip(t *testing.T) {
```
acookin, line 13: what's goin on here.....

acookin (changes requested):
looks like quite a few tests are still uncommented

pkg/api/getambassador.io/v3alpha1/roundtrip_test.go
kflynn: Right -- the `v3alpha1` roundtrip only tests the new things in `v3alpha1`. e.g. `AuthService` exists only in `v2`, so it's covered by the `v2` roundtrip test.

I've updated comments in #3681.

acookin (approved):

Issue #3380, opened 2021-05-03
Helm chart should be installable in two different namespaces with the same release name


acookin:
If you have an ambassador installed with helm in a namespace:
```
helm status -n one  ambassador
NAME: ambassador
LAST DEPLOYED: Mon May  3 12:35:58 2021
NAMESPACE: one
STATUS: deployed
```
Then you try to install the chart with the same release name in a different namespace:
```
helm install --set scope.singleNamespace=true ambassador -n two datawire/ambassador
```

The cluster-wide resources conflict with each other, and you get this error:
```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRole "ambassador-agent" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "two": current value is "one"
```

The chart should have a value that templates out the cluster wide resources so that we can enable this behavior.

hendrikhalkow:
Thanks for bringing this up, @acookin. I'd go even one step further and say that in single namespace mode, Ambassador must not create any non-namespaced objects like `ClusterRole` or `ClusterRoleBinding`. Instead, `Role` and `RoleBinding` should be used. This would automatically solve the naming conflict issue.

PR #3689, opened 2021-08-11
unique clusterrole names when singleNamespace=true

njayp:
Signed-off-by: njayp <nickjaypowell@gmail.com>

## Description
Added flow control (if-else) to give ClusterRoles unique names when singleNamespace=true.

## Related Issues
https://github.com/emissary-ingress/emissary/issues/3380

## Testing
A singleNamespace version of emissary was installed in the same cluster as a standard emissary installation. Both clusters performed "quote" functionality. A standalone singleNamespace emissary also performed "quote" functionality.

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 

CHANGELOG.md
```diff
@@ -69,6 +69,7 @@ Please see the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest
 [HTTP_JSON_V1]: https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/trace/v2/zipkin.proto#envoy-api-field-config-trace-v2-zipkinconfig-collector-endpoint-version
 
 ## RELEASE NOTES
+- Bugfix: A singleNamespace version of emissary can now be installed alongside a standard installation in the same cluster.
```
acookin, line 72: this changelog entry should go into chart/emissary-ingress/CHANGELOG.md

acookin: i now see it's already there, you can remove this entry :)

acookin (changes requested):
I think this makes sense, but we should make sure to push this in a minor version bump because this is a somewhat breaking change :)

acookin (approved):

PR #3712, opened 2021-08-20
[v1.14.0] Chart and manifest image updates

aidanhahn:
Update charts and manifests to use the newly release images for 1.14.0


kflynn (approved):
Hit it! 🙂👍

PR #3718, opened 2021-08-23
add job-image to github-actions, update to use github-actions only secrets

anodelman:
## Description
Another day, another test.  Plus, switching over to github-actions only secrets to mitigate risk.

## Related Issues
List related issues.

## Testing
Tested green on github.

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 
### DIFF
--- a/.github/workflows/execute-pytests.yml
+++ b/.github/workflows/execute-pytests.yml
@@ -83,8 +83,8 @@ jobs:
         if: ${{ matrix.test != 'unit' }}
         uses: docker/login-action@v1
         with:
-          username: ${{ secrets.DOCKER_BUILD_USERNAME }}
-          password: ${{ secrets.DOCKER_BUILD_PASSWORD }}
+          username: ${{ secrets.GH_DOCKER_BUILD_USERNAME }}
+          password: ${{ secrets.GH_DOCKER_BUILD_TOKEN }}
       - name: "Configure Kubernetes cluster and Docker registry"
         if: ${{ matrix.test != 'unit' }}
         run: |
--- a/.github/workflows/job-image.yml
+++ b/.github/workflows/job-image.yml
@@ -1,0 +1,51 @@
+name: job-image
+on:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - 'releng/**'
+      - 'charts/**'
+      - 'manifests/**'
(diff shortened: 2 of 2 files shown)


.github/workflows/execute-pytests.yml
```diff
@@ -83,8 +83,8 @@ jobs:
         if: ${{ matrix.test != 'unit' }}
         uses: docker/login-action@v1
         with:
-          username: ${{ secrets.DOCKER_BUILD_USERNAME }}
-          password: ${{ secrets.DOCKER_BUILD_PASSWORD }}
+          username: ${{ secrets.GH_DOCKER_BUILD_USERNAME }}
+          password: ${{ secrets.GH_DOCKER_BUILD_TOKEN }}
```
acookin, line 87: looks good assuming these are in keybase and are easy to find (which im sure you posted about in slack but honestly i can't remember)

.github/workflows/job-image.yml
```diff
@@ -0,0 +1,51 @@
...
+
+jobs:
+  execute-generate:
+    runs-on: ubuntu-latest
+    name: job-image
+    env:
+      AWS_ACCESS_KEY_ID: ${{ secrets.GH_AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.GH_AWS_SECRET_ACCESS_KEY }}
+      AWS_EC2_METADATA_DISABLED: true
+      DEV_REGISTRY: ${{ secrets.DEV_REGISTRY }}
```
acookin, line 30: non blocker but is there a way to do env variables set in github that aren't secrets? cuz we don't actually care about exposing this value, but we want to set it at the github repo level if that makes sense (circle didn't have a way to do this)

acookin (approved):
looks good just some genuine questions about things in github actions :)

.github/workflows/execute-pytests.yml
anodelman: Yup, totally in keybase.

.github/workflows/job-image.yml
anodelman: I'll look things over - but I think that there are only secrets same as CircleCI.

PR #3731, opened 2021-08-25
[1.14] Allow configuring logical dns service resolution

AliceProxy:
Signed-off-by: AliceProxy <aliceproxy@protonmail.com>

## Description
Allows you to configure the DNS type from `strict_dns` to `logical_dns` for service resolution
Configuration is exposed via a new field in the `Mapping` CRD `dns_type`
Currently the only supported types are `logical_dns` and `strict_dns` (The default)
If the endpoint resolver is enabled, this setting will be ignored.

## Related Issues
List related issues.

## Testing
Need to add automated tests

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 

python/ambassador/envoy/v2/v2cluster.py
```diff
@@ -51,6 +51,17 @@ def __init__(self, config: 'V2Config', cluster: IRCluster) -> None:
...
             ctype = cluster.type.upper()
+            # For now we are only allowing Logical_dns for the cluster since it is similar enough to strict_dns that we dont need any other config changes
+            # It should be easy to add the other dns_types here in the future if we decide to support them
+            allowedDnsTypes = ['STRICT_DNS', 'LOGICAL_DNS']
+            isTypeAllowed = False
+            for dnsType in allowedDnsTypes:
+                if ctype == dnsType:
+                    isTypeAllowed = True
+                    break
+            if not isTypeAllowed:
```
kflynn, line 62: This can just be

```
if dnsType not in [ "STRICT_DNS", "LOGICAL_DNS" ]:
    # scream here
```

The `in` operator, when applied to a list, does a linear search, so it's a poor choice when you have a really long list... but with such a small number, it'll be literally as fast as looking into a Python map, and way faster than doing the iteration in Python. 😉

python/ambassador/envoy/v3/v3cluster.py
```diff
@@ -51,6 +51,17 @@ def __init__(self, config: 'V3Config', cluster: IRCluster) -> None:
...
             ctype = cluster.type.upper()
+            # For now we are only allowing Logical_dns for the cluster since it is similar enough to strict_dns that we dont need any other config changes
+            # It should be easy to add the other dns_types here in the future if we decide to support them
+            allowedDnsTypes = ['STRICT_DNS', 'LOGICAL_DNS']
+            isTypeAllowed = False
+            for dnsType in allowedDnsTypes:
+                if ctype == dnsType:
+                    isTypeAllowed = True
+                    break
+            if not isTypeAllowed:
```
kflynn, line 62: Same here, of course.

kflynn (changes requested):
Minor implementation point, but again, overall looks good -- thanks! Worth writing a test for it, of course...

python/ambassador/envoy/v2/v2cluster.py
AliceProxy: Good feedback, thanks!

kflynn (commented):
Oops. Meant to approve this before merging.

PR #3753, opened 2021-08-31
automate updates to apro

aidanhahn:
## Description
The enclosed modifications consist of the following
- a script that creates the relevant rel/vX.Y branch in apro if need be, bumps the emissary version referenced within, and then pushes the update
- commands calling this script in release/promote-oss/dev-to-rc

The script was tested locally, and is being merged to the shared iterative development branch so that we can improve upon it as we continue working on the other bits that need to be altered for the release pipeline.

### DIFF
--- a/.github/workflows/promote-rc.yml
+++ b/.github/workflows/promote-rc.yml
@@ -14,6 +14,8 @@ jobs:
       AWS_EC2_METADATA_DISABLED: true
       DEV_REGISTRY: ${{ secrets.DEV_REGISTRY }}
       RELEASE_REGISTRY: ${{ secrets.RELEASE_REGISTRY }}
+      APRO_REPO_TARGET: ${{ secrets.APRO_REPO_TARGET }}
+      GIT_API_KEY: ${{ secrets.GH_API_KEY }}
     steps:
       - uses: actions/checkout@v2
         with:
@@ -23,6 +25,12 @@ jobs:
         with:
           username: ${{ secrets.GH_DOCKER_RELEASE_USERNAME }}
           password: ${{ secrets.GH_DOCKER_RELEASE_PASSWORD }}
+      - name: "make release/prep-rc"
+        run: |
+          git config --global user.email "dev@datawire.io"
+          git contig --global user.name "d6e-automaton"
+          git config --global "url.https://$GIT_API_KEY@github.com.insteadOf" https://github.com
+          make release/prep-rc
       - name: "make release/promote-oss/dev-to-rc"
         run: |
           make release/promote-oss/dev-to-rc
--- a/.github/workflows/promote-rc.yml
+++ b/.github/workflows/promote-rc.yml
@@ -27,10 +27,12 @@ jobs:
           password: ${{ secrets.GH_DOCKER_RELEASE_PASSWORD }}
       - name: "make release/promote-oss/dev-to-rc"
         run: |
+          make release/promote-oss/dev-to-rc
+      - name: Update emissary tag in apro
+        run: |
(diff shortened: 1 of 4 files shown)


releng/01-release-rc-update-apro
```diff
@@ -0,0 +1,48 @@
...
+    exit -1
+fi
+
+if [ -z "$2" ]; then
+    echo "Second argument must be a GA tag"
+    exit -1
+fi
+
+if [ -z "$APRO_REPO_TARGET" ]; then
+    echo "APRO_REPO_TARGET not set. This should be set in the environment"
```
acookin, line 14: exit non zero?

builder/builder.mk
```diff
@@ -793,6 +793,8 @@ release/promote-oss/dev-to-rc:
 		$(MAKE) VERSION_OVERRIDE=$${veroverride} push-manifests  ; \
 		$(MAKE) VERSION_OVERRIDE=$${veroverride} publish-docs-yaml ; \
 		$(MAKE) clean-manifests ; \
+		gavers=$$(echo $(RELEASE_VERSION) | sed 's/-rc.*//'); \
```
acookin, line 796: you can also use $(VERSIONS_YAML_VERSION), and it's actually more reliable for -ea releases

acookin (changes requested):
smol comments but otherwise looks good

.github/workflows/promote-rc.yml
```diff
@@ -26,11 +28,16 @@ jobs:
       - name: "make release/promote-oss/dev-to-rc"
         run: |
           make release/promote-oss/dev-to-rc
+      - name: Update emissary tag in apro
```
acookin, line 31: you can probably make this a different job in the workflow. that way if this step fails we still get RC info and crap in the slack message

acookin: also this is double true becuase at this point the tree is dirty so RELEASE_VERSION won't be something you like i don't think

acookin (changes requested):
one minor thing about moving that step into it's own job in the workflow

.github/workflows/promote-rc.yml
aidanhahn: Good point!

.github/workflows/promote-rc.yml
```diff
@@ -47,3 +46,18 @@ jobs:
...
             ]
+
+  update-apro-with-rc:
+    runs-on: ubuntu-latest
+    needs: promote-to-rc
+    name: update-apro-with-rc
+    env:
+      GIT_API_KEY: ${{ secrets.GH_API_KEY }}
+      APRO_REPO_TARGET: ${{ secrets.APRO_REPO_TARGET }}
+    steps:
```
acookin, line 57: you need to make sure to do the git credentials step in here too (and checkout)

acookin (changes requested):

.github/workflows/promote-rc.yml
```diff
@@ -47,3 +46,24 @@ jobs:
...
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: "setup git config"
+        run: |
+          set -x
+          git config --global user.email "dev@datawire.io"
```
anodelman, line 64: Let's get this into a secret along with the user.name - be easier if we ever have to change to a different account.

kflynn (changes requested):
Looks sane, I think! I agree with @anodelman's commentary about secrets, though.

anodelman (approved):

PR #3758, opened 2021-08-31
fix flakey kat test

aidanhahn:
## Description
A test in CI is breaking because it is receiving a slightly different error message than expected. This test expects "connection reset by peer" but is instead getting "broken pipe". The actual error message it is getting means roughly the same thing and this difference can happen if the server never receives TCP FIN.

## The Fix
Add "broken pipe" to the list of expected error messages.


acookin (approved):

kflynn (approved):
Hit it!

PR #3772, opened 2021-09-07
Try Alpine 3.14.

kflynn:
Try Alpine 3.14. This permits updating a lot of Python dependencies (yay), but it also requires actually compiling orjson (augh) because it's just simpler that way.


LukeShu:
Not going in to a 1.x; see-also https://github.com/emissary-ingress/emissary/pull/3779

PR #3779, opened 2021-09-10
[2.x] Upgrade to Alpine 3.14

kflynn:
- Switch to Alpine 3.14 instead of 3.12.
- Update Python dependencies where we can.
   - This includes actually compiling orjson (augh) because it's just simpler that way.
- Don't use the builder shell for pip-compile or for mypy.
- Make mypy pass on the code, given that Python 3.9 gets more strict.
   - Don't run mypy on the tests, because that's currently a disaster.


LukeShu:
Thanks for this branch, I'm stealing it.

PR #3786, opened 2021-09-13
[1.14] DevPortal: allow preserving the server from the OpenAPI spec

AliceProxy:
## Description
Adds a CRD to the `DevPortal` resource that is needed for a feature that will allow you to toggle whether the DevPortal uses a Server definition specified in the opapi docs or if the server is built from Ambassador's Host.


kflynn (approved):
These really need to move. 😐  Landing this one, then we can sort the move out later...

PR #3789, opened 2021-09-14
[2.0] DevPortal: allow preserving the server from the OpenAPI spec

AliceProxy:
## Description
New addition in the DevPortal CRD to configure making the DevPortal respect the server definitions in provided openapi docs rather than generating them.


kflynn (approved):
Hit it!

PR #3792, opened 2021-09-14
Envoy build fixes

kflynn:
Couple of small, but relevant, build fixes around Envoy:

- `make clobber` needs to drop the `envoy-build` Docker volume.
   - `make generate` ends up using Bazel to drive generating Golang code from the Envoy protobuf definitions.
   - Bazel uses a cache to try to speed things up, but if you switch Envoy versions (perhaps by switching between Emissary branches) without dropping the cache, things can get out of sync.
   - If the cache is out of sync, `make generate` can fail in weird ways, like leaving around generated Go files that should've been removed.

- Quieten down the `rsync`s that are part of building Envoy.
   - The `rsync`s are necessary, because we need to run Envoy-build stuff in Docker to make life easier -- however, they were producing (literally) tens of thousands of lines of output in the CI logs. Removing `-v` and `--progress` makes life _much_ better.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -47,11 +47,11 @@ ever find anything missing from this list.
  - docker (make sure you can run docker commands as your dev user without sudo)
  - bash
  - rsync (with the --info option)
- - golang 1.13
- - python 3.7+
+ - golang 1.15
+ - python 3.8 or 3.9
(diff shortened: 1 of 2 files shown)


aidanhahn (approved):
Looks good; Send it!

PR #3803, opened 2021-09-17
Upgrade dtest

LukeShu:
## Description
A newer better dtest lives at github.com/datawire/dtest, upgrade to use that.

Related: Backporting a change from here to there: https://github.com/datawire/dtest/pull/3

## Related Issues
https://github.com/datawire/dtest/pull/3

## Checklist

    + Any new features
    + Any deprecations
 
 - [x] This is unlikely to impact how Ambassador performs at scale. - should only affect tests
 
      (This will be fixed soon.)
 

LukeShu:
From Slack:

Flynn:
> I’m not entirely sure I follow — are you folding in changes to our local `dtest` in early commits and then swapping out our local `dtest` for the One True Version in later commits?

Luke:
> Yes, first I merge the commits that were originally https://github.com/datawire/apro/pull/2666 / https://github.com/emissary-ingress/emissary/pull/3460 and were split off in to the separate dtest repo since they weren't getting merged.  This serves two purposes:
> * help inform me chances that did happen in emissary.git since the split happened (hence dtest PR#3)
> * update all of the callers, since a few of the function signatures changed, and dtest.git obviously doesn't contain those changes
>
> Then, after that merge, I swap it out for the One True Version, now that all of the callers have been updated to the One True Version's function signatures.

kflynn (approved):
Looks OK, I think?

PR #3804, opened 2021-09-17
Add the "this should not happen in CI" checks back in

LukeShu:
## Description

In https://github.com/emissary-ingress/emissary/pull/3644/commits/855c03e6e5f36616a27bf77040afe79d5315d6ea, Alix removed checks that the images we're testing don't change _while we're testing them_ (the "this should not happen in CI" checks).  Without that, we are probably not testing what we think we're testing, and could very easily put out a broken release.

This PR:
- Reverts removing those checks
- Fixes the `.dockerignore` so that those checks don't fail
- Rewrites `dbuild.sh` in Go as `dsum`, because I couldn't follow the `dbuild.sh` output in order to debug the above; it was often not clear which command's output I was looking at.

## Testing
CI should cover it; this is just making CI more strict.

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 
 - [x] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - yes I did


kflynn (approved):
Hit it!

anodelman (approved):
Seems reasonable.

Issue #3593, opened 2021-07-02
Ambassador Agent fails to report snapshot to Ambassador Cloud on "large" snapshots


acookin:
In larger installations of ambassador, the agent hits the default grpc max message size when trying to relay the snapshot to the cloud.

Error:
```
time="2021-07-01 15:22:33" level=warning msg="failed to report: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (8884732 vs. 4194304)" 
```

We should probably stream the snapshot in chunks (we could bump the default message size, but that's a bandaid, not a solution.)

This error was reported by an ambassador cloud user in slack.

andrewnazarov:
I was that reporter. Feel free to ping me here if you need some additional information:)

ankit-ls:
@acookin I am observing the same issue consistently.

> ```{"log":"time=\"2021-07-20 07:16:01\" level=warning msg=\"failed to report: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (11909674 vs. 4194304)\" func=\"github.com/datawire/ambassador/pkg/agent.(*Agent).MaybeReport.func1.1\" file=\"/go/pkg/agent/agent.go:474\" CMD=agent PID=1\n","stream":"stderr","time":"2021-07-20T07:16:01.996242695Z"}```

Any timeline for a fix on this? We use helm charts to install the agent, so a parameter to increase the message size [here](https://github.com/emissary-ingress/emissary/blob/master/charts/emissary-ingress/values.yaml#L361-L373) (potentially as an `agent.env[0]` variable) should also be sufficient for us, as a short term fix.

acookin:
@ankit-ls we're working on this issue right now, will update here when we've solved it

PR #3805, opened 2021-09-17
Large snapshot fix

LukeShu:
## Description

- First, un-delete the agent tests that got moved to apro.git (sibling PR deleting them from apro.git: https://github.com/datawire/apro/pull/2796).  To do this, I grabbed their full history from apro.git, because I has going crazy trying to look at the diffs each time they bounced around.  Because the preserves original commit information, I had to manually set the DCO to pass.
- Then a stream of commits updating those tests to work again.
- The penultimate commit changes the test to fail by triggering the large-snapshot problem.
- The ultimate commit fixes the large-snapshot problem.

## Related Issues
- Fixes https://github.com/emissary-ingress/emissary/issues/3593
- Depends on https://github.com/datawire/saas_app/pull/1745
- Sibling PR deleting moved code from apro.git https://github.com/datawire/apro/pull/2796

## Testing

I verified that `HEAD^^` passes locally (that I successfully updated the tests, that `HEAD^` fails locally and the large-snapshot problem is triggered, and finally that `HEAD` passes.

## Checklist

 - [x] I made sure to update `CHANGELOG.md`. - yes I did
   
    + Any new features
    + Any deprecations
 
 - [x] This is unlikely to impact how Ambassador performs at scale. - it shouldn't affect the ingress at all
 
      (This will be fixed soon.)
 

cmd/kat-server/services/grpc-agent.go
```diff
@@ -118,34 +128,89 @@ func (d *director) GetLastSnapshot() *agent.Snapshot {
...
+		}
+	}
+
+	var snapshot agent.Snapshot
+	err = json.Unmarshal(data, &snapshot)
+	if err != nil {
+		return err
+	}
+
+	log.Print("Recieved snapshot")
```
kflynn, line 206: Probably better to fix this too, since you know people will search for it...

kflynn (approved):
Two minor typos -- I'm gonna approve as-is but would love it if the typos got fixed. Thanks!!

kflynn (approved):
Hit it -- I can't find the other thing that I was thinking of, so let's get this landed.

PR #3836, opened 2021-09-29
make generate, generate-clean: Be more robust

LukeShu:
## Description

- The "clean" rules weren't staying in-sync with the productive rules,
  solve that by having them share a source of truth.

- `make update-yaml` is increasingly a misnomer; rename it to `make
  generate-fast`.

- Directories not getting emptied and having now-orphaned output files
  could be a source of errors; do a better job cleaning (see above).

- Add more comments and organize generate.mk better.

- The symlinks in `generate.tmp` would trigger warnings from `go`;
  rename it to `_generate.tmp` so that Go ignores it.

- Move code out of `charts.mk` to `generate.mk`.

Running `make generate-fast` produces only one change (included in this
commit): have the Helm chart README.md not be executable.  Besides this
file-permission change, all of the changes in this commit were made
manually by me.

## Related Issues

I'm needing to fuss with `make generate` in https://github.com/emissary-ingress/emissary/pull/3834, and got frustrated doing that.  Hence this PR.

## Testing
Shouldn't be a runtime change.  CI should validate that `make generate` works as intended.  I verified that `make generate-fast` is indeed fast.

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 
 - [ ] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - ugg, it didn't mention `make update-yaml` before...


kflynn (dismissed):
HIt it.

kflynn (approved):
🙄 🙂

PR #3837, opened 2021-09-29
Less `pkg/supervisor`, audit `context.Background` and `context.TODO`

LukeShu:
### Migrate away from pkg/supervisor everywhere except in watt
`supervisor.Command`'s logging\no newline
is aggres\no newline
sively terrible.  It logs each `read(2)` as\no newline

Also a lot of the surrounding code logged directly to stdout, instead of using dlog or `testing.TB.Log`, so when running as part of a test the log output is out of order with the test logs and is likely associated with the wrong test.

a separate log line, rather than buffering t\no newline
hem until it hits a `\n` or a sane max line length.

I didn't fuss with moving `watt` over because watt is complex and actually takes advantage of supervisor's more sophisticated readiness functionality.

### Audit uses of `context.Background()` and `context.TODO()`

And then proceed to audit uses of `context.Background()` and `context.TODO()`.  `context.Background()` should be used exceedingly sparingly (directly inside of `main()`, or in an `Example` test); I pruned it down to be used only in the cases where it's actually appropriate.  And `context.TODO()` is a literal TODO; I did most of them but there are a short handful remaining that would have taken more care/time/effort to remove.

This is pursuant to the above thing about "test logs are in the wrong place".

Is there a good place for me two write down style guide type things? I'd put it in the linter configuration, except I can't do that until we upgrade golangci-lint, and we can't do that until we upgrade Go.  So for now it needs to be a human-level thing.  Eh, I'll just wait until I upgrade Go next week during cooldown.

## Testing
I didn't do any testing other than making sure that it compiles and such, that `make golint` is clean for each commit.  CI should cover it.

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 

kflynn (approved):
Hit it!

PR #3838, opened 2021-09-29
Lint/format the `python/schema/` files

LukeShu:
## Description
This is kind of a frustrating PR for me because it's largely just re-doing work from https://github.com/datawire/apro/pull/1627 that form some reason got reverted (there's no blame there: I'm the one who reverted it https://github.com/datawire/apro/pull/1644 ... if only I could the conversation that surely went with that...).

But this time it's different, because instead of me just doing the work, I'm _teaching a computer to do the work_.  And even more important, that computer is CI.

## Related Issues
This is in prep for https://github.com/emissary-ingress/emissary/pull/3834 because it will be fussing with the schema files a lot.

## Testing
I haven't done any testing.

## Checklist

    + Any new features
    + Any deprecations
 
 - [ ] This is unlikely to impact how Ambassador performs at scale. - unless there are some rather weird perf characteristics of the JSON schema validator
 
 - [x] My change is adequately tested. - if CI doesn't cover this, then our test suite is worthless and we should rethink our lives
 
      (This will be fixed soon.)
 
### DIFF
--- a/build-aux-local/lint.mk
+++ b/build-aux-local/lint.mk
@@ -3,19 +3,23 @@
 lint/go-dirs = $(OSS_HOME)
 
 lint:
-	@PS4=; set +x; r=0; { \
-		printf "$(CYN)==>$(END) Linting $(BLU)Go$(END)...\n" ;\
-		go_status=FAIL; go_color="$(RED)" ;\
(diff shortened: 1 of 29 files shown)


kflynn (approved):
Hit it!

PR #3840, opened 2021-09-29
[v1.14.2] Chart and manifest image updates

w-h37:
Update charts and manifests to use the newly release images for 1.14.2


kflynn (approved):
Hit it!

PR #3817, opened 2021-09-21
Move v3alpha1 CRDs back into getambassador.io.

kflynn:
...and drop the `Ambassador`- prefix.

I tested this manually before pushing the PR. I think there's still more coverage to be added just to make sure that mixing `v2` and `v3alpha1` is OK, and for things like making sure a `v3alpha1` `Mapping` honors `host` and `host_regex`...

 - [ ] I made sure to update `CHANGELOG.md`.
 - [x] This is unlikely to impact how Ambassador performs at scale.
 - [ ] My change is adequately tested.
 - [x] I didn't need to update `DEVELOPING.md`.

Issue #3735, opened 2021-08-26
Error installing Emissary 2.0 on k8s v1.22 (no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1beta1")


jar3b:
Hello. I have the k8s cluster created with kubeadm (v1.22.1):
```bash
:~$ kubectl version
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.1", GitCommit:"632ed300f2c34f6d6d15ca4cef3d3c7073412212", GitTreeState:"clean", BuildDate:"2021-08-19T15:45:37Z", GoVersion:"go1.16.7", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.1", GitCommit:"632ed300f2c34f6d6d15ca4cef3d3c7073412212", GitTreeState:"clean", BuildDate:"2021-08-19T15:39:34Z", GoVersion:"go1.16.7", Compiler:"gc", Platform:"linux/amd64"}
```

and I try to fresh install emissary-ingress 2.0 (devel) as described here (https://www.getambassador.io/docs/emissary/2.0/tutorials/getting-started/):

```bash
helm repo add datawire https://app.getambassador.io
helm repo update
helm install emissary-ingress datawire/emissary-ingress --namespace=ambassador --devel --set nodeSelector.role=master    
```

but got the following error:
```
Error: failed to install CRD crds/getambassador.io_authservices.yaml: unable to recognize "": no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1beta1"
```

The problem is related to #3406 

When I manually fixied all the crd's (https://github.com/emissary-ingress/emissary/tree/master/charts/emissary-ingress/crds) to use the new apiVersion `apiextensions.k8s.io/v1` the problem was resolved. But my fixes are a little ugly, because it's need to properly update descriptions, fields types (due to errors like `oneOf[0].type: Forbidden: must be empty to be structural`) and strictly define other types like `BoolOrString`.

I don’t know why most of users don’t face this problem (bcs i don't see new related issued), but I think it’s quite important. If anyone wants i can publish my crd fixes.

Thanks.

Foltik:
Running into this as well. @jar3b, could you publish the patches?

Foltik:
[...]

PR #3841, opened 2021-09-29
Beat the v3 CRDs in to "shape"

LukeShu:
## Description

There are 3 big problems with the v2/v3alpha1 CRD situation in and of itself (regardless of external things like Kubernetes 1.22):
1. A lot of changes for Emissary 2.0 were made to the v2 CRDs, which is wrong, they should be effectively frozen and all changes should be made to the v3alpha1 CRDs
2. v3alpha1 is incomplete; all of the CRDs need to be copied from v2 to v3alpha1
3. v3alpha1 has apiGroup=`x.getambassador.io` (it shouldn't have the `x.` prefix), and the CRDs in it have an `Ambassador` prefix on their names.

This PR addresses those.  It does not address anything related to Kubernetes 1.22 or other things about the CRDs that might need changed.

- The first 3 commits (which are branched from a point back in May before any v3 CRD commits happened) copy v2→v3alpha1 and make some tooling changes related to that.
  + `gitk origin/master ^origin/release/v1.14 -- pkg/api/getambassador.io/` tells me that c53948d6073ad0e4d4bfee414ce9d2721528c85c is the first commit making CRD changes for Emissary 2.0; so I branched from [c53948d^](https://github.com/emissary-ingress/emissary/commit/c53948d6073ad0e4d4bfee414ce9d2721528c85c^) (the parent commit of c53948d6073ad0e4d4bfee414ce9d2721528c85c).
- The next 13 commits incrementally merge the next 4 months of history changing CRDs for Emissary 2.0.  In each of these merge commits, I enforced several rules:
  + Each merge commit is clean with regard to `make update-yaml`/`make generate-fast`
  + Each merge commit is clean with regard to `make go-mod-tidy`
  + Each merge commit passes `make golint` (i.e. all of the Go code compiles, even if it does not work correctly)
  + No merge commit admits any changes to the v2 CRDs (except for a couple of doc-comment fixups), instead making those changes to the v3alpha1 CRDs.
 + No merge commit admits any changes to the v3alpha1 CRDs adding the `x.` prefix to the api group name or adding the `Ambassador` prefix to existing CRD names.
- The final 5 commits adjust things to work with the above; incorporating the work from https://github.com/emissary-ingress/emissary/pull/3817

I must apologize that each commit only leaves things in a state valid as describe above, but not entirely valid; that each commit until the final one has things in a broken state.

So, in all, this (1) rolls back the v2 CRDs to the point they were at when they diverged from v1.14 in Git, and (2) syncs all of the missing stuff in to v3alpha1 from v2.

For reviewers: I usually suggest looking at things commit-by-commit, but in this case I think it's probably most helpful to look at a combination of the combined final diff and the commit-by-commit history.  For viewing the commits in this PR (since it's a complex history with a bunch of merges), I use `gitk origin/lukeshu/v3-crd-cleanup ^origin/master`, sometimes with ` -- pkg/api/getambassador.io/` to view only the CRD changes.  For the merges, it's also often helpful to right-click→`Diff this → selected` to compare 2 commits directly, rather than looking at the magic 3-way-diff that it shows by default for merge commits.

# Related Issues
- In prep for a later PR addressing #3735
- Subsumes PR #3817

## Testing

I checked the invariants above, ensuring that the CRDs are in a sane state.  Beyond that, most of the work has gone in to making sure that this doesn't break anything; via CI.

## Checklist

    + Any new features
    + Any deprecations
 
 - [x] This is unlikely to impact how Ambassador performs at scale. - I don't think so
 
 - [x] My change is adequately tested. - if this isn't adequate, idk what is
 
      (This will be fixed soon.)
 
 - [ ] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - turns out that debugging `make chart-test` is a pain, but I didn't write down any tricks.  My tricks are probably bad and there is probably a better way. But still I shoulda written something down.  But I'm running out of steam.


python/ambassador/fetch/k8sobject.py
```diff
@@ -40,7 +40,7 @@ def domain(self) -> str:
     @classmethod
     def for_ambassador(cls, kind: str, version: str = 'v2') -> KubernetesGVK:
         if 'alpha' in version:
-            return cls(f'x.getambassador.io/{version}', kind)
+            return cls(f'getambassador.io/{version}', kind)
         else:
             return cls(f'getambassador.io/{version}', kind)
```
kflynn, line 45: This is a little silly now...

CHANGELOG.md
```diff
@@ -167,9 +167,10 @@ installations, reduce memory footprint, and improve performance. We welcome feed
 
 ## Emissary-ingress
 
-- Feature: Emissary-ingress 2.0.0 introduces API version `x.getambassador.io/v3alpha1` for configuration
-  changes that are not backwards compatible with the 1.X family. API versions `getambassador.io/v0`
-  and `getambassador.io/v1` are deprecated. Further details are available in the <a
+- Feature: Emissary-ingress 2.0.0 introduces API version `getambassador.io/v3alpha1` for configuration
+  changes that are not backwards compatible with the 1.X family.  API versions
+  `getambassador.io/v0`, `getambassador.io/v1`, and `getambassador.io/v2` are deprecated.  Further
+  details are available in the <a
```
kflynn, line 173: This actually needs to be `x.getambassador.io` since it's talking about 2.0.0.

kflynn (approved):
Two comments. I'm approving, though: let's get this landed and fix the comments in a separate PR.

PR #3843, opened 2021-09-30
Use a mirror of the helm-test Docker image…

kflynn:
…instead of continuing to rely on quay.io.


LukeShu:
What we really need to do is just stop running it in Docker.

PR #3844, opened 2021-09-30
Only support the v3 CRDs

LukeShu:
## Description
If https://github.com/emissary-ingress/emissary/pull/3841 is "make sure everything works if the v3 CRDs are present", then this PR is "switch everything over to use the v3 CRDs, to make sure they work".  To ensure that everything is switched over, I go ahead and delete the v2 CRDs, which is sorta controversial but will end up helping facilitate Kubernetes 1.22 support, since we haven't yet figured out how to write down the v2 types in a way that Kubernetes 1.22 won't reject.

## Related Issues
Yet another PR working toward #3735.

## Testing

Nothing beyond CI.

## Checklist

 - [ ] I made sure to update `CHANGELOG.md`. - I didn't write anything... we've already said "deprecated". We need to figure out what the messaging here is.
   
    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 
 - x ] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - no tricks


pkg/testutils/utils.go
```diff
@@ -74,17 +74,21 @@ func GlobMatch(what string, text string, pattern string, regex bool) (bool, stri
...
+	if mapping.Spec.DeprecatedHost != "" {
+		mappingHost := mapping.Spec.DeprecatedHost
 
-	if mappingHost != "" {
+		mappingHostRegex := false
+		if mapping.Spec.DeprecatedHostRegex != nil {
+			mappingHostRegex = *mapping.Spec.DeprecatedHostRegex
+		}
+
+		fmt.Printf("HostMatch: hostName %s mappingHost %s\n", hostName, mappingHost)
```
LukeShu, line 91: NB: This change is a little more involved than most?

kflynn (approved):
Hit it!

PR #3847, opened 2021-10-01
Improve website CI

aidanhahn:
Signed-off-by: Aidan Hahn <aidanhahn@datawire.io>

Pretty simple stuff, getting the short version with pattern matching instead of an indexed substring means we don't need to worry if we someday want to have a 20.30.40 release.

I just wanted to get this in before I forgot about it.


kflynn (approved):
Hit it!

Issue #3868, opened 2021-10-07
Helm Chart - Module not created without extra configuration when specifing AMBASSADOR_ID


ghost:
# Overview

The helm chart has incorrectly configured requirements for generating the module.yaml file as without settings module params it is not created but is required when using AMBASSADOR_ID and therefore must be able to be configured with only this param set.

# Problem
The *Module* template (emissary/charts/emissary-ingress/templates/module.yaml) starts with the following:

```
{{- if .Values.module }} 
```
> requiring values within Values.module to be set to create file

Therefor, if you only specify AMBASSADOR_ID: 
```
helm install datawire/ambassador \
  --set env.AMBASSADOR_ID="local"
```
> **The Module CRD is not created!!**

# Temp Fix
```
helm install datawire/ambassador \
  --set env.AMBASSADOR_ID="local" \
  --set module.allow_chunked_length=true
```
> set any **module.** value

# Fix 

Ensure module file is created when AMBASSADOR_ID **OR** *module.* params

PR #3869, opened 2021-10-07
Update module.yaml

crocinsocks:
fix for #3868


crocinsocks:
All done! Just a quick note when testing on my setup I had to use (module.yaml)

```
```
not 
```
```

Assume this is correct and just due to the fact I don't have getambassador.io/v3alpha1 CRDs installed?

kflynn (approved):
Taking this one. 🙂

PR #3888, opened 2021-10-18
migrate release/v1.14 ci from circleci to github-actions

anodelman:
## Description
Towards the EOL of circleci for the edgestack team.

Current status:
Anything other than the dev test workflows are untested (repatriate, promote, website-update, etc were untouched and untested).

Works, except for go tests which result in:
```
✓  cmd/entrypoint (48.834s)
✓  pkg/kubeapply (3m54.767s)
✖  pkg/k8s (3m55.566s)
✓  internal/pkg/nat (4m9.252s)
✓  pkg/kates (5m29.905s)
✖  pkg/gateway (10m3.33s)

=== Failed
=== FAIL: pkg/gateway TestGatewayMatches (unknown)

=== FAIL: pkg/k8s TestWatchQuery (15.15s)
    watcher_test.go:124: 
        	Error Trace:	watcher_test.go:124
        	Error:      	Not equal: 
        	            	expected: []string{"kubernetes.default", "kubernetes.default"}
        	            	actual  : []string{"kubernetes.default"}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,2 @@
        	            	-([]string) (len=2) {
        	            	- (string) (len=18) "kubernetes.default",
        	            	+([]string) (len=1) {
        	            	  (string) (len=18) "kubernetes.default"
        	Test:       	TestWatchQuery

DONE 650 tests, 2 failures in 665.466s
```

## Testing
Tested against github.

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 

LukeShu:
If we do this, it'll be by doing a copy of the more-recent 2.x CI, which has evolved since this was filed.

PR #3906, opened 2021-11-03
metrics: Add a ambassador_log_level metric

jfrabaute:
## Description
A new metric ambassador_log_level{level="debug"} is added to report if
the current log level of the instance is debug or not (warning/error).

THis can be useful to monitor that debug logs are not enabled on
production instances, or correctly disabled after some debugging.

## Testing
Tested on kind and deploying ambassador with helm chart.

## Checklist
  
### DIFF
--- a/python/ambassador_diag/diagd.py
+++ b/python/ambassador_diag/diagd.py
@@ -233,10 +233,17 @@ class DiagApp (Flask):
                                  namespace='ambassador', registry=self.metrics_registry)
         self.diag_notices = Gauge(f'diagnostics_notices', f'Number of configuration notices',
                                  namespace='ambassador', registry=self.metrics_registry)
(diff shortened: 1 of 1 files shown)


kflynn (approved):
Looks good from here! Let's see about getting CI on this one..

kflynn (changes requested):
Actually, I need to take that back -- looking over the Prometheus docs, it makes more sense to be able to say that the gauge is always valued 1, and the label tells you what the current loglevel is.

kflynn (approved):
Chatting with @jfrabaute, I'm going to reverse myself again (😂 ) and land this one as the foundation for further improvements. Thanks, @jfrabaute! 

(Since GitHub seems a bit confused by my rebase in the CI branch, I'm landing this as an admin, then pulling in release notes separately.)

PR #3907, opened 2021-11-03
Whitelist command operators in IRErrorResponse

aidanhahn:
## Description
A configuration including any token delimited by "%" that is not a correct Envoy command operator will result in a bad Envoy configuration. This PR includes a check in the IRErrorResponse validation to match all command operators and check them against the whitelist. Mappings that contain an invalid mapping in the error_response_overrides section will be discarded.

## Related Issues
Apro issue #2661

## Testing
Manual testing was done to verify that files, strings, and json containing invalid command operators were rejected, and ones that did were accepted.

## Checklist

 - [x] I made sure to update `docs/releaseNotes.yml`.
 

python/ambassador/ir/irerrorresponse.py
```diff
@@ -180,10 +215,19 @@ def _generate_mappers(self) -> Optional[List[Dict[str, Any]]]:
                     continue
 
                 body_format_override["text_format_source"] = ir_text_format_source
+                try:
+                    fmt_file = open(ir_text_format_source["filename"], mode='r')
```
kflynn, line 219: This will raise an exception if, somehow, `ir_text_format_source` has no `filename` element...

python/ambassador/ir/irerrorresponse.py
```diff
@@ -232,6 +278,20 @@ def _generate_mappers(self) -> Optional[List[Dict[str, Any]]]:
...
                 body_format_override["content_type"] = ir_content_type
 
+            # search the body for command tokens
+            # TODO: remove this code when envoy supports escaping "%"
+            token_finder = re.compile(ENVOY_FMT_TOKEN_REGEX)
+            matches = token_finder.findall(format_body)
+
+            bad_token = False
+            for i in matches:
+                if not i[0] in ALLOWED_ENVOY_FMT_TOKENS:
```
kflynn, line 288: This is probably worth a comment, just in case the reader isn't familiar with what a match looks like. (NGL, I don't remember what the tuple contains.)

kflynn (changes requested):
A couple of minor nits -- overall, yeah, looks good! 🙂 

This should have a test, too, probably something where you feed in a bad token and make sure that the error-response isn't in the generated Envoy? That can be a compiler test.

python/ambassador/ir/irerrorresponse.py
aidanhahn: That case is ruled out earlier in the following code block:
```python
 209             if ir_text_format_source is not None:                                                                                                                                                                                      
 210                 # Verify that the text_format_source field is an object with a string filename.                                                                                                                                        
 211                 if not isinstance(ir_text_format_source, dict) or \                                                                                                                                                                    
 212                         not isinstance(ir_text_format_source.get('filename', None), str):                                                                                                                                              
 213                     self.post_error(                                                                                                                                                                                                   
 214                             f"IRErrorResponse: text_format_source field must be an object with a single filename field, found \"{ir_text_format_source}\"")                                                                            
 215                     continue                                                                                                                                                                                                           
 ```
 
 We can assume that that field exists and is a string.

File: python/ambassador/ir/irerrorresponse.py (hunk 2 above)
Comment (aidanhahn) [reply python/ambassador/ir/irerrorresponse.py:288]: Will add a comment!

File: python/ambassador/ir/irerrorresponse.py (hunk 1 above)
Comment (kflynn) [reply python/ambassador/ir/irerrorresponse.py:219]: Works for me!

File: python/tests/unit/test_error_response.py [hunk 3]
```diff
@@ -124,7 +124,7 @@ def _test_errorresponse(yaml, expectations, expect_fail=False):
     assert body_format is None
 
     mappers = ir_conf.get('mappers', None)
-    assert mappers
+    assert mappersx
```
kflynn, line 127: This seems like a typo?

python/tests/unit/test_error_response.py
```diff
@@ -169,11 +170,12 @@ def _test_errorresponse_onemapper_onstatuscode_textformat_contenttype(
 
 def _test_errorresponse_onemapper_onstatuscode_textformat_datasource(
         status_code, text_format, source, content_type):
+    open(source, "x").close()
```
kflynn, line 173: I don't really follow this?

kflynn (dismissed):
Hit it!

kflynn (approved):
Hit it!

PR #3908, opened 2021-11-03
Make selector and mappingSelector synonyms for now

### LINKED ISSUE #3902: Host.Spec.MappingSelector is documented, but not supported
The documentation for Emissary 2.0.4 describes `Host.Spec.MappingSelector` -- however, the code still uses the old `Host.Spec.Selector` element.

Immediately, we need to support both `Selector` and `MappingSelector` as synonyms. Later, the deprecated `Selector` can be removed.

kflynn:
Make Host.Spec.Selector and Host.Spec.MappingSelector synonyms for now; later,
we'll drop Host.Spec.Selector since we don't need both.

Fixes #3902.

- py-mkopensource: Some things got added to the static list that shouldn't have
- Upgrade to github.com/datawire/go-mkopensource
- ambassador-consul-connector.yaml: Update for the new CRDs
- Switch mapping_selector to mappingSelector in Host JSON
- Make Host.Spec.Selector and Host.Spec.MappingSelector synonyms
- Update tests, with thanks to @AliceProxy!
- Release notes and CHANGELOG


AliceProxy (approved):
Approving the CRD and Test file changes specifically since I made the same changes in another PR and tested them successfully. Leaving the remaining changes for another collaborator to review.

AliceProxy (approved):

python/ambassador/ir/irhost.py
```diff
@@ -72,6 +73,16 @@ def setup(self, ir: 'IR', aconf: Config) -> bool:
         self.insecure_action = insecure_policy.get('action', 'Redirect')
         self.insecure_addl_port: Optional[int] = insecure_policy.get('additionalPort', None)
 
+        # If we have no mappingSelector, check for selector.
+        mapsel = self.get('mappingSelector', None)
+
+        if not mapsel:
+            mapsel = self.get('selector', None)
+
+            if mapsel:
```
LukeShu, line 82: Stupid curiosity: You and I have very different styles for where to put blank lines. So, to pick on this as a typical example (and not that I think I could articulate my style): Why's a blank line belong here?

python/ambassador/ir/irhost.py
```diff
@@ -363,18 +380,20 @@ def matches_httpgroup(self, group: 'IRHTTPMappingGroup') -> bool:
             host_match = True
             self.logger.debug("-- hostname %s group regex => %s", self.hostname, host_match)
         else:
-            group_glob = group.get('host') or None
+            # It is NOT A TYPO that we use group.get("host") here -- whether the Mapping supplies
+            # "hostname" or "host", the Mapping code normalizes to "host" internally.
```
LukeShu, line 384: gross :P

LukeShu (approved):
lgtm

python/ambassador/ir/irhost.py
kflynn: Yeah. It saved an _insane_ amount of code churn at the time, but it's probably time to cause that churn for everyone's sanity.

python/ambassador/ir/irhost.py
kflynn: Uh. 🤔 

... 🤔  ...

...I think it comes down to where I think the breaks in thoughts are. The whitespace is there to make it simpler to read later: groups of lines with no blank lines are things I think are OK to chunk together and either consider or skip as a single unit, but the blank lines mark "oh no, this is a new concept – here's a place where you need to pay more attention even if you were writing the previous chunk off as not relevant for your current study."

I think I should think a bit more about how to articulate that...

PR #3910, opened 2021-11-04
Implement circuit breakers for AuthServices

LukeShu:
## Description
Add a `circuit_breakers` field to AuthServices.

## Related Issues
datawire/apro#2727

## Testing
This PR contains no tests, but I have tested it manually in load testing.  By setting `AuthService.spec.circuit_breakers[0].max_requests` to 4096 (up from the Envoy default of 1024), I can now handle ~2500 concurrent requests, where before it would start 403'ing them at exactly 1024 requests.

## Checklist

    + Any new features
    + Any deprecations
 
 - [x] This is unlikely to impact how Ambassador performs at scale. - unless you set it, it shouldn't change
 
      (This will be fixed soon.)
 

python/ambassador/ir/irauth.py
```diff
@@ -126,6 +127,11 @@ def _load_auth(self, module: Resource, ir: 'IR'):
             if add_linkerd_headers is None:
                 self["add_linkerd_headers"] = ir.ambassador_module.get('add_linkerd_headers', False)
 
+        if module.get('circuit_breakers', None):
+            self['circuit_breakers'] = module.get('circuit_breakers')
+        else:
+            self['circuit_breakers'] = ir.ambassador_module.circuit_breakers
```
kflynn, line 133: Feels like this needs to use `ir.ambassador.module('circuit_breakers', None)` to avoid a crash if the `ambassador` ``Module` has no `circuit_breakers` key... ?

kflynn (changes requested):
One place I'm concerned about, looks pretty smooth overall.

python/ambassador/ir/irauth.py
LukeShu: Huh. Yeah, it does seem like that.

But this works, and I don't have an `ambassador` `Module`, which I think means it's safe?

kflynn: Looking over the code in `irambassador.py`, I think that this is an accident of the current `irambassador` implementation. I'm rebasing this onto `master`, so I'll tweak this as well.

kflynn (approved):
Approved! 🙂

aidanhahn (approved):

PR #3918, opened 2021-11-04
[v2.0.5] Release Branch

aidanhahn:
All commits that are going out in 2.0.5 release **must** be on rel/v2.0.5.
This will allow the appropriate CI to run.

Reviewers, please review the changelog and commits in this branch to make sure everything that needs to go out in the release is on this branch.

Each push to this branch will generate a new RC tag.


LukeShu:
DCO is upset because of commits from d6e-automaton. We should look in to that.

LukeShu:
Convince me why I shouldn't be concerned about the `chart-publish` and `job-chart-changelog` failures.

kflynn:
`job-chart-changelog` failed because the job itself is borked.
`chart-publish` failed because I already published chart 7.1.10. Once we land this, I can get on with making these jobs not run when they shouldn't...

LukeShu (approved):

PR #3927, opened 2021-11-12
Upgrade to Go 1.17

LukeShu:
## Description

This PR does three top-level things:
 - Merge `build-aux-local/` in to `build-aux/`
 - Upgrade to Go 1.17.1 (and a bunch of changes to make this possible)
 - Upgrade to the latest `golangci-lint`.

## Related Issues
- https://github.com/datawire/apro/pull/2837
- https://github.com/datawire/go-mkopensource/pull/5

## Testing
CI should cover it.

## Checklist

 - [ ] I made sure to update `CHANGELOG.md`. - I don't think we usually mention Go upgrades? (unless it's to address a CVE)
   
    + Any new features
    + Any deprecations
 
 - [ ] This is unlikely to impact how Ambassador performs at scale. - honestly, it might. I'd be surprised if that "impact" isn't "improve" though, we're getting a lot of GC improvements here.
 
 - [x] My change is adequately tested. - except for perf testing, CI should cover it
 
      (This will be fixed soon.)
 

tools/src/testcert-gen/main.go
```diff
@@ -0,0 +1,302 @@
...
+
+	"github.com/datawire/dlib/dlog"
+)
+
+func djb2Hash(str string) int64 {
+	hash := uint64(5381)
+	for _, c := range str {
+		hash = (hash * 33) + uint64(c)
+	}
+	return int64(hash)
```
kflynn, line 30: Shouldn't this be `uint64`? I guess it doesn't matter if it's just handed off to `math_rand.NewSource` but, hm.

tools/src/testcert-gen/main.go
```diff
@@ -0,0 +1,302 @@
...
+
+type randReader struct {
+	inner io.Reader
+}
+
+func (r *randReader) Read(d []byte) (int, error) {
+	if len(d) == 1 {
+		// We want deterministic keys.  But rsa.GenerateMultiPrimeKey calls
+		// crypto/internal/randutil.MaybeReadByte() to make them non-deterministic even when
+		// the source of randomness is deterministic.  Shut that down!
```
kflynn, line 47: This should duplicate, or at minimum refer to, your comment in `Main` that talks about why this isn't just a godawful terrible bad idea. 🙂

tools/src/testcert-gen/main.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+func ParseArgs(argStrs ...string) (CLIArgs, error) {
+	var args CLIArgs
+	argparser := pflag.NewFlagSet(os.Args[0], pflag.ContinueOnError)
+
+	argparser.StringSliceVar(&args.Hosts, "hosts", nil, "comma-separated list of hostnames and IPs to generate a cert for")
+	argparser.BoolVar(&args.IsCA, "is-ca", false, "whether this cert should be a Certificate Authority cert")
+	argparser.BoolVar(&args.IsServer, "is-server", true, "whether this cert should server cert")
+	argparser.BoolVar(&args.IsClient, "is-client", false, "whether this cert should server cert")
```
kflynn, line 86: This duplicated sentence no verb.

tools/src/testcert-gen/main.go
```diff
@@ -0,0 +1,302 @@
...
+			Country:            []string{"US"},
+			Province:           []string{"MA"},
+			Locality:           []string{"Boston"},
+			Organization:       []string{"Ambassador Labs"},
+			OrganizationalUnit: []string{"Engineering"},
+
+			CommonName: args.Hosts[0],
+		},
+		NotBefore: time.Date(2021, 11, 10, 13, 12, 0, 0, time.UTC),
+		NotAfter:  time.Date(2099, 11, 10, 13, 12, 0, 0, time.UTC),
```
kflynn, line 252: This will give MacOS browsers fits. I'm pretty sure we don't care – is there ever a reason we'd want to hit a server using these certs from a browser?

tools/src/testcert-gen/main.go
```diff
@@ -0,0 +1,302 @@
...
+
+		// If you ever extend this program to generate certs with non-RSA keys, be aware
+		// that x509.KeyUsageKeyEncipherment is an RSA-specific thing.
+		KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment,
```
kflynn, line 256: Huh. I thought key encipherment was broader than that, and quickly searching seems to imply that it is -- got a source?

kflynn (changes requested):
I don't think it's safe to completely delete the test Docker images. 😞  But it looks great overall (even given that I have a couple of other nits).

tools/src/testcert-gen/main.go
LukeShu: Yeah, it's specifically to hand off to `math_rand.NewSource`.  If I were OK with returning a `uint64`, there wouldn't need to be a cast at all.

tools/src/testcert-gen/main.go
LukeShu: We don't care. I'll add a comment saying that.

tools/src/testcert-gen/main.go
LukeShu: https://github.com/golang/go/blob/go1.17.3/src/crypto/tls/generate_cert.go#L88-L93

kflynn: Now I wonder what _their_ source is... 😂

kflynn (approved):
Hit it! 🙂  The comments in `testcert-gen` make me feel much better, many thanks!

PR #3937, opened 2021-11-16
pkg/kubeapply: Better diagnostics

LukeShu:
## Description
I'm sick of `pkg/kubeapply` having bad diagnostics; missing object names, poorly formatted output, misleading error messages. Do what I can to improve that.

## Related Issues
- https://github.com/datawire/dlib/pull/31

## Testing
I used this when debugging https://github.com/datawire/apro/pull/2837

## Checklist

 - [x] I made sure to update `CHANGELOG.md`. - not a user-visible change
   
    + Any new features
    + Any deprecations
 
 - [ ] My change is adequately tested. - eh
 
      (This will be fixed soon.)
 

pkg/kubeapply/kubeapply.go
```diff
@@ -110,63 +110,56 @@ func (collection YAMLCollection) ApplyAndWait(
...
-		if err != nil {
-			msgs = append(msgs, fmt.Sprintf("%s: %v\n", n, err))
-			valid[n] = false
-		} else {
-			valid[n] = true
+	var scanErrs derror.MultiError
+	for _, filename := range expandedFilenames {
+		valid[filename] = true
+		if err := waiter.Scan(ctx, filename); err != nil {
+			scanErrs = append(scanErrs, fmt.Errorf("watch %q: %w", filename, err))
```
kflynn, line 142: Hmmm. It's a little jarring to see that `scanErrs` is a `MultiError`, then see the `append` call screaming "it's an array!" Could we maybe have `MultiError.AddError` or the like, to avoid exposing the implementation?

pkg/kubeapply/kubeapply.go
```diff
@@ -110,63 +110,56 @@ func (collection YAMLCollection) ApplyAndWait(
...
-		err = kubectlApply(ctx, kubeinfo, dryRun, expanded)
-	}
-
 	if !debug {
-		for _, n := range expanded {
-			if valid[n] {
-				err := os.Remove(n)
-				if err != nil {
-					dlog.Print(ctx, err)
+		defer func() {
```
kflynn, line 147: Probably worth a comment about intentions here. (Admittedly, some of my tripping-up here might be from reading the _diff_ rather than the source, and wondering what the point of adding the deferral was.)

pkg/kubeapply/kubeapply.go
```diff
@@ -110,63 +110,56 @@ func (collection YAMLCollection) ApplyAndWait(
...
+			for _, filename := range expandedFilenames {
+				if valid[filename] {
+					if err := os.Remove(filename); err != nil {
+						dlog.Error(ctx, err)
```
kflynn, line 151: This feels like we should include the filename in the `dlog.Error`. Does `err`'s expansion already have it?

pkg/kubeapply/waiter.go
```diff
@@ -87,32 +93,38 @@ func (w *Waiter) isEmpty() bool {
...
-				if err != nil {
-					dlog.Println(ctx, err)
-					continue
-				}
-				if last.Before(start) {
-					continue
-				}
+		for _, _event := range list {
+			var event corev1.Event
+			if err := kates_internal.Convert(_event, &event); err != nil {
```
kflynn, line 103: Worth a comment here about what type `_event` is.

pkg/kubeapply/waiter.go
```diff
@@ -87,32 +93,38 @@ func (w *Waiter) isEmpty() bool {
...
-				dlog.Printf(ctx, "event: %s %s\n", name, r["message"])
-				printed[r.QName()] = true
+			if event.LastTimestamp.Time.Before(start) && !event.LastTimestamp.IsZero() {
+				continue
+			}
+			eventQName := fmt.Sprintf("%s.%s", event.Name, event.Namespace)
+			if !printed[eventQName] {
+				involvedQKind := func(gv, k string) string {
+					var g, v string
+					if slash := strings.IndexByte(gv, '/'); slash < 0 {
```
kflynn, line 114: I am going to continue to complain about this until my dying breath. Burying the !*@&#!@ lede...

pkg/kubeapply/waiter.go
```diff
@@ -87,32 +93,38 @@ func (w *Waiter) isEmpty() bool {
...
+						g = ""
+						v = gv
+					} else {
+						g = gv[:slash]
+						v = gv[slash+1:]
+					}
+					return strings.Join([]string{k, v, g}, ".")
+				}(event.InvolvedObject.APIVersion, event.InvolvedObject.Kind)
```
kflynn, line 122: And here, too, even worse! Why is there a lambda here _anyway_? Why isn't this just a proper function in its own right? It's not like it's only ever going to be useful here.

kflynn (changes requested):
This will make me _so_ much happier about `kubeapply` that I hate to request changes. But request changes I must. 🙂  (They're pretty minor; I'm just really feeling the burn on some of the places where we've written opaque Golang.)

pkg/kubeapply/kubeapply.go
LukeShu: I'm nearly certain that `os.Remove` returns an `*os.PathError` that already includes the filename.

pkg/kubeapply/waiter.go
LukeShu: An `interface{}`!

pkg/kubeapply/waiter.go
LukeShu: But it's only a handful of lines.

pkg/kubeapply/waiter.go
LukeShu: Fine.

pkg/kubeapply/kubeapply.go
LukeShu: I've added a comment, but I think a significant part of tripping up is looking at the diff.

pkg/kubeapply/kubeapply.go
LukeShu: It is an array! A named type can be anything, not just a struct. Wrapping it in a struct and making it less convenient seems wrong; KISS!

Compare to [`k8s.io/apimachinery/pkg/util/errors.Aggregate`](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/errors#NewAggregate) which does avoid exposing the implementation.  In a comment in https://github.com/datawire/dlib/pull/31 one of the things I dinged it for is "Isn't a simple []error type, and so is clunky to work with."  The one thing I think the opaque-wrapping approach has going for it is that it can (and apimachinery does) make it impossible to have an invalid `"(0 errors; BUG: this should not be reported as an error)"` error.

kflynn (approved):
Hit it! Thanks for the changes. 🙂 

I kinda disagree with you on the merits of thinking about `MultiError` as an array rather than as an opaque type, but that's a longer-term thing. 🙂

PR #3946, opened 2021-11-24
Don't log pointless errors about ACME private-key certs

kflynn:
ACME private-key secrets are not TLS secrets, and they're not supposed to be, so we shouldn't
log errors about them not being	TLS secrets.

The test case here loads a snapshot from a cluster where I ran ACME and	requires that we get
no errors. The certs in	question have been replaced with self-signed certs, though.

- Don't throw errors because ACME private-key secrets aren't TLS secrets. They're not supposed to be.
- Unit test for ACME private-key secrets
- Release notes

 - [x] I didn't need to update `DEVELOPING.md`... yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,10 @@ Please see the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest
   the same `prefix` ("canary"ing multiple `Mapping`s together). This has been corrected, so that all
   such updates correctly take effect. ([3945])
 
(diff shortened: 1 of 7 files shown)


LukeShu:
Nitpick: We usually don't put periods at the end of the commit subject line. You did in e5c59070dc46267df16e205a41e7d1ad0ad4f2ee

LukeShu (approved):
Hit it!

(PS: Part of why I didn't notice this: you added me as an _assignee_, not as a _reviewer_.)

LukeShu (approved):

PR #3948, opened 2021-11-30
Implement CRD version conversions

LukeShu:
Diffstat, excluding `vendor/`:

    92 files changed, 20958 insertions(+), 1222 deletions(-)

## Description

- The first 9 commits are adding the definitions of the v2 CRDs back in, and dealing with everything that comes with that.
- The next 14 commits are cleaning up the CRDs. Improving the tests, fixing mistakes in the v2 schema, that sort of thing.
- The final 3 commits are actually implementing conversions:
  1. Add `conversion-gen` (in its own commit because of all the noise in `vendor/`)[1]
  2. Get `conversion-gen` to run.
  3. Fill in all of the conversion stubs, adding tests for the conversions. The v2 types gained a bunch of `v3*` fields so that they can store things that are new to v3alpha1. The v3alpha1 types gained a bunch of `v2*` fields for storing how something was written in v2 for all the places where there v2 had multiple ways to write something but v3alpha1 now only has one way to write it.

[1]: Also, I had to make some bug-fixes and an enhancement to `converter-gen`. View those changes at https://github.com/datawire/code-generator/compare/v0.20.2...pinned-v0.20.5-rc.0.0.20211127183116-16d402be64a9 or clone the code-generator repo and look at that commit range locally.

This does NOT implement the ambassador_id mangling to cause 1.x to not see v3alpha1 resources. As discussed on Slack, I'm not sure that's the right way to go. And even if it is, I want to get this unit of work in first.

## Related Issues
List related issues.

## Testing

This is very near to 100% test coverage: The only non-generated file that doesn't have 100% coverage is `handwritten.conversion.go`, and the lines not covered are
 - a bunch of `if err != nil { return err }` checks
 - a couple of `if out.SomeMap == nil { out.SomeMap = make(MapType) }` checks
 - A line handling `DeprecatedSelector` in the v3alpha1→v2 conversion; because it's deprecated, it *intentionally* won't losslessly round-trip back to v3alpha1.

## Checklist

 - [ ] I made sure to update `CHANGELOG.md`. - I'll leave that for (Aidan's) webhook PR
   
    + Any new features
    + Any deprecations
 
 - [x] My change is adequately tested. - near 100% test coverage
 
      (This will be fixed soon.)
 
 - [x] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - not in DEVELOPING.md, but comments in the source files


pkg/api/getambassador.io/v2/handwritten.conversion.go
```diff
@@ -9,26 +9,246 @@
...
+		}
+
+		// 2. outSvc
+		//
+		//                  | explicitScheme="https://" | explicitScheme="http://" |
+		//  | originateTLS  | inSvc                     | force-https://           |
+		//  | !originateTLS | not possible              | inSvc                    |
+		//
+		// Because an https:// scheme forces originateTLS=true, the bottom-right corner
+		// isn't possible
```
kflynn, line 81: It's the bottom left that's not possible, right?

pkg/api/getambassador.io/v2/handwritten.conversion.go
```diff
@@ -9,26 +9,246 @@
...
+		var outExplicitScheme *string
+		if (strings.ToLower(explicitScheme) == "https://") == originateTLS {
```
kflynn, line 83: This is really easy to misread, until you look past it and read the "cover the diagonal" comment. Just sayin'. 🙂

pkg/api/getambassador.io/v2/handwritten.conversion.go
```diff
@@ -9,26 +9,246 @@
...
+		originateTLS = true
+	case strings.HasPrefix(strings.ToLower(inSvc), "http://"):
+		bareSvc = inSvc[len("http://"):]
+	default:
+		bareSvc = inSvc
+	}
+	if inTLS != "" {
+		originateTLS = true
+	}
+	// OK, now re-serialize that in v3alpha1 syntax.
```
kflynn, line 130: v3alpha1 syntax, or v2?

pkg/api/getambassador.io/v3alpha1/common.go
```diff
@@ -25,6 +25,42 @@ import (
...
+// names.  But it should not affect any end-user observable behavior.
+type V2ExplicitTLS struct {
+	// TLS controls whether and how to represent the "tls" field when
+	// its value could be implied by the "service" field.
+	//
+	//  | Value        | Representation                        | Meaning of representation          |
+	//  |--------------+---------------------------------------+------------------------------------|
+	//  | ""           | omit the field                        | defer to service (no TLSContext)   |
+	//  | "null"       | store an explicit "null" in the field | defer to service (no TLSContext)   |
+	//  | "string"     | store an empty string in the field    | defer to service (no TLSContext)   |
```
kflynn, line 41: I thought `string` was a _non_-empty string. It'd be nice to maybe have this be `empty` or some such.

kflynn (approved):
Hit it! My only comments are things that aren't worth messing with until the next time you need to be messing with things anyway.

pkg/api/getambassador.io/v2/handwritten.conversion.go
LukeShu: Oops!

pkg/api/getambassador.io/v2/handwritten.conversion.go
LukeShu: It's also really easy to mis-think. IDK what to do about that.

LukeShu: By the end I was just mashing on the keyboard trying to get the code-coverage and unit tests happy. "What do you mean the `else` case isn't getting covered! The tests are extensive! Ohhh... the `else` case is impossible... how does that make sense?"

pkg/api/getambassador.io/v2/handwritten.conversion.go
LukeShu: Oops!

pkg/api/getambassador.io/v3alpha1/common.go
LukeShu: A _non_-empty string means that there's a TLSContext, and the V2ExplicitTLS.TLS field is ignored.  This field is mostly "how to represent an 'empty' value" (except for `bool:true`, which is special).

LukeShu: What about
```
	// TLS controls whether and how to represent the "tls" field when
	// its value could be implied by the "service" field.  In v2, there
	// were a lot of different ways to spell an "empty" value, and this
	// field specifies which way to spell it (if the value will indeed
	// be empty).
	//
	//  | Value        | Representation                        | Meaning of representation          |
	//  |--------------+---------------------------------------+------------------------------------|
	//  | ""           | omit the field                        | defer to service (no TLSContext)   |
	//  | "null"       | store an explicit "null" in the field | defer to service (no TLSContext)   |
	//  | "string"     | store an empty string in the field    | defer to service (no TLSContext)   |
	//  | "bool:false" | store a Boolean "false" in the field  | defer to service (no TLSContext)   |
	//  | "bool:true"  | store a Boolean "true" in the field   | originate TLS (no TLSContext)      |
```

kflynn: Yeah, that works.

pkg/api/getambassador.io/v2/handwritten.conversion.go
LukeShu: what about
```go
		schemeIsHTTPS := strings.ToLower(explicitScheme) == "https://"
		if schemeIsHTTPS == originateTLS {
			// cover the diagonal
			*outSvc = inSvc
		} else {
			// cover the top-right
```

kflynn: 👍

LukeShu:
v2: Made changes per Flynn's feedback
```patch
diff --git a/pkg/api/getambassador.io/v2/handwritten.conversion.go b/pkg/api/getambassador.io/v2/handwritten.conversion.go
index cb0a12cc1..1e94c6238 100644
--- a/pkg/api/getambassador.io/v2/handwritten.conversion.go
+++ b/pkg/api/getambassador.io/v2/handwritten.conversion.go
@@ -77,10 +77,11 @@ func convert_v2_TLS_To_v3alpha1_TLS(
 		//  | originateTLS  | inSvc                     | force-https://           |
 		//  | !originateTLS | not possible              | inSvc                    |
 		//
-		// Because an https:// scheme forces originateTLS=true, the bottom-right corner
-		// isn't possible
+		// Because an https:// scheme forces originateTLS=true, the bottom-left corner isn't
+		// possible.
 		var outExplicitScheme *string
-		if (strings.ToLower(explicitScheme) == "https://") == originateTLS {
+		schemeIsHTTPS := strings.ToLower(explicitScheme) == "https://"
+		if schemeIsHTTPS == originateTLS {
 			// cover the diagonal
 			*outSvc = inSvc
 		} else {
@@ -88,6 +89,7 @@ func convert_v2_TLS_To_v3alpha1_TLS(
 			outExplicitScheme = &explicitScheme
 			*outSvc = "https://" + bareSvc
 		}
+
 		// 3. outExplicit
 		if outExplicitTLS == "" && outExplicitScheme == nil {
 			*outExplicit = nil
@@ -127,7 +129,7 @@ func convert_v3alpha1_TLS_To_v2_TLS(
 	if inTLS != "" {
 		originateTLS = true
 	}
-	// OK, now re-serialize that in v3alpha1 syntax.
+	// OK, now re-serialize that in v2 syntax.
 	tlsIsTruthy := false
 	if inTLS != "" {
 		*outTLS = &BoolOrString{
diff --git a/pkg/api/getambassador.io/v3alpha1/common.go b/pkg/api/getambassador.io/v3alpha1/common.go
index 0e151ea3a..b1102937e 100644
--- a/pkg/api/getambassador.io/v3alpha1/common.go
+++ b/pkg/api/getambassador.io/v3alpha1/common.go
@@ -32,7 +32,10 @@ import (
 // names.  But it should not affect any end-user observable behavior.
 type V2ExplicitTLS struct {
 	// TLS controls whether and how to represent the "tls" field when
-	// its value could be implied by the "service" field.
+	// its value could be implied by the "service" field.  In v2, there
+	// were a lot of different ways to spell an "empty" value, and this
+	// field specifies which way to spell it (and will therefore only
+	// be used if the value will indeed be empty).
 	//
 	//  | Value        | Representation                        | Meaning of representation          |
 	//  |--------------+---------------------------------------+------------------------------------|
```

kflynn (approved):
Hit it! and thanks. 🙂

Issue #3945, opened 2021-11-23
Mapping.weight should not be modified by weight normalization


kflynn:
When normalizing the weights of the `Mapping`s in a `MappingGroup` (so that every `Mapping` has a weight, and all the weights in the `Group` sum to 100), the code currently modifies `Mapping.weight`. This can cause mis-weighted `Mapping`s when the `Group` is reconfigured with the cache active.

This should, instead, do calculations in a different attribute, so that the user's original assigned `weight` isn't altered in the cache.

kflynn:
Fixed by https://github.com/emissary-ingress/emissary/pull/3949.

khussey:
This was fixed in Emissary-ingress 2.1.0, which is now available.

PR #3949, opened 2021-11-30
Fix cache bugs

kflynn:
#3945 talks about a problem with the incremental-reconfigure cache. tl;dr: don't try
to edit weights in place during normalization, and make sure that a changing `Mapping`
will invalidate any cached `Group` that could contain the `Mapping`.

The first bit is straightforward: if you edit `Mapping.weight` during normalization,
you'll cache the edits, and that's a problem when the `Group` reconfigures, since the
assumption is that things in `Mapping.weight` are supplied by the user. Instead, save
the calculations as `Mapping._weight`.

The second bit is less straightforward. Imagine that you have `Mapping1` and `Mapping2`
that're part of a given `Group`. The cache will contain all three of those. If `Mapping2`
then changes, invalidating `Mapping2` will also invalidate the `Group`, due to the links
in the cache, and then the `Group` will correctly reconfigure. Great.

But now suppose that instead of `Mapping2` changing, what happens is that a new `Mapping3`
arrives that will become a part of the same `Group`. The arrival of `Mapping3` must also
invalidate the `Group` - even though the `Mapping3` isn't currently part of the `Group` -
or the code will see the cached `Group` and decide not to mess with it, which will cause
`Mapping3`s arrival to be ignored. Obviously that's a problem.

Instead, we do multiple passes through the `Mapping`s so that we can correctly invalidate
the `Group`s as needed.

Best reviewed commit-by-commit.

- [ci-skip] Add a test for the bad weight bug
- More debugging
- How was this compiling?
- Don't change the actual Mapping when normalizing weights.
- When we invalidate a Mapping due to a Delta, invalidate any Groups containing it too.
- Make sure internal mappings are cached, and that cached Mappings don't duplicate in their Groups.
- Better cache tests.
- Release notes

### DIFF
--- a/cmd/capabilities_wrapper/capset_other.go
+++ b/cmd/capabilities_wrapper/capset_other.go
@@ -1,3 +1,4 @@
+//go:build !linux
 // +build !linux
 
 package main
--- a/cmd/entrypoint/testutil_fake_test_test.go
+++ b/cmd/entrypoint/testutil_fake_test_test.go
@@ -59,6 +59,87 @@ func TestFake(t *testing.T) {
 
 }
 
+func TestWeightWithCache(t *testing.T) {
+	get_envoy_config := func(f *entrypoint.Fake, want_foo bool, want_bar bool) (*v3bootstrap.Bootstrap, error) {
+		return f.GetEnvoyConfig(func(config *v3bootstrap.Bootstrap) bool {
+			c_foo := FindCluster(config, ClusterNameContains("cluster_foo_"))
+			c_bar := FindCluster(config, ClusterNameContains("cluster_bar_"))
+
+			has_foo := c_foo != nil
+			has_bar := c_bar != nil
+
+			return (has_foo == want_foo) && (has_bar == want_bar)
+		})
+	}
+
+	f := entrypoint.RunFake(t, entrypoint.FakeConfig{EnvoyConfig: true}, nil)
+	assert.NoError(t, f.UpsertYAML(`
+---
+apiVersion: getambassador.io/v3alpha1
+kind: Mapping
+metadata:
+  name: mapping-foo
+  namespace: default
+spec:
+  prefix: /foo/
+  service: foo.default
+`))
+
+	f.Flush()
+
+	// We need an Envoy config that has a foo cluster, but not a bar cluster.
+	envoyConfig, err := get_envoy_config(f, true, false)
+	require.NoError(t, err)
+	assert.NotNil(t, envoyConfig)
+
+	// Now add a bar mapping.
+	assert.NoError(t, f.UpsertYAML(`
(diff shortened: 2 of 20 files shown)


cmd/entrypoint/testutil_fake_test_test.go
```diff
@@ -59,6 +59,87 @@ func TestFake(t *testing.T) {
 
 }
 
+func TestWeightWithCache(t *testing.T) {
+	get_envoy_config := func(f *entrypoint.Fake, want_foo bool, want_bar bool) (*v3bootstrap.Bootstrap, error) {
```
LukeShu, line 63: Nit: Style: Prefer camelCase to snake_case for Go identifiers.

python/ambassador/envoy/v2/v2listener.py
```diff
@@ -562,7 +562,7 @@ def finalize_tcp(self) -> None:
                 # First up, which clusters do we need to talk to?
                 clusters = [{
                     'name': mapping.cluster.envoy_name,
-                    'weight': mapping.weight
+                    'weight': mapping._weight or 0  # It's not actually possible for _weight to be unset
```
LukeShu, line 565: ... so why is it `Optional[int]` and why does this `or 0` need to be here?

LukeShu: Is it saying that it's not possible for it to be unset at this point in the lifecycle? And the `or 0` is just to shut up mypy?

If so, I'd prefer to instead on line 562ish, stick a `assert(isinstance(mapping._weight, int))`. This'll satisfy mypy, check that belief at low `python -O` levels, and optimize that check out at high `python -O` levels.

cmd/capabilities_wrapper/capset_other.go
```diff
@@ -1,3 +1,4 @@
+//go:build !linux
 // +build !linux
```
LukeShu, line 2: Lol, this should have been in my Go 1.17 PR. But my `make golint` didn't check it because I have GOOS=linux, so golint ignored that file!

kflynn: That really makes me feel much, much better. 🙂

LukeShu (commented):
I'm going to be honest, I don't really understand how the cache works, so I don't feel equipped to review this very deeply.

python/ambassador/envoy/v2/v2listener.py
kflynn: I had a long rationale for this typed up, but realized that really, the long rationale can be avoided and we can just set it to 0 at instantiation and that's not worse than what's happening right now. 🤷‍♂️

LukeShu (approved):

PR #3957, opened 2021-12-01
Webhook server code

aidanhahn:
## Description
This PR contains the following:
- edits to entrypoint to refactor the webhook server out into its own module
- edits to entrypoint to load or store CA certificates and keys depending on the existence of a hardcoded k8s secret
- edits to entrypoint to edit all CRDs to leverage its own webhook server
- edits to the charts and manifests to give entrypoint the ability to update secrets and customresourcedefinitions
- edits to the charts and manifests to expose the webhook server through the ambassador-admin service
- edits to the charts and manifests to expose the service name to entrypoint

## Testing
Deployment testing and manual testing were performed to verify operation of this code.

## Checklist

    + Any new features
    + Any deprecations
 
      (This will be fixed soon.)
 

LukeShu:
Commit history could stand to be cleaned up.

charts/emissary-ingress/templates/rbac.yaml
```diff
@@ -110,6 +115,10 @@ rules:
     resources: [ "ingresses/status" ]
     verbs: ["update"]
 
+  - apiGroups: [ "apiextensions.k8s.io" ]
+    resources: [ "*" ]
+    verbs: ["get", "list", "watch", "update", "patch", "create", "delete" ]
+
```
LukeShu, line 121: 1. Does this really need to be `resources: [ "*" ]` instead of just `resources: [ "CustomResourceDefinitions" ]`?
2. Does this really need all of those verbs, and not just get/list/update?

charts/emissary-ingress/templates/rbac.yaml
```diff
@@ -78,10 +78,15 @@ rules:
...
     - services
-    - secrets
     - endpoints
     verbs: ["get", "list", "watch"]
 
+  - apiGroups: [""]
+    resources:
+    - secrets
+    - customresourcedefinitions
+    verbs: ["get", "list", "watch", "update", "patch", "create", "delete" ]
```
LukeShu, line 88: Can't we just give it permission to get/create/update the _one_ secret, instead of giving it everything for all of them?

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,275 @@
...
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/conversion"
+	k8sApiExtClient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
+	k8sApiExtTypes "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+)
+
+const WEBHOOK_PORT int = 8043
+// TODO: automatic cert regeneration
+const CERT_VALID_DAYS int = 365
+const CA_SECRET_NAME string = "emissary-ingress-webhook-ca"
```
LukeShu, line 36: Nit: Style: It's generally against Go style to have underscores in identifiers. Also, these are exported when they don't need to be.

.gitignore
```diff
@@ -1,6 +1,9 @@
 # Docker build pin files
 docker/*.pin
 
+pkg/k8scerts/*.key
+pkg/k8scerts/*.pem
+
```
LukeShu, line 6: Not necessary anymore, right?

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,275 @@
...
+		// Create and write the secret
+		_, err = secrets.Create(ctx, &k8sTypesCoreV1.Secret{
+			ObjectMeta: k8sTypesMetaV1.ObjectMeta{
+				Name: CA_SECRET_NAME,
+				Namespace: GetAmbassadorNamespace(),
+			},
+			Type: k8sTypesCoreV1.SecretTypeTLS,
+			Data: map[string][]byte{
+				"tls.key": caPrivKeyPEM.Bytes(),
+				"tls.crt": caPEM.Bytes(),
```
LukeShu, line 125: Nit: Style: Prefer to use the `corev1.TLSCertKey` and `corev1.TLSPrivateKeyKey` constants instead of string literals.

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	// CA Secret already exists, so load data from it
+	if caPEM == nil || caTemplate == nil || caPrivKey == nil {
+		if caSecret == nil {
+			return fmt.Errorf("Couldnt get or generate CA secret")
+		}
+
+		// Parse CA Key
+		caPrivKeyPEMBytes, ok := caSecret.Data["tls.key"]
```
LukeShu, line 147: Likewise.

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,275 @@
...
+				caPrivKey = key
+			} else {
+				return fmt.Errorf("Bad key loaded in CA secret: %s", err.Error())
+			}
+		} else {
+			return fmt.Errorf("No key found in CA secret!")
+		}
+
+		// parse ca cert
+		caPEMBytes, ok := caSecret.Data["tls.crt"]
```
LukeShu, line 160: Likewise.

charts/emissary-ingress/templates/admin-service.yaml
```diff
@@ -45,6 +45,10 @@ spec:
       targetPort: {{ .Values.adminService.snapshotPort }}
       protocol: TCP
       name: ambassador-snapshot
+    - port: {{ .Values.adminService.webhookPort }}
+      targetPort: {{ .Values.adminService.webhookPort }}
```
LukeShu, line 49: The port number isn't configurable in `webhooks.go`, is it?

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,275 @@
...
+				return err
+			}
+		} else {
+			return fmt.Errorf("No cert found in CA secret!")
+		}
+	}
+
+	// these need to be pointer vals for some reason
+	webhk_prt := int32(WEBHOOK_PORT)
+	webhk_path := "/crdconvert"
```
LukeShu, line 179: Nit: Style: Avoid underscores in identifiers. camelCase!

python/README.md
```diff
@@ -30,6 +30,7 @@ Ambassador uses several TCP ports while running. All but one of them are in the
 | 8002 | `watt`  | Internal `watt` snapshot access; not exposed outside pod |
 | 8003 | `ambex` | Internal `ambex` snapshot access; not exposed outside pod |
 | 8004 | `diagd` | Internal `diagd` access when `AMBASSADOR_FAST_RECONFIGURE` is set; not exposed outside pod |
+| 8043 | `entrypoint` | CR Conversion API exposed by `entrypoint` |
```
LukeShu, line 33: Why 8043? Instead of, say, 8005?

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,275 @@
...
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
+	"net/http"
+	"context"
+	"time"
+	b "bytes"
+	"fmt"
+
+	"github.com/datawire/ambassador/v2/pkg/k8s"
```
LukeShu, line 17: Prefer to use pkg/kates instead of pkg/k8s

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,275 @@
...
+		Addr: addr,
+		Handler: mux,
+		TLSConfig: &tls.Config {
+			GetCertificate: func(clientHello *tls.ClientHelloInfo) (*tls.Certificate, error) {
+				return getCert(clientHello.ServerName, *caTemplate, caPrivKey)
+			},
+		},
+	}
+
+	return srv.ListenAndServeTLS("", "")
```
LukeShu, line 227: For servers, prefer to use github.com/datawire/dlib/dhttp instead of net/http.

charts/emissary-ingress/values.yaml
```diff
@@ -197,6 +197,7 @@ adminService:
   loadBalancerIP:
   # Passed to cloud provider load balancer if created (e.g: AWS ELB)
   loadBalancerSourceRanges:
+  webhookPort: 8043
```
LukeShu, line 200: As I said in the other file, I don't think this is necessary, since `webhooks.go` can't be configured to use a different port. But, if you are going to add something to `values.yaml`, be sure to add a comment; the comment gets used to generate documentation in `charts/emissary-ingress/README.md`.

LukeShu:
1. It'd be good if the webhook goroutine function were exported, so that it can be reused by apro.
2. I'm not sure `cmd/entrypoint` is the best package for it? Maybe it should go in a new package?

cmd/entrypoint/webhooks.go
aidanhahn: k8s here is only used for NewKubeInfo() which kates does not seem to replicate.

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,312 @@
...
+			NotAfter:              notAfter,
+			IsCA:                  true,
+			ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+			KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+			BasicConstraintsValid: true,
+		}
+		caPrivKey, err = rsa.GenerateKey(rand.Reader, 4096)
+		if err != nil {
+			return err
+		}
```
LukeShu, line 116: This is long enough that it probably makes sense to factor out separate `genKey` and `genCert` functions or something.

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,312 @@
...
+		if err := pem.Encode(caPEM, &pem.Block{
+			Type:  "CERTIFICATE",
+			Bytes: caBytes,
+		}); err != nil {
+			return err
+		}
+		caPrivKeyPEM := new(b.Buffer)
+		if err := pem.Encode(caPrivKeyPEM, &pem.Block{
+			Type:  "RSA PRIVATE KEY",
+			Bytes: x509.MarshalPKCS1PrivateKey(caPrivKey),
```
LukeShu, line 133: Using "PRIVATE KEY" / PKCS8 feels more future-proof.

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,312 @@
...
+		// PEM encode certificate and key
+		caPEM = new(b.Buffer)
+		if err := pem.Encode(caPEM, &pem.Block{
+			Type:  "CERTIFICATE",
+			Bytes: caBytes,
+		}); err != nil {
+			return err
+		}
+		caPrivKeyPEM := new(b.Buffer)
+		if err := pem.Encode(caPrivKeyPEM, &pem.Block{
```
LukeShu, line 131: Tip: Just use `pem.EncodeToMemory` instead of managing the buffer yourself.

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,312 @@
...
+			if k8sErrors.IsAlreadyExists(err) {
+				caSecret, err = secrets.Get(ctx, caSecretName, k8sTypesMetaV1.GetOptions{})
+				if err != nil {
+					return err
+				}
+			} else {
+				return err
+			}
+		}
+	}
```
LukeShu, line 160: Unless I'm reading this wrong, `caSecret` ends up `nil` in the case where it actually creates the secret.

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,312 @@
...
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"fmt"
+	"math/big"
+	"net/http"
+	"time"
+
+	k8sTypesCoreV1 "k8s.io/api/core/v1"
+	k8sApiExtTypes "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	k8sApiExtClient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
```
LukeShu, line 19: These should both say `v1`, not `v1beta1`.

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,312 @@
...
+		if ok {
+			keyBlock, _ := pem.Decode(caPrivKeyPEMBytes)
+			if key, err := x509.ParsePKCS1PrivateKey(keyBlock.Bytes); err == nil {
+				caPrivKey = key
+			} else {
+				return fmt.Errorf("bad key loaded in CA secret: %s", err.Error())
+			}
+		} else {
+			return fmt.Errorf("no key found in CA secret!")
+		}
```
LukeShu, line 179: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

cmd/entrypoint/webhooks.go
LukeShu: Also, `API` not `Api` https://github.com/golang/go/wiki/CodeReviewComments#initialisms

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,312 @@
...
+
+		// Parse CA Key
+		caPrivKeyPEMBytes, ok := caSecret.Data[k8sTypesCoreV1.TLSPrivateKeyKey]
+		if ok {
+			keyBlock, _ := pem.Decode(caPrivKeyPEMBytes)
+			if key, err := x509.ParsePKCS1PrivateKey(keyBlock.Bytes); err == nil {
+				caPrivKey = key
+			} else {
+				return fmt.Errorf("bad key loaded in CA secret: %s", err.Error())
```
LukeShu, line 175: Use `%w` when wrapping an underlying error, like
```go
fmt.Errorf("bad key loaded in CA secret: %w", err)
```

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,312 @@
...
+			// Versions is a mandatory field we can rely on
+			// to have at least 1 value. Regardless, we
+			// protect against len=0 in the conditional
+			Version: crd.Spec.Versions[0].Name,
+			Kind:    crd.Spec.Names.Kind,
+		}) {
+			continue
+		}
+
+		count += 1
```
LukeShu, line 241: We do have `count++` in Go :)

cmd/entrypoint/webhooks.go
```diff
@@ -0,0 +1,312 @@
...
+		crd.Spec.Conversion = &webhookConfig
+		_, err := crdInterface.Update(ctx, &crd, k8sTypesMetaV1.UpdateOptions{})
+		if err != nil {
+			errs = append(errs, err)
+		}
```
LukeShu, line 246: 1. This should avoid calling update if the webhook config already matches.
2. This should ignore `k8sErrors.IsConflict(err)` errors, since it is racing the other replicas.

manifests/emissary/ambassador.yaml
```diff
@@ -164,6 +171,10 @@ rules:
 - apiGroups: [extensions, networking.k8s.io]
   resources: [ingresses/status]
   verbs: [update]
+
+- apiGroups: [apiextensions.k8s.io]
+  resources: [customresourcedefinitions, '*']
```
LukeShu, line 176: Surely this should not include `'*'`.

charts/emissary-ingress/templates/rbac.yaml
```diff
@@ -78,10 +78,13 @@ rules:
...
     - namespaces
     - services
-    - secrets
     - endpoints
     verbs: ["get", "list", "watch"]
 
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get", "update", "patch", "create" ]
+
```
LukeShu, line 87: Trailing whitespace on this line.

charts/emissary-ingress/templates/rbac.yaml
```diff
@@ -110,6 +113,10 @@ rules:
     resources: [ "ingresses/status" ]
     verbs: ["update"]
 
+  - apiGroups: [ "apiextensions.k8s.io" ]
+    resources: [ "customresourcedefinitions", '*' ]
+    verbs: [ "list", "update", "patch" ]
```
LukeShu, line 118: This is in the `Role`/`ClusterRole` for namespaced resources that it has permission to read cluster-wide (unless "single-namespace" is turned on). It is normally a `ClusterRole`, but gets downgraded to a `Role` if single-namespace is turned on.

Because CRDs are non-namespaced resources, they need to go in a `ClusterRole` *even if* single-namespace is turned on; there's an appropriate `ClusterRole` further up in the file.

docs/releaseNotes.yml
```diff
@@ -34,6 +34,14 @@ items:
   - version: 2.1.0
     date: 'TBD'
     notes:
+      - title: CRD conversion webhooks
+        type: feature
+        body: >-
+          The ambassador-admin service now hosts a webhook server that enables the
+          user to seamlessly request CRs in whatever API version is desired. The
+          webhook server leverages its own CA to perform the TLS connections required
+          of it. The CA certificate and private key are stored in a k8s secret.
```
LukeShu, line 43: This isn't a feedback, but just a stupid curiosity question: This text is wrapped to about 85 columns, but the top of the file says `# -*- fill-column: 100 -*-` and you use Emacs. Are you wrapping the text by hand instead of using M-q?  Is your Emacs ignoring the `-*-` line for some reason?

aidanhahn: I wrote that out by hand.

LukeShu:
Closed in favor of https://github.com/emissary-ingress/emissary/pull/3969

Issue #3818, opened 2021-09-21
Emissary sends gzip encoded payloads to backend services


kmott:
### Describe the bug
If we send certain payloads to Emissary, Emissary seems to be gzip encoding the payload and forwarding it to the upstream service.  Most often, we send Content-Types application/json, text/html (and a few others) which all seem to be gzip encoded before forwarding upstream (which is in-line with what the [docs](https://www.getambassador.io/docs/edge-stack/latest/topics/running/gzip/#the-gzip-api) say the default behaviour is).

Maybe we have something mis-configured, but I don't think Emissary should gzip encode the payload before sending to the backend service?

### To Reproduce
1. Deploy Ambassador to your cluster
2. Setup a `Module` like this:

```yaml
---
apiVersion: getambassador.io/v2
kind: Module
metadata:
  name: ambassador
  namespace: ambassador
spec:
  ambassador_id: gzip-test
  config:
    gzip:
      enabled: true
    use_remote_address: true
    xff_num_trusted_hops: 1
    server_name: "gzip-test"
[ ... ]
```

3. Setup a service and mapping to pass the payload from Emissary to the backend.
4. Send a `curl` request like this, notice it is NOT gzip encoded:

```
curl -i -vv -X POST curl --silent -o /dev/null https://my-ingress.example.com -H "Content-Type: application/json" -d '{"foo":"bar"}'
```
Backend service request:
```
{
  body: '{"foo":"bar"}',
  headers: Headers {
    accept: "*/*",
    "content-type": "application/json",
  },
  [ ... ]
}
```

5. Send a `curl` request like this, with a little larger payload, notice the payload body is gzip encoded:

```
curl -i -vv -X POST curl --silent -o /dev/null https://my-ingress.example.com -H "Content-Type: application/json" -d '{"deadbeef":{"foo":150,"bar":100},"ea7beef":242,"foobaz":240}'
```

Backend service request:
```
{
  body: '\x1f�\b\x00\x00\x00\x00\x00\x00\x03�V*-N-R��V [ ... ]',
  headers: Headers {
    accept: "*/*",
    "accept-encoding": "gzip",
    "content-encoding": "gzip",
    "content-type": "application/json",
    "transfer-encoding": "chunked",
  },
  [ ... ]
}
```

### Expected behavior
My understanding is that the gzip encode/decode should only happen between the client <-> Emissary, and Emissary should not forward that gzip encoded content (or gzip encode the data itself), and forward to the backend service.

### Versions
[...]

PR #3967, opened 2021-12-04
Stop sending upstreams encoded data when using gzip

AliceProxy:
Signed-off-by: AliceProxy <aliceproxy@protonmail.com>

## Description
Fixes a bug introduced in `1.14.0` where upstream services would receive gzip encoded data when using gzip compression in the `Module`.

Thanks to @kflynn for finding the fix

## Related Issues
https://github.com/emissary-ingress/emissary/issues/3818

## Notes:
A future version of edgissary may expose the ability to enable sending encoded data to upstream, but the default should be to not do that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,9 +93,9 @@ Please see the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest
   does), an error would always be logged about the Secret not being present, even though it was
   present, and everything was working correctly. This error is no longer logged.
 
-- Bugfix: When using gzip compression upstream services will no longer recieve encoded data. This
-  bug was introduced in 1.14.0. The fix restores the default behavior of  not sending encoded data
-  to upstream services. ([3945])
+- Bugfix: When using gzip compression, upstream services will no longer receive compressed data.
+  This bug was introduced in 1.14.0. The fix restores the default behavior of  not sending
+  compressed data to upstream services. ([3818])
 
 - Security: Update to busybox 1.34.1 to resolve CVE-2021-28831, CVE-2021-42378, CVE-2021-42379,
   CVE-2021-42380, CVE-2021-42381, CVE-2021-42382, CVE-2021-42383, CVE-2021-42384, CVE-2021-42385,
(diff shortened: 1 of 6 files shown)


kflynn (approved):
Ahh, need to update `docs/releaseNotes.yml` and then run `make generate-fast` to get the `CHANGELOG` too. 🙂  Code looks good though!

kflynn (changes requested):
D'oh. Meant to request changes instead of approve. 🙄

kflynn (changes requested):
Sorry, one more minor thing! 😐

kflynn (approved):
Off we go! Thanks @AliceProxy! 🙂 

(Yes, I force-pushed it, that's 'cause I need to sync release notes with the docs. 🙂)

kflynn (approved):
Yes yes, still approved after a typo fix. 🙂

sharkymcdongles:
Any chance you can backport this fix to 1.14.x?

mrnonz:
Hope this view release to v1.14.5 too https://github.com/emissary-ingress/emissary/blob/master/CHANGELOG.md#1145-tbd

AliceProxy:
> Hope this will release to v1.14.5 too https://github.com/emissary-ingress/emissary/blob/master/CHANGELOG.md#1145-tbd

@mrnonz @sharkymcdongles We currently have no plans of officially releasing a `1.14.5`, but we did end up doing a dev build of `1.14` containing the gzip bug fix because a specific customer was unable to upgrade Edge-Stack from a much older version to `1.14.4` in order to get on the new the new 2.x line due to this bug. The `TBD 1.14.5` exists as a placeholder because `1.14` is still in extended maintenance mode.

This means that should Envoy release security patches for discovered CVEs before `1.14` hits its full End of Life, we would port said security patches back into `1.14` necessitating the `1.14.5` release. Given that this has already been finished and tested for a dev-build, we would include it in said `1.14.5` release if that scenario were to happen, but if not then there will not be a `1.14.5` release.

I can provide the image for the dev-build if you need it because you are in a similar situation, but please understand that dev-builds do not go through the full suite of testing that regular release images do. Additionally, I would strongly recommend only using it as a stepping-stone to help upgrade to 2.x since as I mentioned, unless there are high severity security patches that we need to backport, `1.14` will not receive any additional support, updates, or bugfixes before its EOL. 

Edge-stack dev-build: `datawiredev/aes:1.14.5-hf.1-1956`

For emissary I do not have an image on-hand to provide, but you can build your own off of the [release/v1.14](https://github.com/emissary-ingress/emissary/tree/release/v1.14) branch. Let me know if you need any assistance doing so if you are using Emissary instead of Edge-Stack.

PR #3972, opened 2021-12-07
fix promote jobs

aidanhahn:
The promote jobs are broken due to a go dependency. They have been modified in this PR to use the same "Install Go" action that the testing workflow uses.


LukeShu (approved):

kflynn (approved):
Hit it!

PR #3974, opened 2021-12-07
Set Ambassador node name as Envoy's node.id

psalaberria002:
## Description
Set a sensible value for node.id in Envoy's configuration.

When looking at traces, I noticed there was a node.id tag set to "test-id".
It would be very useful to get the ambassador pod name instead in that field.

I am not sure what other implications this change may have, but it seems harmless at a first glance.

## Testing
I have not done any testing so far.

## Checklist

    + Any new features
    + Any deprecations
 

kflynn (approved):
Taking this one -- sorry for the delay checking this out on our end. 🙁

PR #3981, opened 2021-12-17
fix: updated helm template module's api kind to version v3alpha1

rkalkani:
## Description
Emissary-ingress helm chart 7.2 installations with module give error regarding the module version
```
Error: UPGRADE FAILED: unable to recognize "": no matches for kind "Module" in version "getambassador.io/v2"
```

Updated the module's apiVersion to match with the one used in the documentation for version emissary 2.1


kflynn:
With the release of Emissary 2.1, we have also released chart 7.2.0, which I believe subsumes this PR. Thanks!

PR #4004, opened 2022-01-10
Fix matching a Mapping with host_redirect a Host

### LINKED ISSUE #3709: User reports difficulty setting up redirects in Emissary 2.0
From user Tim Rice: 

Hi! Trying out the emissary-ingress 2.0 dev preview... having trouble getting redirects working, even using the basic redirect example from the docs

```
apiVersion: x.getambassador.io/v3alpha1
kind: AmbassadorMapping
metadata:
  name: redirect
spec:
  prefix: /redirect/
  service: httpbin.org
  host_redirect: true
  hostname: ‘*’
```

just returns a 404.
accessing an application in the cluster works just fine...
is this a known problem or are we doing something wrong?

Comment from Cindy:
The user is reporting an issue with redirects for Emissary 2.0 and it seems like he's using the new CRDs... he says it works for 1.13.

gferon:
## Description

While preparing the upgrade from 1.x to 2.x, we noticed that any `Mapping` with `host_redirect` set would lead to a 404, which is the issue described #3709. It turns out that the issue lies in the `matches_httpgroup` function of `irhost.py`.

I'd first like to get some feedback before I proceed with adding a regression test or do anything else with this PR.

## Related Issues
- #3709

## Testing

Manual testing (image rebuild + running the example) proved that this patch works.

## Checklist

    + Any new features
    + Any deprecations
 
Resolves #3709 and #4005

### DIFF
--- a/python/ambassador/ir/irhost.py
+++ b/python/ambassador/ir/irhost.py
@@ -383,7 +383,7 @@ class IRHost(IRResource):
         else:
             # It is NOT A TYPO that we use group.get("host") here -- whether the Mapping supplies
             # "hostname" or "host", the Mapping code normalizes to "host" internally.
-            group_glob = group.get('host') or None  # NOT A TYPO: see above.
+            group_glob = group.get('host') or group.get('host_redirect', {}).get('host') or None  # NOT A TYPO: see above.
 
             if group_glob:
                 host_match = hostglob_matches(self.hostname, group_glob)


AliceProxy:
Hi there @gferon , thank you for your contribution  :rocket: . We are taking a look at this and will get back to you shortly if we have any questions or comments for you :slightly_smiling_face:

AliceProxy:
Hi @gferon, I just wanted to let you know that I'm closing this PR since your fix was cherry-picked and merged as part of  [4034](https://github.com/emissary-ingress/emissary/pull/4034) so we could get this fix into the next release ASAP seeing as it is a regression. 

We actually had a test for this behavior but it was unfortunately disabled. If you'd like to submit future contributions to Emissary-ingress, here is a little info on how some of the tests work:

`python/tests/kat`: These tests work by applying some config to a cluster and then running checks such as queries, etc.

`python/tests/gold`: These tests ensure that the Envoy config built by Emissary matches the Envoy config in the gold test folders. If you make changes that would modify the Envoy config then you will need to update these files to account for the new bits in the Envoy config.

`python/tests/unit`: These tests generate an Envoy config from Emissary's CRDs and then check the resulting Envoy config for specific expected values and values expected to be omitted. 

Normally we would provide feedback to your original PR if the urgency of getting the fix into the next release was lower. Thanks so much for bringing this issue to our attention and providing the fix for it. We really appreciate your contribution!

Issue #4005, opened 2022-01-11
host_redirect doesn't work with v2.1.0


gferon:
**Describe the bug**
`Mapping` resources with `host_redirect: true` do not end up in the Envoy configuration file.

**To Reproduce**
1. Install `emissary-ingress` Helm chart with default parameters
2. Create a wildcard `Host` resource
3. Create a `Mapping` object like [from the example](https://www.getambassador.io/docs/edge-stack/latest/topics/using/redirects/#basic-redirect)
4. Try to access the resource: you get "404 Not Found"

**Expected behavior**
"301: Moved Permanently"

**Versions (please complete the following information):**
 - Ambassador: 2.1.0
 - Kubernetes environment: doesn't matter

gferon:
Probably a duplicate of #3709 but describes the actual problem.

khussey:
This was fixed in Emissary-ingresss 2.1.2, which is now available.

PR #4019, opened 2022-01-19
Apply fix for host_redirection regression and re-enable tests

AliceProxy:
## Description
Cherry picking a community contribution and making a small change to the tests so they actually run now so we can get this out in the next release.

## Related Issues
Original Contribution: https://github.com/emissary-ingress/emissary/pull/4004
Related issues: 
https://github.com/emissary-ingress/emissary/issues/3709
https://github.com/emissary-ingress/emissary/issues/4005

## Testing
We already had tests that should have discovered that this regression occurred, but those tests were set to not run. I've re-enabled them and updated the gold test config to account for the fact that the redirects are now showing up in the Envoy config.

## Checklist

    + Any new features
    + Any deprecations
 

CHANGELOG.md
```diff
@@ -103,6 +103,16 @@ Please see the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest
 
 [3906]: https://github.com/emissary-ingress/emissary/issues/3906
 
+## [2.1.2] December 16, 2021
```
khussey, line 106: This, of course, would need to be updated to reflect the release date (current target is January 20, 2022).

docs/releaseNotes.yml
```diff
@@ -53,7 +53,20 @@ items:
           Support for the Envoy V2 API and the `AMBASSADOR_ENVOY_API_VERSION` environment
           variable have been removed. Only the Envoy V3 API is supported (this has been the
           default since Emissary-ingress v1.14.0).
- 
+
+  - version: 2.1.2
+    date: '2021-12-16'
```
khussey, line 58: Should be 2022-01-20

CHANGELOG.md
AliceProxy: This and the following date are done by `make generate` and keep overwriting my `TBD` entry for the date (I'm assuming because there is already a `2,2` version above listed as `TBD` :cry:

AliceProxy:
Closing this PR and going to re-open another branch since the target branch changed and things are getting a little messy

PR #4024, opened 2022-01-20
Go build cache and GHA

kflynn:
Enable `DOCKER_BUILDKIT` for CI. This is important because mounting the Go build cache in the Docker build requires BuildKit, and we really, really don't want different Dockerfiles in dev and CI.

And if you're curious why this doesn't use GHA `env:`:

- When I tried it, it didn't work.
- Iterating on this is slow as a dog.
- I want it running now, because it's blocking CI for things I care about.

I'll revisit `env:` later this week.

 - [x] This is CI, so it cannot affect how Ambassador performs at scale.
 - [x] CI is literally the point of this change: it's the only test needed.

### DIFF
--- a/.github/actions/execute-job-image/action.yml
+++ b/.github/actions/execute-job-image/action.yml
@@ -14,10 +14,12 @@ runs:
     - name: "make push"
       shell: bash
       run: |
+        export DOCKER_BUILDKIT=1
         make push
     - name: "make push-dev"
       shell: bash
       run: |
+        export DOCKER_BUILDKIT=1
         make push-dev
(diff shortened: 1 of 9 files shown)


.github/actions/execute-job-image/action.yml
```diff
@@ -14,10 +14,12 @@ runs:
     - name: "make push"
       shell: bash
       run: |
+        export DOCKER_BUILDKIT=1
```
LukeShu, line 17: It probably makes more sense to set this globally for everything in the Makefile?

PR #4013, opened 2022-01-18
Versioning refactor; take 2, part 2

LukeShu:
## Description

This PR is the continuation of #4003; which was me taking over @aidanhahn's #3993.  The big goal of this PR is picking up all the goodness from Aidan's original PR that I dropped for #4003. That is: Doing cleanup enabled by the changes in #4003.

## Related Issues
- #3993
- #4003

## Testing
I've used this to drive a v2.1.2 release in the https://github.com/aidanhahn/emissary repo.

## Checklist

 - [x] I made sure to update `CHANGELOG.md`. - no applicable changes
   
   Remember, the CHANGELOG needs to mention:
    + Any new features
    + Any changes to our included version of Envoy
    + Any non-backward-compatible changes
    + Any deprecations
 
 - [x] This is unlikely to impact how Ambassador performs at scale. - minimal runtime changes
 
   Remember, things that might have an impact at scale include:
    + Any significant changes in memory use that might require adjusting the memory limits
    + Any significant changes in CPU use that might require adjusting the CPU limits
    + Anything that might change how many replicas users should use
    + Changes that impact data-plane latency/scalability
 
 - [x] My change is adequately tested. - I'd say so
 
   Remember when considering testing:
    + Your change needs to be specifically covered by tests.
       + Tests need to cover all the states where your change is relevant: for example, if you add a behavior that can be enabled or disabled, you'll need tests that cover the enabled case and tests that cover the disabled case. It's not sufficient just to test with the behavior enabled.
    + You also need to make sure that the _entire area being changed_ has adequate test coverage.
       + If existing tests don't actually cover the entire area being changed, add tests.
       + This applies even for aspects of the area that you're not changing – check the test coverage, and improve it if needed!
    + We should lean on the bulk of code being covered by unit tests, but...
    + ... an end-to-end test should cover the integration points
 
 - [x] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - no tricks

PR #4025, opened 2022-01-20
Enable Docker BuildKit for CI (v2)

kflynn:
Enable Docker BuildKit for CI. This is important because mounting the Go build cache in the Docker build requires BuildKit, and we really, really don't want different Dockerfiles in dev and CI.

My original version (#4024) exported `DOCKER_BUILDKIT` in the inidividual workflows; this one just forces it in `builder.mk` globally. I was originally concerned about some sort of unintended consequence of the global setting, but it seems OK and it's definitely much cleaner, so probably a good thing that @LukeShu called me on it. 😉 

 - [x] This is CI, so it cannot affect how Ambassador performs at scale.
 - [x] CI is literally the point of this change: it's the only test needed.

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -1,3 +1,5 @@
+# syntax = docker/dockerfile:1.3
+
 ###
 # This dockerfile builds all the source code and docker images for the
 # edge stack.
@@ -77,7 +79,9 @@ ADD pkg pkg
 ADD vendor vendor
 ADD go.mod go.mod
 ADD go.sum go.sum
-RUN mkdir -p /go/bin && \
+
+RUN --mount=type=cache,target=/root/.cache/go-build \
+    mkdir -p /go/bin && \
 	time go build -mod=vendor -o /go/bin/ ./cmd/...
 
 ########################################
--- a/builder/builder.mk
+++ b/builder/builder.mk
@@ -146,7 +146,9 @@ INGRESS_TEST_LOCAL_ADMIN_PORT = 8877
 INGRESS_TEST_MANIF_DIR = $(BUILDER_HOME)/../manifests/emissary/
 INGRESS_TEST_MANIFS = emissary-crds.yaml emissary-emissaryns.yaml
 
-# export DOCKER_BUILDKIT := 0
+# DOCKER_BUILDKIT is _required_ by our Dockerfile, since we use Dockerfile extensions for the
+# Go build cache. See https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md.
+export DOCKER_BUILDKIT := 1
 
 all: help
 .PHONY: all
--- a/docs/releaseNotes.yml
+++ b/docs/releaseNotes.yml
@@ -35,7 +35,12 @@ items:
     prevVersion: 2.1.0
     date: 'TBD'
     notes:
-      # next changelog item here
+      - title: Docker BuildKit always used for builds
+        type: change
+        body: >-
+          Docker BuildKit is enabled for all Emissary builds. Additionally, the Go
+          build cache is fully enabled when building images, speeding up repeated builds.
+        docs: https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md
 
   - version: 2.1.1
     date: 'N/A'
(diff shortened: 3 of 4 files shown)


builder/builder.mk
```diff
@@ -146,7 +146,7 @@ INGRESS_TEST_LOCAL_ADMIN_PORT = 8877
 INGRESS_TEST_MANIF_DIR = $(BUILDER_HOME)/../manifests/emissary/
 INGRESS_TEST_MANIFS = emissary-crds.yaml emissary-emissaryns.yaml
 
-# export DOCKER_BUILDKIT := 0
```
LukeShu, line 149: lmao

docs/releaseNotes.yml
```diff
@@ -35,7 +35,12 @@ items:
     prevVersion: 2.1.0
     date: 'TBD'
     notes:
-      # next changelog item here
+      - title: Docker BuildKit always used for builds
+        type: change
+        body: >-
+          Docker BuildKit is enabled for all Emissary builds. Additionally, the Go
+          build cache is fully enabled when building images, speeding up repeated builds.
+        docs: https://docs.docker.com/develop/develop-images/build_enhancements/
```
LukeShu, line 43: We don't usually put tooling changes like this in the changelog, do we?

LukeShu (changes requested):
You should drop my ldflags commit, since you had me make changes to that commit over at https://github.com/emissary-ingress/emissary/pull/4013 (which I think I also want Aidan's review on before merging).

docs/releaseNotes.yml
kflynn: We did when we first forced `DOCKER_BUILDKIT` to 0 in the Makefile (since it was affecting people trying to build) so it seemed like we should include this one.

builder/Dockerfile
```diff
@@ -77,8 +79,11 @@ ADD pkg pkg
 ADD vendor vendor
 ADD go.mod go.mod
 ADD go.sum go.sum
-RUN mkdir -p /go/bin && \
-	time go build -mod=vendor -o /go/bin/ ./cmd/...
+
+RUN --mount=type=cache,target=/root/.cache/go-build \
+    GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
```
LukeShu, line 84: Am I missing something? I Don't think `TARGETOS` or `TARGETARCH` are set anywhere? So this is just setting `GOOS` and `GOARCH` to empty strings, which is the same as leaving them unset?

LukeShu: Also, they're only set for the `mkdir` and not for the `go build`.

docs/releaseNotes.yml
LukeShu, line 43: The buildkit syntax extensions reference (https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md) is probably a better link than Docker's how-to guide. Especially since Docker's guide refers to Docker's Dockerfile syntax reference, which _doesn't_ mention the buildkit extensions.

builder/builder.mk
```diff
@@ -146,7 +146,7 @@ INGRESS_TEST_LOCAL_ADMIN_PORT = 8877
...
+export DOCKER_BUILDKIT := 1
```
LukeShu, line 149: Not that I expect `docker build` to stick around too much longer (yay ocibuild), but there should probably be a comment here about us using buildkit syntax extensions, and a link to https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md

LukeShu (changes requested):
- remove superfluous env-vars from `mkdir`
- update changelog link
- add a comment to `builder.mk`

builder/Dockerfile
kflynn: 🙄  That's what I get for copy & pasting as a quick test, finding it works, and then not trying to trim it down...

docs/releaseNotes.yml
kflynn: 👍

LukeShu (approved):
Hit it! (Well, after you get a second reviewer)

AliceProxy (approved):

PR #4026, opened 2022-01-20
Fixes for 2.1.2

kflynn:
Fixes for 2.1.2. This used to be #4022, but GitHub didn't read my mind when I rebased onto `flynn/dev/buildkit-2` and closed `flynn/dev/buildkit-test`. D'oh!

Best to review this commit-by-commit:
f34cf759d: Disable fail-fast in the `pytest` matrix -- it causes us more harm than good right now
34a438987: Two places had \u00A0 (non-breaking space) where they should've had \u0020 (space).
719814a14: pkg/kates/validation.go: Use modern error packages
35ebdf24f: Make sure that CORS Envoy-config generation doesn't include fields that aren't correct Envoy configuration
6700c1cf2: Upgrade `k8s.io/kube-openapi` to fix support for `rewrite: ""` in a `Mapping`
47c1c442c: Log conversion-request inputs & outputs in `apiext` if debugging is enabled
dd182079d: Use the correct mapping weight when generating Envoy routes (!! -- how did CI not find this??)

 - [x] I made sure to update `CHANGELOG.md`.
 - [x] This is unlikely to impact how Ambassador performs at scale.
 - [ ] My change is adequately tested.
 - [x] I didn't need to update `DEVELOPING.md`.

PR #4028, opened 2022-01-20
Fix redirect mapping regression

AliceProxy:
## Description
Cherry picking a community contribution and making a small change to the tests so they actually run now so we can get this out in the next release.

## Related Issues
Original Contribution: https://github.com/emissary-ingress/emissary/pull/4004
Related issues: 
https://github.com/emissary-ingress/emissary/issues/3709
https://github.com/emissary-ingress/emissary/issues/4005

## Testing
We already had tests that should have discovered that this regression occurred, but those tests were set to not run. I've re-enabled them and updated the gold test config to account for the fact that the redirects are now showing up in the Envoy config.

## Checklist

    + Any new features
    + Any deprecations
 
### DIFF
--- a/python/ambassador/ir/irhost.py
+++ b/python/ambassador/ir/irhost.py
@@ -383,7 +383,8 @@ class IRHost(IRResource):
         else:
             # It is NOT A TYPO that we use group.get("host") here -- whether the Mapping supplies
             # "hostname" or "host", the Mapping code normalizes to "host" internally.
-            group_glob = group.get('host') or group.get('host_redirect', {}).get('host') or None  # NOT A TYPO: see above.
+            redirect_glob = group.get('host_redirect', {}).get('host') if group.get('host_redirect', {}) else None
+            group_glob = group.get('host') or redirect_glob or None  # NOT A TYPO: see above.
 
             if group_glob:
                 host_match = hostglob_matches(self.hostname, group_glob)
--- a/python/ambassador/ir/irhost.py
+++ b/python/ambassador/ir/irhost.py
@@ -383,7 +383,7 @@ class IRHost(IRResource):
         else:
             # It is NOT A TYPO that we use group.get("host") here -- whether the Mapping supplies
             # "hostname" or "host", the Mapping code normalizes to "host" internally.
-            group_glob = group.get('host') or None  # NOT A TYPO: see above.
+            group_glob = group.get('host') or group.get('host_redirect', {}).get('host') or None  # NOT A TYPO: see above.
 
             if group_glob:
                 host_match = hostglob_matches(self.hostname, group_glob)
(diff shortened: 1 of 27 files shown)


LukeShu:
Does this need to be on top of `shared/2.1.2-fixes`? I ask because that PR (#4026) already needs to be rebased on to newer https://github.com/emissary-ingress/emissary/pull/4025 (which also has some changes requested on it). If there's not a reason it needs to be on top of `shared/2.1.2-fixes`, it would probably be simpler to just target `release/v2.1` directly.

LukeShu:
Oops, I accidentally clicked "Close with comment" instead of "Comment".

python/ambassador/ir/irhost.py
```diff
@@ -383,7 +383,8 @@ def matches_httpgroup(self, group: 'IRHTTPMappingGroup') -> bool:
         else:
             # It is NOT A TYPO that we use group.get("host") here -- whether the Mapping supplies
             # "hostname" or "host", the Mapping code normalizes to "host" internally.
-            group_glob = group.get('host') or None  # NOT A TYPO: see above.
+            redirect_glob = group.get('host_redirect', {}).get('host') if group.get('host_redirect', {}) else None
```
LukeShu, line 386: I'm confused, isn't this equivalent to just
```python
redirect_glob = group.get('host_redirect', {}).get('host')
```
?

python/ambassador/ir/irhost.py
```diff
@@ -383,7 +383,8 @@ def matches_httpgroup(self, group: 'IRHTTPMappingGroup') -> bool:
...
+            group_glob = group.get('host') or redirect_glob or None  # NOT A TYPO: see above.
```
LukeShu, line 387: If `redirect_glob` is falsey, doesn't that mean it's `None`? Isn't the `or None` superfluous?

LukeShu:
I would encourage you to squash this in to just 1 commit (and add a bit to the commit message explaining what's different from @gferon's original commit).

python/ambassador/ir/irhost.py
AliceProxy: The main issue with why this looks so gross is that `group.get('host_redirect', {})` is sometimes `None` meaning that a `group.get('host_redirect', {}).get('host')` will throw an error because you cant `NoneType.get('host')` thats why I first check if anything is returned by `group.get('host_redirect', {})` before doing `group.get('host_redirect', {}).get('host')` that said, there is still probably a better way to write that out.

python/ambassador/ir/irhost.py
AliceProxy: This is true, I can definitely change that

python/ambassador/ir/irhost.py
kflynn: I would probably do 

```
  # It's possible for group.host_redirect to be None instead of missing, and it's also
  # conceivably possible for group.host_redirect.host to be "", which we'd rather be 
  # None. Hence we do this two-line dance to massage the various cases.
  host_redirect = group.get('host_redirect') or {}
  redirect_glob = host_redirect.get('host') or None
```

This might not be "better" but at least there's room for the explanation...

kflynn (changes requested):
I agree with @LukeShu that this could be collapsed into one commit, and that some of the Python can be cleaned up a bit. Thanks for taking this one on! 

(Oh, and I think it can go directly onto `release/v2.1`, too.)

LukeShu:
closing in favor of https://github.com/emissary-ingress/emissary/pull/4034

PR #4016, opened 2022-01-19
Move getambassador.io/config annotation unfolding to happen in entrypoint.go (part 2; Python)

LukeShu:
## Description
This PR is part 2 of an update to https://github.com/emissary-ingress/emissary/pull/2271.  It completes the Python-side of the change, and removes a bunch of now-dead code.

As usual for me, I suggest a commit-by-commit review.

## Related Issues
- https://github.com/emissary-ingress/emissary/pull/2271 (original version)
- https://github.com/emissary-ingress/emissary/pull/4014 (part 1)
- https://github.com/emissary-ingress/emissary/issues/4006 (issue bringing apiVersion issues to light)
- https://github.com/emissary-ingress/emissary/pull/4029 (an issue discovered by this work)

## Testing
CI should adequately verify that this doesn't break anything. This PR doesn't add any testing, though it would benefit from the "run the v1.14 tests" work.

## Checklist

 - [x] I made sure to update `CHANGELOG.md`. - yes
   
   Remember, the CHANGELOG needs to mention:
    + Any new features
    + Any changes to our included version of Envoy
    + Any non-backward-compatible changes
    + Any deprecations
 
 - [ ] This is unlikely to impact how Ambassador performs at scale. - actually, it should improve things
 
   Remember, things that might have an impact at scale include:
    + Any significant changes in memory use that might require adjusting the memory limits
    + Any significant changes in CPU use that might require adjusting the CPU limits
    + Anything that might change how many replicas users should use
    + Changes that impact data-plane latency/scalability
 
 - [ ] My change is adequately tested. - no, this would benefit from the "use the v1.14 tests" work
 
   Remember when considering testing:
    + Your change needs to be specifically covered by tests.
       + Tests need to cover all the states where your change is relevant: for example, if you add a behavior that can be enabled or disabled, you'll need tests that cover the enabled case and tests that cover the disabled case. It's not sufficient just to test with the behavior enabled.
    + You also need to make sure that the _entire area being changed_ has adequate test coverage.
       + If existing tests don't actually cover the entire area being changed, add tests.
       + This applies even for aspects of the area that you're not changing – check the test coverage, and improve it if needed!
    + We should lean on the bulk of code being covered by unit tests, but...
    + ... an end-to-end test should cover the integration points
 
 - [x] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - no tricks

PR #4029, opened 2022-01-21
CRDs: Fix being able to have empty-but-set lists and dicts

LukeShu:
## Description

One of the reasons that https://github.com/emissary-ingress/emissary/pull/4016 is failing is that the Go CRD definitions are failing to preserve set-but-empty `add_request_headers: {}` and `remove_request_headers: []` fields.  Preserving these is important because if the `ambassador` `Module` sets a non-empty global default, you must be able to explicitly set it to an empty value in a Mapping to.

So, since both a nil and an empty slice trigger `,omitempty`, I just added `*` in front of the types to change what "empty" means.  Unfortunately, this causes both `conversion-gen` and `controller-gen` to generate invalid code.

So I went to work patching both `conversion-gen` and `controller-gen`.  As much as I hate checking in `vendor/`, I have to admit that in this instance it makes the PR easier to review, since I don't have to link to a diff in an external repo.  So, yeah, unlike most PRs you actually should pay attention to the changes in `vendor/` this time.

Previously, sometimes when it encountered 2 types it couldn't convert, `conversion-gen` would emit `compileErrorOnMissingConversion()` (triggering a compile error), and sometimes it would emit `// WARNING: in.StructMemberName requires manual conversion: inconvertible types (Type1 vs Type2)`, and I'd have told you that I had absolutely zero intuition for when it would do one vs the other.  Well, with this latest change it always does `compileErrorOnMissingConversion()`, so I had to add a few `// +k8s:conversion-gen=false` comments to prevent compile errors.

I split the human-made in-repo changes in emissary.git in to one commit, and held off updating `vendor/` or running `make generate` for a second commit.  It might be helpful to review these separately, IDK.

## Related Issues
(None)

## Testing
This adds a unit test to both v2 and v3alpha1.

## Checklist

 - [x] I made sure to update `CHANGELOG.md`.  - yes
   
    + Any new features
    + Any deprecations
 
 - [x] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - no new tricks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,7 +102,7 @@ Please see the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest
   `add_response_headers`, `remove_request_headers`, or `remove_response_headers`, it is often
   desirable to be able to turn off that setting locally for a specific `Mapping`. For several
   releases this has not been possible for `Mappings` that are native Kubernetes resources (as
-  opposed to annotations), as an emtpy value ("mask the global default") was erroneously considered
+  opposed to annotations), as an empty value ("mask the global default") was erroneously considered
   to be equivalent to unset ("inherit the global default").  This is now fixed.
 
 ## 2.1.1 not issued
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,13 @@ Please see the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest
 
 ### Emissary-ingress and Ambassador Edge Stack
 
+- Bugfix: If the `ambassador` `Module` sets a global default for `add_request_headers`,
+  `add_response_headers`, `remove_request_headers`, or `remove_response_headers`, it is often
+  desirable to be able to turn off that setting locally for a specific `Mapping`. For several
+  releases this has not been possible for `Mappings` that are native Kubernetes resources (as
+  opposed to annotations), as an emtpy value ("mask the global default") was erroneously considered
+  to be equivalent to unset ("inherit the global default").  This is now fixed.
+
 ## 2.1.1 not issued
 
 *Emissary-ingress 2.1.1 was not issued; Ambassador Edge Stack 2.1.1 uses Emissary-ingress 2.1.0.*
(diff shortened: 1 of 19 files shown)


docs/releaseNotes.yml
```diff
@@ -35,13 +35,29 @@ items:
...
+          <code>add_request_headers</code>, <code>add_response_headers</code>,
+          <code>remove_request_headers</code>, or <code>remove_response_headers</code>, it is often
+          desirable to be able to turn off that setting locally for a specific <code>Mapping</code>.
+          For several releases this has not been possible for <code>Mappings</code> that are native
+          Kubernetes resources (as opposed to annotations), as an emtpy value ("mask the global
+          default") was erroneously considered to be equivalent to unset ("inherit the global
+          default").  This is now fixed.
+        docs: topics/using/defaults/
+
+      # next changelog item here
```
AliceProxy, line 59: Why is this left in here?

LukeShu: To reduce conflicts for other outsdanding PRs.

AliceProxy (dismissed):

CHANGELOG.md
```diff
@@ -85,6 +85,13 @@ Please see the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest
 
 ### Emissary-ingress and Ambassador Edge Stack
 
+- Bugfix: If the `ambassador` `Module` sets a global default for `add_request_headers`,
+  `add_response_headers`, `remove_request_headers`, or `remove_response_headers`, it is often
+  desirable to be able to turn off that setting locally for a specific `Mapping`. For several
+  releases this has not been possible for `Mappings` that are native Kubernetes resources (as
+  opposed to annotations), as an emtpy value ("mask the global default") was erroneously considered
```
kflynn, line 95: `s/emtpy/empty/`

docs/releaseNotes.yml
```diff
@@ -35,13 +35,29 @@ items:
...
 
+      - title: Fix overriding global settings for adding or removing headers
+        type: bugfix
+        body: >-
+          If the <code>ambassador</code> <code>Module</code> sets a global default for
+          <code>add_request_headers</code>, <code>add_response_headers</code>,
+          <code>remove_request_headers</code>, or <code>remove_response_headers</code>, it is often
+          desirable to be able to turn off that setting locally for a specific <code>Mapping</code>.
+          For several releases this has not been possible for <code>Mappings</code> that are native
+          Kubernetes resources (as opposed to annotations), as an emtpy value ("mask the global
```
kflynn, line 54: `s/emtpy/empty/`

kflynn (approved):
Approving, thanks! There's a typo in the release notes but I'm OK folding that into a separate PR if you want to just land this.

PR #4037, opened 2022-01-23
Reenable more tests; fix a bug

kflynn:
- Reenable the `HostCRDWildcards` test
- Add a unit test for `hostglob_matches`, which is the internal function that decides whether DNS glob match
- Fix a bug in `hostglob_matches` revealed by said tests
- Also, reenable the `demo` test, which verifies that we can use `docker run` to run the demo

These were XFailed during the 2.0-EA cycle, but should _never_ have been allowed to stay that way.

 - [x] I didn't need to update `DEVELOPING.md`

### DIFF
--- a/python/ambassador/ir/irutils.py
+++ b/python/ambassador/ir/irutils.py
@@ -16,33 +16,148 @@ from typing import Any, Dict
 
 import logging
 
+######
+# Utilities for hostglob_matches
+#
+# hostglob_matches_start has g1 starting with '*' and g2 not ending with '*';
+# it's OK for g2 to start with a wilcard too.
+
+def hostglob_matches_start(g1: str, g2: str, g2start: bool) -> bool:
+    # Leading "*" cannot match an empty string, so unless we have a wildcard
+    # for g2, we have to have g1 longer than g2.
+
+    g1match = g1[1:]
+    g2match = g2[1:] if g2start else g2
+
+    if len(g1) > len(g2match):
+        if not g2start:
+            # logging.debug("  match start: %s is too short => False", g1)
+            return False
+
+        # Wildcards for both, so make sure we do the substring match against
+        # the longer one.
+        g1match = g2[1:]
+        g2match = g1[1:]
+
+    match = g2match.endswith(g1match)
+    # logging.debug("  match start: %s ~ %s => %s", g1match, g2match, match)
+
+    return match
+
+
+# hostglob_matches_end has g1 ending with '*' and g2 not starting with '*';
+# it's OK for g2 to end with a wilcard too.
+
+def hostglob_matches_end(g1: str, g2: str, g2end: bool) -> bool:
+    # Leading "*" cannot match an empty string, so unless we have a wildcard
+    # for g2, we have to have g1 longer than g2.
+    g1match = g1[:-1]
+    g2match = g2[:-1] if g2end else g2
+
+    if len(g1) > len(g2match):
+        if not g2end:
+            # logging.debug("  match end: %s is too short => False", g1)
+            return False
+
+        # Wildcards for both, so make sure we do the substring match against
+        # the longer one.
+        g1match = g2[:-1]
+        g2match = g1[:-1]
+
+    match = g2match.startswith(g1match)
+    # logging.debug("  match end: %s ~ %s => %s", g1match, g2match, match)
+
+    return match
+
+
 ################
-## hostglob_matches is a utility for host globbing.
 
-def hostglob_matches(glob: str, value: str) -> bool:
+def hostglob_matches(g1: str, g2: str) -> bool:
     """
-    Does a host glob match a given value?
+    hostglob_matches determines whether or not two given DNS globs are
+    compatible with each other, i.e. whether or not there can be a hostname
+    that matches both globs.
+
+    Note that it does not actually find such a hostname: a return of True
+    just means that such a hostname could exist.
     """
 
-    rc = False
+    # logging.debug("hostglob_matches: %s ~ %s", g1, g2)
 
-    if ('*' in value) and not ('*' in glob):
-        # Swap.
-        tmp = value
-        value = glob
-        glob = tmp
+    # Short-circuit: if g1 & g2 are equal, we're done here.
+    if g1 == g2:
+        # logging.debug("  equal => True")
+        return True
 
-    if glob == "*": # special wildcard
-        rc=True
-    elif glob.endswith("*"): # prefix match
-        rc=value.startswith(glob[:-1])
-    elif glob.startswith("*"): # suffix match
-        rc=value.endswith(glob[1:])
-    else: # exact match
-        rc=value == glob
+    # Next special case: if either glob is "*", then it matches everything.
+    if (g1 == "*") or (g2 == "*"):
+        # logging.debug("  \"*\" present => True")
+        return True
 
-    # sys.stderr.write(f"hostglob_matches: {value} gl~ {glob} == {rc}\n")
-    return rc
+    # Final special case: if either starts with a bare ".", that's not OK.
+    # (Ending with a bare "." is different because DNS.)
+    if g1[0] == "." or g2[0] == ".":
+        # logging.debug("  exact match starts with bare \".\" => False")
+        return False
+
+    # OK, we don't have the simple-"*" case, so any wildcards must be at
+    # the start or end, and they must be a component alone.
+    g1start = (g1[0]  == "*")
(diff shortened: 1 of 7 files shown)


python/ambassador/ir/irutils.py
```diff
@@ -17,32 +17,48 @@
...
+        # OK, no valid wildcards, do a simple equality check.
+        return g1 == g2
+
+    # If we're here, we have some wildcards. So. Split both globs into
+    # path components.
+    g1_components = g1.split(".")
+    g2_components = g2.split(".")
+
+    # If the number of components don't match, the globs don't match.
+    if len(g1_components) != len(g2_components):
```
AliceProxy, line 41: This breaks a working example such as `hostname: *.example.net` on a `Mapping` with `foo.bar.example.net` as your host where `foo.bar` would be captured by the first `*`

python/tests/unit/test_hostglob_matches.py
```diff
@@ -0,0 +1,38 @@
...
+        ( "*.example.com", "b.example.com",   True  ),
+        ( "*.example.com", "a.example.org",   False ),
+        ( "a.example.*",   "a.example.com",   True  ),
+        ( "a.example.*",   "a.example.org",   True  ),
+        ( "a.example.*",   "b.example.com",   False ),
+        ( "a.*.com",       "a.example.com",   False ),  # wildcards have to be at the beginning or end
+        ( "*.com",         "a.example.com",   False ),  # wildcards cannot span across "."
+        ( "*.com",         "a.example.org",   False ),
+        ( "*.example.com", "*.example.com",   True  ),
+        ( "*.example.com", "a.example.*",     True  ),  # Yeah, this is _ugly_.
```
AliceProxy, line 32: Maybe add another test here for the case I mentioned above?

AliceProxy (changes requested):

python/ambassador/ir/irutils.py
```diff
@@ -17,32 +17,48 @@
...
-        rc=value.startswith(glob[:-1])
-    elif glob.startswith("*"): # suffix match
-        rc=value.endswith(glob[1:])
-    else: # exact match
-        rc=value == glob
-
-    # sys.stderr.write(f"hostglob_matches: {value} gl~ {glob} == {rc}\n")
-    return rc
+## hostglob_matches determines two given DNS globs fit each other,
+## in either direction.
```
LukeShu, line 21: Why isn't this a doc string?

python/ambassador/ir/irutils.py
```diff
@@ -17,32 +17,48 @@
...
+    # Special case: if either glob is "*", then it matches everything.
+    if (g1 == "*") or (g2 == "*"):
+        return True
+
+    # OK, we don't have the simple-"*" case, so any wildcards must be at
+    # the start or end, and they must be a component alone.
+    if not (g1.startswith("*.") or g1.endswith("*") or
+            g2.startswith("*.") or g2.endswith("*")):
+        # OK, no valid wildcards, do a simple equality check.
+        return g1 == g2
```
LukeShu, line 33: This doesn't allow partial-label prefix wildcards like `*-bar.foo.com`, which Envoy allows.

See:
- https://www.envoyproxy.io/docs/envoy/v1.17.3/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-virtualhost-domains
- https://github.com/datawire/envoy/blob/ambassador-1.14.1/api/envoy/config/route/v3/route_components.proto#L65-L86
- https://github.com/datawire/envoy/blob/ambassador-1.14.1/source/common/router/config_impl.cc#L1280-L1305

LukeShu:
Note that I'd written an Envoy-compatible `hostglob_matches` python function in https://github.com/datawire/apro/pull/1923

python/ambassador/ir/irutils.py
```diff
@@ -16,33 +16,139 @@
...
+    # Next special case: if either glob is "*", then it matches everything.
+    if (g1 == "*") or (g2 == "*"):
+        # logging.debug("  \"*\" present => True")
+        return True
+
+    # Final special case: if either starts with a bare ".", that's not OK.
+    # (Ending with a bare "." is different because DNS.)
+    if g1[0] == "." or g2[0] == ".":
+        # logging.debug("  exact match starts with bare \".\" => False")
+        return False
```
LukeShu, line 94: There are lots of other invalid-hostname cases. Why is this one in particular worth checking for?

python/ambassador/ir/irutils.py
```diff
@@ -16,33 +16,139 @@
...
+    g1start = (g1[0]  == "*")
+    g1end =   (g1[-1] == "*")
+    g2start = (g2[0]  == "*")
+    g2end =   (g2[-1] == "*")
+
+    # logging.debug("  g1start=%s g1end=%s g2start=%s g2end=%s", g1start, g1end, g2start, g2end)
+
+    if (g1start and g1end) or (g2start and g2end):
+        # Not a valid DNS glob: you can't have a "*" at both ends.
+        return g1 == g2
```
LukeShu, line 107: FWIW, if this happens Envoy would treat the leading `*` as a wildcard, and the trailing `*` as a literal.

python/tests/integration/test_docker.py
```diff
@@ -113,7 +113,7 @@ def test_grab_snapshots():
     assert check_grab_snapshots(), "grab-snapshots check failed"
 
 def test_demo():
-    pytest.xfail("IHA FIXME")
+    # pytest.xfail("IHA FIXME")
```
LukeShu, line 116: Any reason to keep the line around commented out?

python/ambassador/ir/irutils.py
```diff
@@ -16,33 +16,139 @@
...
+    # the other ends with "*", because those can always overlap as long as 
+    # the overlap between isn't empty -- and in this method, we only need to
+    # concern ourselves with being sure that there is a possibility of a match
+    # to both.
+    if (g1start and g2end) or (g2start and g1end):
+        # logging.debug("  start/end pair => True")
+        return True
+
+    # OK, now we have to actually do some work. Again, we really only have to
+    # convinced that it's possible for something to match, so e.g.
```
LukeShu, line 126: There's a verb missing from this sentence.

python/tests/unit/test_hostglob_matches.py
```diff
@@ -0,0 +1,59 @@
...
+        ( "a.example*",        "a.example.com",    True  ),
+        ( "a.exa*",            "a.example.com",    True  ),
+        ( "a.example.*",       "a.example.org",    True  ),
+        ( "a.example.*",       "b.example.com",    False ),
+        ( "a.example*",        "b.example.com",    False ),
+        ( "a.exa*",            "b.example.com",    False ),
+        # '*' has to appear at the beginning or the end, not in the middle.
+        ( "a.*.com",           "a.example.com",    False ),
+        # TLS SAN and RFC4592 disagree here about whether '*' can cross subdomain boundaries.
+        # hostglob_matches() follows RFC4592, since that's what Envoy seems to do when routing.
```
LukeShu, line 44: I don't think we follow RFC 4592. I think this should say "Regardless of what any RFCs say, hostglob_matches() does what Envoy does, and Envoy allows wildcards to span multiple labels."

python/tests/unit/test_hostglob_matches.py
```diff
@@ -0,0 +1,59 @@
...
+        ( "*.com",             "a.example.com",    True  ),
+        ( "*.com",             "a.example.org",    False ),
+        ( "*.example.com",     "*.example.com",    True  ),
+        # This looks wrong but it's OK: both match e.g. foo.example.com.
+        ( "*example.com",      "*.example.com",    True  ),
+        # These are ugly corner cases, but they should still work!
+        ( "*.example.com",     "a.example.*",      True  ),
+        ( "*.example.com",     "a.b.example.*",    True  ),
+        ( "*.example.baz.com", "a.b.example.*",    True  ),
```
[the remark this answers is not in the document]
LukeShu: (I think the code does handle that case correctly, but let's still have a testcase for it.)

python/ambassador/ir/irutils.py
kflynn: Honestly, mostly because I hit it while writing tests. Have strong opinions here?

LukeShu: Not really.

LukeShu (approved):

AliceProxy (approved):

PR #4046, opened 2022-01-25
[2.2] Envoy 1.21 Upgrade

aidanhahn:
## Description
This PR is best reviewed commit by commit.
It contains the following:
- edits to envoy.mk to use the commit at the top of datawire/envoy.git branch aidanhahn/v1.21-2
- removal of --bootstrap-version from entrypoint
- updates to the invocation of the envoy generate_go_protobuf.py tool
- updates to protobuf generation to match updates made to the protos in envoy
- go_package added to kat and agent protos
- removal of the Envoy V2 API and Envoy V2 tests
- tweaks to IR generation and unit tests to enable proper V3 testing and Envoy API use.


LukeShu:
Before I get to commenting on the content, I can tell you that I'm going to want the history to be cleaned up.

LukeShu:
Please remove the `pytest-kat-envoy2*` targets from `builder.mk` (to be fair: they didn't exist in commit you started from).

LukeShu:
Please remove the `GetEnvoyAPIVersion` function.

LukeShu:
What triggered the cluster name change in `t_grpc_stats.py`?

LukeShu:
I almost want to ask you to split this in to a first PR that drops the v2-api, and then a second PR that upgrades Envoy.

LukeShu:
`python/ambassador/config/config.py` and `python/tests/kat/abstract_tests.py` are still doing things with `AMBASSADOR_ENVOY_API_VERSION`.

LukeShu:
The `config.py:envoy_api_version()` function should go away.

python/ambassador/envoy/v3/v3httpfilter.py
```diff
@@ -262,7 +262,7 @@ def V3HTTPFilter_authv1(auth: IRAuth, v3config: 'V3Config'):
         }
 
     if auth.proto == "grpc":
-        protocol_version = auth.get('protocol_version', 'v2')
+        protocol_version = auth.get('protocol_version', 'v3')
```
kflynn, line 265: Do we have checks elsewhere that rule out passing `v2` here?

aidanhahn:
> What triggered the cluster name change in `t_grpc_stats.py`?

According to [the gRPC statistics documentation](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/grpc_stats_filter) it is simply grouped by method instead of name. 
```
kind: Module
name: ambassador
config:
    grpc_stats:
        # this will be ignored
        all_methods: true
        services:
            - name: echo.EchoService
              method_names: [Echo]
```

The behavior seems correct to me so I fixed the test.

LukeShu:
So the `t_grcp_stats.py` change is related to the Envoy upgrade, not to the v2-api→v3-api switch?

Since users do care about watching stats, should we call out the naming change in the changelog?

kflynn (commented):
There're a couple of minor-ish things I'd like to see here (cleaner history... 😉 ) but overall this looks good!

python/ambassador/envoy/v3/v3httpfilter.py
LukeShu: What's up with this change? we still support ext_authz protocol v2.

LukeShu:
Ugg, I'm not sure why `go` is putting new deps in their own `require` block in `go.mod` rather than adding them to the existing `require` block. Can you fix that?

LukeShu:
It looks like you made made `api/{agent,kat}/*.proto` changes in the `make generate` commit? You should keep the human-made changes separate from the machine-made changes.

aidanhahn:
> It looks like you made made `api/{agent,kat}/*.proto` changes in the `make generate` commit? You should keep the human-made changes separate from the machine-made changes.

I made no changes by hand. the differences you see are the result of commit 701a2c5

LukeShu:
The `AuthenticationGRPCV2Test` should probably not go away, because as I said we aren't killing ext_authz v2.

LukeShu:
Likewise for `RateLimitV2Test`.

LukeShu:
> > It looks like you made made `api/{agent,kat}/*.proto` changes in the `make generate` commit? You should keep the human-made changes separate from the machine-made changes.
> 
> I made no changes by hand. the differences you see are the result of commit [701a2c5](https://github.com/emissary-ingress/emissary/commit/701a2c5d127e35c029c2e2b875fce044e13c1141)

Right, but then one of the later "regenerate" commits makes this change:

```patch
diff --git a/api/agent/director.proto b/api/agent/director.proto
index 97f1fc1d9..182535580 100644
--- a/api/agent/director.proto
+++ b/api/agent/director.proto
@@ -10,7 +10,7 @@ import "google/protobuf/timestamp.proto";
 
 package agent;
 
-option go_package = "github.com/datawire/ambassador/v2/pkg/api/agent;agent";
+option go_package = "./agent";
 
 service Director {
   // Report a consistent Snapshot of information to the CEPC.  This
diff --git a/api/kat/echo.proto b/api/kat/echo.proto
index a53a3630c..8b06faf80 100644
--- a/api/kat/echo.proto
+++ b/api/kat/echo.proto
@@ -2,7 +2,7 @@ syntax = "proto3";
 
 package echo;
 
-option go_package = "github.com/datawire/ambassador/v2/pkg/api/kat;kat";
+option go_package = "./kat";
 
 service EchoService {
     rpc Echo(EchoRequest) returns (EchoResponse) {}
```

LukeShu:
It looks like you changed the default `ENVOY_REPO` back from `https://github.com` to `git://github.com` (I suspect a mistake during a rebase). GitHub killed off `git://` support, which is why I had changed it to `https;//`.

LukeShu:
Closing in favor of https://github.com/emissary-ingress/emissary/pull/4080

PR #4050, opened 2022-01-26
Emissary watch for CloudConnectToken ConfigMap or Secret 

alexgervais:
## Description
Emissary will now watch for ConfigMap or Secret resources specified by the `AGENT_CONFIG_RESOURCE_NAME` environment variable in order to allow all components (and not only the Ambassador Agent) to authenticate requests to Ambassador Cloud.

Notes for reviewers: I could not fully execute `make generate`, but my intention was to:
- Add RBAC permissions to get, list and watch `configmaps`.
- Add the `AGENT_CONFIG_RESOURCE_NAME` environment variable to the emissary deployment.
- Add a release note entry.

Let me know if I need to generate/synchronize anything else between Helm charts and YAML and changlog.

## Related Issues
None

## Testing
Manual tests

## Checklist

    + Any new features
    + Any deprecations
 

LukeShu (approved):
It would make sense to me if we wanted to stick the `secretRef(…)` call inside of the `if isEdgeStack { … }` block, and stick `, ignoreIf: !IsEdgeStack()` on the interesting_types entry.

But I have no strong opinion on the matter.

LukeShu:
It looks like you need to
1. Run `make generate-fast`
2. Teach `cmd/entrypoint/testutil_fake_k8s_store_test.go:canonGVK()` about ConfigMaps.

LukeShu (changes requested):
It looks like you need to
1. Run `make generate-fast`
2. Teach `cmd/entrypoint/testutil_fake_k8s_store_test.go:canonGVK()` about ConfigMaps.

alexgervais:
> It would make sense to me if we wanted to stick the `secretRef(…)` call inside of the `if isEdgeStack { … }` block, and stick `, ignoreIf: !IsEdgeStack()` on the interesting_types entry.

I thought about this, and while it's true the current use case for watching this ConfigMap is to support an Edge Stack feature, the Cloud Connect Token is something all components (Emissary, Emissary-agent) might care about in the future.

LukeShu (approved):
Whoops, I thought I approved this once you got `make generate` working.

Hit it!

PR #4052, opened 2022-01-26
[WIP] Allow including binary request bodies in authservice

AliceProxy:

## Description
Allows for including request body when it contains data for requests to the authservice.

## Related Issues
List related issues.

## Testing

## Checklist

    + Any new features
    + Any deprecations
 

AliceProxy:
Squashed all the above commits

python/tests/unit/test_irauth.py
```diff
@@ -139,3 +139,63 @@ def test_irauth_grpcservice_version_default_v3():
...
+    econf = _get_envoy_config(yaml, version='V3')
+
+    conf = econf.as_dict()
+    ext_auth_config = _get_ext_auth_config(conf)
+
+    assert ext_auth_config
+    request_body = ext_auth_config['typed_config']['with_request_body']
+    assert request_body['allow_partial_message'] == True
+    assert request_body['max_request_bytes'] == 4096
+    assert request_body['pack_as_bytes'] == True
```
LukeShu, line 201: nittiest of nits: I don't like the red do-not-enter symbol that GH is showing here. Please add a trailing newline.

Relatedly, I recommend you configure your editor to obey the [`.editorconfig` file ](https://editorconfig.org/) (which says `insert_final_newline = true`).

python/tests/kat/t_extauth.py
```diff
@@ -1052,3 +1053,63 @@ def check(self):
...
+        assert self.results[0].status == 200
+        assert self.results[0].headers["Server"] == ["envoy"]
+
+        # test that POST request still suceed when pack_as_bytes is set for the AuthService
+        assert self.results[1].backend.request.headers["requested-status"] == ["200"]
+        assert self.results[1].status == 200
+        assert self.results[1].headers["Server"] == ["envoy"]
+
+        # test that the body of the request can still be send to the AuthService successfully
+        # uploading a file as form-data would be a beter test, but it doesnt seem that harness.py supports it
```
LukeShu, line 1112: 1. I don't think that editing kat-backend to be able to handle it would be too difficult
2. But if you don't feel up to it, can you please mark this as `TODO(AliceProxy):` so that we can identify it as somehting needing improvement?

python/ambassador/envoy/v2/v2httpfilter.py
```diff
@@ -262,6 +262,9 @@ def V2HTTPFilter_authv1(auth: IRAuth, v2config: 'V2Config'):
         if 'allow_partial' in raw_body_info:
             body_info['allow_partial_message'] = raw_body_info['allow_partial']
 
+        if raw_body_info.get('body_raw_bytes', False):
+            body_info['pack_as_bytes'] = raw_body_info['body_raw_bytes']
+
```
LukeShu, line 267: ... I'm shocked Envoy supports this in v2.

pkg/api/getambassador.io/v3alpha1/crd_authservice.go
```diff
@@ -30,6 +30,8 @@ type AuthServiceIncludeBody struct {
 
 	// +kubebuilder:validation:Required
 	AllowPartial bool `json:"allow_partial,omitempty"`
+
+	BodyAsRawBytes *bool `json:"body_raw_bytes,omitempty"`
```
LukeShu, line 34: 1. Because the default value (false) and the zero value are the same, no need to make it a pointer.
2. This is a little hypocritical because none of the other fields have comments right now, but can you please add a doc-comment?
3. The name seems funny. Because this is already inside of the `include_body` field, including `body_` in the name is redundant. `body_raw_bytes` sounds to me like it is of type `[]byte`, not `bool`. I'm thinking `as_bytes` is a better name?

So something like

```go
// AsBytes (default=false) controls whether to transmit the body as bytes (true) or as a
// string (false).  Request data that is not valid UTF-8 cannot be transmitted as a string,
// and will result in a 403 UAEX response if AsBytes is not set to true.
AsBytes bool `json:"as_bytes,omitempty"`

```

python/ambassador/envoy/v2/v2httpfilter.py
LukeShu: Actually, it's not in `envoy/config/filter/http/ext_authz/v2/ext_authz.proto`. How the heck is this working?

LukeShu: Where I'm at is that without an explanation of why it works in v2, I'm thinking that means that the tests aren't doing a good-enough job of actually testing it.

mikberg:
I would greatly appreciate seeing this one merged, as I'm trying to use the ext_authz filter with gRPC and need access to the request body. It seems like `include_body` is currently not usable for grpc/protobuf requests.

PR #4053, opened 2022-01-27
[v2.2] Stream Envoy metrics to the cloud

douglascamata:
## Description

This pull request intends to stream metrics from Envoy to Ambassador's cloud. We're interested the following metrics for both stable and canary clusters:

* upstream_rq_total: total amount of requests served upstream
* upstream_rq_5xx: total amount of requests served upstream with a 5xx status code (server error)
* upstream_rq_time: histogram with request duration per cluster

The code currently drops all the other metrics in the agent. In the future, this might be done at Envoy's configuration to completely avoid all the extra network traffic between agent and Emissary Ingress pods. 

The flow of the metrics is:

1. Envoy's metrics server pushes to a K8s Service
2. This K8s Service point to a gRPC server started by the agent
3. This gRPC server streams the metrics to Ambassador's cloud 

## Related Issues

This is based in the work done at #3657.

## Testing
I could test this manually in a local cluster and a server that behaves like Ambassador's cloud.

## Checklist

    + Any new features
    + Any deprecations
 

rp4rk (dismissed):
LGTM

pkg/agent/agent.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	envoyMetrics "github.com/datawire/ambassador/v2/pkg/api/envoy/service/metrics/v3"
+	envoyMetrics "github.com/datawire/ambassador/v2/pkg/api/envoy/service/metrics/v2"
```
kflynn, line 7: Hmmm, I haven't looked at this part of Envoy in a bit – the final `v2` in the package here is not the Envoy API version, IIRC, but the version number for sending the metrics to the metrics service, right?

kflynn (changes requested):
This looks very cool! We need to change the port number to 8006, sadly: sorry about that! Looking forward to getting this in. 🙂

pkg/agent/agent.go
```diff
@@ -599,6 +600,24 @@ func (a *Agent) ProcessSnapshot(ctx context.Context, snapshot *snapshotTypes.Sna
 	return nil
 }
 
+func (a *Agent) MetricsRelayHandler(in *envoyMetrics.StreamMetricsMessage) {
+	ctx := context.Background()
```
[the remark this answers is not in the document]
douglascamata: In a more recent version of this file, the handler has an extra parameter of type `context.Context`. The `metricsServer.StartServer` function receives also a context to start the gRPC server and stores it to be sent by the `metricsServer.handler` function whenever something in received in the metrics stream.

pkg/agent/agent.go
douglascamata: Exactly, @kflynn.

pkg/agent/envoy_metrics_server.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+// StartServer will start the metrics gRPC server, listening on :8123
+// It is a blocking call until grpcServer.Serve returns.
+func (s *metricsServer) StartServer(ctx context.Context) error {
+	grpcServer := grpc.NewServer()
+	envoyMetrics.RegisterMetricsServiceServer(grpcServer, s)
+
+	listener, err := net.Listen("tcp", ":8123")
```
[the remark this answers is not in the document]
douglascamata: Fixed in [e28a361](https://github.com/emissary-ingress/emissary/pull/4053/commits/e28a3616333428765f55693ce27a8e064a809a08). I'll take care of updating the documentation for us. Should it be done after the RC is cut?

pkg/agent/agent.go
douglascamata: I'm timeboxing a few minutes to try to upgrade this to v3, it'll be better to start at a more recent version.

douglascamata: Upgraded to the transport api v3 for metrics in [0a1fccb](https://github.com/emissary-ingress/emissary/pull/4053/commits/0a1fccb65dd241913bfb5252236c07e561988082). We should be more future-proof now. 👍

kflynn (approved):
I think this looks good! Thanks! 🙂

kflynn (changes requested):
This looks very cool! We need to change the port number to 8006, sadly: sorry about that! Looking forward to getting this in. 🙂 

However, that gotest failure looks like it might be real?

kflynn (dismissed):
OK, I'll approve this again with the understanding that we must track down the race in the tests.

AliceProxy (approved):

pkg/agent/envoy_metrics_server.go
```diff
@@ -0,0 +1,57 @@
...
+	envoyMetrics.RegisterMetricsServiceServer(grpcServer, s)
+
+	listener, err := net.Listen("tcp", ":8006")
+	if err != nil {
+		dlog.Errorf(ctx, "metrics service failed to listen: %v", err)
+	}
+
+	dlog.Infof(ctx, "metrics service listening on %s", listener.Addr().String())
+	s.logCtx = ctx
+	return grpcServer.Serve(listener)
```
LukeShu, line 40: This should use github.com/datawire/dlib/dhttp, rather than the google.golang.org/grpc HTTP server. (For an example of how to do this, see the `cmd/example-envoy-metrics-sink/` which this PR also edits.)

python/ambassador/envoy/v3/v3bootstrap.py
```diff
@@ -122,28 +173,39 @@ def __init__(self, config: 'V3Config') -> None:
...
+    parts = value.split(":")
+    if len(parts) == 1:
+        host = parts[0]
+        port = 80
+    elif len(parts) == 2:
+        host = parts[0]
+        port = int(parts[1])
+    else:
+        raise ValueError("too many colons")
+    return host, port
```
LukeShu, line 211: This will definitely raise an exception for IPv6.

cmd/agent/main.go
```diff
@@ -50,6 +50,13 @@ func run(cmd *cobra.Command, args []string) error {
 		snapshotURL = fmt.Sprintf(DefaultSnapshotURLFmt, entrypoint.ExternalSnapshotPort)
 	}
 
+	metricsServer := agent.NewMetricsServer(ambAgent.MetricsRelayHandler)
+	go func() {
+		if err := metricsServer.StartServer(ctx); err != nil {
+			dlog.Error(ctx, err)
+		}
+	}()
```
LukeShu, line 58: Don't ever launch a goroutine that you don't have a way to wait for it to shut down.  github.com/datawire/dlib/dgroup can help with this, but you are free to use other solutions as well.

python/ambassador/envoy/v3/v3bootstrap.py
LukeShu: Also there should be type annotations on the signature.

LukeShu: The simplest way to do this is probably
```
from urllib.parse import urlparse
from typing import Tuple

def split_host_port(value: str) -> Tuple[str, int]:
    parsed = urlparse("//"+value)
    return parsed.hostname, int(parsed.port or 80)
```

LukeShu (commented):
OK, 3 concerns:
- the split_host_port routine
- not keeping track of goroutines
- using the google.golang.org/grpc HTTP server

I'm not super-opposed to merging this for -rc.0 and then fixing those later. But if we weren't pushing for an RC ASAP, this'd be a "request changes".

kflynn (approved):
Reapproving after fixing merge conflicts and pinning `pytest` to version 6.2.5 -- let's land this thing!!

PR #4057, opened 2022-01-27
.github: Automate creation of repatriation PRs

LukeShu:
## Description

This PR
 1. Makes sure that all of the CI jobs run on all of the correct branches and PRs
 2. Automates creation of the repatriation PRs
 
What _should_ happen is that when this PR gets merged in to `release/v2.1`, CI will then create a PR that merges from `release/v2.1` in to `master` (by way of a `ci/repatriation/from-v2.1-to-v2.2` branch so that a human has something to push merge-conflict resolutions to, if there are any).

This is based on pain from the 2.1.2 release.

## Testing
I have not tested this.

## Checklist

    + Any new features
    + Any deprecations
 
 - [ ] My change is adequately tested. - no
 

.github/workflows/repatriate.yml
```diff
@@ -0,0 +1,66 @@
...
+          this_branch=release/${this_ver}
+          echo "::set-output this_branch=${this_branch}"
+
+          next_ver=$(git for-each-ref --format='%(refname:lstrip=4)' 'refs/remotes/origin/release/v*' | sort --version-sort | grep -A1 -Fx "$this_ver" | sed 1d)
+          if test -n "$next_ver"; then
+            next_branch="release/${next_ver}"
+          else
+            next_ver="${this_ver%.*}.$((${this_ver##*.}+1))"
+            next_branch='master'
+          fi
```
kflynn, line 38: So I had to legit copy and paste crap into the shell to figure out what this does...

...that said, I'm not sure how to make it better. Hmm.

kflynn (approved):
One comment but ohhhh yeah, let's land this thing. 🙂

PR #4059, opened 2022-01-28
Fix ga-check

kflynn:
- Split things up to be more clear about what's breaking exactly.
- Don't get confused about what the chart version is: require it as an argument.
- Clear the remaining TODO items (mostly, GitHub releases).

Note that I dropped the separate GitHub tag check -- I check that the release has the
tag we expect, and I'm trusting GitHub not to allow a release to have a nonexistent tag.

### DIFF
--- a/releng/lib/check_artifacts.py
+++ b/releng/lib/check_artifacts.py
@@ -109,38 +109,6 @@ def main(ga_ver: str, chart_ver: str, include_docker: bool = True,
                     if b != a:
                         subcheck.ok = False
 
-    # def do_check_binary(checker: Checker, name: str, txt: bool, private: bool) -> None:
-    #     with checker.check(name=f'Executable: {name}', clear_on_success=False) as checker:
-    #         for platform in ['linux/amd64/{}', 'darwin/amd64/{}', 'windows/amd64/{}.exe']:
-    #             rc_body: Optional[bytes] = None
-    #             with do_check_s3(checker, f'{name}/{rc_ver}/{platform.format(name)}',
-    #                              private=private, bucket=s3_bucket) as (subcheck, body):
-    #                 if body is not None:
-    #                     rc_body = body
-    #                     # TODO: Validate the binary somehow
-    #             if ga:
-    #                 with do_check_s3(checker, f'{name}/{ga_ver}/{platform.format(name)}',
-    #                                  private=private, bucket=s3_bucket) as (subcheck, body):
-    #                     if body is not None:
-    #                         assert body == rc_body
-    #         if txt:
-    #             if include_latest:
-    #                 with do_check_s3(checker, f'{name}/latest.txt', private=private, bucket=s3_bucket) as (subcheck, body):
-    #                     if body is not None:
-    #                         subcheck.result = body.decode('UTF-8').strip()
-    #                         if is_private:
-    #                             assert subcheck.result != rc_ver
-    #                         else:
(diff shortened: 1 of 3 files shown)


releng/lib/check_artifacts.py
```diff
@@ -164,9 +164,13 @@ def do_check_docker(checker: Checker, name: str) -> None:
...
-    with checker.check(name='Git tags') as check:
-        check.result = 'TODO'
-        raise NotImplementedError()
+    # This is redundant since we now look at the tag for which the GitHub release was
+    # created -- we're trusting GitHub not to allow a release that points to a tag that
+    # doesn't exist.
+    # with checker.check(name='Git tags') as check:
+    #     check.result = 'TODO'
+    #     raise NotImplementedError()
+
```
LukeShu, line 173: Can we just delete this and add a comment on the GH release check saying that "the existence of a release implies the existence of the tag (though you can actually then delete the tag without deleting the release)"?

releng/lib/check_artifacts.py
```diff
@@ -179,20 +183,26 @@ def do_check_docker(checker: Checker, name: str) -> None:
...
-        if line.startswith("version:"):
-            chart_version = line.replace('version:', '').strip()
+
     with checker.check(name="Check Helm Chart"):
+        # chart_version = ""
+        # for line in fileinput.FileInput("charts/emissary-ingress/Chart.yaml"):
+        #     if line.startswith("version:"):
+        #         chart_version = line.replace('version:', '').strip()
+        chart_version = chart_ver
+
```
LukeShu, line 203: Hmm. I think the right thing here is to just drop setting `chart_version` at all and to replace uses of it with `chart_ver`.

releng/lib/check_artifacts.py
```diff
@@ -202,12 +212,24 @@ def do_check_docker(checker: Checker, name: str) -> None:
         if release_channel != '':
             check_tag = f"{check_tag}-{release_channel}"
         assert_eq(check.result, check_tag)
+
     with checker.check(name='ambassador.git GitHub release for chart') as check:
-        check.result = 'TODO'
-        raise NotImplementedError()
+        tag = run_txtcapture([
+            "gh", "release", "view",
```
LukeShu, line 218: Like I said above, let's add a comment here about how this doubles as checking the tag.

releng/lib/check_artifacts.py
kflynn: Done.

releng/lib/check_artifacts.py
kflynn: Done.

releng/lib/check_artifacts.py
kflynn: Done.

releng/release-ga-check
```diff
@@ -12,19 +12,24 @@ if __name__ == '__main__':
     parser.add_argument('--no-docker', dest='docker', default=True, action='store_false')
     parser.add_argument('--promote-path', default="")
     parser.add_argument('--ga-version', required=True)
+    parser.add_argument('--chart-version', required=True)
```
LukeShu, line 15: You also need to edit `builder/builder.mk` `release/ga-check` to add this flag when calling it.

kflynn: D'oh! Nice catch. Done.

LukeShu (approved):
Woo!

Issue #3708, opened 2021-08-18
Resolve dependabot security vulnerabilities


efunk:
Dependabot identified some security vulnerabilities that we need to resolve. None of the critical/high ones identified are applicable to our code base, but, we should resolve these to remove any false positives and keep our code clean.

The recently identified issues are here: https://github.com/emissary-ingress/emissary/security/dependabot

khussey:
This has been addressed in Emissary-ingress 2.2, which is now available.

PR #4073, opened 2022-02-01
[2.2] Upgrade dependencies, part 3: Upgrade Alpine

LukeShu:
## Description

Originally this was going to be a 1-commit PR facilitated by #4074 and #4075.  But no, it turns out the build-system improvements in #4074 were inadequate, because I wasn't getting the caching behavior I needed (also @aidanhahn wanted the CI configuration in #4074 to be different, so I promised him that I'd fix it up in this PR). But I didn't expect it to be 11 commits of further build-system improvements!

This PR upgrades us from Alpine 3.12 to 3.15. Interesting things about this upgrade:
 - New Python (3.8→3.9)
 - New PIP
 - New Python/PIP means the hack for installing PyPI's glibc-based `orjson` no longer works, so we compile our own musl-based `orjson`.
 - We can drop the code backporting newer Alpine's busybox to our old busybox.

That `orjson` piece is a big pain, because on my box compiling orjson takes a full half hour, so I really want that build to be cached when possible.  And when it's a cache-miss, I really want decent diagnostics as to why (so I put a decent amount of effort in to the output that `base-python.docker.gen` (which is renamed from `builder/builder.sh`) spits out).

## Related Issues
- #3708
- https://github.com/emissary-ingress/emissary/pull/4074
- https://github.com/emissary-ingress/emissary/pull/4075

## Testing

## Checklist

    + Any new features
    + Any deprecations
 
 - [ ] This is unlikely to impact how Ambassador performs at scale. - not in a big way, but a little... glibc is known in many cases to be faster than musl, so having a musl-based orjson might be a little slower
 
 - [x] My change is adequately tested. - I would say so
 

aidanhahn (approved):
LGTM!

PR #4074, opened 2022-02-01
Upgrade dependencies, part 1: Build-system improvements

LukeShu:
## Description

In the interest of having small focused PRs, here are the build system improvements that I have sifted out of my "upgrade the deps" work. I got real excited when I realized I could make a "Kill off the builder shell" commit!

The final commit is making it so that the base images get cached at [docker.io/emissaryingress/builder-base](https://hub.docker.com/repository/docker/emissaryingress/builder-base); this means giving release credentials to the `generate` job, which I'm in love with, but I needed  the base images to be cached (for orjson work in #4073), and this was the "good enough" solution.

The GitHub diffstat looks like this is *adding* a bunch, but excluding `go.sum` files, it's only `62 files changed, 749 insertions(+), 1344 deletions(-)`; so it really is *deleting* quite a bit.

## Related Issues
- #3708
- #4075
- #4073

## Testing
CI should cover it.

## Checklist

    + Any new features
    + Any deprecations
 

aidanhahn (changes requested):

.github/workflows/execute-tests-and-promote.yml
```diff
@@ -23,12 +23,20 @@ jobs:
 
   generate: ####################################################################
     runs-on: ubuntu-latest
+    env:
+      # Set DEV_REGISTRY to match BASE_REGISTRY.
+      DEV_REGISTRY: "docker.io/emissaryingress"
```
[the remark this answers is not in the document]
LukeShu: What if I promise to rewrite it in https://github.com/emissary-ingress/emissary/pull/4073 ?

aidanhahn (approved):

Makefile
```diff
@@ -33,6 +33,14 @@ ifneq ($(MAKECMDGOALS),$(OSS_HOME)/build-aux/go-version.txt)
   $(info [make] CHART_VERSION=$(CHART_VERSION))
 endif
 
+ifeq ($(SOURCE_DATE_EPOCH)$(shell git status --porcelain),)
```
kflynn, line 36: This is worth a comment. I read this as a typo at first.

build-aux/check.mk
```diff
@@ -10,9 +10,44 @@ push-pytest-images: docker/kat-client.docker.push.remote
...
+	  --prefix=work \
+	  --chown-uid=0 --chown-uname=root \
+	  --chown-gid=0 --chown-uname=root \
+	  docker/kat-client; } | $(tools/write-ifchanged) $@
+docker/.kat-client.img.tar.stamp: $(tools/ocibuild) docker/base.img.tar docker/kat-client.go.layer.tar docker/kat-client.fs.layer.tar
+	{ $(tools/ocibuild) image build \
+	  --base=docker/base.img.tar \
+	  --config.Cmd='sleep' --config.Cmd='3600' \
+	  --tag=emissary.local/kat-client:latest \
+	  <($(tools/ocibuild) layer squash $(filter %.layer.tar,$^)); } > $@
```
kflynn, line 32: I have to say, this is a really rough idiom to read. Hmmm. Maybe `ocibuild image build` should be allowed to take a bunch of layer files on the command line and squash them.

build-aux/main.mk
```diff
@@ -0,0 +1,22 @@
...
+	@$(tools/copy-ifchanged) $< $@
+docker/%: docker/.%.stamp $(tools/copy-ifchanged)
+	$(tools/copy-ifchanged) $< $@
+
+_ocibuild-images  = base
+_ocibuild-images += kat-client
+_ocibuild-images += kat-server
+$(foreach img,$(_ocibuild-images),docker/.$(img).docker.stamp): docker/.%.docker.stamp: docker/%.img.tar
+	docker load < $<
+	docker inspect $$(bsdtar xfO $< manifest.json|jq -r '.[0].RepoTags[0]') --format='{{.Id}}' > $@
```
kflynn, line 19: This is, what, making images built by `ocibuild` available to Docker? Needs a comment.

kflynn (commented):
I have some comments, even though this has already landed and I'm OK with that. 🙂

Makefile
LukeShu: OK

build-aux/main.mk
LukeShu: Correct. OK.

build-aux/check.mk
LukeShu: I disagree that this is hard to read. It sorta follow lispy indentation/brace style rather than C-ish indentation/brace style, which is I can respect as rough-if-unfamiliar, but I think the only problem is familiarity.

PS: web-dev folks may also be familiar-ish with this style because if you don't get to use a preprocessor like SASS or LESS, it's the most sane way to write CSS.

PR #4077, opened 2022-02-01
TLS cert validation

kflynn:
This is fairly straightforward: stick validation into `ReconcileSecrets`, and if it fails, move the bad Secret into `Invalid` rather than `Secrets`. The most complex bit was getting the error message to propagate through the `fetcher` on the Python side, and that was a trivial change, it just took time to find where it was getting dropped.

I did manual testing by creating a good Secret and an invalid Secret, with a `Host` pointing to each, and verifying that:

- The error for the invalid Secret got logged
- The invalid Secret appeared, with its error message, in the `Invalid` section of the snapshot
- The `Host` referencing the invalid Secret was disabled
- The `Host` referencing the good Secret stayed active

I've also added secret checking to the `TestFakeHello` Go test, and fixed `TestFakeIstioCert` which was, uh, using an invalid cert for its testing because it didn't matter until now. 😂

 - [x] This is unlikely to impact how Ambassador performs at scale (it does involve conversion from Secret to Unstructured, but we do that for every other kind of invalid object too...).


python/ambassador/fetch/secret.py
```diff
@@ -43,6 +44,8 @@ def _admit(self, obj: KubernetesObject) -> bool:
         return super()._admit(obj)
 
     def _process(self, obj: KubernetesObject) -> None:
+        # self.logger.debug("processing K8s Secret %s", dump_json(dict(obj), pretty=True))
```
AliceProxy, line 47: Why add this log but leave it commented out? Also, why change the log line in `python/ambassador/fetch/fetcher.py` but still leave it commented out?

aidanhahn (dismissed):

python/ambassador/fetch/secret.py
kflynn: I needed those logs while debugging getting the error text for an invalid Secret to be passed all the way through to the user, which argues for leaving them in – but I don't want to have `dump_json()` called even while debugging is disabled, hence, commenting them out.

aidanhahn (approved):

AliceProxy (approved):

AliceProxy (approved):

PR #4078, opened 2022-02-01
[2.2] Agent & CRD changes for migrations

kflynn:
I plan to use these for cleaning up the migration from 1.X to 2.X:

- apply CRDs
- apply emissary-migration YAML
- test
- scale down 1.14 agent
- apply emissary-agent YAML

I have tested this, so far, only by making sure that emissary-migration + emissary-agent gives you
emissary.yaml. More soon, like actually using it.


LukeShu (changes requested):
First commit (agent manifests) looks good.

The second commit: I don't like doing that in fix-crds, when we could have kubebuilder do it with a `pkg/api/getambassador.io/v1/`

LukeShu (approved):

PR #4079, opened 2022-02-02
[2.1] Agent & CRD changes for migrations

kflynn:
(This is the 2.1 version of #4078)

I plan to use these for cleaning up the migration from 1.X to 2.X:

- apply CRDs
- apply emissary-migration YAML
- test
- scale down 1.14 agent
- apply emissary-agent YAML

I have tested this, so far, only by making sure that emissary-migration + emissary-agent gives you
emissary.yaml. More soon, like actually using it.


LukeShu (approved):
This is just a backport, right?

PR #4081, opened 2022-02-03
Fix test "flakes" from `cmd/entrypoint`

LukeShu:
## Description

Since the removal of `gotestsum` (in #4074), we've been seeing pretty frequent failures from the `cmd/entrypoint` tests. There are 3 causes here:
 - test issue: The consul tests being too sensitive about how snapshots are coalesced (fixed)
 - actual issue: Race in shutdown of the consul watch goroutines (fixed)
 - actual issue: Race in the shutdown of the FS watcher goroutines (fixed it in the tests, but not the actual code)

These three are all intermittently-detected issues, so `gotestsum`'s retries were hiding them.  And now that they're unhidden, they are causing problems for other PRs.

This is a good example of why "retries without metrics" are bad.  The race detector can often only intermittently detect races.  If a test "flakes" because of `-race`, the "flake" is that it sometimes passes when it should fail, not that it sometimes fails when it should pass!

Let me be clear here: I 100% believe that there are still races in `cmd/entrypoint`.  However, I am 95% certain that the failure rate of of the `cmd/entrypoint` tests is now below 3%. So this PR slightly improves the state of the code, and significantly reduces development friction.

I wish there were a linter that allowed me to outlaw the `go` keyword, and force everyone to use `dgroup`; this would avoid races with leaking goroutines and goroutine shutdown.

## Related Issues
Look at any of the several PRs that are red because of `cmd/entrypoint` even though they don't touch `cmd/entrypoint`.

## Testing
I ran the tests a whole bunch of times (n=100), with the race detector turned on.

## Checklist

 - [ ] I made sure to update `CHANGELOG.md`. - since I believe there are still many races in this code, I do not believe this merits a changelog entry
   
    + Any new features
    + Any deprecations
 
 - [ ] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - I'm pretty sure GOTEST_ARGS and GOTEST_PKGS are already in there

### DIFF
--- a/cmd/entrypoint/testutil_fake_hello_test.go
+++ b/cmd/entrypoint/testutil_fake_hello_test.go
@@ -142,30 +142,53 @@ func FindCluster(envoyConfig *v3bootstrap.Bootstrap, predicate func(*v3cluster.C
 	return nil
 }
 
-func deltaSummary(t *testing.T, snap *snapshot.Snapshot) []string {
+func deltaSummary(t *testing.T, snaps ...*snapshot.Snapshot) []string {
 	summary := []string{}
 
 	var typestr string
 
-	for _, delta := range snap.Deltas {
-		switch delta.DeltaType {
-		case kates.ObjectAdd:
-			typestr = "add"
-		case kates.ObjectUpdate:
-			typestr = "update"
-		case kates.ObjectDelete:
-			typestr = "delete"
-		default:
-			// Bug because the programmer needs to add another case here.
-			t.Fatalf("missing case for DeltaType enum: %#v", delta)
-		}
+	for _, snap := range snaps {
+		for _, delta := range snap.Deltas {
+			switch delta.DeltaType {
+			case kates.ObjectAdd:
+				typestr = "add"
+			case kates.ObjectUpdate:
+				typestr = "update"
+			case kates.ObjectDelete:
+				typestr = "delete"
+			default:
+				// Bug because the programmer needs to add another case here.
+				t.Fatalf("missing case for DeltaType enum: %#v", delta)
+			}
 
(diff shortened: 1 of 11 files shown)


cmd/entrypoint/testutil_fake_hello_test.go
```diff
@@ -142,32 +142,55 @@ func FindCluster(envoyConfig *v3bootstrap.Bootstrap, predicate func(*v3cluster.C
...
 
 	sort.Strings(summary)
 
 	return summary
 }
 
+// getSnapshots is like f.GetSnapshot, but returns the list of every snapshot evaluated (rather than
+// discarding the snapshots from before predicate returns true).  This is particularly important if
+// you're looking at deltas; you don't want to discard any deltas just because two snapshots didn't
+// get coalesced.
```
kflynn, line 176: The thing that always bugged me about this test is that I feel like snapshot coalescing, in the `fake` world, should be deterministic. We should talk about that.

We should also consider making `f.GetSnapshot` coalesce deltas when skipping over snapshots... except, when you think about it, that's really dependent on ordering...

kflynn (approved):
Once comment -- but yeah, let's land this. ESPECIALLY because it fixes the most annoying naming in `entrypoint`... 🙂

cmd/entrypoint/testutil_fake_hello_test.go
LukeShu: I mostly feel like you're right, but I don't have enough of entrypoint in my head to discuss it in detail.

See also: The `watt` package in apro, and the "watchable" things in Telepresence.

PR #4082, opened 2022-02-03
[v2.2] Upgrade dependencies, part 4: Upgrade all the libraries and misc

LukeShu:
## Description

OK, yeah, this is a big PR.  But it's not nearly as big as GitHub leads you to believe; excluding `vendor/`, the diffstat is a much more modest `39 files changed, 1557 insertions(+), 853 deletions(-)`.

Other than the cool new `py-list-deps.py` script, this PR is actually pretty boring.  "Upgrade foo, run `make generate`; upgrade bar, run `make generate`..."

The only way to get through this is to walk through it commit-by-commit.  Each commit indicates which area of thing it is upgrading; first the `tools:` commits, then the `pydeps:` commits, then the `godeps:` commits.

## Related Issues
- #3708
- https://github.com/emissary-ingress/emissary/pull/4074
- https://github.com/emissary-ingress/emissary/pull/4075
- https://github.com/emissary-ingress/emissary/pull/4073

## Testing
CI should cover it.

## Checklist

 - [ ] I made sure to update `CHANGELOG.md`. - no I did not, but maybe I should?
   
    + Any new features
    + Any deprecations
 
 - [ ] This is unlikely to impact how Ambassador performs at scale. - :shrug:
 
 - [x] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - yes I did


LukeShu:
Resolved merge conflicts.

tools/src/py-list-deps.py
```diff
@@ -0,0 +1,228 @@
...
+import sysconfig
+from typing import List, NamedTuple, Optional, Set
+
+
+def parse_members(filepath: str) -> Set[str]:
+    with open(filepath, 'r') as filehandle:
+        filecontent = filehandle.read()
+
+    with open(filepath, 'r') as filehandle:
+        filecontent = filehandle.read()
```
kflynn, line 61: One of these should probably go away, right?

kflynn: ...actually, wait, why do we have `parse_members` and `parse_imports`?

tools/src/py-list-deps.py
```diff
@@ -0,0 +1,228 @@
...
+    sys.path = [d for d in sys.path if (
+        os.path.normcase(d).startswith(stdlib_prefix) and
+        ('site-packages' not in d) and
+        ('dist-packages' not in d))]
+
+    in_stdlib = False
+    try:
+        module = importlib.import_module(item.module)
+        if item.member:
+            getattr(module, item.member)
```
kflynn, line 115: Is this just here for the exception if the member is wrong?

OPENSOURCE.md
```diff
@@ -144,12 +144,10 @@ libraries:
     certifi             2020.6.20  Mozilla Public License 2.0
     charset-normalizer  2.0.8      MIT license
     click               8.0.3      3-clause BSD license
-    clize               4.2.1      MIT license
```
kflynn: Aww yeah.

(And going with click is what I was going to do.)

kflynn (approved):
I have some minor nits in `py-list-deps`, and I'm pretty horrified by some of the manual process on the Go updates (!!), but overall this looks like a massive pain that'll greatly benefit us. Approving -- I'm OK either getting the `py-list-deps` stuff now or in a followon PR.

tools/src/py-list-deps.py
LukeShu: > ...actually, wait, why do we have `parse_members` and `parse_imports`?

Good question.

Short answer: IDK, I wrote this code like 2 years ago

Long answer: After reading it, `parse_members` returns all of the things in a file; the names of all functions, the names of all variables, the names of re-exported things imported from elsewhere... well, not the names of all variables; like the comment says, it doesn't support all statement types. But it "should", it's just that variables aren't implemented.

PR #4100, opened 2022-02-08
Fix the agent service name, and move the port number

LukeShu:
## Description

When the agent's metrics collection was going in, there was confusion where it was believed to be a sidecar to the main `emissary` deployment, when it is in fact its own `emissary-agent` deployment.

There are two effects of this:
 1. The YAML was written to have Envoy stream metrics to the `emissary` service rather than the `emissary-agent` service.
 2. The port number it listens on was allocated from Emissary's 8000-8499 range.  Since it's it's own thing, it's free to choose any port it likes.

So fix the service name.  And move the port from 8006 to 8080, since it's just HTTP.

Also fix a test flake.

## Testing
Validated that this gets the e2etests to pass in apro.

## Checklist

 - [x] I made sure to update `CHANGELOG.md`. - no applicable changes (the wrong version has not been released)
   
    + Any new features
    + Any deprecations
 

python/README.md
```diff
@@ -33,7 +33,6 @@ Ambassador uses several TCP ports while running. All but one of them are in the
 | 8003 | `entrypoint/ambex`                    | Internal `ambex` snapshot access; not exposed outside pod                                                          |
 | 8004 | `diagd`                               | Internal `diagd` access when `AMBASSADOR_FAST_RECONFIGURE` is set; not exposed outside pod                         |
 | 8005 | `entrypoint/external_snapshot_server` | Exposes configuration snapshots for integration with other tools, such as the Ambassador Agent                     |
-| 8006 | `entrypoint/agent`                    | Receives GRPC calls made by the Envoy Metrics Server                                                               |
 | 8080 | `envoy`                               | Default HTTP service port                                                                                          |
```
aidanhahn, line 36: Does this need to be removed or reworded if we are going to use 8080 for the metrics server?

aidanhahn (changes requested):

python/README.md
LukeShu: This is the listing of the ports in use on the main `emissary` Deployment; the agent doesn't run on that Deployment, it runs on its own `emissary-agent` deployment.

aidanhahn (dismissed):

charts/emissary-ingress/templates/deployment.yaml
```diff
@@ -173,7 +173,7 @@ spec:
               containerPort: {{ .Values.adminService.port }}
           env:
             - name: AMBASSADOR_GRPC_METRICS_SINK
-              value: {{ include "ambassador.fullname" . }}:8006
+              value: {{ include "ambassador.fullname" . }}-agent:http
```
douglascamata, line 176: @LukeShu we cannot use a non-integer here for the port in this environment variable because it will break the Envoy config generation here: https://github.com/emissary-ingress/emissary/blob/128bce4fe68dc0f7912b7e28acaedae50b449f4c/python/ambassador/envoy/v3/v3bootstrap.py#L205

```
>>> urlparse("//foobar:http").port
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python3.9/urllib/parse.py", line 178, in port
    raise ValueError(message) from None
ValueError: Port could not be cast to integer value as 'http'
```

douglascamata (changes requested):

douglascamata (approved):
LGTM 👍

kflynn (approved):
I'd like some commentary around how the maturin fix works, but I'm not gonna hold this for it 'cause it looks good! Hit it.

PR #4104, opened 2022-02-09
Check use of forbidden licenses in Go and Python dependencies.

aosoriodw:
## Description
Update build scripts to check for forbidden licenses used by both Go and Python dependencies

Changes:
1. Updated `go-mkopensource` to the latest version
2. Updated format of `DEPENDENCY_LICENSES.md` to match the format used on the website documentation.

## Related Issues
List related issues.

## Testing
- Verified that make is able to generate both `DEPENDENCIES.md` and `DEPENDENCY_LICENSES.md`
- Verified that when a forbidden license is used, make fails

## Checklist

    + Any new features
    + Any deprecations
 

DEPENDENCY_LICENSES.md
```diff
@@ -1,14 +1,4 @@
...
-* [ISC license](https://opensource.org/licenses/ISC)
-* [MIT license](https://opensource.org/licenses/MIT)
-* [Mozilla Public License 2.0](https://opensource.org/licenses/MPL-2.0)
-
-
-Emissary-ingress Python code incorporates Free and Open Source software under the following licenses:
+Emissary-ingress incorporates Free and Open Source software under the following licenses:
 
 * [2-clause BSD license](https://opensource.org/licenses/BSD-2-Clause)
 * [3-clause BSD license](https://opensource.org/licenses/BSD-3-Clause)
```
LukeShu, line 4: This is clearly wrong, because there's no longer any mention of the MPL

LukeShu:
If the repo is now more than just `go-mkopensource`, then perhaps it should be renamed to `mkopensource.git`

aosoriodw:
> What other project is using Python that it's worth it to pull `py-mkopensource` out of emissary.git? And by "worth it", I don't just mean the effort you've done of doing it, I mean that because of packages with an "UNKNOWN" license (which we now know to be because we erroneously don't check the classifiers) we have to update py-mkopensource essentially every time we upgrade the Python packages; so moving py-mkopensource out of emissary.git will make it more difficult for us to upgrade our dependencies.

Currently only Emissary and Apro have python dependencies (I saw the removal of apro's build-aux/requirements.txt but scanning is still necessary since Emissary is a dependency of apro).

I see your point about updating `py-mkopesource` will be more complicated, since you'll have to update `go-mkopensource` repo and then update the version, however I wanted to put together all mkopensource tools (Go, Python and Js) in one repo.

I could move it back here, but if there is no strong preference I'd like to keep it there.

DEPENDENCY_LICENSES.md
aosoriodw: Do you mean MPL 2.0? I see it there in line 8
https://github.com/emissary-ingress/emissary/blob/aosorio/check-forbidden-licenses/DEPENDENCY_LICENSES.md

build-aux/tools.mk
```diff
@@ -68,7 +69,8 @@ $(tools.bindir)/%: $(tools.srcdir)/%/pin.go $(tools.srcdir)/%/go.mod
 tools.main-gomod += $(tools/protoc-gen-go)   # ensure runtime libraries are consistent
 tools.main-gomod += $(tools/controller-gen)  # ensure runtime libraries are consistent
 tools.main-gomod += $(tools/conversion-gen)  # ensure runtime libraries are consistent
-tools.main-gomod += $(tools/go-mkopensource) # ensure it is consistent with py-mkopensource
+tools.main-gomod += $(tools/go-mkopensource)
+tools.main-gomod += $(tools/py-mkopensource)
```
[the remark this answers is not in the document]
aosoriodw: Done.

aosoriodw:
> If the repo is now more than just `go-mkopensource`, then perhaps it should be renamed to `mkopensource.git`



> If the repo is now more than just `go-mkopensource`, then perhaps it should be renamed to `mkopensource.git`

It sounds like a good idea. I'll reach out to talk about this.

aosoriodw:
@LukeShu please let me know what do you want to do about py-mkopensource or if you have additional comments.

kflynn (dismissed):
@LukeShu and I talked offline about `py-mkopensource`, and we both feel that it makes more sense to keep it in `emissary.git` right now: changes to the way we use Python in Emissary are the most likely reason to need to edit `py-mkopensource`, and we can see Python changes on the horizon that splitting it out will make more difficult to no real purpose. 

So... apologies for asking for extra work, but, yes, we think that that needs to happen.

aosoriodw:
> @LukeShu and I talked offline about `py-mkopensource`, and we both feel that it makes more sense to keep it in `emissary.git` right now: changes to the way we use Python in Emissary are the most likely reason to need to edit `py-mkopensource`, and we can see Python changes on the horizon that splitting it out will make more difficult to no real purpose.
> 
> So... apologies for asking for extra work, but, yes, we think that that needs to happen.

After deprecation of Python modules in apro, I also think it makes more sense to keep py-mkopensource here.

aosoriodw:
@kflynn I've added back py-mkopensource but it's a newer version with the following changes:
1. Check that forbidden licenses are not used.
2. Added more tests to check license verification logic
3. Update output to be consistent with go-mkopensource
4. Documented how to use py-mkopensource
5. Removed duplicated code in markdownOutput() and jsonOutput()

aosoriodw:
@kflynn Please let me know if changes are good.

kflynn (approved):
Thanks much! I'm OK taking this one. 🙂

PR #4119, opened 2022-02-15
Synthetic v3 AuthService for Edge-Stack

AliceProxy:

## Description
When running Edge-Stack this will remove all AuthServices not using `protocol_version: v3` and then inject a synthetic AuthService pointing at `127.0.0.1:8500` if there are no valid v3 AuthServices. The purpose is to ensure that Edge-Stack always has an available v3 AuthService and that it will not try to use any that are v2 (which is the default).

## Related Issues
List related issues.

## Testing

## Checklist

    + Any new features
    + Any deprecations
 
### DIFF
--- a/cmd/entrypoint/testutil_fake_syntheticauth_test.go
+++ b/cmd/entrypoint/testutil_fake_syntheticauth_test.go
@@ -1,0 +1,124 @@
+package entrypoint_test
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/datawire/ambassador/v2/cmd/entrypoint"
+	v3bootstrap "github.com/datawire/ambassador/v2/pkg/api/envoy/config/bootstrap/v3"
+	v3cluster "github.com/datawire/ambassador/v2/pkg/api/envoy/config/cluster/v3"
+	"github.com/datawire/ambassador/v2/pkg/snapshot/v1"
+)
+
+// This predicate is used to check k8s snapshots for an AuthService matching the provided name and namespace
+func HasAuthService(namespace, name string) func(snapshot *snapshot.Snapshot) bool {
+	return func(snapshot *snapshot.Snapshot) bool {
+		for _, m := range snapshot.Kubernetes.AuthServices {
+			if m.Namespace == namespace && m.Name == name {
+				return true
+			}
+		}
+		return false
+	}
+}
+
+// Tests the synthetic auth generation when a valid AuthService is created
+// This authservice has protocol_Version: v3 and should not be replaced by the synthetic AuthService
+func TestSyntheticAuthWithValid(t *testing.T) {
+	t.Setenv("EDGE_STACK", "true")
+
+	f := entrypoint.RunFake(t, entrypoint.FakeConfig{EnvoyConfig: true}, nil)
+
+	err := f.UpsertYAML(`
+---
+apiVersion: getambassador.io/v3alpha1
+kind: AuthService
+metadata:
+  name: edge-stack-auth-test
+  namespace: foo
+spec:
+  auth_service: 127.0.0.1:8500
+  protocol_version: "v3"
+  proto: "grpc"
+`)
+	assert.NoError(t, err)
+	f.Flush()
+
+	// Use the predicate above to check that the snapshot contains the AuthService defined above
+	// The AuthService has protocol_Version: v3 so it should not be removed/replaced by the synthetic AuthService
+	// injected by syntheticauth.go
+	snap, err := f.GetSnapshot(HasAuthService("foo", "edge-stack-auth-test"))
+	assert.NoError(t, err)
+	assert.NotNil(t, snap)
+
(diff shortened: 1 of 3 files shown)


cmd/entrypoint/testutil_fake_syntheticauth_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/datawire/ambassador/v2/cmd/entrypoint"
+	v3bootstrap "github.com/datawire/ambassador/v2/pkg/api/envoy/config/bootstrap/v3"
+	v3cluster "github.com/datawire/ambassador/v2/pkg/api/envoy/config/cluster/v3"
+	"github.com/datawire/ambassador/v2/pkg/snapshot/v1"
+)
+
+func HasAuthService(namespace, name string) func(snapshot *snapshot.Snapshot) bool {
```
kflynn, line 16: The way the Fake framework uses predicates to find snapshots and Envoy configs makes it pretty much _always_ worth talking about what your predicates do, where they'll be used, etc.

cmd/entrypoint/testutil_fake_syntheticauth_test.go
```diff
@@ -0,0 +1,119 @@
...
+  namespace: foo
+spec:
+  auth_service: 127.0.0.1:8500
+  protocol_version: "v3"
+  proto: "grpc"
+`)
+	assert.NoError(t, err)
+	f.Flush()
+
+	snap, err := f.GetSnapshot(HasAuthService("foo", "edge-stack-auth-test"))
```
kflynn, line 49: Comment-worthy here, too.

cmd/entrypoint/testutil_fake_syntheticauth_test.go
```diff
@@ -0,0 +1,119 @@
...
+	assert.NotNil(t, snap)
+
+	assert.Equal(t, "edge-stack-auth-test", snap.Kubernetes.AuthServices[0].Name)
+
+	// Check for an ext_authz cluster name matching the provided AuthService (Http_Filters are harder to check since they always have the same name)
+	isAuthCluster := func(c *v3cluster.Cluster) bool {
+		return strings.Contains(c.Name, "cluster_extauth_127_0_0_1_8500_foo")
+	}
+
+	// Grab the next envoy config that satisfies our predicate.
```
kflynn, line 60: And very much here -- it makes me sad that this "Grab the next envoy config that satisfies our predicate" comment gets copied around so much, because it doesn't really help. 😢 

"Grab the next Envoy config that has an Edge Stack auth cluster on 127.0.0.1:8500" maybe?

cmd/entrypoint/testutil_fake_syntheticauth_test.go
```diff
@@ -0,0 +1,119 @@
...
+	envoyConfig, err := f.GetEnvoyConfig(func(envoy *v3bootstrap.Bootstrap) bool {
+		return FindCluster(envoy, isAuthCluster) != nil
+	})
+	require.NoError(t, err)
+
+	// Make sure the cluster exists
+	cluster := FindCluster(envoyConfig, isAuthCluster)
+	endpoints := cluster.LoadAssignment.Endpoints
+	require.NotEmpty(t, endpoints)
```
kflynn, line 69: Hmmmm. Is there a point to this secondary check in this test? Repeating the `FindCluster` from the predicate seems duplicative; maybe just make sure `envoyConfig != nil`?

kflynn: (These comments apply to `TestSyntheticAuthInvalid` too.)

kflynn (approved):
I have some comments about comments, but I'm approving this because I'd really like to get it landed so we can play with it. Comment improvements can be a separate PR tomorrow. 🙂

cmd/entrypoint/testutil_fake_syntheticauth_test.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	// Grab the next Envoy config that has an Edge Stack auth cluster on 127.0.0.1:8500
+	envoyConfig, err := f.GetEnvoyConfig(func(envoy *v3bootstrap.Bootstrap) bool {
+		return FindCluster(envoy, isAuthCluster) != nil
+	})
+	require.NoError(t, err)
+
+	// Make sure an Envoy Config containing a extauth cluster for the AuthService that was defined
+	assert.NotNil(t, snap)
```
kflynn, line 72: s/snap/envoyConfig/ 😉

kflynn (changes requested):
Two minor nits which I'm sure you already found. 🙂  Comments are helpful, thank you!! 🙂

cmd/entrypoint/testutil_fake_syntheticauth_test.go
AliceProxy: Copy & paste responsibly...  *facepalm*

kflynn (approved):
Ship it!!! 🙂

PR #4122, opened 2022-02-16
Add delay in the agent between each metrics push

knlambert:
## Description
- By default, envoy sends metrics every 3 4 seconds to the GRPC service from the agent.
- This delay is short, and not necessary since Ambassador cloud only stores data per minutes (last 30 minutes).
- Added some tests.

## Testing
- Tested in my cluster with a locally built version.
- Unit tests

## Checklist


pkg/agent/agent_metrics_test.go
```diff
@@ -0,0 +1,98 @@
+package agent
+
+import (
+	"context"
+	"github.com/datawire/ambassador/v2/pkg/api/agent"
+	envoyMetrics "github.com/datawire/ambassador/v2/pkg/api/envoy/service/metrics/v3"
+	io_prometheus_client "github.com/prometheus/client_model/go"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
```
LukeShu, line 9: Please use vanilla `testing.T.Run` instead of `testify/suite`.

LukeShu:
Can you please add an entry to `docs/releaseNotes.yml` (and then run `make $PWD/CHANGELOG.md`) summarizing these changes?

pkg/agent/agent_metrics_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+}
+
+func (s *AgentMetricsSuite) AfterTest(suiteName, testName string) {
+	return
+}
+
+func (s *AgentMetricsSuite) TestMetricsHandlerWithRelay() {
+	//given
+	ctx := context.TODO()
```
LukeShu, line 67: Use `ctx := dlog.NewTestContext(t, true)` to get a Context to use in tests.

LukeShu (dismissed):
Some minor changes requested
 - change request: Don't use `context.TODO()`
 - change request: Don't use `testify/suite`
 - suggestion: Come up with a better name than `metricsRelayDeadline`
 
But on the whole it looks good!

knlambert:
> Can you please add an entry to `docs/releaseNotes.yml` (and then run `make $PWD/CHANGELOG.md`) summarizing these changes?

Sure, under 2.3.0 ? or does it go in a 2.2.2 ?

knlambert:
@LukeShu I have one question aswell about the [metrics sink](https://github.com/emissary-ingress/emissary/blob/master/python/ambassador/envoy/v3/v3bootstrap.py#L124) (that the agent is listening, and relaying).

Does it always push all the metrics ? Is there a risk I miss something by putting this 30s cooldown period ?

I see some differences between payloads, but only on some internal metrics which are not relevant to me.

image

AliceProxy (approved):

knlambert:
Rebased to fix conflict in the changelog file.

docs/releaseNotes.yml
```diff
@@ -68,6 +68,14 @@ items:
             link: https://github.com/emissary-ingress/emissary/issues/4134
         docs: https://github.com/emissary-ingress/emissary/issues/4134
 
+      - title: Decrease metric sync frequency
+        type: change
+        body: >-
+          The new delay between two syncs is 30s.
```
AliceProxy, line 74: Small nit: Can you change this to `The new delay between two metrics syncs is now 30s.` so that the corresponding entry in `CHANGELOG.md` makes sense without the context that the title adds? :)

AliceProxy (changes requested):

AliceProxy:
setting DCO to pass since @knlambert is with Ambassador Labs

AliceProxy (approved):

khussey:
@AliceProxy I guess this needs to be merged over to `release/v2.3` in order to make the 2.3 release?

AliceProxy:
@khussey I've already cherry-picked and merged these commits into the 2.3 release branch :)

PR #4126, opened 2022-02-17
Fix support for private CI

LukeShu:
## Description
This fixes the machinery for running in a private repository that can only push things to other private locations.  The big part of that is having no S3, and having to make sure that all Docker registry accesses are authenticated.

The first few commits are doing cleanup in prep for the last few commits, which actually fix things. As usual for me, I suggest a commit-by-commit review.

## Testing
This has been tested as https://github.com/datawire/emissary-private/pull/39.  Everything on that PR has turned green.

## Checklist

    + Any new features
    + Any deprecations
 
 - [x] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - it already mentions `DEV_USE_IMAGEPULLSECRET`.


aidanhahn (approved):

.github/workflows/execute-tests-and-promote.yml
```diff
@@ -38,6 +38,7 @@ jobs:
       - name: "Docker Login"
         uses: docker/login-action@v1
         with:
+          registry: ${{ (!startsWith(secrets.RELEASE_REGISTRY, 'docker.io/')) && secrets.RELEASE_REGISTRY || null }}
```
kflynn, line 47: This is worthy of a comment:

- you need to not set the registry if it _does_ start with `docker.io/`
- GHA requires stupid `&&` `||` crap because there's no ternary operator, but really this is a ternary.

kflynn: (It would've been incomprehensible if I hadn't happened to remember the Slack wisecrackery from yesterday. 🙂 )

.github/workflows/execute-tests-and-promote.yml
```diff
@@ -308,18 +305,7 @@ jobs:
...
-          fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.sha }}
-      - name: Install Deps
-        uses: ./.github/actions/setup-deps
-      - name: "promote build to passed"
-        run: |
-          make release/promote-oss/dev-to-passed-ci
+      - name: No-Op
+        if: ${{ false }}
+        run: "echo Pass"
```
kflynn, line 311: Also comment-worthy, 'cause ew. 😂

kflynn (commented):
Some places I'd like to see comments, but yeah, makes sense!! 🙂

PR #4137, opened 2022-02-24
Metrics for invalid resources

kflynn:
Bring the counts of invalid resources to Metriton. Best reviewed commit-by-commit:

- Tweak error message for invalid secret key

   This is a really simple text change.

- Hacky implementation: bring invalid objects all the way from watt to the IR

   The `watt` snapshot has invalid objects in it, but those invalid objects were
   actually being eaten in the fetcher (as of 3.0, this will need to change, but
   in 2.X it's fine). This is the small-surface-area hackish way to bring the
   invalid objects through from the snapshot to the IR: it's hackish because it
   won't currently result in the invalid objects making it through for the CLI
   tools or mockery -- but let's face it, those tools largely bit-rotted the
   moment we started properly handling annotations in entrypoint, so, whatever.

   A later PR needs to fix the hackery and make the CLI stuff work again.

- Include counts of invalid objects in the features array we hand to Metriton

   This is the center of gravity of this change.

At present, I tested this manually with `gmake deploy`, setting debug on,
applying a broken secret, and looking at the debug logs to see what got passed
to Metriton, which (correctly) included:

```
"invalid_counts": {
  "Secret": 1
}
```

There are no automated tests yet because we don't have any automated tests of
the `features` dict at the moment, and adding some to the `Fake` stuff is...
a little weird. Coming soon! 🙂 

 - [x] I don't think that this needs a note in the `CHANGELOG.md`.
 - [x] My change is adequately tested, I think, just... manually.


aidanhahn (approved):

AliceProxy (approved):
Yeet it!

python/ambassador_diag/diagd.py
```diff
@@ -1520,6 +1520,13 @@ def _load_ir(self, rqueue: queue.Queue, aconf: Config, fetcher: ResourceFetcher,
         with self.app.aconf_timer:
             aconf.load_all(fetcher.sorted())
 
+            # TODO(Flynn): This is an awful hack. Have aconf.load(fetcher) that does
+            # this correctly.
+            #
+            # I'm not doing this at this moment because aconf.load_all() is called in a
+            # lot of places, and I don't want to destablize 2.2.2.
+            aconf.load_invalid(fetcher)
```
AliceProxy, line 1528: Actually the resource fetcher in this call does not have `invalid` initialized at the time of this call. It seems to only be initialized by `fetcher.py - parse_watt()` and not by `__init__` (and parse_watt() was evidently not called by this point.

AliceProxy (changes requested):

python/ambassador_diag/diagd.py
AliceProxy: Which leads to the `'ResourceFetcher' object has no attribute 'invalid'` test error when `load_invalid` does `self.invalid = fetcher.invalid`

kflynn: Hmmm, that's in `_load_ir` -- `parse_watt` will be called above, in `load_config_watt`, before `_load_ir` is called. No other paths ~are~should be used any more.

OTOH, it's hardly a bad thing to initialize it in `__init__`. Done.

kflynn: (Of course I wrote that _before_ looking at the test results. 🙄 )

kflynn:
Grrr. We should modify the build job so that dev charts get overwritten.

aidanhahn (approved):

AliceProxy (approved):

PR #4140, opened 2022-02-24
CI for #3626

kflynn:
This is CI for @jfrabaute's #3626:

default).

pool.

admin port.

This listener is disabled by default. the config "read_port" can be used

Signed-off-by: Fabrice Rabaute <fabrice@arista.com>


LukeShu:
What's going on with 7913b559c9ebf18e62a5325126d7ee160a8020e1 and de202b73e6d1027633012990b5f6b9c9331023c5 being separate commits?

python/ambassador/envoy/v2/v2ready.py
```diff
@@ -0,0 +1,153 @@
...
+                    'path': config.ir.ambassador_module.envoy_log_path,
+                    'json_format': log_format
+                }
+            })
+        else:
+            # Use a sane access log spec
+            log_format = config.ir.ambassador_module.get('envoy_log_format', None)
+
+            if not log_format:
+                log_format = 'ACCESS [%START_TIME%] \"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%\" %RESPONSE_CODE% %RESPONSE_FLAGS% %BYTES_RECEIVED% %BYTES_SENT% %DURATION% %RESP(X-ENVOY-UPSTREAM-SERVICE-TIME)% \"%REQ(X-FORWARDED-FOR)%\" \"%REQ(USER-AGENT)%\" \"%REQ(X-REQUEST-ID)%\" \"%REQ(:AUTHORITY)%\" \"%UPSTREAM_HOST%\"'
```
LukeShu, line 142: This is a lot of data to duplicate from `v3listener.py`.

python/ambassador/envoy/v2/v2ready.py
```diff
@@ -0,0 +1,153 @@
...
+                                    'name': 'local_route'
+                                },
+                                'http_filters': [
+                                    {
+                                        'name': 'envoy.filters.http.health_check',
+                                        'typed_config': typed_config
+                                    },
+                                    {
+                                        'name': 'envoy.filters.http.router'
+                                    }
```
LukeShu, line 79: What's the point in including this?

python/ambassador/envoy/v2/v2ready.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @classmethod
+    def generate(cls, config: 'V2Config') -> None:
+        # Inject the ready listener to the list of listeners if enabled
+        rport = config.ir.aconf.module_lookup('ambassador', 'ready_port', -1)
+        if rport <= 0:
+            config.ir.logger.info(f"V2Ready: ==== disabled")
+            return
+
+        rip = config.ir.aconf.module_lookup('ambassador', 'ready_ip', '127.0.0.1')
```
LukeShu, line 33: Kube readiness probes don't come from localhost, do they? I'm not sure this is a sensible default.

python/ambassador/envoy/v2/v2ready.py
```diff
@@ -0,0 +1,153 @@
...
+        rlog = config.ir.aconf.module_lookup('ambassador', 'ready_log', True)
+
+        config.ir.logger.info(f"V2Ready: ==== listen on %s:%s" % (rip, rport))
```
LukeShu, line 36: I know it's just a log-line, but this'll be hard to read if `rip` is an IPv6.  Do we have a handy function in Python that mirrors Go's `net.JoinHostPort`?

LukeShu:
In concept, this seems like a good thing to me.  But this seems like an un-opinionated approach, on something that Emissary really should have an opinion on.

Like, I'm thinking we should just always do this on :8006, go ahead and move everything to use that instead of :8001.  I'm not sure I'm OK with this being a user-facing toggle.

python/ambassador/envoy/v2/v2ready.py
LukeShu: Err, it is a sane default if the only thing we expect to use this is Emissary internally.  But if that's the case, none of this should be in user-facing toggle.s

jfrabaute:
> Like, I'm thinking we should just always do this on :8006, go ahead and move everything to use that instead of :8001. I'm not sure I'm OK with this being a user-facing toggle.

I agree. There is more work to do to integrate it completely with ambassador.
1/ the control plane should move to using this endpoint when it's enabled.
2/ the stats suffers the same problem (hitting the envoy admin thread and can be slowed by other requests) so the same work should be done, if possible on the stats endpoint. This requires envoy work as there is no filter for stats.
3/ This will take some time, so in the meantime, I'm trying to start small and iterate, by 2/ integrating this change, disabled by default, 2/ work on integrating the control plane to hit this specific listener when it's enabled, then in the future, we could:
* switch from disabled by default to enabled by default when it's well tested/validated.
* make it part of the default, no option to disable, invisible to the user.

python/ambassador/envoy/v2/v2ready.py
jfrabaute: There are two things here:
* The idea here is to have a `secure` default (so it's not exposed in the wild by mistake).
* At some point, this endpoint might/should be consumed by the ambassador control plane, replacing the /ready check it's doing on the envoy admin listener to this one. In this case, 127.0.0.1 will be secure and a good sensible default.

python/ambassador/envoy/v2/v2ready.py
jfrabaute: IIRC, that's needed right? (I'm not familiar with envoy config, I just reused something similar to config generated).
IIRC, I also tested without and it was not working. I can try again.

python/ambassador/envoy/v2/v2ready.py
jfrabaute: Yes. This is also duplicated from other places.
I've discussed with @kflynn and I agree this should be extracted and the access log format stuff should be part of a library reused by other parts.
I didn't want to tackle this on this specific review because I wanted to focus on improving the /ready endpoint, also because refactoring the log format would impact other modules, and I didn't want to start mixing things, and finally as v2 is going to go away at some point, having duplicates between v2 and v3 does not sound like a big deal as v2 is just going to be removed eventually.

jfrabaute:
I've updated the initial PR with the small needed fix (for what I broke last week). It should be good to go.

I'd be happy to continue working on the overall feature after this first step is merged.

Ping me anytime if can I do something to help having this first review merged.

Thanks!

the-wondersmith:
🍃 Housekeeping 🧹

Issue #3979, opened 2021-12-14
Enabling proxy protocol on AWS NLB makes Emissary to unable to handle requests.


phlawski:
**Describe the bug**
When I enable proxy protocol on my AWS NLB and I configure Emissary Ingress to use it as described in the documentation I am getting following error when sending a request to the cluster:

```
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*  CAfile: /etc/ssl/cert.pem
*  CApath: none
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* error:1400410B:SSL routines:CONNECT_CR_SRVR_HELLO:wrong version number
* Closing connection 0
curl: (35) error:1400410B:SSL routines:CONNECT_CR_SRVR_HELLO:wrong version number
```

My `ambassador Module` looks as follows:

```
apiVersion: getambassador.io/v3alpha1
kind: Module
metadata:
  name: ambassador
  namespace: api-gateway
  labels:
    app.kubernetes.io/name: emissary-ingress
    app.kubernetes.io/part-of: amb
    helm.sh/chart: emissary-ingress-7.1.10
    app.kubernetes.io/instance: amb
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/component: emissary-ingress-ratelimit
    product: aes
spec:
  config:
    defaults:
      httpmapping:
        add_request_headers:
          x-forwarded-proto:
            append: false
            value: https
    diagnostics:
      enabled: false
    envoy_log_type: json
    reject_requests_with_escaped_slashes: true
    server_name: aws-dev-apigateway
    use_proxy_proto: true
```

My `Service` looks as follows:

```
apiVersion: v1
kind: Service
metadata:
  name: amb-emissary-ingress
  namespace: api-gateway
  labels:
    app.kubernetes.io/name: emissary-ingress
    app.kubernetes.io/part-of: amb
    helm.sh/chart: emissary-ingress-7.1.10
    app.kubernetes.io/instance: amb
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/component: ambassador-service
    product: aes
  annotations:
    a8r.io/owner: "Ambassador Labs"
    a8r.io/repository: github.com/datawire/ambassador
    a8r.io/description: "The Ambassador Edge Stack goes beyond traditional API Gateways and Ingress Controllers with the advanced edge features needed to support developer self-service and full-cycle development."
[...]

Issue #4153, opened 2022-03-04
Invalid configuration generated with using protocol `HTTPSPROXY`


cyrus-mc:
When enabling protocol `HTTPSPROXY` generated configuration results in the following error when connecting via `TLS`:

`error:1400410B:SSL routines:CONNECT_CR_SRVR_HELLO:wrong version number`

This is a result of the `proxyProtocolStack` being set to [ "TLS", "PROXY", "HTTP", "TCP" ], so TLS wrapping proxy wrapping HTTP wrapping TCP.

**To Reproduce**

Create a listener as such:

```
apiVersion: getambassador.io/v3alpha1
kind: Listener
metadata:
  name: http
  namespace: emissary
spec:
  port: 8443
  protocol: HTTPSPROXY
  securityModel: SECURE
  l7Depth: 0
  hostBinding:
    namespace:
      from: SELF
```

If using AWS NLB enable proxy protocol by setting the following annotation:

`service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: "*"`

When attempting to connect to Emissary via `https` you will receive error `error:1400410B:SSL routines:CONNECT_CR_SRVR_HELLO:wrong version number`.

**Expected behavior**

With proxy protocol enabled at the NLB and configured via protocol `HTTPSPROXY` a successful connection should be established.

**Versions (please complete the following information):**
 - Ambassador: 2.2.2
 - Kubernetes environment: AWS EKS 1.21

cyrus-mc:
This is due to the fact that `HTTPSPROXY` is set to `proxyProtocolStack` [ "TLS", "PROXY", "HTTP", "TCP" ] which generates configuration:

```
"listener_filters": [
  {
    "name": "envoy.filters.listener.tls_inspector"
  },
  {
    "name": "envoy.filters.listener.proxy_protocol"
  }
]
```

The filters are in the incorrect order. Therefore `proxyProtocolStack` should be [ "PROXY", "TLS", "HTTP", "TCP" ]

cyrus-mc:
The work around for now is to not set the `protocol` too `HTTPSPROXY` but rather explicitly set the `protocolStack` as shown above. However the documentation seems to point people to using `protocol` instead.

cyrus-mc:
[...]

PR #4154, opened 2022-03-04
fix: set correct protocol stack when using `protocol` HTTPSPROXY

cyrus-mc:
When using `HTTPSPROXY` the protocol stack needs to be proxy wrapping TLS.

This is a fix for issue #4153 

## Testing

Was done locally by provisioning a HTTPS proxy enabled listener and fronting that with an AWS NLB with proxyProtocol v2 support enabled.

### DIFF
--- a/python/ambassador/ir/irlistener.py
+++ b/python/ambassador/ir/irlistener.py
@@ -53,7 +53,7 @@ class IRListener (IRResource):
         "HTTPPROXY": [ "PROXY", "HTTP", "TCP" ],
 
         # HTTPSPROXY: accepts encrypted HTTP/1.1 or HTTP/2 sessions using the HAProxy PROXY protocol over TLS over TCP.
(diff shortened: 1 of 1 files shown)


ipince:
I think this would also fix https://github.com/emissary-ingress/emissary/issues/3979

olemarkus:
Could we please get this one merged?

AliceProxy:
I believe this is an issue where the protocol stack when using Proxy Protocol is different whether you are using NLB vs ALB vs ELB, etc. If we can get confirmation that this configuration works on all the popular Cloud provider LoadBalancer types that support proxy protocol then I would say sure lets merge it, but otherwise I think introducing alternatives to `HTTPSPROXY` such as `NLB_HTTPSPROXY` might be the way to go here.

sagikazarmark:
Holy crap, I just wasted 3 hours on this.

the-wondersmith:
🍃 Housekeeping 🧹

PR #4172, opened 2022-03-24
Swap the order of the CORS filter and the ext_authz filter.

kflynn:
Having auth happen before CORS prevents CORS preflights of authenticated resources from ever having a chance at working. Swapping them allows preflights and "real" requests to both work.


AliceProxy (approved):
LGTM, but lets figure out those commits that are not signed properly. Since those commits are not from this PR, that change doesn't need to happen in this PR.

AliceProxy (approved):

Issue #3905, opened 2021-11-03
Envoy lightstep option propagation_modes not supported


rbtcollins:
**Please describe your use case / problem.**

The default in Envoy for tracing propagation headers is  X-Ot-Span-Context. Using Envoy directly it is possible to change to different propagation_modes. We'd like to standardise on one of the standardised carriers defined in opentelemetry: for instance (and specifically for us) the W3C TraceContext headers - https://opentelemetry.io/docs/go/instrumentation/#propagators-and-context - e.g. OTEL_PROPAGATORS="tracecontext" in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md 

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#global-propagators specifies a default of TraceContext + Baggage. 

Currently we pay a cost of integration because Emissary's default is different to that in language toolkits and frameworks.

**Describe the solution you'd like**
Support the Envoy API for changing propagation method. Supporting the set-of-several approach that the underlying API supports would permit us to gracefully migrate across without a flag moment.

**Describe alternatives you've considered**
Shifting Emissary's default to be TraceContext for tracing drivers that support it would be nice too, but isn't the core concern.

**Additional context**
Add any other context about the feature request here.

PR #4179, opened 2022-03-31
Add propagation_modes for the Lightstep Tracer

psalaberria002:
The Lightstep tracer supports multiple propagation modes.
See https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/trace/v3/lightstep.proto#enum-config-trace-v3-lightstepconfig-propagationmode

With this PR it is now possible to set a list of propagation modes in
the config field of the TracingService.

I have verified it locally end to end. Seems to work.
However, I wasn't able to make tests work locally. Issues with a dummy-pod not coming up due to pull permissions, but it doesn't seem right. The cluster is able to pull images from the same registry...

Fixes https://github.com/emissary-ingress/emissary/issues/3905


pkg/api/getambassador.io/v3alpha1/crd_tracingservice.go
```diff
@@ -29,16 +29,20 @@ type TraceSampling struct {
 	Overall *int `json:"overall,omitempty"`
 }
 
+// +kubebuilder:validation:Enum=ENVOY;LIGHTSTEP;B3;TRACE_CONTEXT
```
LukeShu, line 32: Nit: I know some of the existing enums don't follow this, but I believe k8s style is to use PascalCase.

python/ambassador/ir/irtracing.py
```diff
@@ -79,6 +79,10 @@ def setup(self, ir: 'IR', aconf: Config) -> bool:
                 if not driver_config['collector_endpoint_version'] in ['HTTP_JSON_V1', 'HTTP_JSON', 'HTTP_PROTO']:
                     self.post_error(RichStatus.fromError("collector_endpoint_version must be one of 'HTTP_JSON_V1, HTTP_JSON, HTTP_PROTO'"))
                     return False
+            if 'propagation_modes' in driver_config:
+                if not all(item in ['ENVOY', 'LIGHTSTEP', 'B3', 'TRACE_CONTEXT'] for item in driver_config['propagation_modes']):
+                    self.post_error(RichStatus.fromError("propagation_modes must contain any of 'ENVOY, LIGHTSTEP, B3, TRACE_CONTEXT'"))
+                    return False
```
LukeShu, line 85: It shouldn't be necessary to check this here; the schema-based validator will have already rejected the resource if it doesn't match what's in the `+kubebuilder:validation:Enum=` comment.

(This is only true since 2.1.2, which is why there's still a bunch of now-superflous code that does these type of checks.)

pkg/api/getambassador.io/v3alpha1/crd_tracingservice.go
psalaberria002: I am following the expected values in the Envoy proto files https://github.com/envoyproxy/envoy/blob/96c317837ab7351f6264d25c6e19e09fae3a6054/api/envoy/config/trace/v3/lightstep.proto#L29

python/ambassador/ir/irtracing.py
psalaberria002: Nice!

psalaberria002:
The tests are based on test_shadow.py. If there is a cleaner way of initializing the confis, I am all ears :)

kflynn (dismissed):
This looks good -- thanks, @psalaberria002!! 🙂 I'm going to see about getting CI to run on it...

@LukeShu, I was feeling the same way about the enum values, until remembering that the `driver_config` in the `TracingService` is literally handed to Envoy after validation. 🙁  We need to fix _that_.

kflynn (approved):
We have a clean CI run in #4192, so off we go! Thanks @psalaberria002!

PR #4181, opened 2022-04-01
 Allow setting custom_tags for traces

psalaberria002:
It allows setting trace tags based on literal values, environment variables, or request headers.

Based on https://www.envoyproxy.io/docs/envoy/latest/api-v3/type/tracing/v3/custom_tag.proto#custom-tag

`tag_headers` should be deprecated in favor of `custom_tags`. I haven't worked with CRDs and field deprecations. What's the procedure?

When it comes to updating the documentation, where is that handled? I cannot find any reference to `TracingService` inside the docs folder.


psalaberria002:
> Hey @psalaberria002! As I understand it:
> 
> * Any `TagHeader` can be expressed as a `CustomHeader`, but not the other way around; and
> * Transforming a `TagHeader` to a `CustomHeader` is trivial and mechanical.
> 
> That implies to me that the best answer here is:
> 
> * `v3alpha1` should consider `TagHeader` to be deprecated
>   
>   * if only `CustomHeader` is set, great
>   * if only `TagHeader` is set, translate to `CustomHeader`
>   * if both `TagHeader` and `CustomHeader` are set, log a warning, ignore `TagHeader`, and use `CustomHeader`
>   * you should rename the `TagHeader` field internally to `DeprecatedTagHeader`, but leave the JSON tag as `TagHeader`
> * `v2` should gain a `V3CustomHeader` field, used only for translations
> * When translating from `v3alpha1` to `v2`:
>   
>   * the `v3alpha1` `CustomHeader` field should be saved unaltered as the `v2` `V3CustomHeader` field
> * When translating from `v2` to `v3alpha1`:
>   
>   * if `v2` `V3CustomHeader` is set, it should be copied unaltered to `v3alpha1` `CustomHeader`
>   * otherwise, if `v2` `TagHeader` is set, it should be translated to `v3alpha1` `CustomHeader`
>   * the translation code must never set `v3alpha1` `DeprecatedTagHeader`
> * In `v3` final, the `DeprecatedTagHeader` field will go away.
> 
> How does that sound?

Makes sense. I have tried to follow that approach in the last commit. It would be nice to have CI for this PR :)

LanceEa:
@psalaberria002 - Thanks for being patient on this. With the recent release master has diverged quite a bit. Can you please rebase this back on master and push it. I will then get CI running on it.

Note: it's probably good to run `git fetch origin --prune` to ensure you are not holding onto any old refs for master before re- pulling it. I had that issue ☹️ .

psalaberria002:
Already pulled from master, cherry picked my change, and force pushed.

pkg/api/getambassador.io/v2/handwritten.conversion.go
```diff
@@ -658,3 +659,55 @@ func Convert_v3alpha1_TCPMappingSpec_To_v2_TCPMappingSpec(in *v3alpha1.TCPMappin
 
 	return nil
 }
+
+func Convert_v2_TracingServiceSpec_To_v3alpha1_TracingServiceSpec(in *TracingServiceSpec, out *v3alpha1.TracingServiceSpec, s conversion.Scope) error {
```
LanceEa, line 663: @LukeShu - I looked at the other handwritten conversion functions and it looks like we are using snake case in this file...is there a historical reason? I'm surprised our Linting rules are not yelling about this.

pkg/api/getambassador.io/v2/handwritten.conversion.go
```diff
@@ -658,3 +659,55 @@ func Convert_v3alpha1_TCPMappingSpec_To_v2_TCPMappingSpec(in *v3alpha1.TCPMappin
...
+	if err := autoConvert_v2_TracingServiceSpec_To_v3alpha1_TracingServiceSpec(in, out, s); err != nil {
+		return err
+	}
+	// WARNING: in.TagHeaders requires manual conversion: does not exist in peer-type
+	// if only tag_headers are set, translate to custom_tags.
+	// if both are set, ignore tag_headers.
+	if in.TagHeaders != nil {
```
LanceEa, line 670: lets reduce the nesting since we only apply this logic when these two things are true:

`if in.TagHeaders != nil && in.V3CustomTags == nil {`

python/ambassador/envoy/v3/v3listener.py
```diff
@@ -447,18 +447,21 @@ def base_http_config(self) -> Dict[str, Any]:
...
-
-            if req_hdrs:
-                base_http_config["tracing"]["custom_tags"] = []
-                for hdr in req_hdrs:
-                    custom_tag = {
-                        "request_header": {
-                            "name": hdr,
+            if custom_tags or req_hdrs:
+                finalCustomTags += custom_tags
+                if req_hdrs:
```
LanceEa, line 455: I thought we already translated these in the CRD conversion work from `v2` to `v3alpha1`? 

AFAIK, by the time we get here the values have been converted to `v3alpha1`. Do we need to still translate or can we assume at this point that the `custom_tags` has everything we want already?

pkg/api/getambassador.io/v2/handwritten.conversion.go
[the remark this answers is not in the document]
psalaberria002: The function is autogenerated by running the command that's defined in the 4th line of this file.

python/ambassador/envoy/v3/v3listener.py
psalaberria002: That's probably true. I have simplified it now. Will need to properly test that. Running tests locally is taking ages.

LanceEa:
@psalaberria002 - lets get this rebased on master and update the release notes so that it will land on the upcoming `3.1.0` release. Once that is done I will pull it in and get CI running to make sure all is still green. I want to get a second review from one of the teammates as well.

Again thanks for the patience on this.

psalaberria002:
> @psalaberria002 - lets get this rebased on master and update the release notes so that it will land on the upcoming `3.1.0` release. Once that is done I will pull it in and get CI running to make sure all is still green. I want to get a second review from one of the teammates as well.
> 
> Again thanks for the patience on this.

Rebased. I am not able to run tests locally. It would be great if you could run this in CI.

pkg/api/getambassador.io/v2/handwritten.conversion.go
```diff
@@ -658,3 +659,53 @@ func Convert_v3alpha1_TCPMappingSpec_To_v2_TCPMappingSpec(in *v3alpha1.TCPMappin
...
+			for _, tag := range in.DeprecatedTagHeaders {
+				in.CustomTags = append(in.CustomTags, v3alpha1.TracingCustomTag{
+					Tag: tag,
+					Header: &v3alpha1.TracingCustomTagTypeRequestHeader{
+						Name: tag,
+					},
+				})
+			}
+		} else {
+			// TODO: Use dlog logger
```
LanceEa, line 700: Let's use `dlog` here and then just use `context.Background` since this function doesn't have access to any other context. This way we are consistent with the rest of the code base and it gets outputted in a consistent format.

pkg/api/getambassador.io/v3alpha1/crd_tracingservice.go
```diff
@@ -53,11 +91,14 @@ type TracingServiceSpec struct {
...
 	Driver string `json:"driver,omitempty"`
 	// +kubebuilder:validation:Required
-	Service    string         `json:"service,omitempty"`
-	Sampling   *TraceSampling `json:"sampling,omitempty"`
-	TagHeaders []string       `json:"tag_headers,omitempty"`
-	Config     *TraceConfig   `json:"config,omitempty"`
-	StatsName  string         `json:"stats_name,omitempty"`
+	Service  string         `json:"service,omitempty"`
+	Sampling *TraceSampling `json:"sampling,omitempty"`
+	// tag_headers is deprecated. Use custom_tags instead.
```
LanceEa, line 96: It looks kubebuilder has no notion of deprecation validation so let's at least take advantage of Golangs doc comments for deprecated.

https://stackoverflow.com/questions/7849663/how-do-you-mark-code-as-deprecated-in-go

LanceEa (commented):
A couple final last nits and then I'm all good. If you could make those changes and then I will pull them in, verify CI is still green and we can get this merged!

Thanks for all the help and putting up with the painful process!

docs/releaseNotes.yml
```diff
@@ -32,6 +32,19 @@
 
 changelog: https://github.com/emissary-ingress/emissary/blob/$branch$/CHANGELOG.md
 items:
+  - version: 3.2.0
+    date: '2022-08-15'
```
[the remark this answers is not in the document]
psalaberria002: Could not find this documented. I tried to set a placeholder, but it complained because it didn't know how to parse the date. TBD does the trick, thanks.

pkg/api/getambassador.io/v3alpha1/crd_tracingservice.go
psalaberria002: This one is a little bit tricky. The comments endup in the CRD specs. The Deprecated convention in Go is for the struct fields, not the json fields. I am not sure if this is worth it. If so, can you write down here the expected comment string?

LanceEa: Correct, the reason I suggested the Go Doc comment is that these objects can be used in the future where we have a transformation engine that is written in Go. This way if objects are being instantiated via Go code the IDE is going to yell you.

However, I was reading through the docs this morning and it looks like there is a deprecated special marker. It says on-type so it might not work on fields but ideally it can be applied to a field too.
https://book.kubebuilder.io/reference/markers/crd.html

psalaberria002: `+kubebuilder:deprecatedversion` doesn't seem to do anything on fields. I am gonna add the Go deprecation comment and use the json fields in the description since those are visible both in the CRD schema and the Go code.

LanceEa:
@psalaberria002 - Landed this in https://github.com/emissary-ingress/emissary/pull/4268 per our discussion in slack. Closing this one out.

Issue #4184, opened 2022-04-06
Working Multi-Arch AMD64/ARM64 Images


the-wondersmith:
Just posting this here in case [anyone](https://github.com/emissary-ingress/emissary/issues/3831) [else](https://github.com/emissary-ingress/emissary/issues/2652) [has run into](https://github.com/emissary-ingress/emissary/issues/2647) [the same roadblock](https://github.com/emissary-ingress/emissary/issues/2121) I did - the lack of multiarch (specifically `arm64`) images for [emissary](https://github.com/emissary-ingress/emissary) and [telepresence](https://github.com/telepresenceio/telepresence).

I've put together a fully functional, `docker buildx build`-able Dockerfile that will builds and assembles a functional `emissary` image. I've been using a private GCR repo while I work this process out, but I will push the working multi-arch images (along with the same for `telepresence`) to my public Docker Hub within the next day or so.

**DISCLAIMER:** _While the binaries built by this process are absolutely identical to those from "official" images, the final image produced by this build process is **not** tested or supported by the Datawire team. DO NOT go whining to them if it doesn't work properly or has some issue._

If you'd like to build your own, (or don't feel like waiting for me to get around to pushing the images to Docker Hub) just run:

```bash
# Prebuild Preparation Command
mkdir -p /tmp/emissary-build && cd /tmp/emissary-build && curl https://raw.githubusercontent.com/the-wondersmith/emissary/multiarch-build/Dockerfile.multiarch > /tmp/emissary-build/Dockerfile

# Single-Arch Build
docker build --tag='your.personal.repo/emissary:your-tag-here' -f /tmp/emissary-build/Dockerfile . && docker push your.personal.repo/emissary:your-tag-here

--OR--

# Multi-Arch Build
docker buildx build --push --tag='your.personal.repo/emissary:your-tag-here' --platform='linux/arm64/v8,linux/amd64' -f /tmp/emissary-build/Dockerfile .

# Deployment Command
[...]

PR #4188, opened 2022-04-13
Enable contributors developing on Apple Silicon (M1 - Arm64)

LanceEa:
## Description
The Emissary-Ingress build system supports and tests 'linux/amd64' containers which presents a problem when developing locally on newer M1 Macs which have Apple Silicon (arm64). To enable developers to build locally on Apple Silicon we have introduce a new environment variable `BUILD_ARCH` which defaults to `linux/amd64` so that running `make images` will produce the correct artifacts.

**Important** - This does NOT introduce a multi-architecture image but rather to decrease the friction for contributors on Apple Silicon. The multi-architecture support is a separate stream which some of the discussion can be found in #4184.

## Related Issues
N/A

## Testing

- Manually verified that `make images` is able to properly build the containers
- Manually verified that `make deploy` can push containers out to development K8s cluster. 
- Verified that Emissary-Ingress ran successfully within cluster after forcing linux/amd64 architecture
- Tested that Make variable `BUILD_ARCH` defaults `linux/amd64` but can be overridden by setting the Environment Variable with the same name

## Checklist

I haven't because this doesn't really apply to a specific release...however, it might be good to cherry pick these changes to any release branch that is still supported?
 
It might be good if someone without an M1 Mac or another developer on the M1 Mac tries it out as well but so far there doesn't appear to be any issues with using this approach in the short-term.
 
 - [] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -40,6 +40,9 @@ requirements and provide a friendly error message. If you ever find
 that this is not the case please file a PR with a fix. Likewise if you
 ever find anything missing from this list.
 
+> To enable contributors using Apple Silicon we ensure that the artifacts are built for `linux/amd64`
+> rather than the host `linux/arm64` architecture. This can be overriden using the `BUILD_ARCH` environment variable.
+
 ### Requirements:
 
  - git
@@ -727,3 +730,10 @@ upgrade everything else, then
     for those you want to pin.
  2. Delete `python/requirements.in` (if it exists).
  3. Run `make generate`.
+
+How do I develop on an Mac with Apple Silicon?
+----------------------------------------------
+
+To ensure that developers using a Mac with Apple Silicon can contribute, the build system ensures
+the build artifacts are `linux/amd64` rather than the host architecture. This behavior can be overriden
+using the `BUILD_ARCH` environment variable (e.g. `BUILD_ARCH=linux/arm64 make images`).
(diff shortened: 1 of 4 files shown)


LanceEa:
@kflynn @LukeShu 

1. Are you cool with the variable name I introduced and location or is their a more preferred location?
2. Should I update the `/docs/releaseNotes.yml` since this is more of a contributor facing PR?
3. Do we cherry-pick this kind of stuff into release branches?

kflynn (changes requested):
@LanceEa Looks good from here! One request: please drop a note into `DEVELOPING.md` about this? but yeah, I'm good with the content. (I would've been OK with `BUILD_ARCH` for the name as well, but I'm good with `CONTAINER_BUILD_ARCH`.)

AliceProxy:
@LanceEa Normally before the next release we will create the release branch off of master so landing that change here is fine. We normally wouldn't cherry-pick it into release branches such as 2.1 that have already been released since the state of those branches should be everything that was in that release.

LanceEa:
> @LanceEa Looks good from here! One request: please drop a note into `DEVELOPING.md` about this? but yeah, I'm good with the content. (I would've been OK with `BUILD_ARCH` for the name as well, but I'm good with `CONTAINER_BUILD_ARCH`.)

I originally had `BUILD_ARCH` so since you will accept that I changed it to keep it smaller 😄 . I have added a few changes to `DEVELOPING.md` so let me know if that works for you.

AliceProxy (approved):
Awesome stuff :raised_hands:

kflynn (approved):
Looks good, hit it! 🙂

PR #4194, opened 2022-04-15
Begone with version-hack.mk and other build-sys messiness

LukeShu:
## Description

This is a big PR that maybe I should have split up, but here we go!

Oh, and I want to call out the misleading `756 insertions(+), 555 deletions(-)` diffstat; the growth is because of a big ol' `go.sum` file; if I exclude the `go.sum` file, the diffstat is `346 insertions(+), 555 deletions(-)`.  That is: things are in fact getting smaller and simpler, not bigger and more bloated.

----

Way back when, we'd ended up in a situation where there were a bunch of output-files that were being checked in to Git, and other parts of the build process wore consuming them.  Because they were checked in to Git, many of the things that consumed them  didn't actually declare dependencies on them, as they would count on them already being there from Git.  And I don't mean `make generate`-type-infrequently-updated files, these files were changing ~all the time. It made (among other things) releasing a nightmare.  So `version-hack.mk` was an awful hack to get us from that bad place to a marginally less-bad place: It stopped checking the files in to Git, and instead caused them to get generated/updated _at Makefile parse time_, before the rest of the Makefile starts running. Then later, we can clean up those files and properly declare dependencies and such.

Well, "later" is now.

There are a few parts to this:
 - The `version-hack.mk` version of the recipes all declared dependencies on `FORCE`, because they effectively depended on the values of variables rather than the contents of files; and that's not a dependency you can express in Make.
   - So, as best I can, I encode those variable values in to filenames.  Instead of a file "mysteriously" depending on `$(VERSION)`, the filename is `build-output/whatever-$(VERSION)`; and then the recipes for `build-output/whatever-%` use `$*` to get the value of VERSION, rather than looking at `$(VERSION)` directly.
   - But, there still were a couple of variables I couldn't figure out a good way to do that for (specifically `DEV_REGISTRY` and `RELEASE_REGISTRY`).  Well, it turns out that many moons ago I invented a way to depend on variable values in Make: `build-aux/var.mk`. (It's really neat how still, several years later, we're spending time fixing problems that we already had solutions to long ago but threw away because reasons™.)  Anyway, those targets can declare dependencies on `$(var.)DEV_REGISTRY` and `$(var.)RELEASE_REGISTRY` instead of having to declare a dependency on `FORCE`.
 - Move all of the files around; put them all in `build-output/`.  This is partially to make it clear to humans that they're outputs, but also in order to identify all of the places that need to be updated to depend on them.
 - It's not _just_ updating filenames and declaring dependencies; a key thing is to look at whether the thing is actually needed; whether the dependency _aught_ to exist. Generating the Helm chart depends on `$(VERSION)`, and generating the checked-in YAMLs depends on the Helm chart; but the checked-in YAMLs have the version stripped out, so having `$(VERSION)` in that dependency chain turns out to be silly.  So in some cases that means instead of depending on `build-output/chart-$(VERSION)` you depend on `build-output/chart-2.0.0-bogus` to signal that you don't actually care about the VERSION.  In some cases (like `agent_test.go`) it means going for the source `*.yaml.in` files directly instead of the output `*.yaml` files.

As usual for me, I suggest a commit-by-commit review:

- The first 11 commits (the `misc cleanup:` and `consolidate:` commits) are cleaning up the things that I'm going to be touching (or cleaning up messiness I noticed while touching them), to make them easier to edit.
- The 12th and 13th commits are doing something big and cool that maybe should have been a separate PR: (1) Don't use Docker to run `ct` for testing the chart; run it outside of Docker!, and (2) integrate the chart stuff in to the main build system, finally killing the vestiges of the fact that the chart used to be in a separate repo.
- Then we have 3 commits that move files out of `version-hack.mk`, turning them in to properly-depended-on outputs (or in the case of `versions.yml` just removing it entirely). (see the discussion above)
- Then we have 1 commit actually removing `version-hack.mk`.
- Then we have 2 commits that that clean up the docs-yaml stuff to match the new scheme.  This is made possible by the aforementioned removal of `versions.yml`.

## Testing
CI turns green. I haven't (yet?) tested the release machinery with this.

## Checklist

    + Any new features
    + Any deprecations
 
 - [ ] My change is adequately tested. - release machinery could stand to be tested
 
 - [ ] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently. - arguably there are some Make tricks, as discussed above. But I'm not sure DEVELOPING.md is the place for them.


build-aux/main.mk
```diff
@@ -124,3 +124,25 @@ PHONY: chart
...
+	rm -rf $@
+else
+	@if test -d $@; then \
+	  echo 'This should not happen in CI: $@ should not need to change' >&2; \
+	  echo 'Files triggering the change are: $?' >&2; \
+	  exit 1; \
+	fi
+endif
+	mkdir -p $@
+	$(foreach src,$(filter %.yaml.in,$^),$(foreach dst,$(patsubst $(CURDIR)/manifests/emissary/%.yaml.in,$@/%.yaml,$(src)),\
```
kflynn, line 140: 😂  Nobody other than you is ever going to dare read this thing, much less edit it. Especially given that your second `foreach` isn't even a loop, unless I'm badly misreading something, since the first `foreach` handles all of that. Why not just do it in the shell, so that it's more clear where the iteration is happening and what the loop values are?

```shell
for src in $(filter %.yaml.in,$^); do \
  dst=$(echo "$$src" | sed -e 's!^$(CURDIR)/manifests/emissary/!$@!' -e 's!\.in$$!!') ;\
  if [[ "$$dst" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-rc\.[0-9]+|-ea)?$$ ]]; then \
    registry=$(RELEASE_REGISTRY) ;\
  else \
    registry=$(DEV_REGISTRY) ;\
  fi ;\
  sed -e 's/\$$version\$$/$*/g' -e 's,\$$imageRepo\$$,'"$${registry}"'/emissary,g' <$(src) >$(dst) ;\
done
```

Less elegant Make-fu, but probably easier for most to reason about...

build-aux/main.mk
```diff
@@ -124,3 +124,25 @@ PHONY: chart
...
+	  { \
+	    if [[ '$*' =~ ^[0-9]+\.[0-9]+\.[0-9]+(-rc\.[0-9]+|-ea)?$$ ]]; then \
+	      registry=$(RELEASE_REGISTRY); \
+	    else \
+	      registry=$(DEV_REGISTRY); \
+	    fi; \
+	    sed -e 's/\$$version\$$/$*/g' -e 's,\$$imageRepo\$$,'"$${registry}"'/emissary,g' <$(src) >$(dst); \
```
kflynn, line 147: Looks like we changed the charts to `@var@` but left this with `$var$`? Any particular reason?

docs/yaml.mk
```diff
@@ -1,22 +1,22 @@
...
+else
+	@if test -d $@; then \
+	  echo 'This should not happen in CI: $@ should not need to change' >&2; \
+	  echo 'Files triggering the change are: $?' >&2; \
+	  exit 1; \
+	fi
+endif
+	$(foreach src,$(filter %.yaml,$^),$(foreach dst,$(patsubst docs/yaml/%,$@/%,$(src)),\
+	  mkdir -p $(dir $(dst))$(NL)\
+	  sed -e 's/\$$version\$$/$*/g' -e 's/\$$quoteVersion$$/0.4.1/g' <$(src) >$(dst)$(NL)))
```
kflynn: Same comment as for the other `foreach-foreach` place.

build-aux/main.mk
```diff
@@ -146,3 +146,17 @@ endif
...
+ifeq ($(CI),)
+	rm -rf $@
+else
+	@if test -d $@; then \
+	  echo 'This should not happen in CI: $@ should not need to change' >&2; \
+	  echo 'Files triggering the change are: $?' >&2; \
+	  exit 1; \
+	fi
+endif
+	$(foreach src,$(filter %.yaml,$^),$(foreach dst,$(patsubst docs/yaml/%,$@/%,$(src)),\
```
kflynn, line 160: Same comment as the other `foreach-foreach` above.

kflynn (approved):
OK, I have some things to talk to you offline about, and a couple of oh-dear-god-the-Make-fu! comments, but: oh dear god let's land this thing!! This is _so_ much better. Love it. 🙂

build-aux/main.mk
LukeShu: I'm certain that _somewhere_ where I do this (maybe not in emissary.git... maybe telepresence.git? or getambassador.io.git?) I have a comment explaining it.

```make
$(foreach key,val,\
    body)
```
is just Make's funny way of spelling the Lisp
```lisp
(let ((key val))
    body)
```

kflynn:
Landing, with any other work to be in separate PRs.

build-aux/main.mk
LukeShu: The reason I prefer to do things with Make's loops is so that I can use make --debug to see what it's going to do before I actually run it.

kflynn: > "Make expressions are just Lisp with dollar signs, commas, and backslashes."

...that doesn't make Lisp any easier for non-Lisp speakers. 😂

PR #4202, opened 2022-04-25
deps: upgrade protoc to enable building on M1 mac

### LINKED ISSUE #4201: unable to run `make generate` on M1 Mac
**Describe the bug**

On an M1 Mac when running the `make generate` command it will fail when trying to download the latest `protoc` binary. There are two issues:

1. M1 Mac not supported on 3.8 version of protoc
2. The protoc binary is named differently than how we determine the binary in the `build-aux/tools.mk`

Due to these two issues it will fail to download the binary and Make will dump and not finishing building the targets.

**To Reproduce**
1. Clone repository on a M1 Mac (arm64) 
2. run cmd `make generate` 

**Expected behavior**
I would expect it to just work on an M1 Mac and that the correct protoc binary would be downloaded for local development so that contributors developing on an M1 Mac.

LanceEa:
Prior to this change the version of protoc being used was `3.8.0`. An arm64 version of protoc for osx is not available in `3.8.0` so running commands like `make generate` do not work for contributors running on M1 macs.

This upgrades the repository to the latest version of protoc and adjusts the `tools.mk` so that it can download protoc correctly. The provided protoc binaries that are released for osx use the suffix naming scheme of `osx_aarch_64`. But, `uname` returns `arm64` causing it to fail on download. This PR includes a change to address this difference between what is reported by uname and what is expected by the precompiled protoc binaries.

## Description
Support M1 Mac and upgrade to latest dep.

## Related Issues
Closes #4201

## Testing

manually ran `make clean && make clobber && make generate` and successfully finished.


LanceEa:
@LukeShu @kflynn  - This came up in Alice's meeting with the ACE team which I hadn't come across yet because I had never ran the `make generate` command.

A couple questions:
1. Do we feel that we have adequate testing to ensure the protoc upgrade isn't an issue?
2. I'm guessing we will need to do the same for EdgeStack?
3. There was a change to the `director.pb.go` due to it appears to be a namespace change for duration/timestamp of the well known types (I will investigate). Does something in Ambassador Cloud need to be potentially adjusted or is this only used within Emissary?

Thanks,
Lance

kflynn (approved):
Hit it!!

AliceProxy (approved):

Issue #4203, opened 2022-04-25
testing helm charts on Apple Silicon fails due to helm chart testing cli not supporting arm64


LanceEa:
**Describe the bug**
When running `make test-chart` on Apple Silicon it will fail due to `v3.0.0-rc.1` of helm `chart-testing` due to it having a transitive dependency on an older version of `golang.org/x/sys/unix` not supporting arm64.

**To Reproduce**
1. clone repository
2. run `make test-chart` on an M1 Mac
3. see error 

```bash
golang.org/x/sys/unix
../../../../../go/pkg/mod/golang.org/x/sys@v0.0.0-20200327173247-9dae0f8f5775/unix/syscall_darwin.1_13.go:25:3: //go:linkname must refer to declared function or variable
```

LanceEa:
Closed via PR #4205

PR #4205, opened 2022-04-25
tools: upgrade helm chart-testing dependency to support Apple Silicon

LanceEa:
## Description

This change upgrades `chart-testing` cli that is used to test and verify our helm charts. The project was using `v3.0.0-rc.1` which didn't have pre-built binaries for arm64. This change updates it to the last minor release of v3.5.1

## Related Issues
closes issue #4203

## Testing
After change was able to run `make test-chart` on an M1 Mac.

## Checklist

 - [x ] My change is adequately tested.


LukeShu:
I don't oppose upgrading `ct`, but I'm confused by the reasoning. We aren't using pre-built binaries; we're building it from source. So why does it matter whether they published pre-built binaries of that version?

LanceEa:
@LukeShu - This is the same issue we are seeing with `yq` due to transitive dependency on an older version of `golang.org/x/sys/unix` which isn't compatible with arm64. 

error shown when running `make test-chart`

```bash
golang.org/x/sys/unix
../../../../../go/pkg/mod/golang.org/x/sys@v0.0.0-20200327173247-9dae0f8f5775/unix/syscall_darwin.1_13.go:25:3: //go:linkname must refer to declared function or variable
```

> "The project was using v3.0.0-rc.1 which didn't have pre-built binaries for arm64"

I took a look at `tools.mk` and see that it is being built from source so yeah I'm not sure why I wrote this sentence in the description. I can edit it or ensure it is cleaned up on merge. Let me know which you prefer!

LanceEa:
No problem, I will get it merged and described correctly.

PR #4210, opened 2022-04-28
Test empty rewrites

kflynn:
We had an issue a little while back where empty rewrites (`rewrite: ""` in a `Mapping`) caused a regression, and we realized that we weren't testing that specific case. 

This PR changes two gRPC tests (which actually require a `rewrite`, and which can use an empty rewrite effectively) to test that case, and also switches one of them to use a CRD rather than an annotation so that we cover both annotations and CRDs. I also verified that the `rewrite` is indeed necessary for the gRPC `Mapping`s: if you set it to something other than the `prefix` value or `""`, the gRPC calls fail.

### DIFF
--- a/python/ambassador/utils.py
+++ b/python/ambassador/utils.py
@@ -169,7 +169,7 @@ def parse_bool(s: Optional[Union[str, bool]]) -> bool:
 
     # OK, we got _something_, so try strtobool.
     try:
-        return strtobool(s)
+        return bool(strtobool(s)) # the linter does not like a Literal[0, 1] being returned here
     except ValueError:
         return False
 
--- a/python/tests/kat/t_grpc.py
+++ b/python/tests/kat/t_grpc.py
@@ -25,7 +25,7 @@ kind: Mapping
 grpc: True
 hostname: "*"
 prefix: /echo.EchoService/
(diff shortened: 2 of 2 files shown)


AliceProxy (approved):

python/ambassador/utils.py
```diff
@@ -169,7 +169,7 @@ def parse_bool(s: Optional[Union[str, bool]]) -> bool:
 
     # OK, we got _something_, so try strtobool.
     try:
-        return strtobool(s)
+        return bool(strtobool(s)) # the linter does not like a Literal[0, 1] being returned here
```
LanceEa, line 172: :) I just ran into this one as well. The type narrowing from mypy doesn't seem to be working. is it an old version of mypy that we lint with?

LanceEa (approved):

python/ambassador/utils.py
kflynn: I think it's that we had `distutils` get updated. I just cherry-picked @AliceProxy's fix for this to include in the PR.

AliceProxy (approved):
Nice! (curse you mypy)

LanceEa (approved):

PR #4212, opened 2022-04-29
dep: upgrade yq tool to latest major version to support apple silicon

LanceEa:
## Description
We are upgrading to the latest major version for the following reasons:

- `yq` v3 is now deprecated per the following <https://mikefarah.gitbook.io/yq/v/v3.x/#deprecated-notice> and will no longer receive security patches.
- `yq` v3 has a transitive dependency on an older version of `golang.org/x/sys/unix` that is not compatible with arm64. Upgrading to v4 allows developers on Apple Silicon computers to contribute.


AliceProxy (approved):
Looks fine to me. I'll defer to @kflynn on whether we should move it to apro

LanceEa:
after discussing with the team, since `yq` is no longer used in this repository we can close this PR.

PR #4217, opened 2022-05-02
[3.0] Repatriate from v2.3

LukeShu:
## Description

CI uses the presence of the `ci/repatriate/from-XXX-to-YYY` branch do determine whether a PR already exists. So what happened is it created the branch, but failed to create the PR, so it hasn't done a PR for any of the repatriate jobs since then.
https://github.com/emissary-ingress/emissary/runs/6197419432


kflynn (dismissed):
Go for it!

AliceProxy (approved):
I approve this message

PR #4232, opened 2022-05-13
[WIP] Fix 503s on secret rotation

d6e-automaton:
Fix 503s on secret rotation by switching to SDS, rather than reconfiguring clusters every time. Additionally, clean up the `ambex` snapshot-dumping code so that it properly deals with typed structures in the Envoy configuration.

Things that need tackling before we ship:
- Comments, comments, some more comments, and then finally some comments.
- Restore fingerprinting of secrets, so that we can log "hey, we're now using a secret with this fingerprint" and have it be meaningful.
- Nice-to-have: log ambex snapshots using the actual snapshot version number so that it's trivial to match up a log message with a snapshot on disk.
- Nice-to-have: do that for the watt snapshots too.

I tested this by running ~10RPS through a `Mapping` using TLS origination, and then rotating the secret it's using every five seconds. _Zero_ 503s in a forty-minute test run, which is good, and the ambex logs show the secret being handed over SDS changing. I want to modify the upstream to show the fingerprint of the offered cert, though, for paranoia.

 - [x] This is unlikely to impact how Ambassador performs at scale; in fact, it may well improve performance at scale.


the-wondersmith:
🍃 Housekeeping 🧹

PR #4235, opened 2022-05-18
Allow 'make clobber' to run

kflynn:
This is the simple fix to allow 'make clobber' to run. `make` seems unhappy with rules like

```
file.ext1.ext2: %.ext2
  command
```

where it's a two-colon rule that defines a pattern (`%.ext2`) that doesn't actually depend on anything: the `command` runs, but then `make` tries to build `%.ext2` and finds nothing.

For this particular case, using '$@' instead of '$*' works and doesn't require the two-colon rule format.

Why did I call this the "simple fix"? Because `make clobber` is supposed to also invoke `make clean`... and it doesn't, because that rule is defined only in `common.mk`, which the `Makefile` doesn't import. So there's another PR coming.

 - [x] I didn't need to update `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently.

### DIFF
--- a/_cxx/envoy.mk
+++ b/_cxx/envoy.mk
@@ -115,8 +115,8 @@ $(OSS_HOME)/_cxx/envoy: FORCE
 	        git checkout origin/master; \
 	    fi; \
 	}
-$(OSS_HOME)/_cxx/envoy.clean: %.clean:
-	$(if $(filter-out -,$(ENVOY_COMMIT)),rm -rf $*)
+$(OSS_HOME)/_cxx/envoy.clean:
+	$(if $(filter-out -,$(ENVOY_COMMIT)),rm -rf $@)
 .PHONY: $(OSS_HOME)/_cxx/envoy.clean
 clobber: $(OSS_HOME)/_cxx/envoy.clean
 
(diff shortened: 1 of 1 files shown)


_cxx/envoy.mk
```diff
@@ -203,8 +207,8 @@ $(OSS_HOME)/docker/base-envoy/envoy-static-stripped: %-stripped: % FORCE
 	        rsync -a$(RSYNC_EXTRAS) --partial --blocking-io -e 'docker exec -i' $$(cat $(OSS_HOME)/_cxx/envoy-build-container.txt):/tmp/$(@F) $@; \
 	    fi; \
 	}
-$(OSS_HOME)/docker/base-envoy/envoy-static.clean $(OSS_HOME)/docker/base-envoy/envoy-static-stripped.clean: %.clean
```
LukeShu, line 206: This is the only line that's the problem; by typo it's missing the `:` at the end of the line.

LukeShu (changes requested):

LukeShu:
See also: https://github.com/emissary-ingress/emissary/pull/4238

kflynn:
Closing in favor of #4238.

PR #4238, opened 2022-05-19
Fix `make clobber`, and have CI check that it works

LukeShu:
## Description
It isn't fixed if CI isn't checking it!

Also, a hack to `py-list-deps` to get it to work with newer `setuptools`, since my laptop has newer `setuptools`.

## Related Issues
List related issues.

## Testing
I had CI check that `make clobber` removes things after each and every CI job.

It's possible that I made a mistake with one of the CI jobs that doesn't run on PRs.

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/.github/actions/after-job/action.yml
+++ b/.github/actions/after-job/action.yml
@@ -1,0 +1,37 @@
+name: 'After-job actions'
+description: >-
+  Perform generic actions that should be performed at the end of every
+  job.
+inputs:
+  jobname:
+    default: ${{ github.job }}
+runs:
+  using: "composite"
+  steps:
+    # collect logs
+    - uses: ./.github/actions/collect-logs
+      if: always()
+      with:
+        jobname: ${{ inputs.jobname }}
+
+    # dirty check
+    - name: "Check that the job didn't change any files"
+      uses: ./.github/actions/git-dirty-check
+      if: ${{ ! failure() }}
+
+    # remote test cluster
+    - run: make ci/teardown-k3d
+      shell: bash
+      if: ${{ ! failure() }}
+
+    # `make clobber`
+    - run: make clobber
+      shell: bash
+      if: ${{ ! failure() }}
+    - name: "Check that `make clobber` didn't change any files"
+      uses: ./.github/actions/git-dirty-check
+      if: ${{ ! failure() }}
+    - name: "Check that `make clobber` did everything it's supposed to"
+      shell: bash
+      run: $GITHUB_ACTION_PATH/check-clobber.sh
+      if: ${{ ! failure() }}
--- a/.github/workflows/promote-ga.yml
+++ b/.github/workflows/promote-ga.yml
@@ -29,6 +29,8 @@ jobs:
       - name: "make release/promote-oss/to-ga"
         run: |
           make release/promote-oss/to-ga
+      - uses: ./.github/actions/after-job
+        if: always()
       - name: Slack notification
         if: always()
         uses: edge/simple-slack-notify@master
@@ -44,6 +46,8 @@ jobs:
              { "title": "Branch", "value": "${env.GITHUB_REF}", "short": true },
              { "title": "Action URL", "value": "${env.GITHUB_SERVER_URL}/${env.GITHUB_REPOSITORY}/actions/runs/${env.GITHUB_RUN_ID}"}
             ]
+      - uses: ./.github/actions/after-job
+        if: always()
   create-gh-release:
     runs-on: ubuntu-latest
     needs: [promote-to-ga]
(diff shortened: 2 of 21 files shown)


build-aux/docs/intro.md
```diff
@@ -95,11 +92,8 @@ Most (but not all) of the snippets include `common.mk`, which
     See [`testing.md`](./testing.md) for information about writing the
     recipe for `my-test.tap`.
 
-    Because of the special semantics around `$(GOOS)` and `$(GOARCH)`
-    in `bin_OS_ARCH/` directories, `common.mk` goes ahead and has
-    `make clean` run `rm -rf -- bin_*`.  Because of `make check`'s use
-    of `test-suite.tap`, `common.mk` also goes head and has `make
-    clean` run `rm -f test-suite.tap`.
+    Because of `make check`'s use of `test-suite.tap`, `common.mk`
+    also goes head and has `make clean` run `rm -f test-suite.tap`.
```
kflynn, line 96: Nit: should be "goes *a*head".

.github/workflows/promote-ga.yml
```diff
@@ -29,6 +29,7 @@ jobs:
       - name: "make release/promote-oss/to-ga"
         run: |
           make release/promote-oss/to-ga
+      - uses: ./.github/actions/end-job
```
kflynn, line 32: Shouldn't this be `after-job`? or even better `post-job`?

.github/actions/after-job/action.yml
```diff
@@ -0,0 +1,37 @@
+name: 'Post-job actions'
```
kflynn, line 1: So the name is "Post-job actions" -- shouldn't it live in `post-job` rather than `after-job`?

build-aux/generate.mk
```diff
@@ -366,12 +366,14 @@ $(OSS_HOME)/k8s-config/%/helm-expanded.yaml: \
...
 	helm template --namespace=$(helm.namespace.$*) --values=$(@D)/values.yaml $(or $(helm.name.$*),$*) $(boguschart_dir) >$@
-
 $(OSS_HOME)/k8s-config/%/output.yaml: \
   $(OSS_HOME)/k8s-config/%/helm-expanded.yaml \
   $(OSS_HOME)/k8s-config/%/require.yaml \
   $(tools/filter-yaml)
 	$(tools/filter-yaml) $(filter %/helm-expanded.yaml,$^) $(filter %/require.yaml,$^) >$@
+k8s-config.clean:
+	rm -f k8s-config/*/helm-expanded.yaml k8s-config/*/output.yaml
+clean: k8s-config.clean
```
kflynn, line 376: (This pattern where we have a recipe for something that matches `%.clean`, then explicitly have to do `clean: thing-that-matches-%.clean`, kind of bugged me, given ce9a8d9f1c3cd89ac99e9e9179d4ed755c2c2be. But the alternative:

```
clean: $(patsubst %,%.rm-f,$(wildcard k8s-config/*/helm-expanded.yaml))
clean: $(patsubst %,%.rm-f,$(wildcard k8s-config/*/output.yaml))
```

seems like it might well be worse.)

build-aux/main.mk
```diff
@@ -48,6 +55,7 @@ $(foreach img,$(_ocibuild-images),docker/.$(img).docker.stamp): docker/.%.docker
 # that don't need Emissary-specif stuff.
```
kflynn, line 55: Nit: "Emissary-specif*ic*"

build-aux/main.mk
```diff
@@ -171,3 +184,14 @@ endif
...
 	  mkdir -p $(dir $(dst))$(NL)\
 	  sed -e 's/\$$version\$$/$*/g' -e 's/\$$quoteVersion$$/0.4.1/g' <$(src) >$(dst)$(NL)))
+
+#

+# Destructive rules
+
+clobber: clean
+
+clean: build-output.rm-r
```
kflynn, line 193: You know, we have defaults for this in common.mk... I was considering importing that in the `Makefile`.

kflynn (commented):
Looking like a very nice cleanup so far! Shall we add

```maintainer-clean: clobber```

in there somewhere?

File: .github/workflows/promote-ga.yml (hunk 2 above)
Comment (LukeShu) [reply .github/workflows/promote-ga.yml:32]: I didn't like `post` because `post` can be a verb.  But yeah, it shouldn't be `end-job`.

File: build-aux/generate.mk (hunk 4 above)
Comment (LukeShu) [reply build-aux/generate.mk:376]: That's bad because it has to do the wildcard at Makefile-parse-time.

File: build-aux/main.mk (hunk 6 above)
Comment (LukeShu) [reply build-aux/main.mk:193]: builder.mk will get upset with common.mk

File: .github/workflows/promote-ga.yml (hunk 2 above)
Comment (kflynn) [reply .github/workflows/promote-ga.yml:32]: Totally OK with `after-job` if the directory name gets synced up. 🙂

File: build-aux/generate.mk (hunk 4 above)
Comment (kflynn) [reply build-aux/generate.mk:376]: Yup, won't work. And neither does this:

```
clean:
    rm -rf k8s-config/*/helm-expanded.yaml k8s-config/*/output.yaml
```

...so overall I just kinda have to accept this one. 😉

AliceProxy (approved):

kflynn (approved):
Hit it! 🙂 Looking forward to this making life easier.

PR #4241, opened 2022-05-20
[DO NOT MERGE] upgrade to envoy 1.22.2

LanceEa:
## Description
This upgrades Envoy Proxy to the latest release which provides bug fixes, performance enhancements and new features to Emissary Ingress.

## Testing
CI is green and I have been manually smoke testing in multiple test clusters.

Here is the verification that `make check-envoy` ran through the envoy test suite and ensure they are green.

envoy-1 22 2-tests

## Checklist
 - [x] This is unlikely to impact how Ambassador performs at scale. - This should improve the story
 - [x] My change is adequately tested. - _Other than upgrading envoy the fixes were mostly getting build green due to deprecations and changes to how filters are looked up by name vs @type_
 - [x] No need to update `DEVELOPING.md`


python/ambassador_cli/ambassador.py
```diff
@@ -352,7 +336,7 @@ def validate(config_dir_path: str):
     Validate an Ambassador configuration. This is an extension of "config" that
     redirects output to devnull and always exits on error.
 
-    :param config_dir_path: Configuration directory to scan for Ambassador YAML files
+    :param config_dir_path: Configuration directory to scan for Ambassador YAML filesZW
```
AliceProxy, line 339: Why the rename on this line?

python/tests/gold/clientcertificateauthentication/bootstrap-ads.json
```diff
@@ -31,13 +31,8 @@
       {
         "name": "static_layer",
         "static_layer": {
-          "envoy.deprecated_features.allow_deprecated_gzip_http_filter": true,
-          "envoy.deprecated_features:envoy.api.v2.route.HeaderMatcher.regex_match": true,
-          "envoy.deprecated_features:envoy.api.v2.route.RouteMatch.regex": true,
-          "envoy.deprecated_features:envoy.config.trace.v2.ZipkinConfig.HTTP_JSON_V1": true,
```
AliceProxy, line 37: Hmmm, do we need to document the removal of these deprecated features anywhere other than the changelog? Thinking about the gzip and Zipkin ones specifically, but I cant say for sure if that is an issue for users.

python/tests/unit/test_cache.py
```diff
@@ -71,7 +71,7 @@ def apply_yaml_string(self, yaml_data: str, allow_updates=True) -> None:
             name = metadata['name']
             namespace = metadata['namespace']
 
-            key = f"{kind}-v2-{name}-{namespace}"
```
AliceProxy, line 74: Just leaving a comment here as a mental note that my changes to the cache for the regression around cache keys will need to be checked before we repatriate 2.3 to ensure they dont overwrite anything here.

AliceProxy (approved):
This seems safe as far as I can tell. Excited to see this landed.

python/tests/gold/clientcertificateauthentication/bootstrap-ads.json
LanceEa: Great question...let me investigate these further. They were removed by aidan, so I followed suit. I do agree that we should document them.

python/ambassador_cli/ambassador.py
LanceEa: fat fingering it probably and didn't realize it...I will fix that

python/tests/unit/test_cache.py
LanceEa: I will repatriate first and rebase on your changes before officially merging it. That way we have a clean commit. That will also give Luke a chance to give a final review.

LanceEa: I have re-based and pushed so this should be resolved...unless you think otherwise?

AliceProxy (dismissed):

LanceEa:
> nit: [5adc3ea](https://github.com/emissary-ingress/emissary/commit/5adc3eaad9ff375f4fe0b1f9174e32834cfb404d) has a bad commit message subject line

Agree, I wasn't sure of protocol there since it was a cherry-picked commit. I'm happy to amend it if it we feel it is ok to do.

python/tests/gold/clientcertificateauthentication/bootstrap-ads.json
LanceEa: Do we still think we need to call these out beyond the removal V2 which was already added. I talk about it in the migration guide. https://github.com/datawire/getambassador.io/blob/a3cfb2147dea099d3ab61072fdba8943e4a5b41d/ambassador-docs/docs/edge-stack/pre-release/about/changes-3.y.md

Which needs some attention before release.

pkg/envoytest/harness.go
```diff
@@ -68,7 +68,9 @@ func SetupEnvoy(t *testing.T, adsAddress string, portmaps ...string) {
 		for _, pm := range portmaps {
 			args = append(args, "-p", pm)
 		}
-		args = append(args, "--rm", "--entrypoint", "envoy", "docker.io/datawire/aes:1.6.2", "--config-yaml", yaml)
+		// TODO: this should be marked as an end-to-end test after we have built an image
```
LanceEa, line 71: Not sure how to handle this one because we have hardcoded it, on one hand the test was always testing against a really old version...

LukeShu:
- What happens if the user says `protocol_version: v2`?
- What happens if the user omits `protocol_version`?

In both cases, the CRD schemas need to allow this (so that it can coexist with a 2.x installation), but it needs to be rejected and marked as an error.  It would be wrong to simply pretend that the user said `protocol_version: v3`.

I grepping this PR, there are no tests for the `protocol_version: v2` case.  Skimming the diff, I believe that there are few tests for the missing-`protocol_version` case, but that they check that it pretends like the user said `protocol_version: v3` (which is wrong).

pkg/envoytest/harness.go
LukeShu: ffs

LukeShu: https://github.com/emissary-ingress/emissary/pull/4291

LukeShu (changes requested):
- What happens if the user says `protocol_version: v2`?
- What happens if the user omits `protocol_version`?

In both cases, the CRD schemas need to allow this (so that it can coexist with a 2.x installation), but it needs to be rejected and marked as an error.  It would be wrong to simply pretend that the user said `protocol_version: v3`.

I grepping this PR, there are no tests for the `protocol_version: v2` case.  Skimming the diff, I believe that there are few tests for the missing-`protocol_version` case, but that they check that it pretends like the user said `protocol_version: v3` (which is wrong).

LanceEa:
Let's talk about it when you are online. In my opinion the `transport_protocol_version` is a no-op at this time because we only support v3 and I do not see us supporting any other version in the near future. 

So, regardless of what the user has in the CRD, it will be ignored. This is the exact behavior we are doing with the `regex_type` so I think there is precedence here. Now, we could add a log message to to indicate as such for the user as a nice-to-have.

Now, this PR has been sitting here for a few weeks so I had to take a look at it again and I do see in one spot where the no-op is not true so I will need to fix that.

As for test, I think its probably good to add one that verifies the no-op so I will re-review the current tests that are in place to make sure they capture it.

LanceEa:
adjusted to draft for now to make sure its not merged... this PR has been superseded by #4292.

LanceEa:
Closing because this was superseded and already merged into master

PR #4242, opened 2022-05-20
.github: Fix the repatriation logic to deal with merged/closed PRs

LukeShu:
## Description

If there's an old PR that's already been merged, that means we should be creating a new one, rather than assuming that pushing to the branch will update the old one.

As seen on https://github.com/emissary-ingress/emissary/runs/6523998225

## Testing

I used `gh` on the CLI to try out these commands interactively. But I noticed some minor differences in `gh pr view` than what we see in CI, so apparently versions are different; so who knows if this will actually work in CI?

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/.github/workflows/repatriate.yml
+++ b/.github/workflows/repatriate.yml
@@ -49,14 +49,15 @@ jobs:
             git push --set-upstream origin "$merge_branch"
           fi
 
-          if url=$(gh pr view --json=url --jq=.url); then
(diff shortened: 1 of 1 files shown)


kflynn (commented):
Looks good! although it makes me sad that `gh` doesn't output the URL in a clean way.

kflynn (approved):
Whoops, meant to approve instead of comment.

PR #4246, opened 2022-05-25
[v3.0] add downstream support for HTTP/3

LanceEa:
## Description
This PR adds downstream support for HTTP/3 and is able to take advantage of the current developer experience by using the `protocolStack` field of the `Listener` CRD. 

HTTP/2 and HTTP/1 traffic will continue to be served using a TCP Listener configured for HTTP and now a second Listener configured for HTTP/3 can be done by setting the `protocalStack: [ "TLS", "HTTP", "UDP"]` on a Listener and binding to the **same address:port** as the TCP Listener

```yaml
# Listener that leverages TCP and will be used to server HTTP/2 and HTTP/1.1 traffic
kind: Listener
metadata:
  name: emissary-ingress-listener-8443
  namespace: emissary
spec:
  port: 8443
  protocol: HTTPS
  securityModel: XFP
  hostBinding:
    namespace:
      from: ALL
---
# Listener that leverages HTTP & UDP, will be used to serve HTTP/3 traffic
# NOTE: this does not add support for raw UDP traffic
kind: Listener
metadata:
  name: emissary-ingress-listener-udp-8443
  namespace: emissary
spec:
  port: 8443
  # order matters here with UDP required to be the last item
  protocolStack:
    - TLS
    - HTTP
    - UDP
  securityModel: XFP
  hostBinding:
    namespace:
      from: ALL
```

When a TCP Listener binds to the same address and port as the UDP Listener then it will inject the `alt-svc` header into the responses returned on the TCP Listener. This header advertises HTTP/3 support to the client and tells the client how to upgrade itself to http/3.

**NOTE - the alt-svc header is not configurable for the initial release but in the future we should make it available to the user**

> More info here on the header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Alt-Svc.

## Related Issues
None 

## Testing
- additional unit tests were added to capture http/3 behavior
- Existing Kat tests were updated to handle and changes introduced

Deployed to multiple clusters and tested setting up in AKS, EKS and GKE.

## Checklist

_It is a new feature and most of the heavy lifting is done in Envoy. Also, fallback to http/2 and http/1.1 is still possible. One potential thing we will need to keep an eye is that the routes are duplicated on both the TCP and UDP listeners. The UDP listener was able to exclude the http (non-tls) redirect routes due to QUIC requiring TLS so its not an exact duplication._
 - [x] I updated `DEVELOPING.md` 

### DIFF
--- a/python/ambassador/envoy/v3/v3listener.py
+++ b/python/ambassador/envoy/v3/v3listener.py
@@ -124,11 +124,13 @@ class V3Listener(dict):
         super().__init__()
 
         self.config = config
+        self.http3_enabled = irlistener.http3_enabled
+        self.socket_protocol = irlistener.socket_protocol
         self.bind_address = irlistener.bind_address
         self.port = irlistener.port
-        self.bind_to = f"{self.bind_address}-{self.port}"
+        self.bind_to = irlistener.bind_to()
 
-        bindstr = f"-{self.bind_address}" if (self.bind_address != "0.0.0.0") else ""
+        bindstr = f"-{irlistener.socket_protocol.lower()}-{self.bind_address}" if (self.bind_address != "0.0.0.0") else ""
         self.name = irlistener.name or f"ambassador-listener{bindstr}-{self.port}"
 
         self.use_proxy_proto = False
@@ -177,9 +179,13 @@ class V3Listener(dict):
                 # listener is OK with TLS-y things like a termination context, SNI,
                 # etc.
                 self._tls_ok = True
-                self.listener_filters.append({
-                    'name': 'envoy.filters.listener.tls_inspector'
-                })
+
+                ## When UDP we assume it is http/3 listener and configured for quic which has TLS built into the protocol
+                ## therefore, we only need to add this when socket_protocol is TCP
+                if self.isProtocolTCP():
+                    self.listener_filters.append({
+                        'name': 'envoy.filters.listener.tls_inspector'
+                    })
 
             if proto == "TCP":
                 # TCP doesn't require any specific listener filters, but it
@@ -193,7 +199,6 @@ class V3Listener(dict):
                     # ...and make sure the group in question wants the same bind
                     # address that we do.
                     if irgroup.bind_to() != self.bind_to:
-                        # self.config.ir.logger.debug("V3Listener %s: skip TCPMappingGroup on %s", self.bind_to, irgroup.bind_to())
                         continue
 
                     self.add_tcp_group(irgroup)
@@ -384,6 +389,12 @@ class V3Listener(dict):
             'normalize_path': True
         }
 
+        # Instructs the HTTP Connection Mananger to support http/3. This is required for both TCP and UDP Listeners
+        if self.http3_enabled:
+            base_http_config['http3_protocol_options'] = {}
+            if self.isProtocolUDP():
+                base_http_config['codec_type'] = "HTTP3"
+
         # Assemble base HTTP filters
         from .v3httpfilter import V3HTTPFilter
         for f in self.config.ir.filters:
@@ -551,7 +562,7 @@ class V3Listener(dict):
             "socket_address": {
                 "address": self.bind_address,
                 "port_value": self.port,
-                "protocol": "TCP"
+                "protocol": self.socket_protocol ## "TCP" or "UDP"
             }
         }
 
@@ -837,6 +848,12 @@ class V3Listener(dict):
 
             filter_chain: Optional[Dict[str, Any]] = None
 
+            # http/3 is built on quic which has TLS built-in. This means that our UDP Listener will only ever need routes
+            # that match the TLS Filter chain and will diverge from the TCP listener in that it will not support redirect
+            # therefore, we can exclude duplicating the filterchain and routes so hitting this endpoint using non-tls http will fail
+            if (chain.type == "http") and self.isProtocolUDP() and self.http3_enabled:
+                continue
+
             if chain.type == "http":
                 # All HTTP chains get collapsed into one here, using domains to separate them.
                 # This works because we don't need to offer TLS certs (we can't anyway), and
@@ -885,8 +902,8 @@ class V3Listener(dict):
                 if (len(chain_hosts) > 0) and ("*" not in chain_hosts):
                     filter_chain_match['server_names'] = chain_hosts
 
-                # Likewise, an HTTPS chain will ask for TLS.
-                filter_chain_match["transport_protocol"] = "tls"
+                # Likewise, an HTTPS chain will ask for TLS or QUIC (when udp)
+                filter_chain_match["transport_protocol"] = "quic" if self.isProtocolUDP() and self.http3_enabled else "tls"
 
                 if chain.context:
                     # ...uh. How could we not have a context if we're doing TLS?
@@ -894,13 +911,33 @@ class V3Listener(dict):
                     # filter_chain_match.
                     envoy_ctx = V3TLSContext(chain.context)
 
-                    filter_chain['transport_socket'] = {
+                    envoy_tls_config = {
                         'name': 'envoy.transport_sockets.tls',
                         'typed_config': {
                             '@type': 'type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.DownstreamTlsContext',
                             **envoy_ctx
                         }
                     }
+
+                    if self.isProtocolUDP():
+                        envoy_tls_config = {
+                            'name': 'envoy.transport_sockets.quic',
+                            'typed_config': {
+                                '@type': 'type.googleapis.com/envoy.extensions.transport_sockets.quic.v3.QuicDownstreamTransport',
+                                'downstream_tls_context':{
+                                    **envoy_ctx
+                                }
+                            }
+                        }
+
+                    filter_chain['transport_socket'] = envoy_tls_config
+
+                else:
+                    # Envoy doesn't like having a UDP Listener with QUIC filter chain that doesn't have a "transport_socket" set with a TLS
+                    # certificate. If the fall-back cert is removed or no certificate is provided then we should not add the quic filter chain
+                    if self.isProtocolUDP() and self.http3_enabled:
+                        self._irlistener.logger.warn(f"Listener: quic network protocol requires a TLSContext to be provided, no tls context found for Listener: {self._irlistener.bind_to()}")
+                        continue
 
                 # Finally, stash the match in the chain...
                 filter_chain["filter_chain_match"] = filter_chain_match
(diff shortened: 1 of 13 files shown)


LanceEa:
@AliceProxy @LukeShu  - This PR is based off from #4241 so you can ignore the first commits because they are being reviewed as part of that PR.

You can start your review at commit: 2c1bcd3af3d6d94df34bb03f39c798309ebef725

I will rebase on the upgrade once merged, update the `changelog` and squash and commits that are unnecessary once you have had a chance to review and we feel its in a good place.

pkg/api/getambassador.io/v3alpha1/crd_listener.go
```diff
@@ -132,6 +132,13 @@ type HostBindingType struct {
 	Selector  *metav1.LabelSelector `json:"selector,omitempty"`
 }
 
+//HTTP3Options provides options to set when a Listener is enabled to understand HTTP/3
+type HTTP3Options struct {
+	//AltSvc allows modifying the default `alt-svc` header that is attached to all HTTP/1 & HTTP/2 response indicating HTTP/3 support.
+	// +kubebuilder:default="h3=\":443\"; ma=86400, h3-29=\":443\"; ma=86400"
+	AltSvc string `json:"altSvc,omitempty"`
+}
```
LukeShu, line 140: 1. I don't like exposing this as a raw string. This feels like something where we really want to "own". That we want the user to just tell us what port H/3 is on, and maybe max age. That we should take on the "expert-knowledge" of forming that in to a valid `Alt-Svc` header. And The user really shouldn't be configuring the list of protocol IDs themselves; what's correct there just depends on what our Envoy supports, and so that's something we should be responsible for tracking, not shifting that on to the user.

2. If we are going to add a raw-string like this, we should probably have the doc comment include a link to the relevant spec or docs to understand the semantics of that header.

3. style nits: please add a space between `//` and the start of the text, and please add a blank `//` line between the main text and the `+kubebuilder` line.

pkg/api/getambassador.io/v3alpha1/crd_listener.go
```diff
@@ -178,6 +190,7 @@ type ListenerSpec struct {
 // +kubebuilder:printcolumn:name="StatsPrefix",type=string,JSONPath=`.spec.statsPrefix`
 // +kubebuilder:printcolumn:name="Security",type=string,JSONPath=`.spec.securityModel`
 // +kubebuilder:printcolumn:name="L7Depth",type=string,JSONPath=`.spec.l7Depth`
+// +kubebuilder:printcolumn:name="HTTP3Options",type=string,JSONPath=`.spec.http3Options`
```
LukeShu, line 193: 1. Given the types in this PR, that syntax is wrong I think? I think it'd need to be `JSONPath=`.spec.http3Options.altSvc` in order to have `type=string`.

2. I'm thinking that's an awful lot to include in the default `kubectl get listeners` table. I do think the H/3 port would be good to include though. Assuming you take my suggestion of splitting `http3Options.altSvc` in to structured fields, I think it would be good to rename the "Port" column and then add an "H3 Port" next to it.
```go
// +kubebuilder:printcolumn:name="TCP/H1/H2 Port",type=string,JSONPath=`.spec.port`
// +kubebuilder:printcolumn:name="H3 Port",type=string,JSONPath=`.spec.http3Options.port`
```

python/ambassador/envoy/v3/v3listener.py
```diff
@@ -853,6 +866,12 @@ def finalize_http(self) -> None:
 
             filter_chain: Optional[Dict[str, Any]] = None
 
+            # http/3 is built on quic which has TLS built-in. This means that our UDP Listener will only ever need routes
+            # that match the TLS Filter chain and will diverge from the TCP listener in that it will not support redirect
+            # therefore, we can exclude duplicating the filterchain and routes so hitting this endpoint using non-tls http will fail
+            if (chain.type == "http") & self.isProtocolUDP() & self.http3_enabled:
```
LukeShu, line 872: `&` is bitwise-and; `and` is boolean-and. This line of code seems to do the right thing, but I'm honestly shocked that it does.

python/ambassador/envoy/v3/v3listener.py
```diff
@@ -990,13 +1045,25 @@ def finalize_http(self) -> None:
             self._filter_chains.append(filter_chain)
 
     def as_dict(self) -> dict:
-        listener = {
+        listener: dict = {
```
LukeShu, line 1048: This is a weird type annotation to add. It can plainly see that it's a dict without an annotation; the reason you'd add the annotation is to tell it the types of the keys and values in the dict; something like `Dict[str, Any]` (or preferably something more specific than `Any`, but looking at it it seems to me that there are at least 4 different types, so... that `Union[…]` would be a little large).

python/ambassador/ir/irlistener.py
```diff
@@ -21,20 +21,25 @@ class IRListener (IRResource):
     IRListener is a pretty direct translation of the Ambassador Listener resource.
     """
 
+    socket_protocol: str # tcp or udp
```
LukeShu, line 24: For enum variables, instead of typing it as `: str`, they can be typed as `: Literal["tcp", "udp"]` (given `from typing import Literal`).

python/ambassador/envoy/v3/v3listener.py
```diff
@@ -990,13 +1045,25 @@ def finalize_http(self) -> None:
...
             "name": self.name,
             "address": self.address,
+            "udp_listener_config": {},
             "filter_chains": self._filter_chains,
             "traffic_direction": self.traffic_direction
         }
 
+        if self.reuse_port == True:
+            listener['enable_reuse_port'] = self.reuse_port
```
LukeShu, line 1057: It looks like you're setting this if both TCP and UDP are set to the same port number.  That isn't necessary; TCP port numbers and UDP port numbers are totally separate in the Linux kernel; no need to set SO_REUSEPORT to listen on both TCP port X and UDP port X.

(Just to be sure, I verified this: I launched `socat TCP-LISTEN:8080,fork STDIO` in one terminal, `socat UDP-LISTEN:8080,fork STDIO` in another, and verified in a 3rd terminal that I could talk to each.)

LukeShu: See Envoy's docs on what SO_REUSEPORT does on different platforms: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#envoy-v3-api-field-config-listener-v3-listener-enable-reuse-port

Also, Envoy defaults it to "true", so the whole thing's double-pointless, I think.

python/ambassador/envoy/v3/v3listener.py
LanceEa: my lack of python knowledge biting me. I will get those fixed

python/ambassador/envoy/v3/v3listener.py
LanceEa: interesting...I see the docs say that and agree. I will double check but I thought the reason I added it explicitly was because it was crashing...I will try it with out and see.

pkg/api/getambassador.io/v3alpha1/crd_listener.go
LanceEa: valid feedback!

This has been removed per our discussion offline and future enhancements will take into account your feedback.

python/ambassador/envoy/v3/v3listener.py
LanceEa: I have gone ahead and updated to use `and` instead so we should be all good here.

python/ambassador/envoy/v3/v3listener.py
LanceEa: So, I tested and it worked so not sure where I got the idea that it needed to be set to make sure it worked. 🤷  I have removed it, built a test image and it came up with no problems in a cluster so we should be good without explicitly setting it.

AliceProxy (approved):

LukeShu:
```
> HTTP/2 and HTTP/1 traffic will continue to be served using a TCP Listener configured for HTTP and now a second Listener configured for HTTP/3 can be done by setting the `protocalStack: [ "TLS", "HTTP", "UDP"]` on a Listener and binding to the **same address:port** as the TCP Listener
```
That protocol stack doesn't make sense to me. Shouldn't that be `protocolStack: ["HTTP", "TLS", "UDP"]` (or better: `protocolStack: ["HTTP", "QUIC", "UDP"]`)?

Edit: Nope, `protocolStack` makes even less damn sense than I thought. Which was already pretty low.

LukeShu:
I'm not sure how I feel about magically associating 2 listeners together if they have matching `addr:port`.  In particular, the rules about how Hosts get attached to Listeners (and then Mappings to Hosts) is complex and I'm worried about locking ourselves in to funny behavior if the 2 Listeners disagree about which Hosts should be bound to them.

LukeShu:
PS: e2e tests would be good too, since this is a very integration-y problem space.

python/ambassador/envoy/v3/v3listener.py
LukeShu: Not a blocker, it's just bizarre.

LukeShu:
On the other hand, I do think that I like having a 2nd listener better than packing 2 ports in to 1 listener.

LukeShu:
You know what, I'm prepared to call any weird listener-host binding behavior at this point a bug. So fine for v0, and we can break it later because it was always a bug.

python/ambassador/ir/ir.py
```diff
@@ -271,8 +271,6 @@ def __init__(self, aconf: Config,
         self.groups = {}
         self.grpc_services = {}
         self.hosts = {}
-        # self.invalidate_groups_for is handled above.
-        # self.k8s_status_updates is handled below.
```
LukeShu, line 275: Why?

python/ambassador/envoy/v3/v3listener.py
```diff
@@ -974,13 +1025,22 @@ def finalize_http(self) -> None:
...
         }
 
+        if self.isProtocolUDP():
+            listener['udp_listener_config'] = {
+                'quic_options': {},
+                'downstream_socket_config': { 'prefer_gro': True }
+            }
+        else:
+            del(listener['udp_listener_config'])
+
```
LukeShu, line 1043: I'm not sure what's to like with this pattern of "initialize with a bogus value, then either overwrite with a valid value or remove it".  Why not drop the initial value and drop the `else` case?

LanceEa: agree that is cleaner

AliceProxy (approved):

PR #4247, opened 2022-05-25
Allow Host and TLSContext to configure a CRL (v2.3)

alexgervais:
Signed-off-by: alex <alex@datawire.io>

## Description
Allow Host and TLSContext to configure a CRL through the `crl_secret` property on both resources.

## Related Issues
https://github.com/datawire/apro/issues/2619

## Testing
Manual tests and automated KAT tests for the Host resource.

## Checklist

    + Any new features
    + Any deprecations


AliceProxy (approved):
Thanks for making this change! Looks good to me, but do we maybe want to change `crl_secret` to something like:
```
crl:
  secretName: string
```
Instead so that way we are leaving ourselves room for additional properties/fields should we add additional fields around the CRLs? I know Envoy exposes multiple options for handling them, so I could see us wanting to add more CRL fields to the CRDs in the future. @LanceEa Thoughts?

python/ambassador/ir/ir.py
```diff
@@ -729,13 +729,17 @@ def save_secret_info(self, aconf):
...
+            # should not generate errors.)
+            if aconf_secret.get('tls_crt') or aconf_secret.get('cert-chain_pem') or aconf_secret.get('user_key') or aconf_secret.get('crl_pem'):
                 secret_info = SecretInfo.from_aconf_secret(aconf_secret)
                 secret_name = secret_info.name
                 secret_namespace = secret_info.namespace
 
                 self.logger.debug('saving "%s.%s" (from %s) in secret_info', secret_name, secret_namespace, secret_key)
                 self.secret_info[f'{secret_name}.{secret_namespace}'] = secret_info
+            else:
+                self.logger.debug('not saving secret_info from %s because there is no public half', secret_key)
```
LanceEa, line 742: 1. should this be Info or Warning? I see the  previous line above is debug but I would think the fact that this doesn't exist would want to be seen by a developer rather than hidden until they turned on debug logging?

2. Is `secret_key` something we should be print to logs? Again, I see it was on the previous log line but just want to make sure there is no security concern there?

python/tests/kat/t_tls.py
```diff
@@ -273,6 +274,87 @@ def requirements(self):
             yield (r[0], query)
 
 
+class ClientCertificateAuthenticationContextCRL(AmbassadorTest):
+
+    def init(self):
+        self.xfail = "FIXME: IHA"  # This test should cover TLSContext with a crl_secret
```
LanceEa, line 280: For my understanding, what is this line saying? Is it disabling the test? Does this need be fixed now or in the future?

LanceEa (commented):
Overall, it looks pretty good. Just a few questions and comments.

LanceEa:
> Thanks for making this change! Looks good to me, but do we maybe want to change `crl_secret` to something like:
> 
> ```
> crl:
>   secretName: string
> ```
> 
> Instead so that way we are leaving ourselves room for additional properties/fields should we add additional fields around the CRLs? I know Envoy exposes multiple options for handling them, so I could see us wanting to add more CRL fields to the CRDs in the future. @LanceEa Thoughts?

Good call @AliceProxy, I would agree that we should set ourselves up for future enhancements. In general, we need to start planning out our road to a proper CRD V3.  

I would also say at a minimum we should be consistent with the casings of our fields. Based on our docs it should be camelCasing, correct?

alexgervais:
I agree with you @AliceProxy and @LanceEa. `tls.crl.secretName` would be desirable both in terms of casing and consistency with how CRDs should be defined. We should aim for this in v3. However, I took a look at the other existing TLS fields:
```
tls:
  cert_chain_file:    # string
  private_key_file:   # string
  ca_secret:          # string
  cacert_chain_file:  # string
  alpn_protocols:     # string
  cert_required:      # bool
  min_tls_version:    # string
  max_tls_version:    # string
  cipher_suites:      # array of strings
  ecdh_curves:        # array of strings
  sni:                # string
```

So I went for consistency in the way current resources are defined... snake case and `_secret` suffix.
Does it make sense?

python/ambassador/ir/ir.py
alexgervais: Good questions!
1. I went with `debug` since this line did not exist before and missing secrets are reported later in code execution. It was useful while debugging KAT tests (which run in debug).
2. I believe the `secret_key` in this context is the unique Key in the secrets dict. It references the secret name, not the potential sensitive secret private key.

AliceProxy:
@alexgervais @LanceEa 

> I would also say at a minimum we should be consistent with the casings of our fields. Based on our docs it should be camelCasing, correct?

So one of our goals for the V3 CRDs is to move everything to camelCase since we currently have a mix of snake_case and camelCase all over the CRDs and camelCase is more "kubelike". Having it be snake_case for this PR not a big problem because with the V3 CRDs we can convert it to camelCase.

If we want to go with snake_case for this version, then I still think it would be beneficial to do something like:
```
crls:
   secret_name: string
```

because if we add new fields in the V3 CRDs, then we have to support those fields in the older CRD versions for round tripping conversion and that would make it simpler to add values into the object instead of having a V3 CRD that looks like:
```
  crls:
    secretName: string
    newField1: string
   newField2: string
```

and then have the v3alpha1/v2 CRDs look like this after conversion if we merge this as `crl_secret: string`:
```
  crl_secret: string
  new_field_1:  string
  new_field_2: string
```

Does that reasoning make sense?

alexgervais:
Thanks for all your thoughts. I'm very much aligned with the CRD design for v3.
As for the current version, I still think we should keep the `tls` config block flat, as other keys also reference either a `_secret` or `_file`.
I am slightly afraid people would think of using the following example as an array of CRLs, even tho only 1 entry can be specified (per CA/Host/TLSContext).
```
crls:
   secret_name: string
```

Seeing how things are handled in the python code, I saw examples where we would take the CA chain from either a `_secret` or `_file` reference. If we need to extend the CRL to file, we could reuse that logic.

python/tests/kat/t_tls.py
LanceEa: ok thanks...yeah, we need to dig into our testing and make sure we unblock things to ensure we have good test coverage. I will add it to the list 😄 .

LanceEa:
@AliceProxy - I'm ok with @alexgervais  reasoning but I haven't felt all the pains the team might have in the past so If you feel strong about this then try to convince him otherwise 😄 . If not let's move forward and get this merged so we can focus on release activities.

LanceEa (approved):

PR #4250, opened 2022-05-27
Cache: fix an issue where the cache key for mappings is wrong when mu…

AliceProxy:
…ltiple mappings have the same name

## Description

## Related Issues
List related issues.

## Testing

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/python/tests/unit/test_cache.py
+++ b/python/tests/unit/test_cache.py
@@ -30,7 +30,7 @@ class Builder:
     def __init__(self, logger: logging.Logger, yaml_file: str,
                  enable_cache=True) -> None:
         self.logger = logger
-        
+
         self.test_dir = os.path.join(
             os.path.dirname(os.path.abspath(__file__)),
             "test_cache_data"
@@ -149,7 +149,7 @@ class Builder:
 
             if kind not in watt['Kubernetes']:
                 watt['Kubernetes'][kind] = []
-            
+
             watt['Kubernetes'][kind].append(rsrc)
 
         watt_json = json.dumps(watt, sort_keys=True, indent=4)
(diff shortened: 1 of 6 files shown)


python/tests/unit/test_cache.py
```diff
@@ -465,6 +465,54 @@ def test_long_cluster_1():
...
+    builder = Builder(logger, "cache_test_4.yaml")
+    b = builder.build()
+    econf = b[1]
+    econf = econf.as_dict()
+
+    # loop through all the clusters in the resulting envoy config and pick out two Mappings from our test set (first and lase)
+    # to ensure their clusters were generated properly.
+    cluster1_ok, cluster2_ok = False
+    for cluster in econf['static_resources']['clusters']:
+        cname = cluster.get('per_connection_buffer_limit_bytes', None)
```
LanceEa, line 480: Why `per_connection_buffer_limit_bytes`?

python/tests/unit/test_cache.py
```diff
@@ -465,6 +465,54 @@ def test_long_cluster_1():
...
+    # Update the yaml for these Mappings to simulate a reconfiguration
+    # We should properly remove the cache entries for these clusters when that happens.
+    builder.apply_yaml("cache_test_5.yaml")
+    b = builder.build()
+    econf = b[1]
+    econf = econf.as_dict()
+
+    cluster1_ok, cluster2_ok = False
+    for cluster in econf['static_resources']['clusters']:
+        cname = cluster.get('per_connection_buffer_limit_bytes', None)
```
LanceEa, line 501: same as above?? not sure I understand this part.

python/tests/unit/test_cache.py
```diff
@@ -465,6 +465,54 @@ def test_long_cluster_1():
...
 
+def test_mappings_same_name_delta():
+    # Tests that multiple Mappings with the same name (but in different namespaces)
+    # are properly added/removed from the cache when they are updated.
+    builder = Builder(logger, "cache_test_4.yaml")
+    b = builder.build()
+    econf = b[1]
+    econf = econf.as_dict()
+
+    # loop through all the clusters in the resulting envoy config and pick out two Mappings from our test set (first and lase)
```
LanceEa, line 476: Do we have guarantee's that our clusters are generated in consistent order? Clusters is a list so I would assume it would be just thought I would ask the question..

LanceEa (commented):
Everything looks good...two quick questions on the test to make sure I understand it.

python/tests/unit/test_cache.py
AliceProxy: Whoops, good catch. That should just be the `name`. I was hasty in copying a section I wrote in the buffer limit bytes test that I happened to know looped over the envoy config. The fact that this did not fail revealed more disturbingly that none of these cache tests actually run :woman_facepalming:.

python/tests/unit/test_cache.py
AliceProxy: same answer as above

python/tests/unit/test_cache.py
AliceProxy: I'm "pretty sure" that the order of clusters is deterministic. That said, for test it doesn't really matter if they are or not. Its fine if there are other clusters, we just assert that they each have a name since that is a blatant failure regardless.

LanceEa (approved):
lgtm and CI is green so go for it. Great work identifying and getting this fixed!

PR #4251, opened 2022-05-28
Alicewasko/reconcile filter secrets

AliceProxy:
## Description

## Related Issues
List related issues.

## Testing

## Checklist

    + Any new features
    + Any deprecations


cmd/entrypoint/secrets.go
```diff
@@ -281,6 +282,21 @@ func ReconcileSecrets(ctx context.Context, sh *SnapshotHolder) error {
 		// ...and for Edge Stack, we _always_ have an implicit reference to the
 		// license secret.
 		secretRef(GetLicenseSecretNamespace(), GetLicenseSecretName(), false, action)
+		// We also want to grab any secrets referenced by edge-stack filters for use in edge-stack
+		// the Filters are unstructured because emissary does not have their type definition
+		for _, f := range sh.k8sSnapshot.Filters {
+			secretName, secretNamespace := findFilterSecret(f)
+			if secretName != "" {
+				if secretNamespace != "" {
```
LanceEa, line 290: I would move this statement into the `findFilterSecret` func since you are already passing in the Filter and you are already checking whether the namespace exist there. This way you can contain the parsing logic to a single function that can be easily unit tested.

```golang
if secreteName != nil && secretNames != nil { 
    secretRef(secretNamespace, secretName, false, action)
}
```

cmd/entrypoint/secrets.go
```diff
@@ -313,6 +329,43 @@ func ReconcileSecrets(ctx context.Context, sh *SnapshotHolder) error {
...
+	// Just making extra sure this is actually a Filter
+	if filter.GetKind() != "Filter" && filter.GetKind() != "Filters" {
+		return "", ""
+	}
+	// Only Oauth2 Filters have secrets, although they dont need to have them.
+	// Yes this is disgusting. It is what it is...
+	filterContents := filter.UnstructuredContent()
+	filterSpec := filterContents["spec"]
+	if filterSpec != nil {
+		oauthFilter := filterSpec.(map[string]interface{})["OAuth2"]
```
LanceEa, line 344: my only question here, is there a risk that we panic due to type assertion failing? Do we have guarantees that the unstructured content within the Spec field is already valid "yaml" thus ensuring this doesn't panic.

I don't have an answer but just want to make sure we think about that.

cmd/entrypoint/secrets.go
```diff
@@ -17,6 +18,7 @@ import (
 	"github.com/datawire/dlib/derror"
 	"github.com/datawire/dlib/dlog"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
```
LukeShu, line 21: Style: Please order the imports in groups of
```go
import (
    "stdlib"
    
    "3rd-party"
    
    "github.com/datawire/*"
)
```
(I see that the `"k8s.io/api/core/v1" there is already breaking this.)
(We really should turn that linter back on.)

cmd/entrypoint/secrets_test.go
```diff
@@ -0,0 +1,156 @@
+package entrypoint
+
+import (
+	"testing"
+
+	snapshotTypes "github.com/datawire/ambassador/v2/pkg/snapshot/v1"
+	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
```
LukeShu, line 8: likewise

cmd/entrypoint/testutil_fake_filtersecrets_test.go
```diff
@@ -0,0 +1,117 @@
+package entrypoint
+
+import (
+	"testing"
+
+	"github.com/datawire/ambassador/v2/pkg/snapshot/v1"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
LukeShu, line 8: likewise

cmd/entrypoint/secrets.go
```diff
@@ -281,6 +283,14 @@ func ReconcileSecrets(ctx context.Context, sh *SnapshotHolder) error {
 		// ...and for Edge Stack, we _always_ have an implicit reference to the
 		// license secret.
 		secretRef(GetLicenseSecretNamespace(), GetLicenseSecretName(), false, action)
+		// We also want to grab any secrets referenced by edge-stack filters for use in edge-stack
```
LukeShu, line 286: Style: `Edge-Stack` should be capitalized.

cmd/entrypoint/secrets.go
```diff
@@ -281,6 +283,14 @@ func ReconcileSecrets(ctx context.Context, sh *SnapshotHolder) error {
...
+		// the Filters are unstructured because emissary does not have their type definition
```
LukeShu, line 287: As should `Emissary`.

cmd/entrypoint/secrets.go
```diff
@@ -313,6 +323,59 @@ func ReconcileSecrets(ctx context.Context, sh *SnapshotHolder) error {
 	return nil
 }
 
+// Returns secretName, secretNamespace from a provided (unstructured) filter if it contains a secret
+// Returns empty strings when the secret name and/or namespace could not be found
+func findFilterSecret(filter *unstructured.Unstructured, action func(snapshotTypes.SecretRef)) error {
+	// Just making extra sure this is actually a Filter
+	if filter.GetKind() != "Filter" && filter.GetKind() != "Filters" {
+		return errors.New("Non-Filter object in Snapshot.Filters")
```
LukeShu, line 331: 1. It would be good to use `fmt.Errorf` in order to include the kind in the error message
2. error messages start with a lower-case letter
3. why would it ever be `"Filters"` instead of `"Filter"`?

cmd/entrypoint/secrets.go
```diff
@@ -313,6 +323,59 @@ func ReconcileSecrets(ctx context.Context, sh *SnapshotHolder) error {
...
+	}
+	// Only Oauth2 Filters have secrets, although they dont need to have them.
```
LukeShu, line 333: Style: the `a` in `OAuth` is capitalized

cmd/entrypoint/secrets.go
```diff
@@ -313,6 +323,59 @@ func ReconcileSecrets(ctx context.Context, sh *SnapshotHolder) error {
...
+		if oauthFilter != nil {
+			secretName, secretNamespace := "", ""
+			// Check if we have a secretName
+			mapOauth, ok := oauthFilter.(map[string]interface{})
+			if !ok {
+				return errors.New("Filter object detected with a bogus \"OAuth2\" field")
+			}
+			oauthSecretName := mapOauth["secretName"]
+			// This is a weird check, but we have to handle the case where secretName is not provided, and when its explicitly set to ""
+			if sName, ok := oauthSecretName.(string); ok && sName != "" {
```
LukeShu, line 355: tip: It should be safe to access the map and check the type in one go
```go
if secretName, isString := mapOAuth["secretName"].(string); isString && secretName != "" {
```

cmd/entrypoint/secrets.go
```diff
@@ -313,6 +323,59 @@ func ReconcileSecrets(ctx context.Context, sh *SnapshotHolder) error {
...
+				secretName = sName
+			} else {
+				if !ok && oauthSecretName != nil {
+					return errors.New("Filter object detected with a bogus \"secretName\" field")
+				}
```
LukeShu, line 360: IMO having a specific error here is making the code hard to read for little benefit. The apiserver should catch+report this. apiext should catch+report this. amb-sidecar should catch+report this. No need to report it again here.

cmd/entrypoint/secrets_test.go
```diff
@@ -0,0 +1,156 @@
...
+	type subtest struct {
+		inputSpec    map[string]interface{}
+		expectedRefs []snapshotTypes.SecretRef
+		expectedErr  string
+	}
+	// Iterate each test for all our API Versions
+	apiVersions := []string{"getambassador.io/v2", "getambassador.io/v3alpha1"}
+	subtests := map[string]subtest{
+		"namespaced": {
+			map[string]interface{}{
```
LukeShu, line 52: Please use `{key1: val1, key2: val2}` r(use keys) rather than `{val1, val2}` (don't use positions) when declaring a struct literal.

cmd/entrypoint/secrets_test.go
```diff
@@ -0,0 +1,156 @@
...
+			},
+			[]snapshotTypes.SecretRef{},
+			``,
+		},
+	}
+
+	for _, apiVersion := range apiVersions {
+		for name, subtest := range subtests {
+			t.Run(name, func(t *testing.T) {
+				t.Parallel()
```
LukeShu, line 124: Hafta stick a `subtest := subtest // capture loop variable` between the `for` and the `t.Run`.

Otherwise the `func(t *testing.T)` will get a reference to the loop's variable, which will change the next iteration through, and so by the time the new goroutines get scheduled the subtest workers will probably all just run the same testcase.

One of the biggest foot-guns in Go, IMO.

cmd/entrypoint/testutil_fake_filtersecrets_test.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func TestFindFilterNamespacedSecret(t *testing.T) {
+	subTests := []subtest{
+		{"TestFindFilterNamespacedSecret V2", "getambassador.io/v2"},
+		{"TestFindFilterNamespacedSecret V3alpha1", "getambassador.io/v3alpha1"},
+	}
+
+	for _, subTest := range subTests {
+		t.Run(subTest.name, func(t *testing.T) {
```
LukeShu, line 35: likewise about needing to capture the loop variable

cmd/entrypoint/testutil_fake_filtersecrets_test.go
```diff
@@ -0,0 +1,117 @@
+package entrypoint
```
LukeShu, line 1: I'm confused why this file is named `testutil_fake_filtersecrets_test.go` instead of just putting it in `secrets_test.go`.

LukeShu (changes requested):
OK, so I left a bunch of super nit-picky things that you can ignore if you really want to.

But the one change I'm putting my foot down on is needing to capture the loop variable in the test.

cmd/entrypoint/secrets_test.go
LanceEa: @LukeShu - Agree on the foot gun! 

Just for curiosity, Is it an issue in this scenario though? I thought it was more to do with referencing pointers which I guess in this case we are iterating a map and grabbing the struct out...golang is probably not copying the struct ?? thus requiring the capture?

LanceEa (approved):
@AliceProxy - lgtm once you address feedback from Luke.

cmd/entrypoint/secrets.go
AliceProxy: As long as there is a defined style I can stick to :slightly_smiling_face:  :+1:

cmd/entrypoint/secrets_test.go
AliceProxy: Ahhh interesting, I would not have guessed that at all :skull:

LukeShu (approved):
Thanks!

LukeShu (changes requested):
It appears I gave you bad advice
```
2022-06-02T17:34:04.2938749Z --- FAIL: TestFindFilterSecrets (0.00s)
2022-06-02T17:34:04.2942841Z     --- FAIL: TestFindFilterSecrets/noSecret (0.00s)
2022-06-02T17:34:04.2943392Z panic: interface conversion: interface {} is nil, not string [recovered]
2022-06-02T17:34:04.2943816Z 	panic: interface conversion: interface {} is nil, not string
2022-06-02T17:34:04.2944016Z 
```

LukeShu (approved):
It's green, hit it!

PR #4257, opened 2022-06-01
[v3.0] Remove all xDS v2 support

LukeShu:
## Description
This is a cleaned-up version of Ava's f175de7b99d697746a649ee9af184d4f460a259b which @LanceEa incorporated in to #4241. It is cleaned up in that
 - It spells out the implications of this in the changelog
 - It prunes a bunch of now-dead weight with passing the `envoy_api_version` around. And with now-dead `regex_type`.
 - It leaves in the `AMBASSADOR_ENVOY_API_VERSION=V2` tests, since those affect the default zipkin API version.
 
There may still be more things that need to be pruned for newer Envoy (esp. transport protocols); but the implications of those things need to be spelled out separately in another commit.

As usual for me, I suggest a commit-by-commit review.  There are 2 commits:
 - The first is the minimal "remove the stuff and get it passing" commit.  It just deletes the `v2` directory and does the minimum necessary to get CI passing with that. I did indeed check that CI passes; though you won't see that on the commit status because I noticed a typo in a comment and so fixed that :)
 - The second commit prunes a bunch of dead weight and cleans things up now that we're not wrangling 2 xDS versions.

## Testing
CI should mostly cover it. I'm not actually sure that we test how the zipkin api version defaults.

## Checklist

    + Any new features
    + Any deprecations

 - [x] This is unlikely to impact how Ambassador performs at scale. - if anything, things will improve

 - [ ] My change is adequately tested. - mostly, but honestly I'm nervous about zipkin and tracing; see above

 - [ ] The changes in this PR have been reviewed for security concerns and adherence to security best practices. - what does this even mean?


CHANGELOG.md
```diff
@@ -105,6 +84,32 @@ it will be removed; but as it won't be user-visible this isn't considered a brea
...
+  **No `regex_type: unsafe`**: The `regex_type` field will be removed from the `ambassador`
+  `Module`, meaning that it will not be possible to instruct Envoy to use the <a
+  href="https://en.cppreference.com/w/cpp/regex/ecmascript">ECMAScript Regex</a> engine rather than
+  the default <a href="https://github.com/google/re2">RE2</a> engine.
+  Users who rely on the specific
+  ECMAScript Regex syntax will need to rewrite their regular expressions with RE2 syntax before
+  upgrading to Emissary-ingress 3.0.0.
+  Note that the `AMBASSADOR_ENVOY_API_VERSION` environment
+  variable is now a misnomer, as it no longer configures which xDS API version is used, but it still
+  affects what the default protocol used for a `TracingService` that points at Zipkin.
```
LanceEa, line 111: Why is this still true? Isn't this completely removed in V3 as well and no longer is necessary?

docs/releaseNotes.yml
```diff
@@ -56,6 +56,41 @@ items:
   - version: 2.3.0
     date: '2022-06-06'
     notes:
+      - title: Envoy xDS v2 API removal, removal of ECMAScript regexes and v2 transport protocols
```
LanceEa, line 59: Why are these in this PR? They should have already been merged?

python/ambassador/compile.py
```diff
@@ -70,7 +67,6 @@ def Compile(logger: logging.Logger, input_text: str,
     out: Dict[str, Union[IR, EnvoyConfig]] = { "ir": ir }
 
     if ir:
-        out[envoy_version.lower()] = EnvoyConfig.generate(ir, envoy_version.upper(), cache=cache)
+        out['xds'] = EnvoyConfig.generate(ir, cache=cache)
```
LanceEa, line 70: same comment as above...i'm not against this but it does mean that if envoy introduces a v4 then we would be back to including a version. So, I don't see a downside to leaving this

docs/releaseNotes.yml
LukeShu: No? I wanted to get them in to 2.3, but because of `regex_type: unsafe` we couldn't.

LukeShu: But the changelog shouldn't mention v2 transport protocol removal, as this PR doesn't do that.

CHANGELOG.md
LukeShu: I don't know, we'll see.

docs/releaseNotes.yml
LukeShu: Oh, but this is in the wrong version's section.

python/tests/unit/test_irratelimit.py
```diff
@@ -80,31 +64,9 @@ def _get_ratelimit_default_conf_v2():
...
-
-    assert conf
-
-    assert conf.get('typed_config') == default_config
-
-
 @pytest.mark.compilertest
 def test_irratelimit_defaults():
-    default_config = _get_ratelimit_default_conf_v2()
+    default_config = _get_ratelimit_default_conf_v3()
```
AliceProxy, line 69: Doesnt this need to be `_get_ratelimit_default_conf()` I don't see a function for `_v3` anywhere in this file (nor do we need to specify v3/v2 anymore) so I'm curious how this is passing.

LanceEa (approved):
lgtm

LanceEa:
I'm merging so I can get my branches rebased on these changes.

AliceProxy (approved):

PR #4258, opened 2022-06-01
[v3.0] Reconcile filtersecrets repatriation

AliceProxy:
## Description
Just a cherry-pick of the commits from the following PR into `master`
See following PR for a proper description
https://github.com/emissary-ingress/emissary/pull/4251


cmd/entrypoint/secrets.go
```diff
@@ -313,6 +323,67 @@ func ReconcileSecrets(ctx context.Context, sh *SnapshotHolder) error {
 	return nil
 }
 
+// Returns secretName, secretNamespace from a provided (unstructured) filter if it contains a secret
+// Returns empty strings when the secret name and/or namespace could not be found
+func FindFilterSecret(filter *unstructured.Unstructured, action func(snapshotTypes.SecretRef)) error {
```
LanceEa, line 328: exporting  for testing purposes?

I had a rant on my last review on this topic but keeping it short this time. In general, we should probably be cleaner about how what we export as our public API. However, I doubt there are people depending on emissary as a library...

cmd/entrypoint/secrets.go
```diff
@@ -313,6 +323,67 @@ func ReconcileSecrets(ctx context.Context, sh *SnapshotHolder) error {
...
+		oauthFilter := mapOauth["OAuth2"]
+		if oauthFilter != nil {
+			secretName, secretNamespace := "", ""
+			// Check if we have a secretName
+			mapSecretName, success := oauthFilter.(map[string]interface{})
+			if !success {
+				return errors.New("Filter object detected with a bogus \"secretName\" field")
+			}
+			oauthSecretName := mapSecretName["secretName"]
+			if oauthSecretName != nil {
```
LanceEa, line 354: no reason for switch statement and we should bail early since the logic below requires a name provided and we ignore filter if not.

```golang
if sName, ok := oauthSecretName.(string); ok  && sName != "" {
secretName = sName
} else {
// bail early since no name provided and not an error
   return nil
}
```

cmd/entrypoint/secrets.go
```diff
@@ -313,6 +323,67 @@ func ReconcileSecrets(ctx context.Context, sh *SnapshotHolder) error {
...
+					secretName = sName
+				}
+			}
+			// Check if we have a secretNamespace
+			mapSecretNS, success := oauthFilter.(map[string]interface{})
+			if !success {
+				return errors.New("Filter object detected with a bogus \"secretNamespace\" field")
+			}
+			oauthSecretNS := mapSecretNS["secretNamespace"]
+			if oauthSecretNS != nil {
```
LanceEa, line 366: no need for switch statement and move namespace fallback from filter here

```golang
if sNs, ok := oauthSecretNS.(string); ok &&  sNS != "" {
secretNamespace = sNS
} else {
secretNamespace = filter.GetNamespace()
}
```

cmd/entrypoint/testutil_fake_filtersecrets_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+	// After we flush the above config then secrets should be reconciled which should
+	// result in the above secret that is referenced by our filter being added to the snapshot if everything succeeds
+	snap, err := f.GetSnapshot(HasSecret("namespaced-secret", "foo"))
+	require.NoError(t, err)
+	assert.NotNil(t, snap)
+	t.Setenv("EDGE_STACK", "")
+}
+
+func TestFindFilterSecretNamespaceV2(t *testing.T) {
```
LanceEa, line 101: What do you think about moving these to a `cmd/entrypoint/secrets_test.go`?

The reason being is that these are true UnitTest where is the ones above are closer to integration test since they depend on `diagd` being available on the host. In general, I have seen this in other places in the code base where we have mixed these which I find a little murky and as a developer and surprising.

Same for all the Test below.

LanceEa (changes requested):

LukeShu:
Can't we just wait for the main rautomated epatriation PR?

LanceEa:
> Can't we just wait for the main rautomated epatriation PR?

I had already merged the regular repatriation and then it recreated this branch https://github.com/emissary-ingress/emissary/pull/4252 which included everything. I put a label DO NO MERGE until we could discuss. So, my assumption is that @AliceProxy made this commit due to do that.

LanceEa:
@AliceProxy - This was already part of the repatriation due to us rolling back master, so we can close this PR?

PR #4265, opened 2022-06-05
Make stats_name work in tracing, ratelimit and auth services

psalaberria002:
## Description
Setting stats_name in Tracing, RateLimit and Auth services did no have any effect. alt_stat_name in Envoy was not being updated.

## Testing
I have tested it with `make deploy`.

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/python/tests/unit/test_irauth.py
+++ b/python/tests/unit/test_irauth.py
@@ -5,6 +5,7 @@ import sys
 import pytest
 
 from kat.harness import EDGE_STACK
+from tests.utils import econf_foreach_cluster
 
 logging.basicConfig(
     level=logging.INFO,
@@ -104,6 +105,40 @@ spec:
     assert ext_auth_config["typed_config"]["transport_api_version"] == "V3"
 
     assert "mycoolauthservice.default.1" not in econf.ir.aconf.errors
+
+
+def test_cluster_fields():
+    yaml = """
+---
+apiVersion: getambassador.io/v3alpha1
+kind: AuthService
+metadata:
+  name:  mycoolauthservice
+  namespace: default
+spec:
+  auth_service: someservice
+  protocol_version: "v3"
+  proto: grpc
+  stats_name: authservice
+"""
+
+    econf = _get_envoy_config(yaml)
+
+    conf = econf.as_dict()
+    ext_auth_config = _get_ext_auth_config(conf)
+
+    cluster_name = "cluster_extauth_someservice_default"
+
+    assert ext_auth_config
+    assert (
+        ext_auth_config["typed_config"]["grpc_service"]["envoy_grpc"]["cluster_name"]
+        == cluster_name
+    )
+
+    def check_fields(cluster):
+        assert cluster["alt_stat_name"] == "authservice"
+
+    econf_foreach_cluster(econf, check_fields, name=cluster_name)
 
 
 @pytest.mark.compilertest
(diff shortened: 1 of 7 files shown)


LanceEa:
@psalaberria002 - Thanks for being patient on this. With the recent release master has diverged quite a bit. Can you please rebase this back on master and push it. I will then get CI running on it.

psalaberria002:
I did a cherry pick on top of the new master after seeing the notice in emissary-dev slack channel.

LanceEa:
@psalaberria002 - Sorry for the slow response on this. This is definitely a regression that we would like to have land in the v3 and v2 branch.

1. rebase on master to pull in all the recent changes for v3
2. can you add a unit tests for `AuthService`, `RateLimitService` and `TraceService` to make sure that the `stats_name` is being passed correctly to the cluster so that we make sure the regression doesn't happen again.
3. let's add a release note making people aware of the bug fix. - I can help direct on this item

Once we land it in master we can back port it to the `release/v2.3` branch.

Thanks and appreciate the patience on this!

python/tests/unit/test_irauth.py
```diff
@@ -106,6 +107,40 @@ def test_irauth_grpcservice_version_v3():
...
+apiVersion: getambassador.io/v3alpha1
+kind: AuthService
+metadata:
+  name:  mycoolauthservice
+  namespace: default
+spec:
+  auth_service: someservice
+  protocol_version: "v3"
+  proto: grpc
+  stats_name: authservice
```
LanceEa, line 122: What happens if a user places a super long one? I know that cluster names are limited to I think 60 characters. Based on reading here: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto is says it follows the conventions of `upstream_cluster.{alt_stat_name}`....so if someone put a super long one would that blow up envoy or would envoy just trim itself?

Can you test that? Because if it is then I think we need to add some logic when the alt_stat_name is being set to trim it and log an error to the diagnostics.

LanceEa (commented):
One quick question...then do you think we should have a KAT test for `t_stats.py`  that verifies this. I just glanced at those tests and not sure if they are easy to mimic and just test to verify alt_stat_name is working as expected??

Take a look and let me know if you think it would be easy to add a similar test or not.

python/tests/unit/test_irauth.py
psalaberria002:
```
$ k exec -it emissary-cdf-69db9d9c74-tlv9v -- /bin/cat envoy/envoy.json | grep alt_stat
        "alt_stat_name": "fffffffffffffffffffffffffffffffffddddddddddddddddddddddddddddddddddddddddddsssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxffffffffffffffffoooooooooooooooooooooooooooooooooooooooopppppppppppppppppppppppppppppppppppppppqqqqqqqqqqqqqqqqqq",
```
No limitations there. Envoy picks it up with no issues. The limit of chars is in the cluster name, not the stats name.

psalaberria002:
> One quick question...then do you think we should have a KAT test for `t_stats.py` that verifies this. I just glanced at those tests and not sure if they are easy to mimic and just test to verify alt_stat_name is working as expected??
> 
> Take a look and let me know if you think it would be easy to add a similar test or not.

I think those kat tests are overkill for testing this feature.

cindymullins-dw:
Many thanks for this awesome contribution, @psalaberria002!  If this works as expected in release [2.4.0](https://www.getambassador.io/docs/emissary/2.4/release-notes/) please let us know if we can close out this PR.

LanceEa:
Already landed so closing out.

PR #4271, opened 2022-06-09
[v2.3.1] Release PR

AliceProxy:


LanceEa:
@AliceProxy @LukeShu - what was the outcome of the make chart-test failing, I believe we understood why but unsure.

AliceProxy:
> @AliceProxy @LukeShu - what was the outcome of the make chart-test failing, I believe we understood why but unsure. 

I believe Luke can better explain, but If the tag for a release is pushed  before the CRDs are made available then it will fail.

LanceEa (approved):
lgtm, since this already is released. The only up question for me is what do we need to do to address the chart check. Is there anything we need to do or just merge?

LukeShu:
I expect the chart check to fail on a commit that has been tagged as an RC.

Issue #4277, opened 2022-06-13
Option to build envoy in FIPS mode.


bathina2:
**Please describe your use case / problem.**
There are some clients that would like emissary ingress to support FIPS approved encryption algorithms. 

**Describe the solution you'd like**
Since there is already a way to build envoy, some additional flags to build it in FIPS mode would be a nice addition.

**Describe alternatives you've considered**
Using envoy by itself.

bathina2:
I have a reasonable enhancement to support this. I would like to work on this issue and contribute if I can.

Thanks,
Sirish

ddymko:
Closing since this is resolved

PR #4279, opened 2022-06-13
Enable fips mode when building envoy

bathina2:
## Description
The following PR enables building envoy/emissary in FIPS mode. This will ensure that the envoy uses FIPS certified cryptographic algorithms when decrypting data.
Developers can set this by using the `export FIPS_MODE=true` option before doing `make update-base`.

## Related Issues
https://github.com/emissary-ingress/emissary/issues/4277

## Testing
After adding my changes I was able to to build envoy using `make update-base`. The resulting image was tagged as a FIPS enabled image-

```
# docker run -it --entrypoint envoy-static-stripped  emissaryingress/base-envoy:envoy-0.-.opt --version

envoy-static-stripped  version: 4ce93dc3ace00ae9108b179d0afaceac13f4602a/1.17.4/Modified/RELEASE/BoringSSL-FIPS
```

## Checklist

 - [x] I made sure to update `CHANGELOG.md`. _Not required_

    + Any new features
    + Any deprecations

    Not sure exactly how I can Unit test my changes. I have automations in place that follow the described build steps to create an Envoy image. These images do contain the FIPS suffix at the end of the version. 
    When omitting the option, it generates the appropriate image with out the suffix.

### DIFF
--- a/_cxx/envoy.mk
+++ b/_cxx/envoy.mk
@@ -19,8 +19,11 @@ RSYNC_EXTRAS ?=
   # You may reset BASE_ENVOY_RELVER when adjusting ENVOY_COMMIT.
   BASE_ENVOY_RELVER ?= 0
 
+  # Set to non-empty to enable compiling Envoy in FIPS mode.
+  FIPS_MODE ?=
+
   ENVOY_DOCKER_REPO ?= $(if $(IS_PRIVATE),quay.io/datawire-private/base-envoy,docker.io/emissaryingress/base-envoy)
-  ENVOY_DOCKER_VERSION ?= $(BASE_ENVOY_RELVER).$(ENVOY_COMMIT).$(ENVOY_COMPILATION_MODE)
+  ENVOY_DOCKER_VERSION ?= $(BASE_ENVOY_RELVER).$(ENVOY_COMMIT).$(ENVOY_COMPILATION_MODE)$(if $(FIPS_MODE),.FIPS)
   ENVOY_DOCKER_TAG ?= $(ENVOY_DOCKER_REPO):envoy-$(ENVOY_DOCKER_VERSION)
   ENVOY_FULL_DOCKER_TAG ?= $(ENVOY_DOCKER_REPO):envoy-full-$(ENVOY_DOCKER_VERSION)
 # END LIST OF VARIABLES REQUIRING `make update-base`.
@@ -206,7 +209,7 @@ $(OSS_HOME)/docker/base-envoy/envoy-static: $(ENVOY_BASH.deps) FORCE
 	            exit 1; \
 	        fi; \
 	        $(call ENVOY_BASH.cmd, \
-	            $(ENVOY_DOCKER_EXEC) bazel build --verbose_failures -c $(ENVOY_COMPILATION_MODE) --config=clang //source/exe:envoy-static; \
+	            $(ENVOY_DOCKER_EXEC) bazel build $(if $(FIPS_MODE), --define boringssl=fips) --verbose_failures -c $(ENVOY_COMPILATION_MODE) --config=clang //source/exe:envoy-static; \
 	            rsync -a$(RSYNC_EXTRAS) --partial --blocking-io -e 'docker exec -i' $$(cat $(OSS_HOME)/_cxx/envoy-build-container.txt):/root/envoy/bazel-bin/source/exe/envoy-static $@; \
 	        ); \
 	    fi; \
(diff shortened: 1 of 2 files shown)


_cxx/envoy.mk
```diff
@@ -7,6 +7,8 @@ YES_I_AM_OK_WITH_COMPILING_ENVOY ?=
 ENVOY_TEST_LABEL ?= //test/...
 # Set RSYNC_EXTRAS=Pv or something to increase verbosity.
 RSYNC_EXTRAS ?=
+# Set to non-empty to enable compiling Envoy in FIPS mode.
+FIPS_MODE ?=
```
LukeShu, line 11: If it affects the Envoy build, then it's gotta be included in the image tag and in the section of variables that `IF YOU MESS WITH ANY OF THESE VALUES, YOU MUST RUN make update-base.`

_cxx/envoy.mk
```diff
@@ -206,7 +208,7 @@ $(OSS_HOME)/docker/base-envoy/envoy-static: $(ENVOY_BASH.deps) FORCE
 	            exit 1; \
 	        fi; \
 	        $(call ENVOY_BASH.cmd, \
-	            $(ENVOY_DOCKER_EXEC) bazel build --verbose_failures -c $(ENVOY_COMPILATION_MODE) --config=clang //source/exe:envoy-static; \
+	            $(ENVOY_DOCKER_EXEC) bazel build $$(if [ -n '$(FIPS_MODE)' ]; then echo "--define boringssl=fips"; fi) --verbose_failures -c $(ENVOY_COMPILATION_MODE) --config=clang //source/exe:envoy-static; \
```
LukeShu, line 211: This can be better written with Make macros: `$(if $(FIPS_MODE), --define boringssl=fips )`.

_cxx/envoy.mk
bathina2: Updated

_cxx/envoy.mk
bathina2: Thanks for the feedback, not super familiar with using make macros, but tested with the change and it did infact set it correctly!

bathina2:
@LukeShu I updated `DEVELOPING.md` with better messaging around what is supported. Also removed the release note.

_cxx/envoy.mk
```diff
@@ -23,6 +23,9 @@ RSYNC_EXTRAS ?=
   ENVOY_DOCKER_VERSION ?= $(BASE_ENVOY_RELVER).$(ENVOY_COMMIT).$(ENVOY_COMPILATION_MODE)
   ENVOY_DOCKER_TAG ?= $(ENVOY_DOCKER_REPO):envoy-$(ENVOY_DOCKER_VERSION)
   ENVOY_FULL_DOCKER_TAG ?= $(ENVOY_DOCKER_REPO):envoy-full-$(ENVOY_DOCKER_VERSION)
+
+  # Set to non-empty to enable compiling Envoy in FIPS mode.
+  FIPS_MODE ?=
```
kflynn, line 28: I'm pretty sure that you also must include `$(FIPS_MODE)` in `ENVOY_DOCKER_VERSION` for things to work correctly:

```
ENVOY_DOCKER_VERSION ?= $(BASE_ENVOY_RELVER).$(ENVOY_COMMIT).$(ENVOY_COMPILATION_MODE).$(if $(FIPS_MODE),FIPS,noFIPS)
```

or the like. I'll defer to @LukeShu here though.

kflynn (changes requested):
I think you're going to have to update `ENVOY_DOCKER_VERSION` too, or things won't work.

_cxx/envoy.mk
bathina2: Thanks for the suggestion @kflynn!

ddymko:
@bathina2 @kflynn 

I had made a change in https://github.com/emissary-ingress/emissary/pull/4377/commits/38f9fd62d324efddb151e9d61b6e6000d1df25c4 in order for CI to pass.

I noticed that if `.$(if $(FIPS_MODE),FIPS,noFIPS)`  was set we would be looking for builds with `.noFIPS` which we don't have. I modified this to `$(if $(FIPS_MODE),.FIPS)` to prevent the the CI failures. 

@bathina2 Could you please pull in this change and also revise your commit history to a single commit?

bathina2:
@ddymko Thanks for the review and suggestions. I added your changes and squashed everything into a single commit.

ddymko (approved):
Changes look good - CI passed in https://github.com/emissary-ingress/emissary/pull/4377

PR #4288, opened 2022-06-16
Improve tests, improve mypy strictness

LukeShu:
## Description
While doing other work, I was frustrated that `mypy` wasn't catching things that (IMO) it should.  So then I was relying on actual running tests instead of static analysis. And that was frustrating, because a bunch of the Python unit tests make assumptions about `/tmp` that make them not run on my system.  So get the unit tests to obey `$TMPDIR`, and crank up the mypy strictness with `--check-untyped-defs`.

## Testing
To ensure that the Python unit tests fully obey `$TMPDIR` by entering a new mount namespace (`sudo unshare --mount sudo -u $USER bash -i`), and bind-mounting `/var/emtpy` over `/tmp` (`sudo mount -o bind /var/empty /tmp`), then running `make pytest-unit` in this modified environment.  In this setup, anything that disobeys `$TMPDIR` and tries to write to `/tmp` will fail.

## Checklist

    + Any new features
    + Any deprecations


python/ambassador/config/config.py
```diff
@@ -345,25 +346,17 @@ def post_notice(self, msg: str, resource: Optional[Resource]=None, log_level=log
...
-
-        if isinstance(msg, RichStatus):
-            return 'RichStatus'
-        elif isinstance(msg, str):
-            return 'string'
-        else:
-            return type(msg).__name__
+    @singledispatchmethod
+    def post_error(self, msg: Union[RichStatus, str], resource: Optional[Resource]=None, rkey: Optional[str]=None, log_level=logging.INFO) -> None:
+        assert False
```
LanceEa, line 351: This is here as a fail safe to catch bugs correct? It should never happen if type checking stops user from calling post_error with anything other than RichStatus and str, correct? If it does then we know there is bug...

python/ambassador_diag/diagd.py
```diff
@@ -957,6 +961,7 @@ def show_overview(reqid=None):
     # to compute -- we don't want to call it more than once here, so we cache
     # its value.
     diag = app.diag
+    assert diag
```
LanceEa, line 964: In python, does any of these assertions crash Emissary? Are we adding them as a way to detect bugs?

python/tests/utils.py
```diff
@@ -223,17 +223,28 @@ def econf_foreach_cluster(econf, fn, name='cluster_httpbin_default'):
             break
     assert found_cluster
 
-def assert_valid_envoy_config(config_dict, v2=False):
-    with tempfile.NamedTemporaryFile() as temp:
-        temp.write(bytes(json.dumps(config_dict), encoding = 'utf-8'))
-        temp.flush()
-        f_name = temp.name
-        cmd = [ENVOY_PATH, '--config-path', f_name, '--mode', 'validate']
+def assert_valid_envoy_config(config_dict, extra_dirs=[], v2=False):
```
LanceEa, line 226: This was one of the items on my list that I wanted us to address. I like the fact that this is now not dependent on local installs. However, the part I still do not like which we can address in the future is that its mixed into the "unit" test which I found unexpected when trying to develop.

python/watch_hook.py
```diff
@@ -213,6 +213,10 @@ def load_yaml(self, yaml_stream):
                         # At the moment, we stuff the resolver's datacenter into the association
                         # ID for this watch. The ResourceFetcher relies on that.
 
+                        assert resolver.datacenter
```
LanceEa, line 216: Same comment as earlier, what happens if these assertions fail, is this a bug if they fail or should we be handling these gracefully?

LanceEa (commented):
overall, everything looks pretty good. I have a few comments I want to discuss just to make sure I'm understanding how python will handle a few situations.

LanceEa (approved):
lgtm

AliceProxy (approved):

PR #4293, opened 2022-06-20
Fix cooldown feature to properly manage several instances of emissary

knlambert:
## Description

Previously, we added a cooldown step (30s) for the metrics[ to avoid to send too many requests](https://github.com/emissary-ingress/emissary/pull/4122) to the ambassador cloud API.

Although, some values were not consistent, and we figured out that the way the metrics are sent to the agent are **one request per emissary node**, and not **one for all**.

It means that the previous iterations of the code were wrong for two reasons : 
* If we forward the metrics directly to the cloud, we won't be able to identify what node corresponds to what metric.
* If we put a simple cooldown of 30s, for 3 nodes, we'll miss 2 out of 3 sets of metrics.

As a workaround, with minimal changes, this PR changes the logic to make the metric relay handler to accumulate metrics in a map per nodes' IP, and unload them after the cooldown period.

image

## Testing

Adapted existing tests, & tested it in a dev cluster.

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -824,7 +824,7 @@ func (a *Agent) MetricsRelayHandler(
 		p, ok := peer.FromContext(ctx)
 
 		if !ok {
-			dlog.Errorf(ctx, "peer not found in context")
+			dlog.Warnf(ctx, "peer not found in context")
 			return
 		}
 
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -14,6 +15,7 @@ import (
 	"time"
 
 	io_prometheus_client "github.com/prometheus/client_model/go"
+	"google.golang.org/grpc/peer"
 	"google.golang.org/protobuf/types/known/timestamppb"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
@@ -114,6 +116,9 @@ type Agent struct {
 	// Cloud API.
 	metricsBackoffUntil time.Time
 
+	// Used to accumulate metrics for a same timestamp before pushing them to the cloud.
+	metricStack map[string][]*io_prometheus_client.MetricFamily
+
 	// Extra headers to inject into RPC requests to ambassador cloud.
 	rpcExtraHeaders []string
 
@@ -181,8 +186,9 @@ func NewAgent(directiveHandler DirectiveHandler, rolloutsGetterFactory rolloutsG
 		directiveHandler:             directiveHandler,
 		reportRunning:                atomicBool{value: false},
 		agentWatchFieldSelector:      getEnvWithDefault("AGENT_WATCH_FIELD_SELECTOR", "metadata.namespace!=kube-system"),
-		metricsBackoffUntil:          time.Now(),
+		metricsBackoffUntil:          time.Now().Add(defaultMinReportPeriod),
 		rpcExtraHeaders:              rpcExtraHeaders,
+		metricStack:                  map[string][]*io_prometheus_client.MetricFamily{},
 	}
 }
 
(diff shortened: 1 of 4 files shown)


LanceEa:
@knlambert - I will have some time to look at this more later and think about it but just as a quick glance.

1. does this need to land in v2.3 branch as Alex mentioned?  I would rather it land in the v3 branch which is `master` then if need be we can backport it.
2. Again only taking a few minutes to look at it, it appears that you delete metrics before they have successfully been sent which means on errors they will be dropped. Is this the intended behavior? Keeping them around would introduce memory pressure too so not advocating for any particular solution but I just wanted to understand better 😄.

knlambert:
> @knlambert - I will have some time to look at this more later and think about it but just as a quick glance.
> 
> 1. does this need to land in v2.3 branch as Alex mentioned?  I would rather it land in the v3 branch which is `master` then if need be we can backport it.
> 2. Again only taking a few minutes to look at it, it appears that you delete metrics before they have successfully been sent which means on errors they will be dropped. Is this the intended behavior? Keeping them around would introduce memory pressure too so not advocating for any particular solution but I just wanted to understand better 😄.

1. The next one to be published :o) it's a bug fix.
2. It's ok, that's not critical data and we don't want to spam cloud if the link is broken, or saturate the memory as you said :)

pkg/agent/agent.go
```diff
@@ -111,6 +112,9 @@ type Agent struct {
 	// Cloud API.
 	metricsBackoffUntil time.Time
 
+	// Used to accumulate metrics for a same timestamp before pushing them to the cloud.
+	metricStack map[string][]*io_prometheus_client.MetricFamily
```
LanceEa, line 116: bike shed: `metricStack` doesn't quite capture the variables purpose. Can we maybe name it something like `aggregatedMetrics` or `queuedMetrics`?  Other ideas?

pkg/agent/agent.go
```diff
@@ -169,8 +173,9 @@ func NewAgent(directiveHandler DirectiveHandler, rolloutsGetterFactory rolloutsG
 		directiveHandler:             directiveHandler,
 		reportRunning:                atomicBool{value: false},
 		agentWatchFieldSelector:      getEnvWithDefault("AGENT_WATCH_FIELD_SELECTOR", "metadata.namespace!=kube-system"),
-		metricsBackoffUntil:          time.Now(),
+		metricsBackoffUntil:          time.Now().Add(defaultMinReportPeriod),
```
LanceEa, line 176: Just so I understand, was this added to make sure that it doesn't try to report instantly when it starts up and delays the initial report?

pkg/agent/agent.go
```diff
@@ -635,56 +640,70 @@ var allowedMetricsSuffixes = []string{"upstream_rq_total", "upstream_rq_time", "
...
+					newMetrics = append(newMetrics, metricFamily)
 					break
 				}
 			}
 		}
 
-		outMessage := &agent.StreamMetricsMessage{
-			Identity:     a.agentID,
-			EnvoyMetrics: outMetrics,
+		p, ok := peer.FromContext(ctx)
```
LanceEa, line 665: Can you move this up above line 650 (`a.ambassadorAPIKeyMutex.Lock()`). 

Since it is required, we can avoid obtaining a lock on the apiKey and not allocate newMetrics if we are just going to return.

pkg/agent/agent.go
```diff
@@ -635,56 +640,70 @@ var allowedMetricsSuffixes = []string{"upstream_rq_total", "upstream_rq_time", "
...
+				delete(a.metricStack, key)
+			}
 
-			if err := a.comm.StreamMetrics(logCtx, outMessage, apikey); err != nil {
-				dlog.Errorf(logCtx, "error streaming metric(s): %+v", err)
+			if relayedMetricCount := len(outMessage.GetEnvoyMetrics()); relayedMetricCount > 0 {
+
+				dlog.Infof(ctx, "Relaying %d metric(s)", relayedMetricCount)
+
+				if err := a.comm.StreamMetrics(ctx, outMessage, apikey); err != nil {
```
LanceEa, line 698: Just confirming again, If this err's out we are comfortable dropping this set of metrics, correct?

pkg/agent/agent.go
```diff
@@ -635,56 +640,70 @@ var allowedMetricsSuffixes = []string{"upstream_rq_total", "upstream_rq_time", "
...
-			dlog.Infof(logCtx, "Relaying %d metric(s)", relayedMetricCount)
+		a.metricsRelayMutex.Lock()
+		defer a.metricsRelayMutex.Unlock()
+		// Collect metrics until next report.
+		if time.Now().Before(a.metricsBackoffUntil) {
+			dlog.Infof(ctx, "Append %d metric(s) to stack from %s",
+				len(newMetrics), instanceID,
+			)
+			a.metricStack[instanceID] = newMetrics
+		} else {
```
LanceEa, line 682: style suggestion: can we just `return` after capturing the logs in the metricStack. Then the `else` block can be removed.

pkg/agent/agent_metrics_test.go
```diff
@@ -46,33 +48,44 @@ func agentMetricsSetupTest() (*MockClient, *Agent) {
...
+		metricStack: map[string][]*io_prometheus_client.MetricFamily{},
 	}
 
 	return clientMock, stubbedAgent
 }
 
 func TestMetricsRelayHandler(t *testing.T) {
 
-	t.Run("will relay the metrics", func(t *testing.T) {
+	t.Run("will relay metrics from the stack", func(t *testing.T) {
```
LanceEa, line 59: Two test cases that I think still need coverage are:

1. backoff time was met but no metrics available in `metricStack`
2. unable to get peer IP address from context, thus no metrics sent

LanceEa (changes requested):
@knlambert - Overall it looks good, I have left a few style suggestions and a few test cases we should add coverage for.

Once addressed we can get this merged into master and cherry-picked to the `release/v2.3` branch as well.

pkg/agent/agent.go
knlambert: It was added because otherwise it sends at least 3 requests every 5 seconds, which is too much for ambassador cloud. More-other, we only need to have data every 30s.

pkg/agent/agent.go
knlambert: Yep !

LanceEa (approved):
lgtm

alexgervais (approved):
Lets cherry pick this on `release/v2.3` as soon as it lands on master. We don't want the changelog to be lying!

haq204 (commented):
out of curiosity, what's the expected rate of metrics per second?

Would it make sense to set a fixed buffer size and push if its either full or after 30 seconds?

LanceEa (dismissed):
lgtm...now that CI is green and merge conflicts are all good

LanceEa:
Let's squash the commits and clean up commit message when we merge since they are all related.

haq204 (approved):

PR #4292, opened 2022-06-19
[3.0] upgrade envoy

LanceEa:
WIP

## Related Issues
List related issues.

## Testing
A few sentences describing what testing you've done, e.g., manual tests, automated tests, deployed in production, etc.

## Checklist

 - [ ] I made sure to update `CHANGELOG.md`.

   Remember, the CHANGELOG needs to mention:
    + Any new features
    + Any changes to our included version of Envoy
    + Any non-backward-compatible changes
    + Any deprecations

 - [ ] This is unlikely to impact how Ambassador performs at scale.

   Remember, things that might have an impact at scale include:
    + Any significant changes in memory use that might require adjusting the memory limits
    + Any significant changes in CPU use that might require adjusting the CPU limits
    + Anything that might change how many replicas users should use
    + Changes that impact data-plane latency/scalability

 - [ ] My change is adequately tested.

   Remember when considering testing:
    + Your change needs to be specifically covered by tests.
       + Tests need to cover all the states where your change is relevant: for example, if you add a behavior that can be enabled or disabled, you'll need tests that cover the enabled case and tests that cover the disabled case. It's not sufficient just to test with the behavior enabled.
    + You also need to make sure that the _entire area being changed_ has adequate test coverage.
       + If existing tests don't actually cover the entire area being changed, add tests.
       + This applies even for aspects of the area that you're not changing – check the test coverage, and improve it if needed!
    + We should lean on the bulk of code being covered by unit tests, but...
    + ... an end-to-end test should cover the integration points

 - [ ] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently.

 - [ ] The changes in this PR have been reviewed for security concerns and adherence to security best practices.

PR #4299, opened 2022-06-21
[v3.0] Drop support for `transport_protocol: v2`

LukeShu:
## Description
This PR is split off from https://github.com/emissary-ingress/emissary/pull/4292.

The goal here is to establish a sane behavior for what to do when we see `transport_protocol: v2`, which is no longer supported. For `LogService` and `RateLimitService`, we post an error, and discard the resource.  For `AuthService` that's not quite safe; so instead of discarding it, we post the error and then when emitting the Envoy config if any errors have been posted to that resource instead of emitting an `ExtAuthz` filter, we emit a `Lua` filter that says HTTP 500 and says that there's a misconfigured `AuthService`; similar to how Edge Stack reports a misconfigured `Filter`.

The first 3 commits are cleaning up the `kat-server` to make headers easier to reason about in that final commit that does the real work we care about.

## Testing
The changes to the tests have been carefully considered.

## Checklist

    + Any new features
    + Any deprecations


AliceProxy (approved):
lgtm, send it!

PR #4300, opened 2022-06-21
ready: Use envoy listener to expose endpoint from worker for ambassador

tomasbanet:
Covering for @jfrabaute 
This is a rebased version of https://github.com/emissary-ingress/emissary/pull/4253

===

default).

This change is fixing the slowness of /ready endpoint.
The /ready endpoint is now exposed in the worker pool by adding a
pool.
Related Issues

Testing

Manual tests


jfrabaute:
Replacing https://github.com/emissary-ingress/emissary/pull/4253

python/ambassador/envoy/v3/v3ready.py
```diff
@@ -0,0 +1,159 @@
...
+if TYPE_CHECKING:
+    from . import V3Config # pragma: no cover
+
+# The defaults can be changed by using those 2 env vars:
+# AMBASSADOR_READY_PORT: Port number (default 8002)
+# AMBASSADOR_READY_LOG: true/false (default true)
+ambassador_ready_port = int(os.getenv("AMBASSADOR_READY_PORT", "8002"))
+if ambassador_ready_port not in range(1, 32767):
+    ambassador_ready_port = 8002
+ambassador_ready_log = os.getenv("AMBASSADOR_READY_LOG", "true") in ("yes", "true", "t", "1")
```
kflynn, line 30: In `ambassador/utils.py` there's a `parse_bool` that does exactly this, so

```
ambassador_ready_log = parse_bool(os.getenv("AMBASSADOR_READY_LOG", "true"))
```

is probably a better choice.

kflynn (changes requested):
Code looks good apart from one small note. I'd love to see a test for this, though -- should be easy enough to make sure that things become ready using the port 8002 server as well as the default port 8001, right?

Thanks for sticking with this!

pkg/acp/envoy.go
```diff
@@ -144,3 +151,19 @@ func (w *EnvoyWatcher) IsAlive() bool {
...
+	strReadyPort := os.Getenv("AMBASSADOR_READY_PORT")
+	if strReadyPort != "" {
+		readyPort, err = strconv.Atoi(strReadyPort)
+		if err != nil {
+			dlog.Infof(context.Background(), "Unable to parse AMBASSADOR_READY_PORT: %s", err)
+		}
+	}
+	if readyPort < 1 || readyPort > 32767 {
+		readyPort = 8002
+	}
```
LukeShu, line 167: Suggestions `strconv.ParseUint(strReadyPort, 10, 15)` would do this bounds checking for you.

tomasbanet: Yeah I would prefer to have an explicit bounds check. @LukeShu should I resolved this comment?

tomasbanet:
Docs change: https://github.com/datawire/ambassador-docs/pull/683 (kept it in the pre-release section. Not sure where to add the docs for the upcoming release that this change is targetted to go into)

kflynn (changes requested):
Hey, @tomasbanet, are you the Emissary [Slack](https://slack.a8r.io/)? I'd love to get this landed, but I still don't see a test change to make sure that the ready listener as configured actually functions as intended. 🙁 I'd love to chat with you about this!

It feels like it should be as easy as picking one of the tests - I might just pick `Empty` in `t_basics.py` - and setting the correct environment variable, something like modifying the test's `init` (not `__init__` in this case) to include

```
        self.manifest_envs += """
    - name: AMBASSADOR_READY_PORT
      value: "8002"
"""

**Comment by tomasbanet** [comment]
> Hey, @tomasbanet, are you the Emissary [Slack](https://slack.a8r.io/)? I'd love to get this landed, but I still don't see a test change to make sure that the ready listener as configured actually functions as intended. 🙁 I'd love to chat with you about this!
> 
> It feels like it should be as easy as picking one of the tests - I might just pick `Empty` in `t_basics.py` - and setting the correct environment variable, something like modifying the test's `init` (not `__init__` in this case) to include
> 
> ```
>         self.manifest_envs += """
>     - name: AMBASSADOR_READY_PORT
>       value: "8002"
> """
> ```

Hi Flynn. Modified the Empty kat test to use a non default port (decided to go with port 8500 - going with the port ranges described in - python/README.md#Ambassador Components and Ports. That said, should we be using 8002 by default? It seems to be reserved for the watcher, but the tests are working fine so maybe it's not always started at that port). All the other tests should be using port 8002 by default. Readiness is tested by checking that the k8s pods are ready (done by the test framework automatically).

File: python/ambassador/envoy/v3/v3ready.py [hunk 3]
```diff
@@ -0,0 +1,171 @@
...
+    from . import V3Config  # pragma: no cover
+
+# The defaults can be changed by using those 2 env vars:
+# AMBASSADOR_READY_PORT: Port number (default 8006)
+# AMBASSADOR_READY_LOG: true/false (default true)
+ambassador_ready_port = int(os.getenv("AMBASSADOR_READY_PORT", "8006"))
+if ambassador_ready_port not in range(1, 32767):
+    ambassador_ready_port = 8006
+ambassador_ready_log = parse_bool(os.getenv("AMBASSADOR_READY_LOG", "true"))
+ambassador_ready_ip = "127.0.0.1"
```
LanceEa, line 31: Just some clarifying questions:

1. The default Listeners are listening on `0.0.0.0` and this is listening on loopback (127.0.0.1). This was a conscious decision to ensure it is only available for internal implementation use, correct?
2. How does `127.0.0.1` work in an IPV6 only scenario? 


On **2**, we introduced the ability to change the public facing health check bind address, port and IP family in this commit: https://github.com/emissary-ingress/emissary/commit/507bf91f95e489d7d5cb437ca7cd1d8f320d4a81 to support IPV6 only networks. Maybe we need to check the `AMBASSADOR_HEALTHCHECK_IP_FAMILY` and adjust the ready ip to be `::1` for ipv6?

pkg/acp/envoy.go
```diff
@@ -144,3 +151,19 @@ func (w *EnvoyWatcher) IsAlive() bool {
...
+	if strReadyPort != "" {
+		readyPort, err = strconv.ParseUint(strReadyPort, 10, 15)
+		if err != nil {
+			dlog.Infof(context.Background(), "Unable to parse AMBASSADOR_READY_PORT or port is out of bounds: %s", err)
+		}
+	}
+	if readyPort < 1 {
+		readyPort = 8006
+	}
+	return fmt.Sprintf("http://localhost:%d/ready", readyPort)
```
LanceEa, line 168: I think we are fine with `localhost` here based on this https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/net/ipsock.go;drc=38cfb3be9d486833456276777155980d1ec0823e;l=310.

@kflynn - Do you know if Golang will correctly resolve localhost to an ipv6 address, when in ipv6 only?

LanceEa (commented):
@tomasbanet @kflynn - I pulled this into Edge Stack code base and verified all is green and installed it into a test cluster. 

I just have a few clarifying questions around IPV6 and making sure this doesn't break the IPV6 only clusters.

pkg/acp/envoy.go
kflynn: The resolver should do the right thing with `localhost`, always forever and for true – if not, that's a serious bug.

python/ambassador/envoy/v3/v3ready.py
kflynn: I believe that using 127.0.0.1 is deliberate, but: it won't work on an IPv6-only host. That's a great catch: we'll need to switch the IP address based on family, or see if we can use `localhost` here (I suspect not, but worth checking).

kflynn (commented):
Evidently my comments got stuck behind a review, so let me post a "comment" review here...

python/tests/kat/t_basics.py
```diff
@@ -16,6 +16,12 @@ class Empty(AmbassadorTest):
     def init(self):
         if EDGE_STACK:
             self.xfail = "XFailing for now"
+        # Specify a non-default readiness port for test coverage purposes.
+        # All other tests will use the default 8006 port.
```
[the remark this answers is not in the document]
tomasbanet: Added an explicit query for the ready endpoint.

python/ambassador/envoy/v3/v3ready.py
tomasbanet: Good catch. Made a change to use ::1 if AMBASSADOR_HEALTHCHECK_IP_FAMILY is IPV6_ONLY, and 127.0.0.1 in all other cases.

To test this I deployed emissary to an IPv6 kind cluster on my laptop (https://kind.sigs.k8s.io/docs/user/configuration/#networking). First I deployed emissary using the 127.0.0.1 ready address first just to check what kind of issues it causes. For some reason it was working fine without any problems despite the cluster being ipv6 based. 

Afterwards I tried "localhost". Emissary ended up crash looping with this https://gist.github.com/tomasbanet/439fd5dc6d6ea91f9f3171753ea63aea `msg="Post \"http://localhost:8004/_internal/v0/watt?url=http%3A%2F%2Flocalhost%3A9696%2Fsnapshot\": dial tcp [::1]:8004: connect: connection refused" func=github.com/emissary-ingress/emissary/v3/cmd/entrypoint.notifyWebhookUrl file="/go/cmd/entrypoint/notify.go:124"`.

I switched to using either an IPv6 address or IPv4 loopback address explicitly based on the ip family env var. This works for ipv4 but doesn't seem to work when specifying the env var to use ipv6 - https://gist.github.com/tomasbanet/09bcf9659c4ce33b01bfaa6185d2e6e9 `time="2023-01-16 21:02:08.0044" level=error msg="shut down with error error: listen tcp6: address 0.0.0.0: no suitable address found" func=github.com/emissary-ingress/emissary/v3/pkg/busy.Main file="/go/pkg/busy/busy.go:87" CMD=entrypoint PID=1`. I tried deploying emissary from master (without my changes) and it seems to be crashing in the same way. Maybe it's a problem with my cluster set up.

LanceEa: @tomasbanet - Thanks for doing thorough testing on this. As of today, we don't explicitly state support for IPV6_ONLY environments but I think it is important we start making moves towards that. Most importantly, I just want to make sure we don't break the recently added health check feature.

@TimonOmsk - I believe you said that you were running Emissary-ingress in IPV6_ONLY mode, is that correct? Would you be willing to check this PR out and make sure it doesn't break you as you? 

Assuming that it doesn't break anything, then I think we can get this merged in here! Thanks for sticking with this.

kflynn (approved):
Looks good from here! 🙂

LanceEa:
@tomasbanet - Thanks so much for this! I landed https://github.com/emissary-ingress/emissary/pull/4715 with your contributions so it will go out in the next release.

PR #4308, opened 2022-06-27
Improve KAT test reliability

LukeShu:
## Description

This PR aims to cut down on CI flakes.  The main way it does this is by turning backend service deployments in to proper KAT `Node`s with readiness checks (`requirements`) so that the KAT harness can tell whether they're working.  Along the way, there are a few cleanups too.

## Related Issues
I'm kinda surprised we don't seem to have a ticket about KAT reliability. Maybe it's in Notion?

## Testing
That's all this is.

## Checklist

    + Any new features
    + Any deprecations

 - [x] My change is adequately tested. - that's all this is


AliceProxy (dismissed):
Lgtm, thanks for cleaning up the tests :slightly_smiling_face:

.golangci.yml
```diff
@@ -5,7 +5,7 @@ linters:
   enable:
     - gofmt
     - govet
-    - depguard
+    #- depguard # there's a bug in depguard with (or just exacerbated by?) Go 1.19 that causes it to have false positives
```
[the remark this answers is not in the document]
LanceEa: How often is this happening? I saw it once and yes annoying to restart, I would rather leave it on. Because, as we have already shown with t_stats and other things these get forgotten.

Here is the issue in godep that as a potential fix available as of a few days ago:
https://github.com/OpenPeeDeeP/depguard/issues/25

python/kat/harness.py
```diff
@@ -1429,7 +1430,7 @@ def _setup_k8s(self, selected):
                     try:
                         for o in load(n.path, cfg, Tag.MAPPING):
                             parent_config.merge(o)
-                    except YAMLScanError as e:
+                    except (YAMLScanError, YAMLParseError) as e:
```
LanceEa, line 1427: For my understanding:

Does python figure out which type when passing into `Exception` so it can pass friendly messages or is it that since we were only testing for `YAMLScanError` the `YAMLParseError` would get ignored?

python/tests/kat/t_tracing.py
```diff
@@ -144,8 +144,8 @@ def check(self):
         assert self.results[101].backend
         assert self.results[101].backend.name == "raw"
 
-        tracelist = {x: True for x in self.results[101].backend.response}
-
+        tracelist = set(x for x in self.results[101].backend.response)
+        print(f"tracelist = {tracelist}")
```
LanceEa, line 148: Is this going to be too noisy? how big does the set end up being?

LanceEa (commented):
@LukeShu - Thanks looks overall pretty good. I just had a few questions for my understanding. 
The depguard is the only one I would rather we not disable because it seems this project has been plagued by doing that time and time again. Especially, if it is only causing intermittent hiccups (annoying yes).

.golangci.yml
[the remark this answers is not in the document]
LukeShu: In CI, it's not very bad. For actually working on the codebase locally: OMG it happens all the time.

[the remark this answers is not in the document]
LukeShu: In a bit I'll put up a separate PR to turn it back on and bump depguard to that beta release.

python/kat/harness.py
LukeShu: If the exception doesn't match the types listed, then it gets raised as-is; if it does match then we intercept it and turn it in to `Exception("Parse Error: %s, input text:\n%s" % (e, cfg))` and this is useful because "syntax error on line 40" isn't helpful if you can't actually see the YAML that it's talking about.  I suppose there's an argument to be made that we should just always do that and say `except e:` instead of only doing it for certain types.

python/tests/kat/t_tracing.py
LukeShu: The size depends on the test; I think the test with the largest list has 3 items in it.

python/kat/harness.py
LanceEa: ok, that is what I thought but wanted to make sure. I'm not familiar enough with the various errors that could be thrown from load but this is probably sufficient to get by for now.

.golangci.yml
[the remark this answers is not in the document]
LanceEa: Maybe I missed it but do we have the other PR that re-enables this or an issue that tracks it?

ddymko, line 8: Just out of curiosity do we know if this is the case? and if it's been filed against depguard?

ddymko: didn't see @LanceEa comment in https://github.com/emissary-ingress/emissary/pull/4308#discussion_r968606292 my bad

LukeShu:
@LanceEa @ddymko I've added a commit that upgrades to a `golangci-lint` with the depguard fix (references to golangci-lint and depgaurd ticket numbers in the commit message) and un-comments the line from `.golangci.yml`.

LanceEa:
> @LanceEa @ddymko I've added a commit that upgrades to a `golangci-lint` with the depguard fix (references to golangci-lint and depgaurd ticket numbers in the commit message) and un-comments the line from `.golangci.yml`.

Thanks appreciate it!

LanceEa (approved):
lgtm, once CI is green again.

LukeShu:
Resolved merge conflicts in `builder.mk` and `t_tcpmapping.py`.

ddymko (approved):

LukeShu:
Fixed a mistake in resolving the merge conflicts; just had to do a search/replace for `self.name.k8s`→`self.path.k8s`.

ddymko (approved):

LanceEa (approved):

PR #4312, opened 2022-06-28
Add capability to find APIKey filter secrets

knlambert:
## Description

We want to add the possibility to create APIKey filters in edge stack. Although, these keys have to be stored in a secret, so it's required to add their discovery into the emissary code base, along with the one available for OAuth2.

This PR implements the lookup for the APIKeys filters, and the registration of the associated secrets.

## Testing

Deployed in a test cluster & added units tests for this specific case.

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,19 @@ it will be removed; but as it won't be user-visible this isn't considered a brea
 
 ## RELEASE NOTES
 
+## [3.1.0] TBD
+[3.1.0]: https://github.com/emissary-ingress/emissary/compare/v3.0.0...v3.1.0
+
+### Emissary-ingress and Ambassador Edge Stack
+
+- Change: Emissary will register a secret into its internal machinery if it is used by an API Key
+  filter, since the value has to be read at filter creation.
+
+### Ambassador Edge Stack only
+
+- Feature: You can now define a new type of filter to enforce API Keys on top of your services, and
+  provide a Kubernetes secrets as a source for the list of all the keys.
+
 ## [3.0.0] June 27, 2022
 [3.0.0]: https://github.com/emissary-ingress/emissary/compare/v2.3.1...v3.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,19 +72,6 @@ it will be removed; but as it won't be user-visible this isn't considered a brea
 
 ## RELEASE NOTES
 
-## [3.1.0] TBD
-[3.1.0]: https://github.com/emissary-ingress/emissary/compare/v3.0.0...v3.1.0
-
-### Emissary-ingress and Ambassador Edge Stack
-
-- Change: Emissary will register a secret into its internal machinery if it is used by an API Key
-  filter, since the value has to be read at filter creation.
-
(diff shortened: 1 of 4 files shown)


cmd/entrypoint/secrets.go
```diff
@@ -338,47 +339,111 @@ func findFilterSecret(filter *unstructured.Unstructured, action func(snapshotTyp
...
+					secretNamespace = filterNamespace
 				}
 			}
+
 			secretRef(secretNamespace, secretName, false, action)
 		}
 	}
-	return nil
+
+	return
```
LanceEa, line 446: unnecessary `return` since the function has no return type

docs/releaseNotes.yml
```diff
@@ -32,6 +32,22 @@
 
 changelog: https://github.com/emissary-ingress/emissary/blob/$branch$/CHANGELOG.md
 items:
+  - version: 3.1.0
```
LanceEa, line 35: In general, I'm thinking that we do not call this out here because it has no relevance to Emissary-ingress users.

@AliceProxy - thoughts on this?

LanceEa (changes requested):

LanceEa:
@knlambert - overall it looks good but we just need to nail down the design so let's finish that discussion in EdgeStack then the feedback above can be addressed.

knlambert:
@AliceProxy ?

LanceEa (approved):
lgtm - let's get a second review and then it should be good.

CHANGELOG.md
```diff
@@ -72,6 +72,19 @@ it will be removed; but as it won't be user-visible this isn't considered a brea
...
+
+### Emissary-ingress and Ambassador Edge Stack
+
+- Change: Emissary will register a secret into its internal machinery if it is used by an API Key
+  filter, since the value has to be read at filter creation.
+
+### Ambassador Edge Stack only
+
+- Feature: You can now define a new type of filter to enforce API Keys on top of your services, and
+  provide a Kubernetes secrets as a source for the list of all the keys.
```
AliceProxy, line 86: Typo but you can address this in a follow-up PR. Lets get this landed.

docs/releaseNotes.yml
AliceProxy: agreed

docs/releaseNotes.yml
```diff
@@ -32,6 +32,22 @@
...
+    date: 'TBD'
+    notes:
+      - title: Register secrets used with an API Key filter.
+        type: change
+        body: >-
+          Emissary will register a secret into its internal machinery if it is used by
+          an API Key filter, since the value has to be read at filter creation.
+
+    edgeStackNotes:
```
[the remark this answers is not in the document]
AliceProxy: I don't think the release notes are required. That said if we do have edgestack relnotes then putting thme under `edgeStackNotes:` is appropriate although not really done anymore since I don't think the docs render this section anymore.

AliceProxy (approved):
LGTM with a follow-up PR for the typo and release notes removal.

LanceEa (approved):

Issue #4316, opened 2022-07-01
Mapping matches more Hosts than it should with mapping selector


f-herceg:
**Describe the bug**
If I leave `host` field empty in my Mapping then instead of Mapping not matching any Hosts, it matches all of the Hosts.

**To Reproduce**
Create any mapping without setting `host` field.
Create multiple Host resources.
Mapping is available on all hostnames.

**Expected behavior**
Mapping unavailable on any hostname.

**Versions (please complete the following information):**
 - Kubernetes environment: kOps cluster
 - Emissary-ingress: 3.0.0 and also 2.3.1

**Additional context**
MappingSelector, also doesn't seem to work. 
This documentation seems completely wrong https://www.getambassador.io/docs/emissary/latest/topics/running/host-crd/#controlling-association-with-mappings since not setting `host` field in Mapping will match any Host there is in cluster regardless of mappingSelector.

My Mapping:
```
apiVersion: getambassador.io/v2
kind: Mapping
metadata:
  labels:
    app: test-nginx
  name: test-nginx
  namespace: test-nginx
spec:
  host: ""
  prefix: /test/
  service: test-nginx
```
My Host1 which should route to Mapping:
```
apiVersion: getambassador.io/v3alpha1
kind: Host
metadata:
  name: test-nginx-fh-correct
  namespace: emissary
spec:
  hostname: test.something.io
  mappingSelector:
    matchLabels:
      app: test-nginx
  requestPolicy:
    insecure:
      action: Route
status: {}
```
My Host2 which shouldn't route to Mapping but it does.
```
apiVersion: getambassador.io/v3alpha1
kind: Host
metadata:
  name: test-nginx-fh-wrong
  namespace: emissary
spec:
  hostname: wrong.something.io
  mappingSelector:
    matchLabels:
      app: definitely-dont-match-this
  requestPolicy:
    insecure:
      action: Route
status: {}
```

f-herceg:
After trying to debug why this is happening. I've realized that it is expected that if host field in mapping is not defined that default is then `"*"`. This is ok behaviour but I think that it should be better documented somewhere and added to the CRD as default value in my opinion.

It turns out that this is issue with `matches_httpgroup` function in irhost and also `selector_matches` function which is in my opinion terribly written. I have managed to fix it (I hope) and I will create a PR for it. 

[...]

PR #4318, opened 2022-07-05
Fix label selector to match standard kubernetes behaviour 

f-herceg:
Signed-off-by: Filip Herceg <filip.herceg@superbet.com>

## Description
This fixes label matching to match all labels instead of any label. As is the usual behaviour in Kubernetes.

I'm putting this as a fix because I expected this label selector to behave as any other Kubernetes label selector, as mentioned in the documentation.

Docs reference - https://www.getambassador.io/docs/emissary/latest/topics/running/host-crd/#controlling-association-with-mappings
```
The selector is a Kubernetes [label selector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#labelselector-v1-meta), but in 2.0, only matchLabels is supported, for example:
```

## Related Issues
Mapping matches more Hosts than it should with mapping selector #4316 

## Testing
I have tested this by creating multiple `Mappings` with different `metadata.labels` and `hostnames` and I've created multiple `Hosts` with different `mappingSelectors` to validate that behaviour is as would be expected. (mappingSelector to match all labels instead of any label).

## Checklist

 - ☑️ I made sure to update `CHANGELOG.md`.

    + Any new features
    + Any deprecations

 - ☑️  This is unlikely to impact how Ambassador performs at scale.

 - ☑️  I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently.

 - ☑️  The changes in this PR have been reviewed for security concerns and adherence to security best practices.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@ refer both to Emissary-ingress and to the Ambassador Edge Stack.
 
 ## UPCOMING BREAKING CHANGES
 
+### Emissary 3.1.0
+ - Changes to label matching will change how Hosts match Mappings. Hosts were matching any labels found in Mappings, now it's changed to match `all labels`. To avoid unexpected behaviour after the upgrade, add all labels that Hosts have in their mappingSelector to Mappings you want it to match.
+
 ### Emissary 3.0.0
 
  - **No `protocol_version: v2`**: Support for specifying `protocol_version: v2` in `AuthService`,
--- a/python/ambassador/ir/irutils.py
+++ b/python/ambassador/ir/irutils.py
@@ -182,14 +182,13 @@ def selector_matches(
         logger.debug("    no incoming labels => False")
         return False
 
-    selmatch = False
-
+    # For every label in mappingSelector, there must be a label with same value in Mapping itself.
     for k, v in match.items():
         if labels.get(k) == v:
             logger.debug("    selector match for %s=%s => True", k, v)
-            return True
+        else:
+            logger.debug("    selector miss for %s=%s => False", k, v)
+            return False
 
-        logger.debug("    selector miss on %s=%s", k, v)
-
-    logger.debug("    all selectors miss => False")
-    return False
+    logger.debug(f"    all selectors match => True")
+    return True


alexgervais:
@f-herceg First of all, thanks for your contribution!
We had a discussion during the last community meeting. Although we all seem to be in favour of your proposed change, we feel changing the behavior in a minor version of Emissary will cause some headaches. We would instead like to make this change happen in the next CRD version update (`v3alpha2` or `v3beta1`). This would give more control to the users rolling out changes while not forcing them to update existing resources before rolling out an update of Emissary.
Happy to hear your thoughts on this!

f-herceg:
@alexgervais I don't know that well how CRD versions are handled in cases such as this. My concern is that this change does not impact the CRD itself but rather how CRD is handled in code. If it's possible to have one way of handling the selector for `v3alpha1` and another type of handling the selector for `v3alpha2` then I would say that would be the best course of action. If it's only one way of handling the selector regardless of the CRD version then I would say that releasing `v3alpha2` or `v3beta1` would be pointless.

I wasn't at the community meeting so I don't know what the general feeling on this matter is but for me, this is more like a fix to me because the selector isn't behaving the way I would expect and the way it's documented with the existing implementation. I'm obviously biased on this because I would like to start using this feature. But as I said, I wasn't at the community meeting so I don't have the full context and we need to take into consideration everyone that has already shaped their setup with the existing behaviour in mind.

LanceEa:
@f-herceg - Thanks for this again. As @alexgervais mentioned since this has been a known behavior for awhile now we don't want to treat this as a bugfix because it would probably break a lot of people depending on that behavior.

One way we are thinking of handling this is by introducing a new field when we go to the next CRD version so that old behavior and new behavior can live side-by-side. The old behavior would get deprecated and eventually removed allowing time for users to convert over to the new set of fields.

We haven't finalized the work for the next CRD version yet but it is top of mind and we appreciate the PR. Once we have something more concrete would you be open to giving feedback?

f-herceg:
@LanceEa I like the idea you proposed. I would be happy to provide feedback :)

underrun:
TL;DR: trusting your docs here without both the change here and in #4499 leads at best to incorrect behavior and at worst to a serious security hole. please treat this with critical urgency as a security issue including backports to all 2.x minor version if at all possible.

as noted in this PR #4499 fixing the problematic behavior fully also requires a change in irhost.py to only match hosts and routes if both the host and the selectors (if provided) match.

this behavior could subject users of emissary to a high severity security problem in how it diverges from the described behavior in the documentation (both in the docs and in comments in the code).

to illustrate, if a user configures:
* two listeners different ports (let's call one tls and one mtls)
* two different upstream services one to handle requests that come in through each listener
* two hosts (one for each listener) with the same hostname but different selectors for mappings
* two mappings (one for each upstream) with different labels
  * where the mapping that is associated with the host selected by the mtls listener routes to a service that expects client certs to have been required at the edge (or to have had other auth measures applied)
  * where the mapping that is associated with the host selected by the tls listener does not require client certs or further auth

given this deployment, users would be able to access the supposedly mtls protected upstream service by coming in through the tls listener's port with any client certs (because they are not verified by envoy at all).

this is because routes for all mappings end up associated with every host with a hostname match even if the hosts specify different selectors. the bug this PR addresses fixes the issue where mapping selectors don't prevent routes being added if the mappings share even one label that matches that is included in the selector. but #4499 is also needed to prevent selectors from being ignored if there is a hostname match.

to be very clear we consider this a very serious bug with emissary for any use case where users expect to be able to route traffic from different ports on the same hostname (not an uncommon pattern at all) to different upstream services especially where those ports have different levels of authentication at the edge (in which case it's a serious security bug).

AliceProxy:
I agree that we need to address this now. It doesn't make sense to wait until a future CRD version bump with an unknown date when there are ways to address the problem now without modifying the CRDs in any way. I'm going to pull the commits from this PR into a new branch along with the changes to `irhost.py` from #4499 and add an environment variable to opt-out of the new (correct) behaviour just in-case there are any users who are both reliant on the current behaviour and also wish to continue using it. A future version of Edgissary can remove the ability to opt-out using the environment variable.

From there I will merge this PR along with #4499 into master and then my own changes following that. We are already planning to release versions `2.4` and `3.2` next week, so I will cherry-pick these commits back into the branch for `2.4` so that both the 3.x and 2.x release lines will get this fix, but porting these changes back on top of all older versions such as 2.1, 2.2, etc. is not something we currently plan on doing.

LanceEa:
@AliceProxy - per our discussion offline I agree with your call and approach on this.

AliceProxy:
Just a quick update: since `2.4` has already been released, these changes will re-target an upcoming `2.5` release to get these fixes into a 2.x release. We still hope to get them in for the version `3.2` releasing next week. Please refer to https://github.com/emissary-ingress/emissary/pull/4506 for any other details or if you have additional feedback.

PR #4320, opened 2022-07-06
Rumsrami/dev/diagnostics

rumsrami:
## Description
Added support for the emissary agent to query admin for diagnostics data and a stream diagnostics grpc endpoint

## Related Issues
List related issues.

## Testing
still in draft, not well tested yet

## Checklist

    + Any new features
    + Any deprecations


cmd/agent/main.go
```diff
@@ -52,6 +56,11 @@ func run(cmd *cobra.Command, args []string) error {
 		snapshotURL = fmt.Sprintf(DefaultSnapshotURLFmt, entrypoint.ExternalSnapshotPort)
 	}
 
+	diagnosticsURL := os.Getenv("AES_DIAGNOSTICS_URL")
```
alexgervais, line 59: Would it be possible to add an environment variable (or rejigger this one) to completely disable diagnostic reporting? It might be a feature some users want disabled. Think of it as a feature flag?

pkg/agent/agent.go
```diff
@@ -113,6 +116,16 @@ type Agent struct {
 
 	// Extra headers to inject into RPC requests to ambassador cloud.
 	rpcExtraHeaders []string
+
+	// diagnostics reporting
+	diagnosticsReportingStopped bool // Director stopped diagnostics reporting
+	//minDiagnosticsReportPeriod  time.Duration // How frequently do we collect diagnostics
```
AliceProxy, line 122: nit: comments should begin with a space

pkg/agent/agent.go
```diff
@@ -113,6 +116,16 @@ type Agent struct {
...
+
+	// The state of reporting
+
+	//diagnosticsReportToSend   *agent.Diagnostics // Diagnostics report that's ready to send
```
AliceProxy, line 126: `diagnosticsReportToSend *agent.Diagnostics` meant to be left commented out and `// Diagnostics report that's ready to send` belongs on the following line?

pkg/agent/agent.go
```diff
@@ -631,6 +745,13 @@ func (a *Agent) ProcessSnapshot(ctx context.Context, snapshot *snapshotTypes.Sna
 	return nil
 }
 
+// ProcessDiagnosticsSnapshot translates ambassadors diagnostics into streamable agent diagnostics
+func (a *Agent) ProcessDiagnosticsSnapshot(ctx context.Context, snapshot *diagnosticsTypes.Diagnostics, ambHost string) (*agent.Diagnostics, error) {
+	// implement
```
g-vega-cl, line 750: Is this on purpose?

pkg/diagnostics/v1/types.go
```diff
@@ -0,0 +1,301 @@
...
+	Kind         string   `json:"kind"`
+	Location     string   `json:"location"`
+	Name         string   `json:"name"`
+	Namespace    string   `json:"namespace"`
+	SecretInfo   struct {
+		CertChainFile  string `json:"cert_chain_file"`
+		PrivateKeyFile string `json:"private_key_file"`
+		Secret         string `json:"secret"`
+	} `json:"secret_info"`
+}
```
g-vega-cl, line 301: Amazing work at writing this structure.
I see `hColor` , `service` , and `weight`, `notices` But I can't find `key`.
As in `"key": "http://staging-app.datawire.io/download/tel2/windows/amd64/nightly/telepresence.zip",`
I think it should be in the Groups struct. (Note, "key" is different from `rKey` and `cache_key`

g-vega-cl, line 301: Amazing work at writing this structure.
I see `hColor` , `service` , and `weight`, `notices` But I can't find `key`.
As in `"key": "http://staging-app.datawire.io/download/tel2/windows/amd64/nightly/telepresence.zip",`
I think it should be in the Groups struct. (Note, "key" is different from `rKey` and `cache_key`

AliceProxy:
Overall looking good. Two things that I wanted to mention:
1. It is important to sign your commits in emissary ingress. You can rebase your commits to sign them or I can force the DCO bot to pass for this PR since you're with Ambassador labs and we trust you. That said, even though we can force DCO it is a good practice to get into especially for future PRs.

2. I'd recommend configuring your editor of choice to respect the editorconfig of whatever project you're working on. If you use VSCode then this extension will do that for you https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig.

pkg/agent/agent.go
rumsrami: yep, still ongoing

pkg/diagnostics/v1/types.go
rumsrami: will add it

pkg/agent/agent.go
rumsrami: it will be removed. i was trying to follow how the snapshots were done, but decided to divert and not add the diagnostics report to the struct, instead pass it explicitly to the reporting function

alexgervais (approved):
One small question we can address in a subsequent PR.

Happy to merge this!

cmd/agent/main.go
```diff
@@ -52,6 +56,16 @@ func run(cmd *cobra.Command, args []string) error {
...
+	if diagnosticsURL == "" {
+		diagnosticsURL = fmt.Sprintf(DefaultDiagnosticsURLFmt, AdminDiagnosticsPort)
+	}
+
+	reportDiagnostics := os.Getenv("AES_REPORT_DIAGNOSTICS_TO_CLOUD")
```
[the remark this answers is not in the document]
AliceProxy: I think the default should be to not report diagnostics to the cloud due to concerns from the types of customers I mentioned in my previous message. Then the `helm upgrade` commands in the cloud that have you setup a cloud connect token can set the environment variable to `opt-in`. That said if there is a more compelling reason to have the default be `opt out` I don't think it will be too much of a problem so long as these changes are included in the Edge-Stack release notes so that any customers who care can preemptively set the variable to `opt-out` when they upgrade.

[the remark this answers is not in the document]
AliceProxy: `reportDiagnostics := os.Getenv("AES_REPORT_DIAGNOSTICS_TO_CLOUD")` should also probably take whatever default value we want to go with whenever `AES_REPORT_DIAGNOSTICS_TO_CLOUD` is not set so we don't mess up the customers using YAML installs.

AliceProxy (approved):
lgtm with a change that @alexgervais already pointed out that can be in a follow-up PR.

cmd/agent/main.go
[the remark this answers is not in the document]
ddymko: Will add defaults in the follow-up PR once we decide which route to take

PR #4324, opened 2022-07-08
Synthetic v3 RateLimitService for Edge-Stack

ddymko:
Signed-off-by: David Dymko <ddymko@datawire.io>

## Description
This adds a pre-compiler hook for injecting a synthetic `RateLimitService` for Edge-Stack similar to what exists for the `AuthService`.  This is required because Edge-Stack only allows for a single RateLimitService to be available.

A synthetic RateLimitService will be injected in the following scenarios:
1. no RateLimitService provided
2. invalid RateLimitService (bad settings, doesn't exist, doesn't point at localhost)

If a valid RateLimitService is provided that points at `127.0.0.1:8500` then the provided one will be used to ensure we capture custom settings from users.. If the service does not provide the `protocol_version` field or it  is NOT set to `v3` then the service will automatically be upgraded to `v3`.

This is to ensure Edge-Stack always has a single RateLimitService available that speaks v3 protocol and points to the EdgeStack provided service.

## Related Issues
None

## Testing
Added integration tests to ensure the proper envoy config is generated.

## Checklist

### DIFF
--- a/cmd/entrypoint/syntheticratelimit.go
+++ b/cmd/entrypoint/syntheticratelimit.go
@@ -1,0 +1,113 @@
+package entrypoint
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/datawire/dlib/dlog"
+	"github.com/emissary-ingress/emissary/v3/pkg/api/getambassador.io/v3alpha1"
+	"github.com/emissary-ingress/emissary/v3/pkg/kates"
+)
+
+func iterateOverRateLimitServices(sh *SnapshotHolder, cb func(
+	rateLimitService *v3alpha1.RateLimitService, // rateLimitService
+	name string, // name to unambiguously refer to the rateLimitServices by; might be more complex than "name.namespace" if it's an annotation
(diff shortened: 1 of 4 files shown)


LanceEa:
Note: a change log entry will be made when doing release later this week.

LanceEa (approved):
lgtm

AliceProxy (changes requested):
lgtm but we should add a changelog for this since it is meant to reduce upgrade friction :+1:

AliceProxy:
> lgtm but we should add a changelog for this since it is meant to reduce upgrade friction +1

> Note: a change log entry will be made when doing release later this week.

Whoops nevermind then, lets get this landed.

AliceProxy (approved):

LanceEa:
@AliceProxy - here or on EdgeStack side?

AliceProxy:
> @AliceProxy - here or on EdgeStack side?

@LanceEa Here, but also for edge-stack. Ideally edge-stack.git's release notes should be kept up to date with emissary.git since changes to  Emissary impact Edge-Stack users.

PR #4325, opened 2022-07-08
Add a secret sync directive to the agent

knlambert:
Signed-off-by: Kévin Lambert <kevin.lambert.ca@gmail.com>

## Description

We want to support the sync of a secret from a third party (here ambassador cloud) to automate the creation of Filters.

This PR adds:
* A new field to the directive object, along the one for the rollouts for this use case.
* A handler to add / remove keys from the secret.
* Edit the chart to make it work. Note i've edited an existing role, I think I should rename it, but the CI prevents me from doing it.

I will probably need some hints about how we should manage the version of all of this, especially regarding system-a :)

## Testing

Tested it by deploying it in a kubernetes cluster & interrogating the service to validate it works properly.
Added also a bunch of unit tests to validate the behavior of the command.

## Checklist

    + Any new features
    + Any deprecations


api/agent/director.proto
```diff
@@ -129,6 +130,19 @@ message RolloutCommand {
   string command_id = 4;
 }
 
+message SecretSyncCommand {
```
LanceEa, line 133: Would we have a specific Result message? I didn't see one...

docs/releaseNotes.yml
```diff
@@ -32,6 +32,16 @@
...
 changelog: https://github.com/emissary-ingress/emissary/blob/$branch$/CHANGELOG.md
 items:
+  - version: 3.1.0
+    date: 'TBD'
+    notes:
+      - title: Add support for secrets sync
+        type: feature
+        body: >-
+          Adding an option to the directive service in the agent to synchronize secrets. The objective is to allow
+          a third party product to automate the creation of CRDs depending on a secret to work
```
LanceEa, line 42: suggestions: 
- use "command" instead of "option" since that is what it is called.
- "synchronize" gives the impression that it is two way. Maybe use "manage" secrets.
- remove "The objective" to more direct language like "This allows"

Maybe something like so:

> Adds a new directive command to the $productName$ Agent service that allows it to manage secrets. This allows a third party product to manage $productName$ CRD's that depend upon a secret.

pkg/agent/directive_handler.go
```diff
@@ -49,15 +50,70 @@ func (dh *BasicDirectiveHandler) HandleDirective(ctx context.Context, a *Agent,
...
+
+	err := cmd.RunWithClientFactory(ctx, dh.secretsGetterFactory)
+	if err != nil {
+		dlog.Errorf(ctx, "error running secret sync command %s: %s", cmd, err)
+	}
+
+	dh.reportCommandResult(ctx, commandID, cmd, err, a)
+
+}
+func (dh *BasicDirectiveHandler) handleRolloutCommand(
```
LanceEa, line 112: nit: put space between end of previous func and new func

pkg/agent/directive_handler.go
```diff
@@ -90,10 +146,17 @@ func (dh *BasicDirectiveHandler) handleRolloutCommand(ctx context.Context, cmdSc
...
 	dh.reportCommandResult(ctx, commandID, cmd, err, a)
 }
 
-func (dh *BasicDirectiveHandler) reportCommandResult(ctx context.Context, commandID string, cmd *rolloutCommand, cmdError error, a *Agent) {
+type command interface {
+	String() string
+}
+
+func (dh *BasicDirectiveHandler) reportCommandResult(
+	ctx context.Context, commandID string, cmd command, cmdError error, a *Agent,
```
LanceEa, line 158: Rather than introducing a custom interface, would just using the standard `Stringer` interface be sufficient?

Alternatively, a more heavy handed approach would be to update the protocol buffers to have a "BaseCommand" message type that was extended and then use that. But, not sure that is necessary at this time.

pkg/agent/directive_handler_test.go
```diff
@@ -18,3 +19,33 @@ func TestHandleDirective(t *testing.T) {
...
 }
+
+func TestHandleSecretSyncDirective(t *testing.T) {
+	// given
+	ctx := dlog.NewTestContext(t, false)
+
+	a := &agent.Agent{}
+	dh := &agent.BasicDirectiveHandler{}
+
+	d := &agentTypes.Directive{
```
LanceEa, line 30: Some additional test coverage would be nice:

1. Set Action with multiple key/value pairs
2. Set Action with no key/value pairs
3. Set command with missing secret, namespace/name
3. Del Action for secret that exists
4. Del Action for secret that does not exist

I would also think some test with Roles missing as well.

pkg/agent/directive_handler_test.go
```diff
@@ -18,3 +19,33 @@ func TestHandleDirective(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}
+
+	// when
+	dh.HandleDirective(ctx, a, d)
+
+	// then
```
LanceEa, line 50: I see this is copying a previous test that already exists but how are we asserting anything to be true? Is it just running this and then manually checking a cluster or ambassador cloud?

pkg/agent/secrets.go
```diff
@@ -0,0 +1,159 @@
...
+func (s *secretSyncCommand) getOps() (ops []map[string]string) {
+	switch s.action {
+	case secretSyncActionDelete:
+		for key := range s.secret {
+			ops = append(ops, map[string]string{
+				"op":   "remove",
+				"path": fmt.Sprintf("/data/%s", key),
+			})
+		}
+	default:
```
LanceEa, line 65: Since this is using an alias on a string there is nothing stopping someone from creating a `secretSyncCommannd` with action "foo". 

Therefore, we should explicitly catch the case `secretSyncActionSet` and then the default case should probably `panic` because that would mean we have a programming error somewhere and we would want to catch that in CI.

pkg/agent/secrets.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to update the secret: %w", err)
+	}
+
+	// if no keys left, we should delete the secret.
+	if len(secret.Data) == 0 {
+		err := client.Delete(ctx, s.name, metav1.DeleteOptions{})
+
+		if err != nil {
+			return fmt.Errorf("failed to clean up the secre %s: %w", s.name, err)
```
LanceEa, line 129: typo: `secre` --> `secret`

pkg/agent/secrets.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+	}
+
+	return nil
+}
+
```
LanceEa, line 135: Exported function should have comment

pkg/agent/secrets.go
```diff
@@ -0,0 +1,159 @@
...
+func NewSecretsGetter(namespace string) (secretInterface, error) {
+	kubeConfig, err := newK8sRestClient()
+	if err != nil {
+		return nil, err
+	}
+
+	clientSet, err := kubernetes.NewForConfig(kubeConfig)
+	if err != nil {
+		panic(err.Error())
```
LanceEa, line 144: I'm guessing we want this to happen because it should never happen unless something was really messed up with the cluster as a whole?

pkg/agent/directive_handler_test.go
LanceEa: After looking at `secrets_test.go` some of these are already covered, correct?

pkg/agent/rollouts.go
```diff
@@ -142,14 +139,3 @@ func NewArgoRolloutsGetter() (argov1alpha1.RolloutsGetter, error) {
 
 	return argoClient, nil
 }
-
-func newK8sRestClient() (*rest.Config, error) {
```
LanceEa, line 146: It looks like you moved this to `secrets.go`. Rather than do that should it move to a new file or probably the `utils.go` since it is shared in the package?

LanceEa (dismissed):
In general, it looks inline with what we did with Rollouts. I have left some feedback and comments. 

Definitely want to get a second set of eyes on the permission changes for the roles and want to make sure we have the correct test coverage.

api/agent/director.proto
```diff
@@ -41,10 +41,10 @@ service Director {
 // no user account specific information should be contained in here
 message Identity {
   // The account ID assigned by the DCP
-  string account_id = 1 [deprecated=true];
+  string account_id = 1 [deprecated = true];
```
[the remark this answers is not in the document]
knlambert: Oh I'm not. That's just the IDE formating the fields.

api/agent/director.proto
knlambert: The commands arrive by the `Retrieve` method which is a stream which is always open. 

We don't have a result because it's asynchronous. The commands could take some time to execute, and arrive un bulk. 

Although, when a command is finished, the status is reported with a `CommandResult` object via the method `ReportCommandResult`.

charts/emissary-ingress/templates/ambassador-agent.yaml
```diff
@@ -65,6 +65,9 @@ rules:
 - apiGroups: [""]
   resources: [ "configmaps" ]
   verbs: [ "get", "list", "watch" ]
+- apiGroups: [""]
```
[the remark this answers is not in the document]
knlambert: - Yes, the roles are either global or by namespace. When it's by namespace, you can't do anything out of it.
- The agent only creates secrets. The API Keys & Filters are created by the end user using GitOps or manual creation.
- We were planning to only create filters in the ambassador namespace, but I don't know if we should support filters created in a different namespace by the end user. In this case, you're right, this permission may not be enough.

pkg/agent/secrets.go
knlambert: Yeah, because the library discovers by itself what are the available connection methods in the pod. Usually that's the kubernetes token mounted on a specific path in the pod.

pkg/agent/rollouts.go
knlambert: Move to k8s.

pkg/agent/directive_handler_test.go
knlambert: 1. Data is passed directly to the function, so doesn't change anything, but I've updated the test case to reflect that.
2. Added test with an empty map, should be allowed though if empty. Not for nil (it will fail in the nil case)
3. It's handled by the directive handler.
4. There is a test case in secrets_test.
5. There is a test case in secrets_test. It doesn't return any error.

pkg/agent/directive_handler_test.go
knlambert: Actually I thought I could actually test something, but the reality is that it is badly designed. I can't replace Agent for an interface or anything which would allow me to test its behavior.

It could for sure be improved, but I don't think that's the purpose of this PR.

charts/emissary-ingress/templates/ambassador-agent.yaml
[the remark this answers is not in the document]
LanceEa: In general, we own the Ambassador namespace so its best if users place stuff in their own namespace. I would think if we were creating the filters on their behalf via the cloud then maybe it makes sense to put them there or some new namespace specific to this use case. 

But in this case if we are creating the `secrets` in our namespace then based on the recommendation of removing the `namespace` from the `APIKeyFilter` will not work if a user creates a Filter in another namespace, correct? Because, we require the secret to be in the same namespace?

pkg/agent/directive_handler_test.go
LanceEa: Ok, that is fine. I don't want to block this but I think that definitely is an area we should return to later to improve our test story here.

LanceEa (commented):
For me the big open question is where are these secrets going to get stored and how are they going to be used.

knlambert:
> For me the big open question is where are these secrets going to get stored and how are they going to be used.

The only place these secrets are stored is in this kubernetes secret. We don't keep them in database in the cloud product or anything. We only add or remove key to the secret :)

api/agent/director.proto
```diff
@@ -129,6 +130,19 @@ message RolloutCommand {
...
+  string name = 1;
+  string namespace = 2;
+  string command_id = 3;
+
+  enum Action {
+    SET = 0;
+    DELETE = 1;
+  }
+  Action action = 4;
+  map<string, bytes> secret = 5;
```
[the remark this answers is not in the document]
LanceEa: Ok thanks for the detailed update and it makes more sense now. I still find it odd to be a map of always one item but understand what you are doing 😄

charts/emissary-ingress/templates/ambassador-agent.yaml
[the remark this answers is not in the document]
LanceEa: understood

LanceEa:
> > For me the big open question is where are these secrets going to get stored and how are they going to be used.
> 
> The only place these secrets are stored is in this kubernetes secret. We don't keep them in database in the cloud product or anything. We only add or remove key to the secret :)
> 
> Typical workflow is :
> 
> * We give the end user the command to apply the filter (and they does).
> * They creates an API Key in the cloud interface, and keep the API Key for themself.
> * The cloud transmits the key to the agent, which adds it as an entry in the secret used by the filter (and creates the secret if it if it doesn't exist).
> * When the end user revokes the API Key, we remove the entry from the secret.
> 
> That's it.

Ok the part that is tripping me up, that I guess I want to clarify is that in this other PR https://github.com/emissary-ingress/emissary/pull/4312/files#diff-ff02720b66a6ee517bcd28beddc007b390071e1ca4a83d3d6bacc5e0883cc3e3R398 we are saying we will only load secrets if the Filter is in the same namespace as the Filter. 

So, does this mean the agent needs access to create secrets in all namespaces or are we just saying to the customer we will create the secret with APIKeys within the `ambassador` namespace and you must setup your gitops to add the Filter to that namespace too?

LanceEa:
BTW...you also have a merge conflict now due to a recent merge into master

knlambert:
> So, does this mean the agent needs access to create secrets in all namespaces or are we just saying to the customer we will create the secret with APIKeys within the ambassador namespace and you must setup your gitops to add the Filter to that namespace too?

It only needs access to create secrets in the ambassador namespace (except if you think we should anticipate that it may be in others), and the clients will configure their GitOps to create the filters in that ambassador namespace too.

LanceEa:
> It only needs access to create secrets in the ambassador namespace (except if you think we should anticipate that it may be in others), and the clients will configure their GitOps to create the filters in that ambassador namespace too (since the secrets are there).

That is fine for a v0 implementation but I do believe that users will want more flexibility in the future. The purpose of most of the Emissary-ingress CRD's is to empower the end developer and provide them with flexibility without having to involve a platform engineer. I could see where platform engineers are more strict about letting developers add things to the `ambassador` namespace while requiring them to create Mappings, Filters, Hosts, etc... in their developer specific namespace. 

However, to enable this you would have to give the Agent access to manage secrets in all namespaces and the cloud interface would have to add the secret to the correct namespace that the filter was added to so this could be a source of friction due to Filters being added to different namespaces.

So, just make sure the v0 implementation is well documented so users understand the current restrictions.

pkg/agent/secrets.go
```diff
@@ -60,14 +60,16 @@ func (s *secretSyncCommand) getOps() (ops []map[string]string) {
...
+	case secretSyncActionSet:
 		for key, value := range s.secret {
 			ops = append(ops, map[string]string{
 				"op":    "add",
 				"path":  fmt.Sprintf("/data/%s", key),
 				"value": string(value),
 			})
 		}
+	default:
+		panic(fmt.Sprintf("action %s is not supported by the secret sync directive", s.action))
```
LanceEa, line 81: Do we really want the agent crashing when this happens?

pkg/agent/secrets.go
```diff
@@ -21,15 +19,15 @@ const (
 	secretSyncActionDelete = secretSyncAction("DELETE")
 )
 
-type secretInterface interface {
+type SecretInterface interface {
```
LanceEa, line 22: Why is this being exported now? 

If it needs to be exported then it should have doc comments

LanceEa (commented):
a few last questions on the recent changes and then I want a second review on this. I will ping someone on the team to take a second look.

pkg/agent/secrets.go
knlambert: Because otherwise `func NewSecretsGetter(namespace string) (secretInterface, error)` returns a private interface, and it triggers a check from the IDE saying `Exported function with the unexported return type`

LanceEa:
:( changelog merge conflicts due to merging your other PR. Fix that up and once we a second review then I will be good on this one.

LanceEa:
Looks like maybe you need to re-run `make generate` after a rebase?

LanceEa (approved):
lgtm...I want to get a second set of eye's on the RBAC changes and then we should be good.

alexgervais (dismissed):
I agree with @LanceEa, this is fine for a v0 implementation but users will want more flexibility in the future. Given Edge Stack already has a ClusterRole to read and write secret values (acme tls), I don't foresee pushback from adding the same permission to the agent component.

LGTM, if:
- The secret namespace limitation is documented in the public docs.
- We capture a story future consideration to remove the limitation.

knlambert:
> I agree with @LanceEa, this is fine for a v0 implementation but users will want more flexibility in the future. Given Edge Stack already has a ClusterRole to read and write secret values (acme tls)

Then, would it be better to just add the agent to this role ? Then we'll be able to update the code to create filters in other namespaces if it's needed without having to release again.

alexgervais:
@knlambert I would keep both roles distinct. It's nice to have the agent role be different than Emissary (especially with what's coming up next cycle). We'd have to add new rules to the existing agent role.

LanceEa:
Does it make more sense to break new commits up into new PR's? IMO, we had already approved this so personally I would have preferred the latest commit to be in a new PR.

pkg/agent/directive_handler.go
```diff
@@ -90,11 +90,6 @@ func (dh *BasicDirectiveHandler) handleSecretSyncCommand(
 		return
 	}
 
-	if secret == nil && cmdSchema.Action == agentapi.SecretSyncCommand_SET {
```
LanceEa: Why is this ok now? I would have thought the secret would need to be required when executing the Set Command?

knlambert: Hello Lance, sorry about that, we should be ok now, we identified a last use case where we need the secret to be empty, and even if we give an empty map which is not `nil`, GRPC converts that into a `nil value`. That's why I've removed this condition.

pkg/agent/secrets.go
```diff
@@ -52,28 +52,27 @@ func (s *secretSyncCommand) RunWithClientFactory(
...
+func (s *secretSyncCommand) getOps(insertRoot bool) ([]map[string]interface{}, error) {
+	var ops = make([]map[string]interface{}, 0)
 	// if the secret is empty, this is required.
 	if insertRoot {
-		ops = append(ops, map[string]string{
+		ops = append(ops, map[string]interface{}{
 			"op":    "add",
 			"path":  "/data",
-			"value": "{}",
+			"value": map[string]interface{}{},
```
knlambert, line 62: I've changed this for interface{} because it needs to be an object if the value is {}. "{}" won't work.

LanceEa (approved):
lgtm

LanceEa (approved):

PR #4343, opened 2022-07-21
backport: Fix cooldown feature to properly manage several instances of emissary

knlambert:
## Description

_This commit comes from [the version 3](https://github.com/emissary-ingress/emissary/pull/4293) and is a backport._

----

Previously, we added a cooldown step (30s) for the metrics[ to avoid to send too many requests](https://github.com/emissary-ingress/emissary/pull/4122) to the ambassador cloud API.

Although, some values were not consistent, and we figured out that the way the metrics are sent to the agent are **one request per emissary node**, and not **one for all**.

It means that the previous iterations of the code were wrong for two reasons : 
* If we forward the metrics directly to the cloud, we won't be able to identify what node corresponds to what metric.
* If we put a simple cooldown of 30s, for 3 nodes, we'll miss 2 out of 3 sets of metrics.

As a workaround, with minimal changes, this PR changes the logic to make the metric relay handler to accumulate metrics in a map per nodes' IP, and unload them after the cooldown period.

image

## Testing

Adapted existing tests, & tested it in a dev cluster.

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,15 @@ it will be removed; but as it won't be user-visible this isn't considered a brea
 
 ## RELEASE NOTES
 
+## [2.3.2] TBD
+[2.3.2]: https://github.com/emissary-ingress/emissary/compare/v2.3.1...v2.3.2
+
+### Emissary-ingress and Ambassador Edge Stack
+
+- Bugfix: A regression was introduced in 2.3.0 causing the agent to miss some of the metrics coming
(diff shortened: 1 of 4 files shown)


LanceEa:
CI isn't running so I'm going to close and re-open to see if it re-triggers it.

LanceEa:
opening to hopefully trigger CI to run.

LanceEa:
Ok that worked, CI is running. Once green we can get it merged 😄

LanceEa (approved):
lgtm

PR #4346, opened 2022-07-22
[v3.0] Repatriate from v2.3

d6e-automaton:
This PR merges release/v2.3 in to master.


LanceEa:
Closing because these commits are all back ports from master and do not need to be repatriated.

PR #4355, opened 2022-07-27
deps: pin python3 to 3.9.13 patch version

LanceEa:
## Description
Alpine 3.15 removed 3.9.7 from the main registry index.
This updates the pin to the latest patch release in apk 3.15

## Related Issues
N/a

## Testing
CI will go green

## Checklist


LanceEa:
Address #4356

Note: We may want to adjust the pin to just be `=~3.9` so that it takes all patches and not just the 3.9.13 but for the PR I wanted to match what is already landed in the release/v2.3 branch.

Once we are finished with the release we can re-visit the pin if it is still causing issues.

PR #4357, opened 2022-07-27
deps: pin curl in optimized container

LanceEa:

## Description
We previously pinned curl in the base-python but in `builder/Dockerfile`
we redownload curl using apk without
it being pinned.

This ensures we are pulling the curl version into the optimized
container and the base-python container.

**Note** - this will need to be cherry-picked to `rel/v2.3.2` so that it will get merged back into the `release/v2.3` after release.

## Testing
CI green and build image and execute into it to verify correct curl version.

## Checklist

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -79,7 +79,7 @@ ADD go.sum go.sum
 
 RUN --mount=type=cache,target=/root/.cache/go-build \
     mkdir -p /go/bin && \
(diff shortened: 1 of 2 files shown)


LanceEa:
@haq204 - CI should go green now that the base-python has been generated. I'm re-running `make generate` locally to make sure there are no dirty changes. I think we should be good but just double checking 😄 .

haq204 (approved):

PR #4366, opened 2022-07-31
Fix cluster naming for really long service names

### LINKED ISSUE #4354: Envoy cluster creation issue with conflicting service names
**Describe the bug**
When emissary is creating envoy clusters, emissary is evidently constrained by a character limit for the length of the cluster and must handle potential naming conflicts. If a mapping and service are named identically in different namespaces, and the name of the service is over a certain number of characters, there seems to be a condition that causes envoy cluster creation for one of those mappings to fail. 

**To Reproduce**
1. Create a service that maps to a pod with a name that is over 30 characters or so, but in different namespaces. I've reproduced this issue in minikube with the following manifest.
```
apiVersion: v1
kind: Service
metadata:
  name: hello-this-name-is-too-long-for-emissary
  namespace: alpha
spec:
  ports:
    - port: 80
      targetPort: 80
      protocol: TCP
  selector:
    app: hello-alpha
---
apiVersion: v1
kind: Service
metadata:
  name: hello-this-name-is-too-long-for-emissary
  namespace: beta
spec:
  ports:
    - port: 80
      targetPort: 80
      protocol: TCP
  selector:
    app: hello-beta
```
I have a simple nginx pod in each namespace with `hello-alpha` and `hello-beta` as the matching label selector.

2. Create a mapping in the beta namespace:

```
kind: Mapping
metadata:
  name: hello
  namespace: beta
spec:
  host: "*"
  prefix: /hello-beta
  service: http://hello-this-name-is-too-long-for-emissary
```

The following route will be created in envoy according to the debug GUI

```
{
  "match": {
    "case_sensitive": true,
    "prefix": "/hello-beta",
    "runtime_fraction": {
      "default_value": {
        "denominator": "HUNDRED",
        "numerator": 100
      },
      "runtime_key": "routing.traffic_shift.cluster_http___hello_this_name_is_too_lo-0"
    }
  },
  "route": {
    "cluster": "cluster_http___hello_this_name_is_too_lo-0",
    "prefix_rewrite": "/",
    "timeout": "3.000s"
  }
}
```

3. Create a similar mapping but in the `alpha` namespace. 
```
kind: Mapping
metadata:
  name: hello
  namespace: alpha
spec:
  host: "*"
  prefix: /hello-alpha
  service: http://hello-this-name-is-too-long-for-emissary
```

Once the mapping is created, it will now "take over" the envoy cluster with the `-0` suffix, and the endpoints for this cluster will be updated to point to the alpha pod. 
 
```
{
  "match": {
    "case_sensitive": true,
    "prefix": "/hello-alpha",
    "runtime_fraction": {
      "default_value": {
        "denominator": "HUNDRED",
        "numerator": 100
      },
      "runtime_key": "routing.traffic_shift.cluster_http___hello_this_name_is_too_lo-0"
    }
  },
  "route": {
    "cluster": "cluster_http___hello_this_name_is_too_lo-0",
    "prefix_rewrite": "/",
    "priority": null,
    "timeout": "3.000s"
  }
}
```
However, the mapping for beta will still be pointing to the envoy cluster named `cluster_http___hello_this_name_is_too_lo-0`. This causes traffic destined for the beta instance to be routed to the alpha instance instead.

Before creating the alpha mapping, if I hit the ClusterIP of the emissary-ingress service, requests for `/hello-beta` are routed to the beta pod correctly.

```
 testpod ~ $ curl -s emissary-ingress.emissary.svc.cluster.local/hello-alpha | grep name
<p><span>Server&nbsp;name:</span> <span>hello-alpha</span></p>
 testpod ~ $ curl -s emissary-ingress.emissary.svc.cluster.local/hello-beta | grep name
<p><span>Server&nbsp;name:</span> <span>hello-beta</span></p>
```

After creating the alpha mapping, requests for `/hello-beta` are send to the `hello-alpha` pod

```
 testpod ~ $ curl -s emissary-ingress.emissary.svc.cluster.local/hello-alpha | grep name
<p><span>Server&nbsp;name:</span> <span>hello-alpha</span></p>

 testpod ~ $ curl -s emissary-ingress.emissary.svc.cluster.local/hello-beta | grep name
<p><span>Server&nbsp;name:</span> <span>hello-alpha</span></p>
```

Additionally, the emissary pod logs the following error:

```
2022-07-26 16:05:18 diagd 3.0.0 [P17TAEW] ERROR: {
...
<trim long JSON dump of envoy config>
...
could not validate the envoy configuration above after 5 retries, failed with error
[2022-07-26 16:05:18.276][140][info][main] [source/server/server.cc:786] runtime: layers:
  - name: static_layer
    static_layer:
      re2.max_program_size.error_level: 200
      envoy.reloadable_features.no_extension_lookup_by_name: false

[2022-07-26 16:05:18.280][140][critical][main] [source/server/config_validation/server.cc:63] error initializing configuration '/ambassador/snapshots/econf-tmp.json': route: unknown cluster 'cluster_http___hello_this_name_is_too_lo-1'

(exit code 1)
Aborting update...
```

This would suggest to me that the `beta` mapping should get a route that points to a new envoy cluster named `cluster_http___hello_this_name_is_too_lo-1`. However, this cluster is not being created correctly. I haven't been able to spot anything in the logs that would suggest why this cluster is not being created.

**Expected behavior**
The new envoy cluster should be created and the "beta" route should be correctly updated to point to the new cluster.

**Versions:**
 - emissary-ingress 3.0.0 (can reproduce on 2.3.1 as well)
 - Kubernetes 1.21, minikube 1.26.0

kflynn:
Without this fix, the collision-handling code for a cluster with a really long service name will fail, because the really really long service name will push all the stuff that's supposed to disambiguate the cluster names off past the point of relevance. 🤦‍♂️ 

With this fix, we stick a cryptographic hash of the whole too-long cluster name into the truncated name, so that the stuff trimmed off the end is still correctly accounted for.

I pretty much have tests running on my M1 Pro Mac after this PR.

Fixes #4354. Best reviewed commit-by-commit:

- 1545a2a4: Clean up a duplicated entry in `docs/releaseNotes.yml`
- c5a6bd83: fix a buglet where using `query_parameters` or `regex_query_parameters` would alter the internal name of the `Mapping` object. Should have no end-user-visible effect, but OMG was it making debugging weird.
- 3a5f6ae1: fix #4354 by changing the way we do hashing for collision resolution.
- 607f005b: big alterations to `TestFakeCollision` to better test everything.
- cb2351e5: more test changes so that my M1 Mac can run tests

 - [x] This is unlikely to impact how Emissary performs at scale.
     It's going to add an SHA1 to every too-long cluster name being generated, but that should be a small fraction of cluster names.


LanceEa (commented):
leaving a comment to make github review like me 😄.

kflynn:
Hey @LanceEa! I ~modified~substantially rewrote the `TestFakeColliision` test to do a better job of testing this; more info in the lead comment above. 🙂

python/ambassador/ir/ir.py
```diff
@@ -480,7 +480,11 @@ def __init__(
             if len(name) > 60:
                 # Too long. Gather this cluster by name prefix and normalize
                 # its name below.
-                short_name = name[0:40]
+                h = hashlib.new("sha1")
+                h.update(name.encode("utf-8"))
+                hd = h.hexdigest()[0:16].upper()
```
LanceEa, line 485: Just trying to make sure I understand this due to my lack of python skills. Envoy limits us to 60 characters so when it is greater than that we need to shorten and merge clusters.

The first 39 characters + 1 static dash ("-") + 15 character hash so we will shorten it to 55 characters...So did I interpret my python code correctly? If so any reason for only allowing 55 characters rather than full 60?

cmd/entrypoint/testutil_fake_collision_test.go
```diff
@@ -22,51 +30,81 @@ func TestFakeCollision(t *testing.T) {
...
+	c1 := FindCluster(envoyConfig, ClusterHasAltName("subway_staging_stable_staging_3001"))
+	assert.NotNil(t, c1)
+
+	// OK. c0 should get its load assignments from EDS with a key of
+	// k8s/staging/subway-staging-stable/3000, and it should have a name that starts with
+	// cluster_subway_staging_stable_staging_30- and ends with -0.
+
+	assert.Equal(t, "k8s/staging/subway-staging-stable/3000", c0.EdsClusterConfig.ServiceName)
+
+	assert.True(t, strings.HasPrefix(c0.Name, "cluster_subway_staging_stable_staging_30-"))
```
LanceEa, line 85: The change was first 40 chars, static dash and then hash. So, the assertion is testing for the initial 40 characters and dash correct? 

There really isn't any assertions or checks for hash but we assume it's there because of cluster counts are correct?

Then what is the suffix? Is that some thing we add to end somewhere else?

LanceEa (commented):
@kflynn - In general, I'm good with the changes. I just had a few questions so I make sure my understanding is correct.

python/ambassador/ir/ir.py
kflynn: `name[0:40]` is actually the first 40 characters rather than the first 39, so it's 40 + "-" + 16 for the hash + `-0` or 59 characters, which is close enough that I didn't want to push my luck. 😉 

(The `-0` is all but certainly unneeded at this point, too, but I still need to finish convincing myself of that.)

cmd/entrypoint/testutil_fake_collision_test.go
kflynn: `TestFakeCollision` does make sure that the hashes are all distinct (see lines 97-105 in `testutil_fake_collision_test.go`, just a bit further down). I don't recompute the hashes because that seems a bit too fragile if we change this later; what's important is that they're all distinct in this case, even when we force the names to collide.

And as noted before, the suffix is really left over from the previous collision handling. I'm hoping to get rid of it, but not just yet.

python/ambassador/ir/ir.py
LanceEa: Ok, that makes sense and understood. Yeah, I would think introducing the hash negates any reason for keeping the suffix around. Do you want to remove that as part of this PR or follow-up when you have had more time to think of the repercussions of that change?

cmd/entrypoint/testutil_fake_collision_test.go
LanceEa: got it thanks

LanceEa:
@kflynn - it looks good to me. Let me know what you think of removing the suffix as part of this PR?

also, I would like to get a second review, mostly for knowledge transfer so the other team members are aware of this naming convention.

LanceEa:
@ddymko - Can you take a look at this, I think it would be good to understand this logic and how we consolidate clusters that have name collisions.

AliceProxy (approved):
lgtm, thanks @kflynn :slightly_smiling_face:

python/ambassador/ir/ir.py
kflynn: The latter, yeah. Hopefully soon. 🤞

ddymko (approved):
looks good

PR #4371, opened 2022-08-01
[v3.2] Repatriate from v3.1

d6e-automaton:
This PR merges release/v3.1 in to master.


LanceEa:
These were already merged into master and the repatriation is not needed.

PR #4376, opened 2022-08-03
Add support for active health checking

AliceProxy:
## Description
Adds support for active healthchecking. If an upstream cluster (envoy cluster) fails its configured active health check threshold then envoy will no longer route requests to it. 

## Related Issues
List related issues.

## Testing

## Checklist

    + Any new features
    + Any deprecations


AliceProxy:
Perhaps a future consideration would be to create a HealthChecking CRD that defines a single set of config for HealthChecking that can be used as a default config for either all or a subset of upstreams. I don't think that is necessary at this time though.

KatieLo:
fixes https://github.com/datawire/apro/issues/2431 and helps with https://github.com/datawire/apro/issues/2911  ?

AliceProxy:
> fixes [datawire/apro#2431](https://github.com/datawire/apro/issues/2431) and helps with [datawire/apro#2911](https://github.com/datawire/apro/issues/2911) ?

@KatieLo For https://github.com/datawire/apro/issues/2431 Envoy's active healthchecking is not the same thing as Kubernetes healthchecking, but they sound like they would both resolve the same problem.

With Envoy's active healthchecking you can specify (on a Mapping) what the healthchecks should look like (what path/port what the interval of checks should be, how many failed checks it takes to mark an upstream as unhealthy, and how many successful checks it takes for an unhealthy upstream to be marked as healthy again.)

If you have 10 pods and 2 of them beome unhealthy for whatever reason (maybe they are overloaded or having transient problems) then Envoy will stop routing requests to those unhealthy pods but continue healthchecking them. Once they become healthy it will start routing requests to them again. This lets those pods have a chance to recover while keeping requests going to only to the pods that are known to be healthy. 


For  https://github.com/datawire/apro/issues/2911 I'm affraid I don't know too much about the graphql endpoing, but if triggering a restart of the Ambassador pods solves the problem then it sounds more like a problem with Ambassador than the upstream pods.

AliceProxy:
Of course I'll write some docs for this to explain the usage and whatnot when we decide it's good to merge :wink:

python/ambassador/ir/irhealthchecks.py
```diff
@@ -0,0 +1,240 @@
...
+                                    "status {} must be an integer >= 100 and < 600".format(endCode)
+                                )
+                            if startCode > endCode:
+                                raise ValueError(
+                                    "status range start value {} cannot be higher than the end {} for range".format(
+                                        startCode, endCode
+                                    )
+                                )
+
+                            # We add one to the end code because by default Envoy expects the start of the rangge to be
```
haq204, line 172:
```suggestion
                            # We add one to the end code because by default Envoy expects the start of the range to be
```

haq204 (commented):
More of a general question: when would one prefer using envoy upstream health checks over liveness probes?

pkg/api/getambassador.io/v3alpha1/common.go
```diff
@@ -112,6 +114,51 @@ type ErrorResponseOverride struct {
...
+	// Timeout for connecting to the health checking endpoint. Defaults to 3 seconds.
+	Timeout *metav1.Duration `json:"timeout,omitempty"`
+	// Interval between health checks. Defaults to every 5 seconds.
+	Interval *metav1.Duration `json:"interval,omitempty"`
+	// Number of non-expected responses for the upstream to be considered unhealthy. A single 503 will mark the upstream as unhealthy. Defaults to 2.
+	UnhealthyThreshold *int `json:"unhealthy_threshold,omitempty"`
+	// Number of expected responses for the upstream to be considered healthy. Defaults to 1.
+	HealthyThreshold *int             `json:"healthy_threshold,omitempty"`
+	HttpHealthCheck  *HttpHealthCheck `json:"http_health_check,omitempty"`
+	GrpcHealthCheck  *GrpcHealthCheck `json:"grpc_health_check,omitempty"`
```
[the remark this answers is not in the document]
AliceProxy: There are a lot of values within the health checking config that I decided on not adding support for at the moment. Typically the goal is not to expose the entire Envoy config set unless needed. Support for L3/L4 and redis health checking could be added as an extension to these changes at a future point as a feature request. I could add support for either of those now as part of this PR if there is a need for them.

python/ambassador/ir/irhealthchecks.py
```diff
@@ -0,0 +1,240 @@
...
+                # Process the expected statuses
+                expected_statuses = http_health_check.get("expected_statuses", None)
+                if expected_statuses is not None:
+                    validStatuses = []
+                    for statusRange in expected_statuses:
+                        try:
+                            startCode = int(statusRange["start"])
+                            endCode = int(statusRange["end"])
+                            if startCode < 100 or startCode >= 600:
+                                raise ValueError(
```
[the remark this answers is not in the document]
AliceProxy: Honestly... I don't really know what I decided to write it that way :slightly_smiling_face:

pkg/api/getambassador.io/v3alpha1/common.go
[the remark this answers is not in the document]
haq204: makes sense. I think we add L3 later but we can keep the scope of this PR limited to HTTP and GRPC healthchecking

pkg/api/getambassador.io/v3alpha1/crd_mapping.go
```diff
@@ -79,6 +79,9 @@ type MappingSpec struct {
 	Timeout     *MillisecondDuration `json:"timeout_ms,omitempty"`
 	IdleTimeout *MillisecondDuration `json:"idle_timeout_ms,omitempty"`
 	TLS         string               `json:"tls,omitempty"`
+	// +kubebuilder:validation:MinItems=1
+	HealthChecks    []HealthCheck `json:"health_checks,omitempty"`
+	HealthCheckPort *int          `json:"health_check_port,omitempty"`
```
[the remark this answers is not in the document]
AliceProxy: I removed health_check_port since Envoy only has config to support it when specifying Endpoints manually. Unfortunately the endpoint resolver does not support this configuration despite the fact that Envoy says that Health checking is meant to be used with the endpoint resolver for the most benefit.

python/ambassador/ir/irhealthchecks.py
```diff
@@ -0,0 +1,240 @@
...
+                    log_level=logging.ERROR,
+                )
+                continue
+
+            timeout = hc.get("timeout", "3s")  # default 3.0s timeout
+            interval = hc.get("interval", "5s")  # default 5.0s Interval
+            healthy_threshold = hc.get("healthy_threshold", 1)
+            unhealthy_threshold = hc.get("unhealthy_threshold", 2)
+
+            mapper: Dict[str, Any] = {
```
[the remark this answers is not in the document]
AliceProxy: I added some more type annotations, hopefully that helps :)

pkg/api/getambassador.io/v3alpha1/common.go
```diff
@@ -112,6 +114,51 @@ type ErrorResponseOverride struct {
...
+}
+
+// HealthCheck for HTTP upstreams. Only one of http_health_check or grpc_health_check may be specified
+type HttpHealthCheck struct {
+	Host string `json:"hostname,omitempty"`
```
LanceEa, line 145: Formatting issue...this should be aligned with the other lines

python/ambassador/ir/irhealthchecks.py
```diff
@@ -0,0 +1,235 @@
...
+    def _generate_mappers(self) -> Optional[List[Dict[str, Union[str, int, Dict]]]]:
+        all_mappers: List[Dict[str, Union[str, int, Dict]]] = []
+
+        # Make sure each health check in the list has config for either a grpc health check or an http health check
+        for hc in self._ir_config:
+
+            grpc_health_check = hc.get("grpc_health_check", None)
+            http_health_check = hc.get("http_health_check", None)
+
+            if grpc_health_check is None and http_health_check is None:
```
LanceEa, line 78: We can't move this logic earlier in the CRD Validation, correct? The only reason we can do that on the Edge Stack side is because we explicitly call our own Validation func?

I know OpenAPI v3 supported it but I think OpenAPIv2 doesn't which is still used internally by K8s, correct?

LanceEa: The reason I ask is because it would be nice from a user perspective if this type of logic was at the apply time and rejected before ever getting persisted.

python/ambassador/ir/irhealthchecks.py
```diff
@@ -0,0 +1,235 @@
...
+                    http_mapper["host"] = host
+
+                # Process the expected statuses
+                expected_statuses = http_health_check.get("expected_statuses", None)
+                if expected_statuses is not None:
+                    validStatuses = []
+                    for statusRange in expected_statuses:
+                        startCode = int(statusRange["start"])
+                        endCode = int(statusRange["end"])
+                        if startCode < 100 or startCode >= 600:
```
LanceEa, line 154: This is unecessary isn't it? The CRD Validation should capture it on apply? Is there ever a time when this could occur short of calling diagd directly?

python/ambassador/ir/irhealthchecks.py
```diff
@@ -0,0 +1,235 @@
...
+                            continue
+                        if startCode > endCode:
+                            self.post_error(
+                                f"IRHealthChecks: expected_statuses: status range start value {startCode} cannot be higher than the end {endCode} for range. Ignoring expected status for health-check {hc}",
+                                log_level=logging.ERROR,
+                            )
+                            continue
+
+                        # We add one to the end code because by default Envoy expects the start of the range to be
+                        # inclusive, but the end of the range to be exclusive. Lets just make both inclusive for simplicity.
```
LanceEa, line 174: Is it better to diverge from Envoy in this case? I mean most people follow our docs which we can call this out but for folks that read envoy docs this would be confusing...no hard opinion on this one but not sure it is a benefit to diverge...

pkg/api/getambassador.io/v2/common.go
```diff
@@ -143,6 +145,51 @@ type ErrorResponseOverride struct {
...
+
+type HealthCheck struct {
+	// Timeout for connecting to the health checking endpoint. Defaults to 3 seconds.
+	Timeout *metav1.Duration `json:"timeout,omitempty"`
+	// Interval between health checks. Defaults to every 5 seconds.
+	Interval *metav1.Duration `json:"interval,omitempty"`
+	// Number of non-expected responses for the upstream to be considered unhealthy. A single 503 will mark the upstream as unhealthy. Defaults to 2.
+	UnhealthyThreshold *int `json:"unhealthy_threshold,omitempty"`
+	// Number of expected responses for the upstream to be considered healthy. Defaults to 1.
+	HealthyThreshold *int             `json:"healthy_threshold,omitempty"`
```
LanceEa, line 169: Is there a reason we are using underscores and not camel casing? I know I'm being slow on nailing down CRD design guidelines but since this is a new one I would think we would change it to `healthyThreshold`...For example, here is the K8s guidelines and when we get to V3 we want to align with it.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions

LanceEa (commented):
I have left some questions and comments. I still need to review a few of the files but before I do that I wanted you to see some of my feedback around naming conventions (sorry I didn't have that to you before the effort). But, I want to get your thoughts on it

pkg/api/getambassador.io/v3alpha1/common.go
```diff
@@ -112,6 +114,51 @@ type ErrorResponseOverride struct {
...
+	// +kubebuilder:validation:Maximum=599
+	End int `json:"end,omitempty"`
+}
+
+type HealthCheck struct {
+	// Timeout for connecting to the health checking endpoint. Defaults to 3 seconds.
+	Timeout *metav1.Duration `json:"timeout,omitempty"`
+	// Interval between health checks. Defaults to every 5 seconds.
+	Interval *metav1.Duration `json:"interval,omitempty"`
+	// Number of non-expected responses for the upstream to be considered unhealthy. A single 503 will mark the upstream as unhealthy. Defaults to 2.
```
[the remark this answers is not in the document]
AliceProxy: No normally any other non-expected (by default only 200 is expected) response like 5xx counts towards that quota. Such as a 504 or a 500. Regardless of how you configure envoy, a single 503 response will result in the upstream being marked as unhealthy.  Think theres room in the comment to capture some more of this information, or should I just explain in more detail when I write the documentation?

python/ambassador/ir/irhealthchecks.py
AliceProxy: We can for fields where one is required all the time, but unfortunately there was no easy way to check for a case when neither field is required but at least one must be present. I was really hoping there would be a way to get that check in earlier so that you would be prevented from applying the CRD :cry:

python/ambassador/ir/irhealthchecks.py
AliceProxy: Yeah, thats just me being paranoid that something would change in the future or if something else were to break. Something just kind of bothers me about trusting CRD Validation with some of these data types.

python/ambassador/ir/irhealthchecks.py
AliceProxy: I'm of the opinion that if you mark something as the end/start it should be included. I was trying to make it easier to understand for our users rather than cater to people who might know what the underlying envoy config is. To me ranges where the start is inclusive and the end is exclusive always irritated me, it might make sense when writing code but I think it makes a lot less sense when specifying configuration. 

I'm open to changing this though if there is benefit to doing so.

pkg/api/getambassador.io/v2/common.go
AliceProxy: This was actually something I was hoping to sync about regarding the next major CRD version, particularly whether it would be better for this whole health_checks object to follow the syntax of the other Mapping fields for now, then change everything all at once with the next CRD version, or if there is a benefit to mixing the casing until everything is camelCase.

pkg/api/getambassador.io/v3alpha1/common.go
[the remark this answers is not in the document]
LanceEa: Ok, your explanation makes sense. Maybe leave it off here and then move the 503 comment to documentation instead to avoid confusion here??

python/ambassador/ir/irhealthchecks.py
LanceEa: Ok that is what I thought, we do not call any additional validation functions in the watch phase prior to passing it along to `diagd`. As you point out ideally it would be handled via openAPI 3's support for oneOf but kubernetes doesn't support it yet due to converting it to openAPI v2.

I will keep this in mind when thinking about ways we can improve this story overall.

python/ambassador/ir/irhealthchecks.py
LanceEa: Is 600 a valid code?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

python/ambassador/ir/irhealthchecks.py
LanceEa: What if a user sets the end code to 599 which passes validation and then the plus 1 logic here would set the end to 600 correct?

If I read the docs it doesn't allow for the value to be 600 so would this throw invalid envoy configuration?

https://www.envoyproxy.io/docs/envoy/v1.23.0/api-v3/config/core/v3/health_check.proto.html?highlight=expected_statuses#config-core-v3-healthcheck-httphealthcheck

The plus one logic should probably move above the range error checks? Can we have a test that verifies the ends to ensure nothing causes issues.

python/ambassador/envoy/v3/v3cluster.py
```diff
@@ -74,6 +74,18 @@ def __init__(self, config: "V3Config", cluster: IRCluster) -> None:
...
+        # Don't feed the raw health_checks object to envoy, grab its config where the econf has already been built
+        if cluster.health_checks is not None:
+            health_checks_config = cluster.health_checks.config()
+            if health_checks_config is not None:
+                if ctype == "EDS":
+                    self["health_checks"] = health_checks_config
+                else:
+                    cluster.ir.logger.error(
+                        "health_checks configuration is only supported when using the endpoint resolver. Ignoring health_checks config for cluster: %s..."
+                        % (cluster.name)
```
LanceEa, line 86: Is the cluster name going to be sufficient for the user to figure out which one is causing the error? Are we able to provide additional information here? I don't have any thing off the top of my head but just wanted to ask?

pkg/api/getambassador.io/v2/common.go
LanceEa: Yes, I think that is a good call for this one and agree. So the rule for now will be:

If adding to an existing CRD as in this case extending the `Mapping` then we should match existing conventions. If it is a new standalone CRD then we should adopt k8s naming guidelines. Then in the next version and ultimately when we get to v3 then all the crds will be aligned on the new naming conventions.

I will document this in the design guide for v3. I swear I'm going to finish it...someday :).

python/ambassador/envoy/v3/v3cluster.py
```diff
@@ -203,19 +215,27 @@ def get_endpoints(self, cluster: IRCluster):
...
-                address = {
-                    "address": target["ip"],
-                    "port_value": target["port"],
-                    "protocol": "TCP",  # Yes, really. Envoy uses the TLS context to determine whether to originate TLS.
+                endpoint = {
+                    "address": {
+                        "socket_address": {
+                            "address": target["ip"],
+                            "port_value": target["port"],
+                            "protocol": "TCP",  # Yes, really. Envoy uses the TLS context to determine whether to originate TLS.
```
LanceEa, line 223: Is this comment still relevant, if not I would say remove it. Because, personally it makes no sense to me so does additional context need to be added or since we are modifying the code we should just remove it.

LanceEa (commented):
I left a few final comments and other than that I think this is looking really great!

python/ambassador/ir/irhealthchecks.py
AliceProxy: Great suggestion!

python/ambassador/ir/irhealthchecks.py
AliceProxy: FWIW the CRD validation will prevent you from applying one that has a value greater than or equal to `600` and the Envoy config is fine with `600` as a code so the main problems here are just that the increment of `end` should happen before the check and that there needs to be test coverage for this.

python/ambassador/ir/irhealthchecks.py
AliceProxy: > What if a user sets the end code to 599 which passes validation and then the plus 1 logic here would set the end to 600 correct?
If I read the docs it doesn't allow for the value to be 600 so would this throw invalid envoy configuration?

nope, the following Envoy config example is perfectly legal and works fine because Envoy is treating `end:  600` as exclusive. `start` can never be 600 because of CRD validation but the IR check for validity on these values needs fixed:
```
            "http_health_check": {
              "expected_statuses": [
                {
                  "end": 600,
                  "start": 200
                }
              ],
              "path": "/health"
            },
````

File: pkg/api/getambassador.io/v2/common.go (hunk 10 above)
Comment (AliceProxy) [reply pkg/api/getambassador.io/v2/common.go:171]: Cool. I'm definitely don't have any strong opinions on this, just want to make sure there is a standard to follow and that we go with whatever eliminates the most friction.

File: pkg/api/getambassador.io/v2/common.go [hunk 14]
```diff
@@ -143,6 +145,54 @@ type ErrorResponseOverride struct {
...
+	// Start of the statuses to include. Must be between 100 and 599 (inclusive)
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=100
+	// +kubebuilder:validation:Maximum=599
+	Start int `json:"start,omitempty"`
+	// End of the statuses to include. Must be between 100 and 599 (inclusive)
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=100
+	// +kubebuilder:validation:Maximum=599
+	End int `json:"end,omitempty"`
```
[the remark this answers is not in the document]
LukeShu: Actually, I think "min"/"max" would be better, since IMO "max" conveys "inclusive" while "end" is ambiguous about whether it's inclusive or exclusive.

python/tests/unit/test_healthchecks.py
```diff
@@ -0,0 +1,517 @@
...
+metadata:
+  name: healthchecktest
+  namespace: default
+spec:
+  hostname: '*'
+  service: coolsvcname
+  prefix: /test
+  health_checks: {}
+"""
+    testcases = [
```
LukeShu, line 79: The real mypy trick is that this needs to say
```python
testcases: List[Dict[str, Any]] = [
```

python/tests/unit/test_healthchecks.py
```diff
@@ -0,0 +1,517 @@
...
+
+    for case in testcases:
+
+        caseYaml = typing.cast(typing.Dict[str, Any], case)["input"]  # seriously mypy???
+        testName = typing.cast(typing.Dict[str, Any], case)["name"]
+        econf = _get_envoy_config(caseYaml)
+        cluster = _get_cluster_config(econf.clusters, "cluster_coolsvcname_default")
+        assert cluster != False
+
+        expectedChecks = typing.cast(typing.Dict[str, Any], case)["expected"]
```
LukeShu, line 415: You should drop these `typing.cast`s in favor of just adding type information to `testcases`. (See the comment I made further up.)

python/ambassador/ir/irhealthchecks.py
LukeShu: @LanceEa 600 is not valid. "All valid status codes are within the range of 100 to 599, inclusive." https://www.rfc-editor.org/rfc/rfc9110#section-15-1

LukeShu (changes requested):
The unit tests for making sure the IR works the way we intend is good. But since this is leveraging an Envoy feature we haven't used before, I think it's also important that we have an integration test that checks that Envoy is actually behaving the way we think this config will make it behave. We've shipped DOA features because we had tests that check that the code generates the Envoy config that the author thinks it should, only for us to learn that Envoy doesn't actually work the way the author thought it works.

python/ambassador/ir/irhealthchecks.py
LanceEa: @LukeShu - Yup, I miss-read the code above. I thought the code was allowing 600 to be a valid value. But after taking a look it was excluding 600 😄..doh

pkg/api/getambassador.io/v2/common.go
```diff
@@ -143,6 +145,54 @@ type ErrorResponseOverride struct {
...
+	AddRequestHeaders    *map[string]AddedHeader `json:"add_request_headers,omitempty"`
+	RemoveRequestHeaders *[]string               `json:"remove_request_headers,omitempty"`
+	ExpectedStatuses     *[]StatusRange          `json:"expected_statuses,omitempty"`
+}
+
+// HealthCheck for gRPC upstreams. Only one of grpc_health_check or http_health_check may be specified
+type GrpcHealthCheck struct {
+	// The service name parameter which will be sent to gRPC service in the health check message
+	// +kubebuilder:validation:Required
+	ServiceName string `json:"service_name,omitempty"`
```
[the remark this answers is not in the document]
AliceProxy: Yep :upside_down_face: 
Thoughts on `upstream_name`/`UpstreamName`? in the econf its `service..`

python/ambassador/envoy/v3/v3cluster.py
AliceProxy: Not sure what else we would be able to provide at this point for debugging unfortunately.

python/tests/kat/t_active_health_check.py
```diff
@@ -0,0 +1,140 @@
...
+
+"2022-10-04T23:24:45Z"
+"2022-10-04T23:26:45Z"
+
+
+class ActiveHealthCheckTest(AmbassadorTest):
+    def init(self):
+        self.target = HealthCheckServer()
+
+    def config(self) -> Generator[Union[str, Tuple[Node, str]], None, None]:
```
LanceEa, line 15: We should deprecate the usage of this because it uses annotations to apply the yaml vs using the `def manifest` function which will actually play yamls directly. It doesn't have to be apart of this PR but just something to keep in mind.

python/tests/kat/t_active_health_check.py
```diff
@@ -0,0 +1,140 @@
...
+service: {self.target.path.fqdn}
+resolver: endpoint
+load_balancer:
+  policy: round_robin
+health_checks:
+- unhealthy_threshold: 1
+  interval: 1s
+  health_check:
+    http:
+      path: /healthcheck/actualcheck/
```
LanceEa, line 33: since we are only picking a path, I'm assuming the host is the `service` correct?

python/tests/kat/t_active_health_check.py
```diff
@@ -0,0 +1,140 @@
...
+        )  # the deployment has 5 pods. This will make one of them start returning errors
+
+        # Make 1000 requests over the course of a minute
+        for _ in range(1000):
+            yield Query(self.url("healthcheck/"), expected=[200, 500])
+
+    def check(self):
+        print(self.results)
+        # Check the manual requests first
+        assert self.results[0].status == 200
```
LanceEa, line 123: Again, I think even when you don't set an expected if it was to return say a 403 randomly then the Query fails expectation and the check function is never run. At least this is what I observed when adding the other tests.

python/tests/kat/t_active_health_check.py
```diff
@@ -0,0 +1,140 @@
...
+        assert self.results[1].status == 200
+        assert self.results[2].status == 200
+
+        # Add up the number of 500 and 200 responses that we got.
+        valid = 0
+        errors = 0
+        for i in range(5, 1005):
+            assert (self.results[i].status == 200 or self.results[i].status == 500) == True
```
LanceEa, line 71: Isn't this unecessary? the `expected` in the Query would fail the test before getting to the check function I believe.

LanceEa (commented):
Overall I think it is looking pretty good. I have a few questions and comments.

My only concern (maybe unwarranted) with the new KAT tests is are we going to experience flakes. I know you implement some logic to hopefully address the variability in it so we can roll with it and see how it goes :).

python/tests/kat/abstract_tests.py
```diff
@@ -466,6 +466,21 @@ def requirements(self):
...
+    skip_variant: ClassVar[bool] = True
+
+    def __init__(self, *args, **kwargs) -> None:
+        self.use_superpod = False
+        # Do this unconditionally, because that's the point of this class.
+        kwargs["service_manifests"] = integration_manifests.load("health_check_server")
+        super().__init__(*args, **kwargs)
+
+    def requirements(self):
+        yield ("deployment", self.path.k8s)
```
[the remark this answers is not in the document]
AliceProxy: The requirement are just waiting for things to be ready before we start the test. So in this example, the deployment needs to be ready and it needs to respond to a few queries before we consider the service ready for use in tests.

python/tests/kat/t_active_health_check.py
AliceProxy: Yep, it will make a request to the same service and that is currently not configurable.

python/tests/kat/t_active_health_check.py
AliceProxy: Hmm weird, from all of my tests it will fail the test immediately when it receives an unexpected response from the queries or fails the asserts.

python/tests/kat/t_active_health_check.py
```diff
@@ -0,0 +1,140 @@
+import time
+from typing import Generator, Tuple, Union
+
+from abstract_tests import AmbassadorTest, HealthCheckServer, Node
+from kat.harness import Query
+
+"2022-10-04T23:24:45Z"
```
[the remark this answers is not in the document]
AliceProxy: Huh, didn't even notice this slip in. No clue what caused it.

python/tests/kat/t_active_health_check.py
AliceProxy: Ya know I think you're right. We do perform an assert on the Query for the expected result. I was just following the pattern in most of the other tests without thinking about it. Really we only need the assert on the number of 200s vs 500s so I removed em.

python/tests/kat/t_active_health_check.py
AliceProxy: Misinterpreted what you meant there for a minute but yep I agree.

cmd/kat-server/server.go
```diff
@@ -114,7 +115,80 @@ func main() {
...
+		HealthyStatusCode, err := strconv.Atoi(os.Getenv("HEALTHY_STATUS_CODE"))
+		if err != nil {
+			HealthyStatusCode = 200
+		}
+		UnhealthyStatusCode, err := strconv.Atoi(os.Getenv("UNHEALTHY_STATUS_CODE"))
+		if err != nil {
+			UnhealthyStatusCode = 500
+		}
 
+		for {
```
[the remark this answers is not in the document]
AliceProxy: I was just copying the setup for the main kat server (since the health checks server is pretty similar) but I removed the loop and it seems fine without it.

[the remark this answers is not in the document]
LanceEa: OK cool :).

LanceEa (dismissed):
lgtm

pkg/api/getambassador.io/v2/crd_mapping.go
```diff
@@ -85,6 +85,10 @@ type MappingSpec struct {
 	// +k8s:conversion-gen=false
 	TLS *BoolOrString `json:"tls,omitempty"`
 
+	// +kubebuilder:validation:MinItems=1
+	// +k8s:conversion-gen:rename=HealthChecks
+	V3HealthChecks []HealthCheck `json:"v3health_checks,omitempty"`
```
LukeShu, line 90: Can we have this just use `v3alpha1.HealthCheck` instead of having to duplicate types in `v2/common.go`?

python/tests/kat/t_active_health_check.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def queries(self):
+        yield Query(self.url("healthcheck/"), phase=1)  # Just making sure things are running
+        yield Query(self.url("ambassador/v0/diag/"), phase=1)
+
+        yield Query(
+            self.url("healthcheck/makeUnhealthy/"), phase=1
+        )  # the deployment has 5 pods. This will make one of them start returning errors
+
+        # These two queries on their own in separate phases are just a hack way of getting the kat client
```
LukeShu, line 42: There are three queries here, not two.

python/tests/kat/t_active_health_check.py
```diff
@@ -0,0 +1,129 @@
...
+            if self.results[i].status == 200:
+                valid += 1
+            elif self.results[i].status == 500:
+                errors += 1
+
+        # with 1000 requests and 1/5 being an error response, we should have the following distribution +/- 10
+        # assert 190 <= errors <= 210
+        # assert 790 <= valid <= 810
+
+        # But since we configure health chekcing we should actually see 0 errors because the health checks noticed
```
LukeShu, line 72: typo: `chekcing`

python/tests/kat/t_active_health_check.py
```diff
@@ -0,0 +1,129 @@
...
+        errors = 0
+        for i in range(6, 1006):
+            if self.results[i].status == 200:
+                valid += 1
+            elif self.results[i].status == 500:
+                errors += 1
+
+        # with 1000 requests and 1/5 being an error response, we should have the following distribution +/- 10
+        # assert 190 <= errors <= 210
+        # assert 790 <= valid <= 810
```
LukeShu, line 70: I know this is commented out, but FWIW: ±10 is probably way too tight. I'd probably give it ±100. We wouldn't want something like this causing CI flakes.

python/tests/kat/t_active_health_check.py
```diff
@@ -0,0 +1,129 @@
...
+        for i in range(3, 1003):
+            if self.results[i].status == 200:
+                valid += 1
+            elif self.results[i].status == 500:
+                errors += 1
+        msg = "Errors: {}, Valid: {}".format(errors, valid)
+
+        # with 1000 requests and 1/5 being an error response, we should have the following distribution +/- some
+        assert 175 <= errors <= 225, msg
+        assert 775 <= valid <= 825, msg
```
LukeShu, line 129: ±25 makes me nervous.  In the TCPMapping tests, which do 1000 requests with a 700:300 split, I used a margin of 150.

PS: what I did in `t_tcpmapping.py`, to make the ± margins more readable (IMO):
```python
        # Probabalistic, margin might need tuned
        margin = 150
        assert abs(counts[self.target70.path.k8s] - 700) < margin
        assert abs(counts[self.target30.path.k8s] - 300) < margin
```
(... and pasting that here, I realize that I spelled "Probabilistic" wrong)

python/tests/kat/t_active_health_check.py
```diff
@@ -0,0 +1,129 @@
...
+"""
+        )  # The round robin load balancer is not necessary for the test but should help make the request distribution even across the pods
+
+    def queries(self):
+        yield Query(self.url("healthcheck/"), phase=1)  # Just making sure things are running
+        yield Query(self.url("ambassador/v0/diag/"), phase=1)
+
+        yield Query(
+            self.url("healthcheck/makeUnhealthy/"), phase=1
+        )  # the deployment has 5 pods. This will make one of them start returning errors
```
LukeShu, line 40: Something needs to reset this at startup, otherwise running the test repeatedly won't work, as more and more pods get marked unheathly each run.  That would be unacceptable for locally running the tests as you hack on the code.

There are two ways that existing tests do such resets:
- Most of them call a "reset" endpoint in `requirements()` (example: `t_circuitbreaker.py`).  This might be problematic because we'd need to be sure to hit each Pod.
- `t_tracing.py`'s `Zipkin` does a thing with a nonce in an env-var, since there's no way to reset Zipkin. This might be the best option.

python/ambassador/ir/irhealthchecks.py
```diff
@@ -0,0 +1,256 @@
...
+                    append = bool(v["append"])
+                headers.append({"header": {"key": k, "value": v["value"]}, "append": append})
+            else:
+                headers.append(
+                    {
+                        "header": {"key": k, "value": v},
+                        "append": append,  # Default append True, for backward compatability
+                    }
+                )
+        return headers
```
LukeShu, line 256: This file is a lot of boilerplate for not very much functionality.

Just something to think about wrt improving the base `IRResource` in the future.

python/ambassador/ir/irhealthchecks.py
LukeShu: I'm confused why this still needs to be here with the

```go
// +kubebuilder:validation:MinProperties=1
// +kubebuilder:validation:MaxProperties=1
type HealthCheckLocation struct {
```

validations.

python/ambassador/ir/irhealthchecks.py
```diff
@@ -0,0 +1,256 @@
...
+                expected_statuses = http_health_check.get("expected_statuses", None)
+                if expected_statuses is not None:
+                    validStatuses = []
+                    for statusRange in expected_statuses:
+                        minCode = int(statusRange["min"])
+                        maxCode = int(statusRange["max"])
+                        # We add one to the end code because by default Envoy expects the start of the range to be
+                        # inclusive, but the end of the range to be exclusive. Lets just make both inclusive for simplicity.
+                        maxCode += 1
+                        if minCode < 100 or minCode > 600:
```
LukeShu, line 178: I'm confused why these checks need to be here, instead of trusting the CRD validations.

LukeShu (changes requested):
This is a lot of commits that fixup earlier commits, which makes reviewing it hard; I just have to look at the total diff as if it were squashed in one commit. Which is fine I guess, but please call that out in the PR description.

Changes requested (most important to least important):
- Have the tests reset themselves in order to allow repeated runs
- Don't duplicate new types in to `v2/common.go`; just import the v3alpha1 types
- (unless there's a reason I'm missing) remove superfluous validation from the Python
- Typos and corrections in comments (number of queries, `chekcing`, if you want to fix my spelling in the tcpmapping tests)

Suggested changes:
- Widen the test margin for error

On the whole: Looks solid.

python/ambassador/ir/irhealthchecks.py
AliceProxy: Yeah, main driver was just that its a ton of checking and translating to jam into the v3cluster.py and there was already kind of a precedent with irerrorresponse.py although that one does do more than this file does.

python/tests/kat/t_active_health_check.py
AliceProxy: Went with the nonce approach for this

python/ambassador/ir/irhealthchecks.py
AliceProxy: removed all the ones that were not necessary for the linting type checks

docs/releaseNotes.yml
```diff
@@ -86,6 +98,14 @@ items:
...
           (Thanks to <a href="https://github.com/olemarkus">Ole Markus</a>!).
 
+      - title: Add support for active health checking configuration.
+        type: feature
+        body: >-
+          It is now possible to configure active healhchecking for upstreams within a <code>Mapping</code>.
+          If the upstream fails its configured health check then Envoy will mark the upstream as unhealthy and no longer send
+          traffic to that upstream. Single pods within a group may can be marked as unhealthy. The healthy pods will continue to receive
+          traffic normally while the unhealthy pods will not receive any traffic until they recover by passing the health check.
+
```
LukeShu, line 108: You've added it to *both* the 3.3.0 release and a new 3.4.0 release.

LukeShu (changes requested):
Minor problem with the changelog. Otherwise, lgtm.

Thanks!

LukeShu (approved):

LanceEa (approved):
🚀 rock and roll

PR #4379, opened 2022-08-04
ci: add container cve scanning to ci build process

LanceEa:
## Description

We already use code scanning for our own code dependencies but we also pull in other packages during our container build process via alpine package manager. This adds a cve scanner to our built containers so that we can proactively investigate cve's to determine if Emissary-ingress is vulnerable to them or not. This will ensure that we keep our dependencies up-to-date and address potential vulnerabilities quickly.

This adds support for two open source scanners Trivy and Grype. When evaluating them both they seem to catch different CVE's. Due to the fact that they run fast and do not have a big impact on CI time
both tools will be used to scan. They are both set to warning for now as well so that we become more familiar with them and how they will fit into our CI processes.

## Related Issues
N/A

## Testing
Tested Github Actions and CI in a personal project to verify setup was correct and to verify how to configure the Github Actions.

## Checklist

### DIFF
--- a/.github/workflows/execute-tests-and-promote.yml
+++ b/.github/workflows/execute-tests-and-promote.yml
@@ -8,7 +8,6 @@ name: job-promote-to-passed
   pull_request: {}
 
 jobs:
-
   lint: ########################################################################
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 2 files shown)


LanceEa:
@AliceProxy & @kflynn  - I just wanted you to be aware of this since so that you can pull it into whatever upcoming changes you have.

@ddymko - I tagged you so you are aware since this goes along with introducing dependabot as well.

Note: I will follow-up with PR's that address some of the dependency upgrades so that we can officially remove the alerts.

AliceProxy (approved):

PR #4408, opened 2022-08-11
[v2.3] Clarify what different branches are for

LukeShu:
## Description

There's been lotsa confusion about `master` vs `release/v3.1` or whatever. This PR does 2 things to address this:
 - Add a section about branch names to the README (see it at https://github.com/emissary-ingress/emissary/blob/e9b8f766083092f669885f85447d92085e75bd67/README.md)
 - Add a CI job that checks that the tag-derived version number, the branch name, and the `releaseNotes.yml` version are all in agreement.
 
One of the error messages it can spit out is
```
echo "You seem to be on the Git branch for v${branch_majorminor}.z, but Git tags indicate that this is work for v${tag_majorminor}.z"
echo "Perhaps you need to go yell at the person who set up the release branch for the last .y bump."
```
It'd be good if that setting-up-the-branch bit was scripted to reduce the likelihood of messing it up. I'll do that when this repatriates to 3.1; with some changes from 2.3 to then, I don't think it's worth writing a version of it for 2.3.

## Testing

I verified that when I try this [against master](https://github.com/emissary-ingress/emissary/pull/4409) it complains that
```
You seem to be on the Git branch for v3.2.z, but Git tags indicate that this is work for v3.0.z
Perhaps you need to go yell at the person who set up the release branch for the last .y bump.
```
Which is correct because branching was screwy for 3.1.0.

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,11 +18,9 @@ Emissary-ingress
 
 <!-- Links are (mostly) at the end of this document, for legibility. -->
 
-----
-
 [Emissary-Ingress](https://www.getambassador.io) is an open-source Kubernetes-native API Gateway +
 Layer 7 load balancer + Kubernetes Ingress built on [Envoy Proxy](https://www.envoyproxy.io).
-Emissary-ingress is a CNCF incubation project (and was formerly known as Ambassador API Gateway.)
+Emissary-ingress is a CNCF incubation project (and was formerly known as Ambassador API Gateway).
 
 Emissary-ingress enables its users to:
 * Manage ingress traffic with [load balancing], support for multiple protocols ([gRPC and HTTP/2], [TCP], and [web sockets]), and Kubernetes integration
--- a/README.md
+++ b/README.md
@@ -34,6 +34,33 @@ Emissary-ingress enables its users to:
 
 See the full list of [features](https://www.getambassador.io/features/) here.
 
+Branches
+========
+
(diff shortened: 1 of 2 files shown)


.github/workflows/check-branch.yml
```diff
@@ -0,0 +1,72 @@
+name: Check branch version
+
+# This is a separate workflow from the main build-and-test workflow
+# because we want it to run when a PR's target branch changes, and
+# it'd be prohibitively to re-run the main workflow every time that
```
AliceProxy, line 5: `# it'd be prohibitively to re-run` -> `# prohibitively expensive/?`

README.md
```diff
@@ -36,6 +34,33 @@ Emissary-ingress enables its users to:
 
 See the full list of [features](https://www.getambassador.io/features/) here.
 
+Branches
```
AliceProxy, line 37: This section is great!

AliceProxy (dismissed):
lgtm other than what I think is a small typo

.github/workflows/check-branch.yml
LukeShu: Ah, I meant to include "time-consuming"

LukeShu: fixed

AliceProxy (approved):

PR #4412, opened 2022-08-12
deps : adding all go.mod locations to dependabots monitoring

ddymko:
Signed-off-by: David Dymko <ddymko@datawire.io>

## Description
The current dependabot setup only watches `go.mod` in `/`. This will have dependabot also monitor the other `go.mod` we have in the `/tool/src/*` folders.

## Related Issues
n/a

## Testing
Validated this works on a fork of emissary 

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,39 @@ updates:
     schedule:
       interval: daily
     open-pull-requests-limit: 10
+
+  - package-ecosystem: gomod
+    directory: "/tools/src/yq/"
+    schedule:
+      interval: daily
+    open-pull-requests-limit: 10
+
+  - package-ecosystem: gomod
+    directory: "/tools/src/chart-doc-gen"
+    schedule:
+      interval: daily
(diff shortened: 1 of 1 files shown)


LanceEa:
lgtm...one word of caution. We will want to double check when updating some of these tools that they still work with Edge Stack. This is due to the tight coupling in edge stack  to the tooling in Emissary-ingress. Alternatively, we could also look to see if there are opportunities to reduce this tight coupling.

LanceEa:
Specifically, the `yq` I know is used in Edge Stack. So, it might end up being moot but just again calling it out as something that could cause fits.

LanceEa (approved):

ddymko:
@LanceEa another example is golangci-lint which is used in edge-stack. So we should definitely keep an eye on these dependencies. I figured it might be best just to be aware of these dep updates as they crop up.

However, I do agree we should look to see if we can reduce the coupling.

PR #4421, opened 2022-08-18
[v1.14] Miscellaneous

LukeShu:
## Description

This is a bunch of miscellaneous things. We might not do another 1.14, but if we do, these are things that we'll want to include and not lose track of.

- (1 commit) Alice's gzip bugfix backport https://github.com/emissary-ingress/emissary/pull/4326
- (2 commits) Release notes for that gzip backport
- (2 commits) Backport README and CI-branch-checking from https://github.com/emissary-ingress/emissary/pull/4408
- (4 commits) Improve the error reporting when validating config resources, to help people who are trying to upgrade
- (1 commit) Backport the repatriation automation
- (1 commit) Fix CI; pin telepresence to 2.6, since 2.7 had a breaking change
- (1 commit) Have the repatriation automation skip over 2.0 through 2.3; skip straight to 2.4.

NB: The dev image that we gave the one user is `docker.io/datawiredev/ambassador:1.14.4-dev.12-gb3354b9d0`; b3354b9d0 is the same as the "backport the repatriation automation" commit here, but was missing an `OPENSOURCE.md` update.

## Checklist

    + Any new features
    + Any deprecations


build-aux/teleproxy.mk
```diff
@@ -23,7 +23,7 @@ $(tools/telepresence): CGO_ENABLED = 1
 endif
 $(tools/telepresence): $(abspath $@)
 	mkdir -p $(@D)
-	cd $(OSS_HOME) && curl -s --fail -L https://app.getambassador.io/download/tel2/$$(go env GOOS)/$$(go env GOARCH)/latest/telepresence -o $(abspath $@) && chmod a+x $(abspath $@)
+	cd $(OSS_HOME) && curl -s --fail -L https://app.getambassador.io/download/tel2/$$(go env GOOS)/$$(go env GOARCH)/2.6.6/telepresence -o $(abspath $@) && chmod a+x $(abspath $@)
```
AliceProxy, line 26: General question about the Telepresence pin: is there a specific reason why we don't use latest such as knowing that the newer versions don't work or that CI is expecting a certain set of features/syntaxes from Telepresence and commands may change in newer versions?

AliceProxy (approved):
I was like "hmmm.... that's a lot of files changed" until I saw the commit upgrading dlib :slightly_smiling_face: 
lgtm

build-aux/teleproxy.mk
LukeShu: As I said in the commit message: I just checked what the last version was that we saw work in Emissary 1.14 CI, which turned out to be Telepresence 2.6.6; at the time I didn't have any idea what had changed in newer Telepresence that would break it.  After communicating with Thomas: Yes, there was an intentional breaking change in Telepresence 2.7.0 (it now requires you to explicitly `telepresence helm install` it in to your cluster, rather than automatically doing that). So assuming that I don't think it's worth the effort to switch switch 1.14's machinery to use Tel 2.7, I suppose I could have pinned it to the latest 2.6, instead of 2.6.6, but... eh.

LanceEa (approved):
lgtm

AliceProxy (approved):

PR #4423, opened 2022-08-19
Support setting secret namespace in a host

AliceProxy:
## Description
Currently it is only possible to use a secret within a `Host` if that secret shares the same namespace as the `Host`. While originally this was a design decision, we've seen users experiencing friction due to this restriction. These changes will allow you to use secrets from any namespace within a `Host` like so:

```
tlsSecret:
  name: tls-cert
  namespace: foobar # new field!
```

## Related Issues
List related issues.

## Testing
Manual tests and added some test cases to the existing `Host` tests.

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/pkg/api/getambassador.io/v3alpha1/crd_host.go
+++ b/pkg/api/getambassador.io/v3alpha1/crd_host.go
@@ -103,15 +103,7 @@ type HostSpec struct {
 	// certificates.  If ACME is enabled (see $acmeProvider), then the
 	// default is $hostname; otherwise the default is "".  If the value
 	// is "", then we do not do TLS for this Host.
-	//
-	// Note that this is a native-Kubernetes-style core.v1.LocalObjectReference, not
-	// an Ambassador-style `{name}.{namespace}` string.  Because we're opinionated, it
-	// does not support referencing a Secret in another namespace (because most native
-	// Kubernetes resources don't support that), but if we ever abandon that opinion
-	// and decide to support non-local references it, it would be by adding a
-	// `namespace:` field by changing it from a core.v1.LocalObjectReference to a
-	// core.v1.SecretReference, not by adopting the `{name}.{namespace}` notation.
-	TLSSecret *corev1.LocalObjectReference `json:"tlsSecret,omitempty"`
+	TLSSecret *corev1.SecretReference `json:"tlsSecret,omitempty"`
 
 	// Request policy definition.
 	RequestPolicy *RequestPolicy `json:"requestPolicy,omitempty"`
--- a/python/ambassador/ir/irhost.py
+++ b/python/ambassador/ir/irhost.py
@@ -1,6 +1,6 @@
 import copy
 import os
-from typing import TYPE_CHECKING, Dict, List, Optional
+from typing import TYPE_CHECKING, Dict, List, Optional, Union
 
 from ..config import Config
 from ..utils import SavedSecret, dump_json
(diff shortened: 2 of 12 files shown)


LukeShu:
Don't we need to target this at 2.x?

pkg/api/getambassador.io/v3alpha1/crd_host.go
```diff
@@ -103,15 +103,7 @@ type HostSpec struct {
...
 	// default is $hostname; otherwise the default is "".  If the value
 	// is "", then we do not do TLS for this Host.
-	//
-	// Note that this is a native-Kubernetes-style core.v1.LocalObjectReference, not
-	// an Ambassador-style `{name}.{namespace}` string.  Because we're opinionated, it
-	// does not support referencing a Secret in another namespace (because most native
-	// Kubernetes resources don't support that), but if we ever abandon that opinion
-	// and decide to support non-local references it, it would be by adding a
-	// `namespace:` field by changing it from a core.v1.LocalObjectReference to a
-	// core.v1.SecretReference, not by adopting the `{name}.{namespace}` notation.
```
LukeShu, line 113: lol @ how relevant this comment was

But also perhaps it's worth keeping the note about it not being a `{name}.{namespace}` string, for people migrating to this from TLSContexts.  Something like "Note that this is a native-Kubernetes-style `core.v1.SecretReference`, not an old-Emissary-style `{name}.{namespace}` string."

AliceProxy: That's a good idea for sure and something to call out when I make a docs update for this for sure.

LukeShu: Ideally the `kubectl describe` docs are all helpful, and One Day™ we'll have website docs generated from it.

AliceProxy:
> Don't we need to target this at 2.x?

It sure does. Since we don't want 3.x to be missing any features this PR will target `master` and once any Feedback is taken care of I'll just cherry pick the commits over to 2.x

Why did I target 3.x first instead of 2.x? No clue :woman_shrugging:

LukeShu:
I said on a call once that I'd add an env-var to make this opt-in. But since it was already possible with a TLSContext, as I think about it more I'm not sure I stand by that. But if we eventually drop TLSContext, it'd be good if things could be locked down. So IDK.

LukeShu (approved):
Implementation looks good to me.

I love how simple this is is a non-legacy-mode world.

python/ambassador/ir/irhost.py
```diff
@@ -197,9 +207,13 @@ def setup(self, ir: "IR", aconf: Config) -> bool:
                             namespace=self.namespace,
                             location=self.location,
                             hosts=[self.hostname or self.name],
-                            secret=tls_name,
+                            secret=tls_full_name,
                         )
 
+                        # Ensure that we handle the secret properly even if tls_secret_namespacing is False in the Module
+                        if tls_namespace:
+                            tls_context_init["secret_namespacing"] = True
```
LanceEa, line 215: Am I understanding this comment correctly...If I was a Platform Admin and set up a Module to disable this behavior. Then me as a developer comes along and says I don't care ignore it because I put? 

Or does this flag mean something else?

LukeShu: The `Host` generates its own internal `IRTLSContext`. The `IRTLSContext` uses the questionable representation of `secret={name}, secret_namespacing=false` for non-cross-namespace secrets, and `secret={name}.{namespace}, secret_namespacing=true` for cross-namespace secrets.

But, you raise a good point that setting `secret_namespacing=False` in the Module effectively turns off the ability to do cross-namespace secrets in a manor similar to my proposed env-var, so perhaps we should disallow cross-namespace secrets if `ir.module.secret_namespacing==False`.

AliceProxy: In a TLSContext you can only specify secrets by providing a string such as `secret-name` or `secret-name.foo` By default secret namespacing is turned on. That tells a TLSContext that hey `secret-name.foo` is `secret-name` in the `foo` namespace. You can turn that behaviour off if you want, commonly because you might have a secret called `example.domain.net` where none of the `.`s are meant to indicate the namespace of the secret. 

When we create the implicit TLSContext for this Host we obviously want our TLSContext to treat the name as a namespace because one was explicitly supplied so I told the TLSContext to force the behaviour on otherwise if you have a Module that sets `tls_secret_namespacing: False` this feature won't work at all.

I think your comment about What if a platform Admin wants to disable the feature for security, not just to make sure that the format of the secrets is handled correctly has a point so I'll see if I can rework it real quick to not be dependent on how TLSContexts handle the secret name.

LukeShu: Oh, but at the `ambassador` `Module` level it's just a default; so you can still override it on a per-`TLSContext` basis. So that's not any change here.

LanceEa: Ok, thanks for details. That makes a ton more sense :)

LanceEa (approved):

PR #4435, opened 2022-08-22
[v3.2] Repatriate from v3.1 and kick of v3.2.0 development

d6e-automaton:
This PR merges release/v3.1 in to master.

----

Notes from @LukeShu:

OK, there's actually enough new stuff in this one that's not purely repatriation that it needs a review (not from me):

OK, so things to look for when reviewing this:

Things being brought forward (for these, you mostly want to look at the sum diff, not individual commits, since many of the brought-forward commits are cherry-picked commits that we already have, or merge commits; this is the typical mode for repatriation PRs):
- OK, so most of the big things that have happened in older releases that are getting forward were backports to 2.3 that were already in 3.1, so this PR is largely no-op; with the biggest changes from that being the added release notes for the releases with those backports.
- It also includes the "are you targeting the correct branch" CI check and the `prepare-y-bump`
- It includes a fix to the `repatriate.yml` CI; there was a missing `if: always()`
- It includes an inconsequential upgrade to dlib because I wanted a newer dlib when doing the one 1.14 dev build.
- It upgrades Telepresence 2.4.2→2.6.6; why I chose 2.6.6 is discussed in https://github.com/emissary-ingress/emissary/pull/4421, but I have since learned that an added bonus of this (that really makes it worth pulling forward) is that Telepresence only started doing darwin/arm64 builds at 2.4.6, so this upgrade helps unblock local development on M1 macs.

New things (for these, you should look at individual commits):
 - 0cdbe812a4c98dd0eca8bc23b85426d45cc4bf6d, 9fd5b384dfb54fc1721d3bec78354680aa42c556, 56f27cd679336248378bc6bac4acac2989eb80b2 : These 3 commits are touching up the `prepare-y-bump` script, and running it; so that `master` knows that it's for v3.2 now instead of v3.1. 9fd5b384dfb54fc1721d3bec78354680aa42c556 is tagged as both  `v3.2.0-dev` and `chart/v8.2.0-dev`.


kflynn (dismissed):
Looks sane, let's get this landed!

haq204:
Looks good to me. Will approve once tests pass.

haq204 (approved):

ddymko (approved):
lgtm

PR #4447, opened 2022-08-24
Fix : Bypass EDS to resolve 503s for cert rotation

ddymko:
## Description
When we rotate a cert this causes envoy to create a new Cluster, which has the same name as the old one, but since it is a new cluster the transport socket changes. This will then cause the cluster to be put in a "warming" state until it gets a ClusterLoadAssignment from xDS which will supply the endpoints from k8. 

The solution here is to manually update the cluster configuration and insert the endpoints instead of relying on EDS when using the endpoint resolver.

**NOTE** This doesn't change the default behavior which is to use EDS when you are using endpoint resolver. To bypass eds you must set a new environmental variable called `AMBASSADOR_EDS_BY_PASS` to true.

## Related Issues
List related issues.

## Testing
Tested this manually on k8 clusters to make sure behavior has stayed the same 

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/pkg/ambex/main.go
+++ b/pkg/ambex/main.go
@@ -138,7 +138,7 @@ type Args struct {
 	snapdirPath string
 	numsnaps    int
 
-	// edsByPass will bypass using EDS and will insert the endpoints into the custer data manually
+	// edsByPass will bypass using EDS and will insert the endpoints into the cluster data manually
 	// This is a stop gap solution to resolve 503s on certification rotation
 	edsByPass bool
 }
@@ -196,10 +196,11 @@ func parseArgs(ctx context.Context, rawArgs ...string) (*Args, error) {
 		dlog.Errorf(ctx, "Invalid AMBASSADOR_AMBEX_SNAPSHOT_COUNT: %s, using %d", numsnapStr, args.numsnaps)
 	}
 
-	// edsByPass will bypass using EDS and will insert the endpoints into the custer data manually
+	// edsByPass will bypass using EDS and will insert the endpoints into the cluster data manually
 	// This is a stop gap solution to resolve 503s on certification rotation
 	edsByPass := os.Getenv("AMBASSADOR_EDS_BY_PASS")
-	if edsByPass == "true" {
+	if strings.ToLower(edsByPass) == "true" {
+		dlog.Info(ctx, "AMBASSADOR_EDS_BY_PASS has been set to true. EDS will not be bypassed and endpoints will be inserted manually.")
 		args.edsByPass = true
 	}
 
--- a/pkg/ambex/main.go
+++ b/pkg/ambex/main.go
@@ -137,6 +137,10 @@ type Args struct {
 
 	snapdirPath string
 	numsnaps    int
+
+	// edsByPass will bypass using EDS and will insert the endpoints into the custer data manually
+	// This is a stop gap solution to resolve 503s on certification rotation
+	edsByPass bool
 }
 
 func parseArgs(ctx context.Context, rawArgs ...string) (*Args, error) {
@@ -190,6 +194,13 @@ func parseArgs(ctx context.Context, rawArgs ...string) (*Args, error) {
 	if (err != nil) || (args.numsnaps < 0) {
 		args.numsnaps = 30
 		dlog.Errorf(ctx, "Invalid AMBASSADOR_AMBEX_SNAPSHOT_COUNT: %s, using %d", numsnapStr, args.numsnaps)
+	}
+
+	// edsByPass will bypass using EDS and will insert the endpoints into the custer data manually
+	// This is a stop gap solution to resolve 503s on certification rotation
+	edsByPass := os.Getenv("AMBASSADOR_EDS_BY_PASS")
+	if edsByPass == "true" {
+		args.edsByPass = true
 	}
 
 	return &args, nil
@@ -413,6 +424,7 @@ func update(
 	ctx context.Context,
 	snapdirPath string,
 	numsnaps int,
+	edsByPass bool,
 	config ecp_v2_cache.SnapshotCache,
 	configv3 ecp_v3_cache.SnapshotCache,
 	generation *int,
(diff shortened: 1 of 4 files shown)


haq204:
Are we only merging this into 2.4?

ddymko:
@haq204 Yeah, for now, it's going to be going into the 2.4 release

pkg/ambex/main.go
```diff
@@ -137,6 +137,10 @@ type Args struct {
 
 	snapdirPath string
 	numsnaps    int
+
+	// edsByPass will bypass using EDS and will insert the endpoints into the custer data manually
```
LanceEa, line 141: spelling: `custer` -> `cluster` no reason to do an extra CI run for it but if you do make other changes.

pkg/ambex/transforms.go
```diff
@@ -246,35 +246,50 @@ func V3ListenerToRdsListener(lnr *apiv3_listener.Listener) (*apiv3_listener.List
...
 
-		var source string
-		ep, ok := edsEndpoints[ref]
-		if ok {
-			source = "found"
+		// This change was introduced as a stop gap solution to mitigate the 503 issues when certificates are rotated.
+		// The issue is CDS gets updated and waits for EDS to send ClusterLoadAssignment.
+		// During this wait period calls that are coming through get hit with a 503 since the cluster is in a warming state.
+		// The solution is to "hijack" the cluster and insert all the endpoints instead of relying on EDS.
+		// Not there will be a discrepancy between envoy/envoy.json and the config envoy has
```
LanceEa, line 268: spell: `Not` -> `Now`

pkg/ambex/main.go
```diff
@@ -192,6 +196,13 @@ func parseArgs(ctx context.Context, rawArgs ...string) (*Args, error) {
 		dlog.Errorf(ctx, "Invalid AMBASSADOR_AMBEX_SNAPSHOT_COUNT: %s, using %d", numsnapStr, args.numsnaps)
 	}
 
+	// edsByPass will bypass using EDS and will insert the endpoints into the custer data manually
+	// This is a stop gap solution to resolve 503s on certification rotation
+	edsByPass := os.Getenv("AMBASSADOR_EDS_BY_PASS")
```
LanceEa, line 201: We do have a helper method `envBool`. It's behavior basically ignores the value of the os.Environment variable and is more of a if the environment variable exists with any value then it is true. I'm not sure I agree with that behavior (need to think about it) but it might be good to be consistent?

take a look and let me know your thoughts...

pkg/ambex/main.go
```diff
@@ -192,6 +196,13 @@ func parseArgs(ctx context.Context, rawArgs ...string) (*Args, error) {
...
+	if edsByPass == "true" {
+		args.edsByPass = true
```
LanceEa, line 203: I think we should add a log entry so that it is known that the feature flag is enabled

pkg/ambex/transforms.go
```diff
@@ -246,35 +246,50 @@ func V3ListenerToRdsListener(lnr *apiv3_listener.Listener) (*apiv3_listener.List
...
+		if edsByPass {
+			if ep, ok := edsEndpoints[ref]; ok {
+				c.LoadAssignment = ep
+				c.EdsClusterConfig = nil
+
+				// Type 1 is STRICT_DNS
+				c.ClusterDiscoveryType = &apiv2.Cluster_Type{Type: 1}
```
LanceEa, line 275: Aren't the endpoints in this case already resolved to IP:Port so no DNS service discover is needed?

I would think we would just set it back to the default of 
 `STATIC`, no?
 https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/service_discovery#static?

LanceEa (commented):
Left some questions and nits

haq204 (approved):

pkg/ambex/transforms.go
ddymko: updated

pkg/ambex/main.go
ddymko: I am not a fan of ignoring the value of the env var. I say we leave it as is but I added in `strings.ToLower` so that if a user puts in True or true (or any variant of true) it will be valid.

pkg/ambex/main.go
ddymko: added

pkg/ambex/transforms.go
ddymko: Good catch! 

I didn't think this one through. I changed it to Static (type 0) and confirmed it is working on a cluster.

pkg/ambex/transforms.go
```diff
@@ -320,14 +320,14 @@ func JoinEdsClustersV3(ctx context.Context, clusters []ecp_cache_types.Resource,
...
+		// Now there will be a discrepancy between envoy/envoy.json and the config envoy.
 		if edsByPass {
 			if ep, ok := edsEndpoints[ref]; ok {
 				c.LoadAssignment = ep
 				c.EdsClusterConfig = nil
 
-				// Type 1 is STRICT_DNS
-				c.ClusterDiscoveryType = &apiv3_cluster.Cluster_Type{Type: 1}
+				// Type 0 is STATIC
+				c.ClusterDiscoveryType = &apiv3_cluster.Cluster_Type{Type: 0}
```
haq204, line 330: in addition to websockets, lets also test with grpc as well to make sure load balancing works correctly. I don't think there will be any issues since the upstream hosts are directly applied, envoy should connect to all of them but still good to check

ddymko: we can do follow up tests once we have a dev build out

haq204 (approved):

LanceEa (approved):
lgtm...let's just confirm the discussion about opt-in vs opt-out

ddymko:
@LanceEa confirmation that the `eds by pass` will be opt in

pkg/ambex/main.go
```diff
@@ -192,6 +196,14 @@ func parseArgs(ctx context.Context, rawArgs ...string) (*Args, error) {
 		dlog.Errorf(ctx, "Invalid AMBASSADOR_AMBEX_SNAPSHOT_COUNT: %s, using %d", numsnapStr, args.numsnaps)
 	}
 
+	// edsByPass will bypass using EDS and will insert the endpoints into the cluster data manually
+	// This is a stop gap solution to resolve 503s on certification rotation
+	edsByPass := os.Getenv("AMBASSADOR_EDS_BY_PASS")
+	if strings.ToLower(edsByPass) == "true" {
```
LukeShu, line 202: This should be equivalent to `strconv.ParseBool`, but in general: Prefer `strconv.ParseBool`.

LukeShu (commented):
a couple of nits, but lgtm

rgs1:
Any plans to make this the default behavior? Since the race between EDS and CDS still exists in ambex, right?

PR #4449, opened 2022-08-24
Migrate ambassador-agent to datawire/ambassador-agent repo

njayp:
## Description
This PR does three things
- modifies the helm repo to pull the image from a new dockerhub repo
- removes the agent from the codebase
- imports the agent from the new repo `datawire/ambassador-agent`


charts/emissary-ingress/templates/_helpers.tpl
```diff
@@ -71,7 +71,7 @@ disabled if fullImageOverride is present
 {{- if hasKey .Values.image "repository" -}}
 {{- printf "%s:%s" .Values.image.repository .Values.canary.image.tag -}}
 {{- else -}}
-{{- printf "%s:%s" "docker.io/emissaryingress/emissary" .Values.canary.image.tag -}}
+{{- printf "%s:%s" "docker.io/ambassador/ambassador-agent" .Values.canary.image.tag -}}
```
LukeShu, line 74: These changes are problematic because it'll also use that image for the main Emissary.  It needs to grow a separate image config for the agent.

cmd/busyambassador/main.go
```diff
@@ -13,11 +13,11 @@ import (
 	"strings"
 
 	// 1st-party libs
+	"github.com/datawire/ambassador-agent/cmd/agent"
```
LukeShu, line 16: So do you expect Emissary to consume the agent as an image, or as a Go library?

njayp: At this point it does both, but I think we want to consume as an image. I don't understand this package enough to know if I could just rip out this import

charts/emissary-ingress/templates/_helpers.tpl
njayp: the helm chart is updated

PR #4454, opened 2022-08-25
[v3.0] Repatriate from v2.4

d6e-automaton:
This PR merges release/v2.4 in to release/v3.0.

PR #4455, opened 2022-08-25
[v3.0] Repatriate from v2.4

LukeShu:
This is a more selective version of https://github.com/emissary-ingress/emissary/pull/4454, in order to avoid pulling in features.


kflynn (approved):
This looks pretty sane.

PR #4456, opened 2022-08-25
[v2.4] .github: Skip over v3.0 and v3.1 when repatriating

LukeShu:
## Description

Besides just the tedium of having to merge a 2.4→3.0 PR then a 3.0→3.1 PR then a 3.1→3.2 PR, even just landing the 2.4→3.0 PR would be problematic because it would contain features, and 3.0 is feature-frozen. So have 2.4 repatriate straight in to 3.2.

### DIFF
--- a/.github/workflows/repatriate.yml
+++ b/.github/workflows/repatriate.yml
@@ -29,7 +29,7 @@ jobs:
           this_branch=release/${this_ver}
           echo "::set-output name=this_branch::${this_branch}"
 
-          next_ver=$(git for-each-ref --format='%(refname:lstrip=4)' 'refs/remotes/origin/release/v*' | sort --version-sort | grep -A1 -Fx "$this_ver" | sed 1d)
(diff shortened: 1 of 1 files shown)


AliceProxy (approved):

.github/workflows/repatriate.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           this_branch=release/${this_ver}
           echo "::set-output name=this_branch::${this_branch}"
 
-          next_ver=$(git for-each-ref --format='%(refname:lstrip=4)' 'refs/remotes/origin/release/v*' | sort --version-sort | grep -A1 -Fx "$this_ver" | sed 1d)
+          next_ver=$(git for-each-ref --format='%(refname:lstrip=4)' 'refs/remotes/origin/release/v*' | sort --version-sort | grep -vFx -e 'v3.0' -e 'v3.1' | grep -A1 -Fx "$this_ver" | sed 1d)
```
haq204, line 32: Would we have to keep updating this line for new 3.y releases?

LukeShu: Presumably no. Assuming that 3.2 will have feature parity with 2.4, everything that goes in to 2.4 will be repatriatable to 3.2. It may reduce the tedium to add new releases here, but it wouldn't be problematic if we don't. If we do a 2.5, we may have to re-assess this list, but please let's not do a 2.5.

PR #4462, opened 2022-08-25
[v3.1] charts: CHANGELOG.md: Fill in missing versions

LukeShu:
## Description

I looked for GA chart Git tags that aren't mentioned in the CHANGELOG.

```bash
comm <(grep '^## v' CHANGELOG.md |sed 's,## ,chart/,'|sort) <(git tag |grep ^chart/v|grep -v -e -|sort)
```

I also noticed that chart/v8.0.0 didn't mention that it had an Emissary update.

I'm just targeting this PR at the latest released branch; I don't think it's particularly worth backfilling these to old branches... *maybe* the `release/v2.4` branch.

### DIFF
--- a/charts/emissary-ingress/CHANGELOG.md
+++ b/charts/emissary-ingress/CHANGELOG.md
@@ -5,9 +5,17 @@ numbering uses [semantic versioning](http://semver.org).
 
 ## Next Release
 
+(no changes yet)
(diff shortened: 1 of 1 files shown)


kflynn (approved):
😱 yes, please, let's land this.

PR #4464, opened 2022-08-26
ratelimit: add failure_mode_deny field

LanceEa:
## Description

By default, Envoy will fail open when it is unable to communicate with the configured service. This can be surprising and potentially lead to taking down upstream services. The `failure_mode_deny` field has been added to the `RateLimitService` resource so that Envoy can be configured to reject request when it is unable to communicate with the service. When this happens Envoy will return a 500.

## Related Issues

https://github.com/emissary-ingress/emissary/issues/4465

## Testing

Updated unit test for diagd and CRD conversion to test new field. CI is green and also manually tested by setting up a cluster with the RateLimitService set to talk with a service that doesn't exist. The default value allows traffic through like existing behavior. When setting `failure_mode_deny: true` then it will return a 500 as seen here:

```shell
curl -v http://xx.xxx.xxx.xxx/ratelimit-v3/
*   Trying xx.xxx.xxx.xxx...
* Connected to xx.xxx.xxx.xxx (xx.xxx.xxx.xxx) port 80 (#0)
> GET /ratelimit-v3/ HTTP/1.1
> Host: xx.xxx.xxx.xxx
> User-Agent: curl/7.85.0-DEV
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< content-type: application/json
< date: Fri, 26 Aug 2022 15:57:05 GMT
< content-length: 158
< x-envoy-upstream-service-time: 0
< server: envoy
<
{
    "server": "jovial-mango-35wuwv6r",
    "quote": "The last sentence you read is often sensible nonsense.",
    "time": "2022-08-26T15:57:05.596169812Z"

}

curl -v http://xx.xxx.xxx.xxx/ratelimit-v3/
* Connected to xx.xxx.xxx.xxx(xx.xxx.xxx.xxx) port 80 (#0)
> GET /ratelimit-v3/ HTTP/1.1
> Host: xx.xxx.xxx.xxx
> User-Agent: curl/7.85.0-DEV
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 500 Internal Server Error
< date: Fri, 26 Aug 2022 15:56:36 GMT
< server: envoy
< content-length: 0
<

```

## Checklist

### DIFF
--- a/pkg/api/getambassador.io/v2/crd_ratelimitservice.go
+++ b/pkg/api/getambassador.io/v2/crd_ratelimitservice.go
@@ -39,6 +39,12 @@ type RateLimitServiceSpec struct {
 
 	// +k8s:conversion-gen:rename=StatsName
 	V3StatsName string `json:"v3StatsName,omitempty"`
+
+	// +k8s:conversion-gen:rename=FailureModeDeny
+	//
+	// FailureModeDeny when set to true, envoy will deny traffic if it
+	// is unable to communicate with the rate limit service.
+	V3FailureModeDeny bool `json:"failure_mode_deny,omitempty"`
 }
 
 // RateLimitService is the Schema for the ratelimitservices API
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,12 @@ it will be removed; but as it won't be user-visible this isn't considered a brea
 - Bugfix: Distinct services with names that are the same in the first forty characters will no
   longer be incorrectly mapped to the same cluster. ([#4354])
 
+- Feature: By default, when Envoy is unable to communicate with the configured RateLimitService then
+  it will allow traffic through. The  `RateLimitService` resource now exposes the  <a
+  href="https://www.envoyproxy.io/docs/envoy/v1.23.0/configuration/http/http_filters/rate_limit_filter">failure_mode_deny</a>
+  option. Set `failure_mode_deny: true`, then Envoy will  deny traffic when it is unable to
+  communicate to the RateLimitService  returning a 500.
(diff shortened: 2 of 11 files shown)


pkg/api/getambassador.io/v2/crd_ratelimitservice.go
```diff
@@ -39,6 +39,12 @@ type RateLimitServiceSpec struct {
 
 	// +k8s:conversion-gen:rename=StatsName
 	V3StatsName string `json:"v3StatsName,omitempty"`
+
+	// +k8s:conversion-gen:rename=FailureModeDeny
```
ddymko, line 43: this is more of a clarifying question.

I am assuming this conversion is needed so it aligns with what we have defined for v3?

ddymko (dismissed):
changes look good

pkg/api/getambassador.io/v2/crd_ratelimitservice.go
LanceEa: Yes, correct. Because we use V2 as our storage version when they are applied it is converted from v3alpha1 to v2. We need a field on the V2 struct to hold this data. We use the convention of adding a V3 at the beginning to indicate it is a v3 only field for storage. The conversion-gen makes sure the generated code will copy the contents of `V3FailureModeDeny` into `FailureModeDeny` and vice-versa.

LanceEa:
@ddymko @AliceProxy - Fixed merge conflicts due to change log and re-pushed. No other changes were made since last time.

PR #4477, opened 2022-09-01
Lukeshu/v1.14 tcpmapping

LukeShu:
## Description

## Related Issues
List related issues.

## Testing

## Checklist

    + Any new features
    + Any deprecations


LanceEa:
@LukeShu - General question on the Gold Tests...I thought we dropped those in v2/v3, correct? What was the reason for dropping these? Did we not feel they were beneficial, duplicates of other tests, or some other reason?

LukeShu:
> General question on the Gold Tests...I thought we dropped those in v3, correct? What was the reason for dropping these? Did we not feel they were beneficial, duplicates of other tests, or some other reason?

The annotation fixes in 2.1.2 (https://github.com/emissary-ingress/emissary/pull/4016) willfully broke our ability to run the gold tests; as the mechanisms it uses to run those tests need substantially reworked to work with the fixes.

That said, in the months since then, I have come to believe that we don't get value from the gold tests.  The gold tests were an hacky optimization; avoiding CI-billing, dev-wait-time, and CI-flakes associated with running the full `KAT_RUN_MODE=envoy` tests.
- However, we had previously (in the 1.x timeframe) had several incidents where we were bitten by a developer mistakenly checking in bad gold files and a regression getting shipped; as a corrective action, we started running the `KAT_RUN_MODE=envoy` tests in CI (you should be able to find postmortems in Notion).  In my opinion; at that point the gold tests are redundant; but potentially still useful for saying "oh, the envoy failed but the gold didn't change; this is almost certainly a CI flake".  I'm not sure anyone intended for "have CI always run both gold and envoy" to be a long term situation; but a temporary CI bloat while we figure out how to safely trim CI back down.
- The value that we'd get by switching back to running the tests with gold is minimal:
  - We have migrated from CircleCI to GitHub Actions; GHA is free for Emissary, so there's no costs to cut there; though there is in Edge Stack.
  - We have split the envoy tests across multiple jobs; this has both cut time dev-wait-time and reduced flakes.  The "slow" envoy tests now run faster than the "fast" gold tests ever did.
  - As we've been forced to live with the Envoy tests in CI, we've dramatically improved their flakiness.

That said, I never managed to convince @kflynn that the gold tests aren't worth bringing back; so the scripts related to them haven't been deleted. Flynn, would you care to weigh in on why you believe that the gold tests are worth bringing back?

> I see these are still available in v2 so we at least kept them around for that.

No? The gold directory does not exist in `release/v2.4`.  Some of the machinery around gold is still there; but because it's not getting run, the gold files are dead weight, so we deleted them.  If we start running them again, all that it would take to bring them back is `make pytest-gold`.

LanceEa (approved):

LukeShu:
I've filed https://github.com/emissary-ingress/emissary/issues/4515

haq204:
do any fixes from https://github.com/emissary-ingress/emissary/pull/4493 need to be ported over?

LukeShu:
> do any fixes from #4493 need to be ported over?

No, the only fixes that #4493 has are 2.x-specific tests.

haq204 (approved):

PR #4478, opened 2022-09-01
Revert "Use the external github.com/emissary-ingress/goversion"

ddymko:
Reverts emissary-ingress/emissary#4463

Reverting this due to build issues


LanceEa (approved):
lgtm...

PR #4488, opened 2022-09-06
pkg/kates: fix bug where Accumulator was not coalescing changes mid-watch 

haq204:
## Description
The Accumulator struct attempts to coalesce changes into a single snapshot update as a way to do graceful load shedding.
However, while this was the behavior on bootstrap, it didn't always happen mid-watch - each event that was received turned into a single snapshot update, thus not really satisfying this requirement.

We add a new option to batch changes for a specified window interval before sending a snapshot update.
The batching behavior is as follows:
 - The Accumulator will receive raw changes up until the window period where it will then send a change, even if new updates are still coming in.
   This is to prevent the potential of a scenario where a change is never sent due to an extremely volatile cluster.
   While there may be a way to be more dynamic in how long to wait before sending this change, this approach is simpler and more predicable.

 - If an isolated updated comes in (e.g. last change was submitted an hour ago but the window period is set to 10s), it may not neccessarily wait until the window period before sending change, it can send immediately.

 - The default interval is set to 1s to be inline with current change velocity.

 - A snapshot update won't be sent until all resources are fully bootstrapped, regardless of what interval is set.
   This is the ensure that the other requirements for the Accumulator are still satisfied.

`AMBASSADOR_RECONFIG_MAX_DELAY` controls the interval to wait before sending snapshot updates when listening for K8s resources, especially when many resources are updated in quick succession.

## Related Issues
Issue is not in this repo.

## Testing
Add new test cases. 

For performance testing, ran a test that concurrently applies namespaces with each namespace applying 35 deployments, host, and mappings each. Each deployment has 2 replicas. We track the number of snapshot versions pushed. Prevously the number of snapshot versions created when applying 1 namespace was ~118. After setting a 10s interval, the snapshot version reduced to 16. For max concurrent namespaces it would previously start OOMing at 6 concurrent namespaces; with a 10 sec interval it's at least >15. Peak memory usage was 600MB so with a higher memory limit, it can likely support much higher.

## Checklist

    + Any new features
    + Any deprecations

 - [ ] This is unlikely to impact how Ambassador performs at scale - load testing shows xxxx.
   

 - [x] I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently - N/A.

### DIFF
--- a/cmd/entrypoint/watcher.go
+++ b/cmd/entrypoint/watcher.go
@@ -5,8 +5,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	gw "sigs.k8s.io/gateway-api/apis/v1alpha1"
 
@@ -33,6 +35,16 @@ func WatchAllTheThings(
 	if err != nil {
 		return err
 	}
+	intv, err := strconv.Atoi(env("AMBASSADOR_RECONFIG_MAX_DELAY", "1"))
+	if err != nil {
+		return err
+	}
+	maxInterval := time.Duration(intv) * time.Second
+	err = client.MaxAccumulatorInterval(maxInterval)
+	if err != nil {
+		return err
+	}
+	dlog.Infof(ctx, "AMBASSADOR_RECONFIG_MAX_DELAY set to %d", intv)
 
 	serverTypeList, err := client.ServerResources()
 	if err != nil {
--- a/docs/releaseNotes.yml
+++ b/docs/releaseNotes.yml
@@ -74,7 +74,6 @@ items:
           deny traffic when it is unable to communicate to the RateLimitService 
           returning a 500.
         docs: https://www.getambassador.io/docs/emissary/latest/topics/running/services/rate-limit-service/
-
       - title: Properly populate alt_state_name for Tracing, Auth and RateLimit Services
         type: bugfix
         body: >-
@@ -82,6 +81,11 @@ items:
           or the <code>AuthService</code> would have no affect because it was not being properly passed to the Envoy cluster
           config. This has been fixed and the <code>alt_stats_name</code> field in the cluster config is now set correctly.
           (Thanks to <a href="https://github.com/psalaberria002">Paul</a>!)
+      - title: Add support for config change batch window before reconfiguring Envoy
+        type: feature
+        body: >-
+          The <code>AMBASSADOR_RECONFIG_MAX_DELAY</code> env var can be optionally set to batch changes for the specified 
+          non-negative window period in seconds before doing an Envoy reconfiguration. Default is "1" if not set.
 
   - version: 3.1.1
     prevVersion: 3.1.0
--- a/pkg/kates/accumulator.go
+++ b/pkg/kates/accumulator.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"reflect"
 	"sync"
+	"time"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 )
@@ -78,6 +79,13 @@ const (
 	ObjectAdd DeltaType = iota
 	ObjectUpdate
 	ObjectDelete
+)
+
+type changeStatus int
+
+const (
+	awaitingDispatch changeStatus = iota
+	dispatched
 )
 
 func (dt DeltaType) MarshalJSON() ([]byte, error) {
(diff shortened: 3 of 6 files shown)


cmd/entrypoint/watcher.go
```diff
@@ -28,11 +29,13 @@ func WatchAllTheThings(
 	fastpathCh chan<- *ambex.FastpathSnapshot,
 	clusterID string,
 	version string,
+	maxInterval time.Duration,
 ) error {
 	client, err := kates.NewClient(kates.ClientConfig{})
 	if err != nil {
 		return err
 	}
+	client.MaxAccumulatorInterval(maxInterval)
```
LanceEa, line 38: Maybe a log line to help support team with debugging and allowing user to see it was set correctly?

pkg/kates/accumulator.go
```diff
@@ -165,47 +173,61 @@ func newAccumulator(ctx context.Context, client *Client, queries ...Query) (*Acc
...
 	return acc, nil
 }
 
+// This coalesces reads from rawUpdateCh to notifications that changes are available to be
+// processed. This loop along with the logic in storeField guarantees the 3
+// Goals/Requirements listed in the documentation for the Accumulator struct, i.e. Ensuring
+// all Kinds are bootstrapped before any notification occurs, as well as ensuring that we
+// continue to coalesce updates in the background while business logic is executing in order
+// to ensure graceful load shedding.
+func (a *Accumulator) Listen(ctx context.Context, rawUpdateCh chan rawUpdate, interval time.Duration) {
```
LanceEa, line 196: I would narrow types arguments to indicate that this channel is only for reading `rawUpdateCh <-chan rawUpdate`

pkg/kates/accumulator.go
```diff
@@ -165,47 +173,61 @@ func newAccumulator(ctx context.Context, client *Client, queries ...Query) (*Acc
...
+	ticker := time.NewTicker(100 * time.Millisecond)
```
LanceEa, line 197: I will ping you tomorrow, to make sure I better understand the ticker portion of the Select below. But, since this runs every 100ms does that mean the user must set their interval greater than this ticker value or else it will behave as though they set it to 0? Actually, thinking about it more as I'm writing this, your parsing logic ensures that the interval is in seconds, correct?

pkg/kates/accumulator_test.go
```diff
@@ -82,3 +93,72 @@ func TestBootstrapNotifyEvenOnEmptyWatch(t *testing.T) {
...
+		}
+		err := cli2.Upsert(ctx, cm, cm, &cm)
+		require.NoError(t, err)
+		cms[i] = cm
+	}
+	wg.Wait()
+
+	// After receiving 2 updates from the Accumulator, we should have 10 ConfigMaps
+	// in our Snapshot due to the Accumulator coalescing changes before sending an update.
+	assert.Equal(t, 10, len(snap.ConfigMaps))
```
LanceEa, line 155: Should there be an assertion after bootstrap and before we get the "mid" stream batched items? Or is this test saying that a single bootstrap will not occur because we will wait the interval and at the end of that interval you will on only get 1 "snapshot notification" (aka reconfigure) which will include all 10 items?

pkg/kates/accumulator_test.go
```diff
@@ -58,6 +60,14 @@ func TestBootstrapNoNotifyBeforeSync(t *testing.T) {
 	// ConfigMaps prior to starting te Watch, all 10 of those ConfigMaps should be present in the
 	// first update.
 	assert.Equal(t, 10, len(snap.ConfigMaps))
+
+	t.Cleanup(func() {
```
LanceEa, line 65: I don't think this is necessary is it? The client is being told to watch for configMaps and includes the LabelSelector as well. 

I guess you could have duplicate LabelSelectors so is that what this is trying to avoid?

LanceEa (commented):
Quite a few comments, questions and suggestions. I will ping you in the morning to go over a few items to make sure I'm understanding them all correctly.

BTW...awesome to see the improved testing results :)

LanceEa:
@ddymko - I think it is important for you to have yours eyes on this just so you familiarize yourself with it a little bit.

cmd/entrypoint/entrypoint.go
```diff
@@ -188,9 +189,14 @@ func Main(ctx context.Context, Version string, args ...string) error {
 
 	if !demoMode {
 		group.Go("watcher", func(ctx context.Context) error {
+			intv, err := strconv.Atoi(env("AMBASSADOR_WATCH_INTERVAL_SEC", "0"))
```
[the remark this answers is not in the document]
haq204: The original reason was to keep the config closer together instead of spread out. But to the point of having a general mechanism for config or a config object of sorts, while I prefer to have one, I'm not super opposed to moving it closer to it actual usage in this instance.

pkg/kates/accumulator_test.go
haq204: more the latter. this came about because when I originally wrote my other test I initially used the same label selectors so it was picking up way more config maps then it's supposed to. A future test may also accidentally pick them up

Also it's just good habit to clean up after yourself.

pkg/kates/accumulator_test.go
haq204: that is an oversight on my part - I should be also verifying that the first change event representing bootstrap is empty

pkg/kates/accumulator.go
haq204: users can only set in seconds

haq204:
Going to make some changes after discussions with @LanceEa and @ddymko on expected batching behavior

pkg/kates/client.go
```diff
@@ -211,6 +213,16 @@ func InCluster() bool {
 		err == nil && !fi.IsDir()
 }
 
+func (c *Client) MaxAccumulatorInterval(interval time.Duration) error {
```
ddymko, line 216: GoDoc would be nice since this is exported

docs/releaseNotes.yml
```diff
@@ -74,14 +74,18 @@ items:
...
         type: bugfix
         body: >-
           Previously, setting the <code>stats_name</code> for the <code>TracingService</code>, <code>RateLimitService</code> 
           or the <code>AuthService</code> would have no affect because it was not being properly passed to the Envoy cluster
           config. This has been fixed and the <code>alt_stats_name</code> field in the cluster config is now set correctly.
           (Thanks to <a href="https://github.com/psalaberria002">Paul</a>!)
+      - title: Add support for config change batch window before reconfiguring Envoy
+        type: feature
+        body: >-
+          The AMBASSADOR_RECONFIG_MAX_DELAY env var can be optionally set to batch changes for the specified
```
ddymko, line 87: I think we might want to wrap AMBASSADOR_RECONFIG_MAX_DELAY in `<code>AMBASSADOR_RECONFIG_MAX_DELAY</code>` just to keep the changelog consistent with how the other env vars are displayed

pkg/kates/accumulator.go
```diff
@@ -165,45 +173,57 @@ func newAccumulator(ctx context.Context, client *Client, queries ...Query) (*Acc
...
+		changed:  changed,
+		mutex:    sync.Mutex{},
+	}
+
+	go acc.Listen(ctx, rawUpdateCh, client.maxAccumulatorInterval)
+
+	return acc, nil
+}
+
+// This coalesces reads from rawUpdateCh to notifications that changes are available to be
```
ddymko, line 190: We just need to define `Listen` here for the godoc

pkg/kates/accumulator.go
```diff
@@ -165,45 +173,57 @@ func newAccumulator(ctx context.Context, client *Client, queries ...Query) (*Acc
...
+	var synced bool
+
+	sendUpdate := func() {
+		a.changed <- struct{}{}
+		changeStatus = dispatched
+		lastChangeSent = time.Now()
+	}
+
+	for {
+		select {
```
ddymko, line 210: Just for clarity sake.

If we set an interval of 30 seconds. We'll have two paths here 

- We get new data and it has been past our set interval since we last updated anything. We go ahead and immediately send that
- We get new data but we just recently set data without our interval limit. We'll now wait until we get our `Tick`

We should maybe put up a brief comment that highlights the intended behavior here?

docs/releaseNotes.yml
LanceEa: Let's make sure we get this in the Docs too. In the release notes and Alice's new Environment page

cmd/entrypoint/watcher.go
```diff
@@ -33,6 +35,16 @@ func WatchAllTheThings(
 	if err != nil {
 		return err
 	}
+	intv, err := strconv.Atoi(env("AMBASSADOR_RECONFIG_MAX_DELAY", "1"))
```
LanceEa, line 38: So, why was 1 second introduced? I thought the default was 0 to maintain existing behavior of sending each update as they are received?

LanceEa (commented):
overall looks good based on our discussion. Just a quick question and I agree with David's feedback.

cmd/entrypoint/watcher.go
haq204: It's mostly because the time.Ticker accepts only a non-zero positive interval and panics otherwise. So if we wanted to keep/support 0 interval you would have to hard code some interval between 0 and 1 seconds but then its not really 0 is it? 

From load tests, it seems with a slightly above 0 interval (e.g. 100ms) interval it would do a bit more reconfigurations that what it was previously so I chose the lowest non-zero second instead.

LanceEa (dismissed):
lgtm

ddymko:
@haq204 I think you have to rerun `make generate` because we cut the `3.2.0-rc.1` yesterday 

Otherwise it looks good!

LanceEa (approved):

ddymko (approved):
looks good

Issue #4476, opened 2022-09-01
Can't use a TLSContext to configure origination without a secret


LukeShu:
**Describe the bug**
`TLSContexts` are deprecated for TLS termination (which requires a secret), but are still the only way to provide advanced configuration for TLS origination (which does not require a secret) beyond "turn it on" such as specifying the TLS version to use, or what hostname to send in SNI.  However, the validator checks that the `TLSContext` has sec

**To Reproduce**
1. apply
```yaml
apiVersion: getambassador.io/v2
kind: TLSContext
metadata:
  name: my-tlsclient
spec:
  sni: my-funny-name
```
2. observe that Emissary discards the TLSContext, saying
```
2022-09-01 03:27:43 diagd 1.14.4-dev.26+g2b01994c2 [P19TAEW] DEBUG: my-tlsclient.default.1: <RichStatus BAD error='TLSContext myt-tlsclient has no certificate information at all?' hostname='tcpmappingtlsoriginationcontexttest' version='1.14.4-dev.26+g2b01994c2'>
```
3. Observe that it does not originate TLS when speaking to the upstream.

**Expected behavior**
I expected it to originate TLS with the given SNI information.

**Versions (please complete the following information):**
 - Ambassador: Tip of `release/v1.14`
 - Kubernetes environment: Kubeception, with a 1.18.7 cluster

**Additional context**
I only tested with 1.14, but it looks like that code is unchanged on `master`.

dethi:
I'm impacted by this too. This was previously working (see previous issue and fix https://github.com/emissary-ingress/emissary/issues/1708).

PR #4493, opened 2022-09-07
[v2.4] Add more tests for TCPMappings

LukeShu:
## Description

This PR adds a bunch of tests for TCPMappings, many of them marked `xfail`. Even so, it isn't *quite* as complete as I would like it to be; see the "TODO" comments at the bottom of `t_tcpmapping.py`.

This builds on the tests that are verified to work with 1.14 from https://github.com/emissary-ingress/emissary/pull/4477.

Cherry-picks some reliability improvements from https://github.com/emissary-ingress/emissary/pull/4308

## Related Issues
- Motivating issue: https://github.com/datawire/apro/issues/3036
- Found issue while writing the tests: https://github.com/emissary-ingress/emissary/issues/4476
- Sibling PR: https://github.com/emissary-ingress/emissary/pull/4477
- Cherry-picks from https://github.com/emissary-ingress/emissary/pull/4308

## Testing
That's all this is.

## Checklist

    + Any new features
    + Any deprecations

 - [x] My change is adequately tested. - that's all this is

### DIFF
--- a/python/tests/kat/t_tcpmapping.py
+++ b/python/tests/kat/t_tcpmapping.py
@@ -1,9 +1,10 @@
-from typing import Generator, Tuple, Union
+from typing import Dict, Generator, Tuple, Union
 
 from kat.harness import Query
 
 from abstract_tests import AmbassadorTest, ServiceType, HTTP, Node
 from tests.selfsigned import TLSCerts
+from kat.harness import abstract_test
 from tests.integration.manifests import namespace_manifest
 
 # An AmbassadorTest subclass will actually create a running Ambassador.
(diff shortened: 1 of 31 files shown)


python/tests/kat/t_tcpmapping.py
```diff
@@ -264,3 +265,735 @@ def check(self):
...
+metadata:
+  name: {self.name.k8s}
+spec:
+  ambassador_id: [ {self.ambassador_id} ]
+  port: 6789
+  service: {self.target.path.fqdn}:80
+''') + super().manifests()
+
+    def queries(self):
+        yield Query(self.url("", port=6789))
```
LanceEa, line 319: For my understanding: 

Does `Query` just send a random ping down a tcp socket? Then the results happens to be an http response which is interpreted and plugged into the results? Or is it sending an http request (since its built on tcp) which is enough to say that envoy is listening for TCP traffic and proxying it to the backend without considerations for the L7 protocol?

python/tests/kat/t_tcpmapping.py
```diff
@@ -264,3 +265,735 @@ def check(self):
...
+            backend = result.json["backend"]
+            counts[backend] = counts.get(backend, 0)+1
+        assert counts[self.target70.path.k8s]+counts[self.target30.path.k8s] == 1000
+        # Probabalistic, margin might need tuned
+        margin = 150
+        assert abs(counts[self.target70.path.k8s]-700) < margin
+        assert abs(counts[self.target30.path.k8s]-300) < margin
+
+
+# TODO: Add tests for all of the config knobs for the upstream connection:
```
LanceEa, line 1123: Rather than have a TODO at the bottom of a file, can you please create an issue for this and reference this PR?  

As we have seen from t_stats being disabled and the numerous other spots in the code base  where these get completely lost and ignored for years. An issue has the same potential but has the benefit of being more visible to the community, the support team, customer experience team, leadership team and ourselves.

haq204: +1

python/tests/kat/t_tcpmapping.py
LukeShu: It's sending an HTTP request.

This test works because the `TCPMapping` points to something that is itself serving HTTP.  Another one of the tests pints at something that is itself serving `HTTPS`, and we check the cert that it uses in order to test that Emissary is passing the raw stream through rather than terminating and re-originating the TLS.

python/tests/kat/t_tcpmapping.py
LukeShu: I've filed https://github.com/emissary-ingress/emissary/issues/4516

That said, it's a problem on its own that TODOs don't get attention.

python/tests/kat/t_tcpmapping.py
```diff
@@ -264,3 +265,868 @@ def check(self):
...
+---
+apiVersion: getambassador.io/v2
+kind: TCPMapping
+metadata:
+  name: {self.name.k8s}
+spec:
+  ambassador_id: [ {self.ambassador_id} ]
+  port: 6789
+  service: {self.target.path.fqdn}:443
+  tls: true
```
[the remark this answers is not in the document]
LukeShu: `tls: true` was how you said "originate TLS without a client cert (and with the default settings)" in v2; in v3alpha1 you do this by adding `https://` to the `service` (since allowing `tls` to be either a bool or a string is a violation of Kubernetes structural typing, which is something that v3alpha1 was going for).

python/tests/kat/t_tcpmapping.py
```diff
@@ -264,3 +265,868 @@ def check(self):
...
+    extra_ports = [ 6789, 6790 ]
+    target: ServiceType
+
+    def init(self) -> None:
+        self.target = HTTP()
+
+    def manifests(self) -> str:
+        return format('''
+---
+apiVersion: getambassador.io/v3alpha1
```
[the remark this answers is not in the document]
LukeShu: Oops, it was a mistake that I left `6790` in `extra_ports`.

The difference is that in v2 it didn't support saying "originate TLS by adding `https://` to the `service`; but in v3alpha1 this is the **primary** way of originating TLS without a client cert (since v3alpha1 doesn't support `tls: true`, which is how you did that in v2.)

python/tests/kat/t_tcpmapping.py
```diff
@@ -264,3 +265,868 @@ def check(self):
...
+---
+apiVersion: getambassador.io/v2
+kind: TCPMapping
+metadata:
+  name: {self.name.k8s}
+spec:
+  ambassador_id: [ {self.ambassador_id} ]
+  port: 6789
+  service: {self.target.path.fqdn}:443
+  tls: {self.name.k8s}-tlsclient
```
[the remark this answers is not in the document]
LukeShu: The answer is poorly defined, and I hate it, but it's OK that it's poorly defined since it's always been poorly defined.

Good question though.

python/tests/kat/t_tcpmapping.py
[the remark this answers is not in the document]
LukeShu: I've added a commit removing `6790` from `extra_ports`.

python/tests/kat/t_tcpmapping.py
LanceEa: Got it makes sense. Thanks!

LanceEa (approved):

ddymko (approved):
looks fine overall

haq204 (approved):

Issue #4175, opened 2022-03-30
Listener and Host selectors are wonky


LukeShu:
**Describe the bug**

You know what, I'm just going to post some things I said in Slack in #team-support

1. `hostBinding.namespace.from=SELECTOR` is accepted by the validator, but doesn't actually work.  When Emissary encounters that value it will

   ```python
   # (K8s validation should prevent this from happening.)
   self.post_error("hostBinding.namespace.from=selector is not yet supported")
   ```

   but, despite what the comment says, the k8s validation actually does accept "SELECTOR":

   ```go
   // +kubebuilder:validation:Enum=SELF;ALL;SELECTOR
   type NamespaceFromType string
   ```

2. When setting `hostBinding.selector` but not `hostBinding.namespace`, it behaves as if `hostBinding.namespace.from=SELF`. IMO this is surprising and isn't documented anywhere sufficiently prominently.

3. Despite it also being accepted by the k8s validator, Emissary only actually supports `hostBinding.selector.matchLabels`; it does not support `hostBinding.selector.matchExpressions`. This is because the selector matching is implemented in Python, rather than using the Go Kubernetes library for matching selectors.

4. Related to selector matching being implemented in Python: I wouldn't be shocked if our Python `selector_matches` is buggy.  Actually, reading the source, I can't believe that it works correctly.  In particular: I believe it will match a Host if *any* of the entries in `matchLabels` are matched, rather than if *all* of the entries are matched.  If we're going to implement the thing in Python, we should do property testing that it behaves identically to the Go Kubernetes library's version.

**To Reproduce**

**Expected behavior**

**Versions (please complete the following information):**
 - Ambassador: [e.g. 0.32.1]
 - Kubernetes environment [e.g. Minikube, bare metal, Google Kubernetes Engine]
 - Version [e.g. 1.8.1]

**Additional context**

PR #4499, opened 2022-09-09
Fix label selector matching and host/mapping pairing

dynajoe:
Fix bug where label matching used OR instead of AND. Fix bug where a Mapping could be paired to a Host only on hostname when selector mappings are also present.

## Description
The logic for pairing hosts and mappings seems incorrect. We witnessed that two separate listeners on different ports but with the same hostname had combined routes. This has significant implication on security by exposing backend services through unintended ports.

## Related Issues
The 4th bullet point on this issue seems related: https://github.com/emissary-ingress/emissary/issues/4175
another pr https://github.com/emissary-ingress/emissary/pull/4318/files

## Testing

Two separate Hosts with the same hostname but different mapping selectors.
Multiple mappings that use the same hostname with labels.

### DIFF
--- a/python/ambassador/ir/irhost.py
+++ b/python/ambassador/ir/irhost.py
@@ -472,8 +472,11 @@ class IRHost(IRResource):
                 dump_json(group.get("metadata_labels")),
                 sel_match,
             )
+        else:
+            # if the host does not have a mapping selector it matches any label set
+            sel_match = True
 
-        return host_match or sel_match
+        return host_match and sel_match
 
     def __str__(self) -> str:
         request_policy = self.get("requestPolicy", {})
--- a/python/ambassador/ir/irutils.py
+++ b/python/ambassador/ir/irutils.py
@@ -181,14 +181,11 @@ def selector_matches(
         logger.debug("    no incoming labels => False")
         return False
 
-    selmatch = False
-
+    # every selector label must exist and be equal
     for k, v in match.items():
-        if labels.get(k) == v:
-            logger.debug("    selector match for %s=%s => True", k, v)
-            return True
+        if labels.get(k) != v:
+            logger.debug("    selector match for %s=%s => False", k, v)
+            return False
 
-        logger.debug("    selector miss on %s=%s", k, v)
-
-    logger.debug("    all selectors miss => False")
-    return False
+    logger.debug("    all selectors match => True")
+    return True


AliceProxy:
Hi @dynajoe, could you please sign off your commit? The DCO bot should provide instructions for how to do so with:
```
git rebase HEAD~1 --signoff
git push --force-with-lease origin mapping-selector-fixes
```

dynajoe:
I signed off. Please feel free to take/copy the fixes even without merging this commit. I just wanted to highlight the way I believed this should work. I honestly didn't put a lot of time into making sure all of the debug statements present were still useful.

ghostsquad:
This is super important to my deployment. I'm implementing something like these label selectors:

```
"a8r.io/hostname": "*"
"a8r.io/requestPolicy": "route-insecure"
```

and

```
"a8r.io/hostname": "*"
"a8r.io/requestPolicy": "redirect-insecure"
```

instead of putting a bunch of data in 1 label. Not sure if this is best practice or not, but this bug affects this working model. Thanks for your time.

AliceProxy:
> This is super important to my deployment. I'm implementing something like these label selectors:
> 
> ```
> "a8r.io/hostname": "*"
> "a8r.io/requestPolicy": "route-insecure"
> ```
> 
> and
> 
> ```
> "a8r.io/hostname": "*"
> "a8r.io/requestPolicy": "redirect-insecure"
> ```
> 
> instead of putting a bunch of data in 1 label. Not sure if this is best practice or not, but this bug affects this working model. Thanks for your time.

Hi @ghostsquad, thanks for the information. I would say that your example is a best practice but unfortunately there are just several bugs with our implementation :disappointed: at the moment. 

You can refer to https://github.com/emissary-ingress/emissary/pull/4506 for the current status that will pull in all the fixes for this issue. My original comments mentioned this being in a 2.4 release which actually happened yesterday, so the fix is being re-targeted at a 2.5 which we hope to have out within the next several weeks and it will include more improvements for the 2.x+ line.

The PR that I just mentioned is targeting master for the upcoming 3.2 release which should happen next week and we hope to still get this in for. Once any feedback is addressed on the PR I linked we will back-port the changes to the 2.5 release branch. 

If you have any questions, comments, or concerns about the implementation of the fixes then please feel free to leave feedback on my PR.

PR #4506, opened 2022-09-13
[v3.2] Fix label selectors Re: PRs #4499 & #4318

AliceProxy:
## Description
Pulls in the commits from #4499 and #4318. There was a little extra work needed on top of those commits. I also added a changelog entry and some test coverage for the new strict matching behaviour as well as the environment variable `DISABLE_STRICT_LABEL_SELECTORS` which restores the old (flawed) behaviour. 

Plan moving forward with this is to first merge #4499 and #4318 once this PR is approved, then merge this PR on top of them immediately afterwards. 

I'll also throw up a PR cherry-picking all of this stuff back into ~~`release/v2.4`~~ `release/v2.5` so we can have the fix in both the upcoming `3.2` and ~~`2.4`~~ `2.5`.

### Issue Context:
Currently there are some problems with `Host` association with `Mappings` particularly regarding `mappingSelector`. 

1. According to all of our documentation, when `mappingSelector` and `hostname` are both present on a Mapping, but the current code treats it as an OR relationship. This means that if a `mappingSelector` is configured that would exclude a `Mapping` from being associated with a `Host`, then the rules enforced by the `mappingSelector` will be disregarded if the `Mapping` has a `hostname` that matches the `hostname` of the Host. This is flat out a bug since it goes against everything we have docummented about their interaction.
2. There is a problem in the logic of `mappingSelector`. Currently, if a `mappingSelector` requires 3 different labels for example, then if only one of those labels is present on a `Mapping` then it will be associated with the `Host` despite the fact that the `mappingSelector` requires the 3 lables. This is also a bug and goes against how label selectors work everywhere else in Kubernetes.

## Related Issues
Sister PR for ~~`v2.4`~~ `v2.5` https://github.com/emissary-ingress/emissary/pull/4512

## Testing
Manual testing, we've already got a bunch of test-coverage for Hosts that are passing and I've also added new tests explicitly for this behaviour.

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/docs/releaseNotes.yml
+++ b/docs/releaseNotes.yml
@@ -43,6 +43,18 @@ items:
           release of 1.23. This provides $productName$ with the latest security patches, performances enhancments,
           and features offered by the envoy proxy.
         docs: https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.23/v1.23.0
+      - title: Fixed <code>mappingSelector</code> associating <code>Hosts</code> with <code>Mappings</code>
+        type: change
+        body: >-
+          Changes to label matching will change how <code>Hosts</code> are associated with <code>Mappings</code>. There was a bug with label
+          selectors that was causing <code>Hosts</code> to be incorrectly being associated with more <code>Mappings</code> than intended.
+          If any single label from the selector was matched then the <code>Host</code> would be associated with the <code>Mapping</code>.
+          Now it has been updated to correctly only associate a <code>Host</code> with a <code>Mapping</code> if <b>all</b> labels required by
+          the selector are present. This brings the <code>mappingSelector</code> field in-line with how label selectors are used
+          in Kubernetes. To avoid unexpected behaviour after the upgrade, add all labels that Hosts have in their
+          <code>mappingSelector</code> to <code>Mappings</code> you want to associate with the <code>Host</code>. You can opt-out of the new behaviour
+          by setting the environment variable <code>DISABLE_STRICT_LABEL_SELECTORS</code> to <code>"true"</code> (default: <code>"false"</code>).
+          (Thanks to <a href="https://github.com/f-herceg">Filip Herceg</a> and <a href="https://github.com/dynajoe">Joe Andaverde</a>!).
       - title: Add support for Host resources using secrets from different namespaces
         type: feature
         body: >-
@@ -67,24 +79,24 @@ items:
         type: feature
         body: >-
           By default, when Envoy is unable to communicate with the configured
-          RateLimitService then it will allow traffic through. The 
-          <code>RateLimitService</code> resource now exposes the 
-          <a href="https://www.envoyproxy.io/docs/envoy/v1.23.0/configuration/http/http_filters/rate_limit_filter">failure_mode_deny</a> 
-          option. Set <code>failure_mode_deny: true</code>, then Envoy will 
-          deny traffic when it is unable to communicate to the RateLimitService 
+          RateLimitService then it will allow traffic through. The
+          <code>RateLimitService</code> resource now exposes the
+          <a href="https://www.envoyproxy.io/docs/envoy/v1.23.0/configuration/http/http_filters/rate_limit_filter">failure_mode_deny</a>
+          option. Set <code>failure_mode_deny: true</code>, then Envoy will
+          deny traffic when it is unable to communicate to the RateLimitService
           returning a 500.
         docs: https://www.getambassador.io/docs/emissary/latest/topics/running/services/rate-limit-service/
       - title: Properly populate alt_state_name for Tracing, Auth and RateLimit Services
         type: bugfix
         body: >-
-          Previously, setting the <code>stats_name</code> for the <code>TracingService</code>, <code>RateLimitService</code> 
+          Previously, setting the <code>stats_name</code> for the <code>TracingService</code>, <code>RateLimitService</code>
           or the <code>AuthService</code> would have no affect because it was not being properly passed to the Envoy cluster
           config. This has been fixed and the <code>alt_stats_name</code> field in the cluster config is now set correctly.
           (Thanks to <a href="https://github.com/psalaberria002">Paul</a>!)
       - title: Add support for config change batch window before reconfiguring Envoy
         type: feature
         body: >-
-          The <code>AMBASSADOR_RECONFIG_MAX_DELAY</code> env var can be optionally set to batch changes for the specified 
+          The <code>AMBASSADOR_RECONFIG_MAX_DELAY</code> env var can be optionally set to batch changes for the specified
           non-negative window period in seconds before doing an Envoy reconfiguration. Default is "1" if not set.
 
   - version: 3.1.1
@@ -268,8 +280,25 @@ items:
           connections using QUIC and the UDP network protocol. It currently only supports for connections
           between downstream clients and $productName$.
 
-  - version: 2.4.0
+  - version: 2.5.0
     date: 'TBD'
+    prevVersion: 2.4.0
+    notes:
+      - title: Fixed <code>mappingSelector</code> associating <code>Hosts</code> with <code>Mappings</code>
+        type: change
+        body: >-
+          Changes to label matching will change how <code>Hosts</code> are associated with <code>Mappings</code>. There was a bug with label
+          selectors that was causing <code>Hosts</code> to be incorrectly being associated with more <code>Mappings</code> than intended.
+          If any single label from the selector was matched then the <code>Host</code> would be associated with the <code>Mapping</code>.
+          Now it has been updated to correctly only associate a <code>Host</code> with a <code>Mapping</code> if <b>all</b> labels required by
+          the selector are present. This brings the <code>mappingSelector</code> field in-line with how label selectors are used
+          in Kubernetes. To avoid unexpected behaviour after the upgrade, add all labels that Hosts have in their
+          <code>mappingSelector</code> to <code>Mappings</code> you want to associate with the <code>Host</code>. You can opt-out of the new behaviour
+          by setting the environment variable <code>DISABLE_STRICT_LABEL_SELECTORS</code> to <code>"true"</code> (default: <code>"false"</code>).
+          (Thanks to <a href="https://github.com/f-herceg">Filip Herceg</a> and <a href="https://github.com/dynajoe">Joe Andaverde</a>!).
+
+  - version: 2.4.0
+    date: '2022-09-19'
     prevVersion: 2.3.2
     notes:
       - title: Add support for Host resources using secrets from different namespaces
@@ -289,7 +318,7 @@ items:
       - title: Properly populate alt_state_name for Tracing, Auth and RateLimit Services
         type: bugfix
         body: >-
-          Previously, setting the <code>stats_name</code> for the <code>TracingService</code>, <code>RateLimitService</code> 
+          Previously, setting the <code>stats_name</code> for the <code>TracingService</code>, <code>RateLimitService</code>
           or the <code>AuthService</code> would have no affect because it was not being properly passed to the Envoy cluster
           config. This has been fixed and the <code>alt_stats_name</code> field in the cluster config is now set correctly.
           (Thanks to <a href="https://github.com/psalaberria002">Paul</a>!)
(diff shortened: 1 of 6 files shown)


python/tests/kat/t_hosts.py
```diff
@@ -716,7 +716,7 @@ def manifests(self) -> str:
     authority: none
   mappingSelector:
     matchLabels:
-      hostname: tls-context-host-1
+      host-one: tls-context-host-1
```
AliceProxy, line 719: I just renamed this from `hostname` to `host-one` since the third `Host` in this test uses the `hostname` label for it's `matchLabels` selector, and with the fixed logic, the shared `Mapping` between those `Hosts` can't have the same label twice which is now a strict requirement. The renaming just avoids collision on the `hostname` label.

LukeShu:
> Plan moving forward with this is to first merge #4499 and #4318 once this PR is approved, then merge this PR on top of them immediately afterwards.

Low key: why? Why not just merge this one and close the other ones commenting that they've been cherry-picked in to this PR? Or have the first commit in this PR be `git merge pull/4499/head pull/4318/head` so that it's their same actual commits are in this PR (and GitHub would even see that merging this PR implicitly merges the other two)?

python/ambassador/ir/irutils.py
```diff
@@ -181,14 +184,21 @@ def selector_matches(
...
-        logger.debug("    selector miss on %s=%s", k, v)
+            if disable_strict_selectors:
+                return True
+        elif not disable_strict_selectors:
+            logger.debug("    selector miss for %s=%s => False", k, v)
+            return False
 
-    logger.debug("    all selectors miss => False")
-    return False
+    logger.debug(f"    all selectors match => True")
```
LukeShu, line 203: This probably shouldn't be an f-string?

docs/releaseNotes.yml
```diff
@@ -282,10 +294,23 @@ items:
...
+          selectors that was causing <code>Hosts</code> to be incorrectly being associated with more <code>Mappings</code> than intended.
+          If any single label from the selector was matched then the <code>Host</code> would be associated with the <code>Mapping</code>.
+          Now it has been updated to correctly only associate a <code>Host</code> with a <code>Mapping</code> if **all** labels required by
+          the selector are present. This brings the <code>mappingSelector</code> field in-line with how label selectors are used
+          in Kubernetes. To avoid unexpected behaviour after the upgrade, add all labels that Hosts have in their
+          <code>mappingSelector</code> to <code>Mappings</code> you want to associate with the <code>Host</code>. You can opt-out of the new behaviour
+          by setting the environment variable <code>DISABLE_STRICT_LABEL_SELECTORS</code> to <code>"true"</code> (default: <code>"false"</code>).
+          (Thanks to <a href="https://github.com/f-herceg">Filip Herceg</a> and <a href="https://github.com/dynajoe">Joe Andaverde</a>!).
+
```
LukeShu, line 309: This PR updates the changelog for both 3.2.0 and 2.4.0; this PR should probably only be touching 3.2.0?

docs/releaseNotes.yml
```diff
@@ -43,6 +43,18 @@ items:
           release of 1.23. This provides $productName$ with the latest security patches, performances enhancments,
           and features offered by the envoy proxy.
         docs: https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.23/v1.23.0
+      - title: Fixed <code>mappingSelector</code> associating <code>Hosts</code> with <code>Mappings</code>
+        type: change
```
LukeShu, line 47: I'm OK calling this a `bugfix`, not a `change`. (But at the same time, doing it in a .z is questionable)

docs/releaseNotes.yml
```diff
@@ -43,6 +43,18 @@ items:
...
+        body: >-
+          Changes to label matching will change how <code>Hosts</code> are associated with <code>Mappings</code>. There was a bug with label
+          selectors that was causing <code>Hosts</code> to be incorrectly being associated with more <code>Mappings</code> than intended.
+          If any single label from the selector was matched then the <code>Host</code> would be associated with the <code>Mapping</code>.
+          Now it has been updated to correctly only associate a <code>Host</code> with a <code>Mapping</code> if **all** labels required by
```
LukeShu, line 52: Gotta use HTML, not `**` markdown.

python/ambassador/ir/irutils.py
```diff
@@ -181,14 +184,21 @@ def selector_matches(
         logger.debug("    no incoming labels => False")
         return False
 
-    selmatch = False
+    # Ambassador (2.0-2.3) & (3.0-3.1) consider a match on a single label as a "good enough" match.
+    # In versions 2.4+ and 3.2+ _ALL_ labels in a selector must be present for it to be considered a match.
+    # DISABLE_STRICT_LABEL_SELECTORS provides a way to restore the old unintended loose matching behaviour
+    # in the event that it is desired. The ability to disable strict label matching will be removed in a future version.
+    disable_strict_selectors = parse_bool(os.environ.get("DISABLE_STRICT_LABEL_SELECTORS", "false"))
```
LukeShu, line 191: Part of me thinks the "correct" fix is to move this code to Go so we can just call the standard Kubernetes library function for checking label selectors.

python/ambassador/ir/irutils.py
```diff
@@ -181,14 +184,21 @@ def selector_matches(
...
         if labels.get(k) == v:
             logger.debug("    selector match for %s=%s => True", k, v)
-            return True
-
-        logger.debug("    selector miss on %s=%s", k, v)
+            if disable_strict_selectors:
+                return True
+        elif not disable_strict_selectors:
+            logger.debug("    selector miss for %s=%s => False", k, v)
+            return False
```
LukeShu, line 201: Maybe it's just me, but I had to twist my brain around to understand this.  Perhaps don't try so hard to share the code?

```python
if disable_strict_selectors:
    old code goes here
else:
    new code goes here
```

python/ambassador/ir/irhost.py
```diff
@@ -473,7 +478,41 @@ def matches_httpgroup(self, group: "IRHTTPMappingGroup") -> bool:
...
+        groupName = group.get("name") or "None"
+
+        # The synthetic Mappings for diagnostics, readiness, and liveness probes always match all Hosts.
+        # They can all still be disabled if desired via the Ambassador Module resource
+        if groupName in [
+            "GROUP: internal_readiness_probe_mapping",
+            "GROUP: internal_liveness_probe_mapping",
+            "GROUP: internal_diagnostics_probe_mapping",
+        ]:
+            return True
```
LukeShu, line 490: Relying on the internal group name feels slimy to me. But, uh, I guess it's the best we have.

LukeShu (approved):
Some comments and suggestions, but on the whole: looks fine to me.

docs/releaseNotes.yml
AliceProxy: Yeah while it is a bugfix, I don't want someone to be skimming the release notes and go "oh bugfix, that's probably just an improvement" then not look into it too much when the change in behaviour may actually be really relevant to them.

docs/releaseNotes.yml
AliceProxy: There is a sister PR for `2.4` which I guess needs to be re-targeted for `2.5` now. This was just my attempt at avoiding any merge-conflicts when the `2.4` PR was merged and triggered repatriation. 

That and if someone is looking at the repo, its probably nice to see that this is going into 2.x as well without having to change the branch from `master`

python/ambassador/ir/irhost.py
AliceProxy: I agree that it does feel weird, but I couldn't think of a way to use a special field that seemed better. Are you suggesting using `ambassador_id` itself, or adding a new field?

If the group names were ever to change, the tests will fail so that at least makes me feel a little better about using them.

AliceProxy:
> > Plan moving forward with this is to first merge #4499 and #4318 once this PR is approved, then merge this PR on top of them immediately afterwards.
> 
> Low key: why? Why not just merge this one and close the other ones commenting that they've been cherry-picked in to this PR? Or have the first commit in this PR be `git merge pull/4499/head pull/4318/head` so that it's their same actual commits are in this PR (and GitHub would even see that merging this PR implicitly merges the other two)?

Bet, I did just that

python/ambassador/ir/irhost.py
```diff
@@ -473,7 +478,41 @@ def matches_httpgroup(self, group: "IRHTTPMappingGroup") -> bool:
...
+
+        # Ambassador (2.0-2.3) & (3.0-3.1) consider a match on a single label as a "good enough" match.
+        # In versions 2.5+ and 3.2+ _ALL_ labels in a selector must be present for it to be considered a match.
+        # DISABLE_STRICT_LABEL_SELECTORS provides a way to restore the old unintended loose matching behaviour
+        # in the event that it is desired. The ability to disable strict label matching will be removed in a future version.
+        disable_strict_selectors = parse_bool(
```
LanceEa, line 496: Looks like we duplicate this twice, maybe we just move it into a util function so the comment/default are maintained in a single place?

LanceEa (commented):
Just one suggestion to remove duplication. Other than that it looks inline with what we discussed and the behavior we want.

docs/releaseNotes.yml
LukeShu: Fair enough.

python/ambassador/ir/irhost.py
LukeShu: I was suggesting adding a new field, or otherwise using a special sentinel value in an existing field.

Good point about the test.

LukeShu (approved):
Thanks!

LanceEa (approved):
lgtm..thanks!

crocinsocks:
Sorry to be the bearer of bad news but it would seem this issue is not fixed. I've been playing around with migrating from defined hostnames on Mappings to a more dynamic solution using labels with Hosts.

I have the following simplistic setup for testing:

```
kind:  Mapping
metadata:
  name: web1
  labels:
    emissary/mapping: web1 ( web2 / web3 )
spec:
  ambassador_id: ["web"]
  bypass_auth: true
  prefix: /
  service: http://web1:8080 ( http://web2:8080 / http://web3:8080 )
```
> no hostname defined (as wanted) this gets translated to v2 apply hostname: _skip_mapping_with_empty_host_ 

```
kind: Host
metadata:
  name: web1
spec:
  ambassador_id: ["web"]
  mappingSelector:
    matchLabels:
      emissary/mapping: web1 ( web2 / web3 )
  hostname: web1.example.com  ( web2.example.com / web2.example.com )
  acmeProvider:
    authority: none
  tlsSecret:
    name: web1  ( web2 / web3 )
```
> The host is in a different namespace to mapping however placing them in the same namespace makes no difference.

When the following is configured there is a random selection of which mapping is chosen. I've tried completely different labels (e.g. so no repetition of any kind), in case of some matching issues in the code however the result is always the same. 

Certificates always correspond to each host correctly on load limiting the possibility of the problem being solely between the host and mapping. 

I believe this has something to do with all hostnames becoming a catch all and then emissary randomly selecting the routes ignoring the labels entirely.

More info:
- sometimes a host(s) will not route at all (e.g. no round robin) simply acts dead - can not find a specific reason why and usually only one host round robins but sometimes 2/3 will. 
- if I delete the other two mappings only the single host will route (the others return 404) and there is no round robin effect so it would seem the hosts do respect the labels. This makes me think the Host is only activated when there is a matched mapping with a label (as expected) but afterwards the logic to route to the correct mapping based on labels is then ignored leaving the round robin effect. 

Was testing on 3.3.0 upgraded to 3.5.0 hoping it would solve the problem but to no avail. Clean cluster on every test.  

The goal here is to remove the use of a hostname on Mapping which I believe is also the purpose mappingSelector though I have seen many examples where both a hostname and labels are used in Mapping. However, I believe the documentation is clear that only one or the other is required and I am therefore using the labels correctly and this is bug still within the code. 

Hope this helps.

ralf-cestusio:
@crocinsocks did you ever hear anything back about your problem. 

I can reproduce the exact same problem you have and it really defeats the purpose of having a label selector

PR #4510, opened 2022-09-14
Service name fix, changed to use envoy name instead of the service name

aremozei:
## Description
Envoy stats are being sent and build with the incorrect name inside the snapshot, this one being formatted as `<name>`, instead of using the envoy name `<cluster>_<name>_<namespace>`. 

This change sends it with the correct envoy name, so that we are able to send the hmetric and other health stats that were not set because we were sending the envoy name to match only to cluster_stats inside envoy_stats.py when it was called from diagnostics.py. 

## Related Issues
List related issues.

## Testing
Manual, built a kubeception cluster, then make a couple of requests and trigger a 5xx error to see the health percent change.  

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/python/ambassador/ir/ircluster.py
+++ b/python/ambassador/ir/ircluster.py
@@ -312,11 +312,10 @@ class IRCluster(IRResource):
         # If we have a stats_name, use it. If not, default it to the service to make life
         # easier for people trying to find stats later -- but translate unusual characters
         # to underscores, just in case.
-
         if stats_name:
             new_args["stats_name"] = stats_name
         else:
-            new_args["stats_name"] = re.sub(r"[^0-9A-Za-z_]", "_", service)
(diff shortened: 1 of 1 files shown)


LanceEa (changes requested):
@aremozei - I wanted to capture our discussion offline here.

As mentioned, I don't think changing the behavior of the cluster is the right call here.

Since we are trying to fix how diagnostics are collected we should fix it in:
https://github.com/emissary-ingress/emissary/blob/785b4dee950355096f0861c49b3810b8b853d7e2/python/ambassador/diagnostics/envoy_stats.py#L95

My guess is that we just need to match the cluster name parsing logic used when building out the cluster stats data structure found here:

https://github.com/emissary-ingress/emissary/blob/785b4dee950355096f0861c49b3810b8b853d7e2/python/ambassador/diagnostics/envoy_stats.py#L323

That way we are not introducing behavior change for users that are using their own Observability tools.

PR #4513, opened 2022-09-15
Resolve mismatched clusters name on envoy_stats without overwriting

alexeikmun:
## Description
Envoy stats are being sent and built with the incorrect name inside the snapshot due to cluster name mismatches, this one being formatted as `<name>`, instead of using the envoy name `<cluster>_<name>_<namespace>`.

This change seeks to make both cluster names similar temporarily in order to check if the given cluster really exists in `envoy_stats`

#### Befere

The old logic creates an aux variable `self.cluster_names` which the only goal is to store the `envoy_name`s mapped from `self.diag.clusters`; then the `self.cstats` variable does a second map over `self.cluster_names` passing `envoy_name` to `self.estat.cluster_stats` and returns a Dict with key-cluster-name formatted as ` <cluster>_<name>_<namespace>`:

```python
self.cluster_names = [cluster.envoy_name for cluster in self.diag.clusters.values()]
self.cstats = {name: self.estat.cluster_stats(name) for name in self.cluster_names}
```

In the following pic, the routes are in orange color, meaning that the cluster names didn't match

Screen Shot 2022-09-16 at 3 10 33 PM

So in consequence the `self.cstats` returns undefined cluster stats

Screen Shot 2022-09-16 at 3 12 28 PM

#### After

I removed the `self.cluster_names` aux variable in favor of directly using `cluster.stats_name` from `self.diag.clusters` as parameter for `self.estat.cluster_stats` because this one has the right format <cluster_name> accepted by the method

```python
self.cstats = {cluster.name: self.estat.cluster_stats(cluster.stats_name) for cluster in self.diag.clusters.values()}
```
In this case, the cluster names are matching (_I've created pod called weby to test traffic shown in green color_)

Screen Shot 2022-09-16 at 2 40 57 PM

So this output is way different from the original, due to cluster names matching avoiding the condition 

https://github.com/emissary-ingress/emissary/blob/77a348c34617acf59b934aa71b74d2f510967b41/python/ambassador/diagnostics/envoy_stats.py#L110
Screen Shot 2022-09-16 at 2 40 10 PM

#### UPDATE:
I've changed `cluster.envoy_name` in favor of `cluster.name` due to issues when `envoy_name` surpasses the 60 characters gets truncated causing undefined stats

Screen Shot 2022-09-19 at 1 07 54 PM

## Related Issues
List related issues.

## Testing
Manual, built a kubeception cluster, then make a couple of requests and trigger a 5xx error to see the health percent change.

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/python/ambassador/diagnostics/diagnostics.py
+++ b/python/ambassador/diagnostics/diagnostics.py
@@ -558,7 +558,7 @@ class Diagnostics:
                     "type": type_name,
                     "_source": svc.location,
                     "name": url,
-                    "cluster": cluster.envoy_name,
+                    "cluster": cluster.name,
                     "_service_weight": svc_weight,
                 }
             )
--- a/python/ambassador/diagnostics/diagnostics.py
+++ b/python/ambassador/diagnostics/diagnostics.py
@@ -119,7 +119,7 @@ class DiagResult:
         # Go ahead and grab Envoy cluster stats for all possible clusters.
         # XXX This might be a bit silly.
         self.cstats = {
-            cluster.envoy_name: self.estat.cluster_stats(cluster.stats_name)
+            cluster.name: self.estat.cluster_stats(cluster.stats_name)
             for cluster in self.diag.clusters.values()
         }
 
--- a/python/ambassador/diagnostics/diagnostics.py
+++ b/python/ambassador/diagnostics/diagnostics.py
@@ -118,8 +118,10 @@ class DiagResult:
 
         # Go ahead and grab Envoy cluster stats for all possible clusters.
         # XXX This might be a bit silly.
-        self.cluster_names = [cluster.envoy_name for cluster in self.diag.clusters.values()]
-        self.cstats = {name: self.estat.cluster_stats(name) for name in self.cluster_names}
+        self.cstats = {
+            cluster.envoy_name: self.estat.cluster_stats(cluster.stats_name)
+            for cluster in self.diag.clusters.values()
+        }
 
         # Save the request host and scheme. We'll need them later.
         self.request_host = request.headers.get("Host", "*")


haq204:
change wise this looks fine with me pending CI

python/ambassador/diagnostics/envoy_stats.py
```diff
@@ -107,7 +107,15 @@ def cluster_stats(self, name: str) -> Dict[str, Union[str, bool]]:
         when = self.last_update
         cstat = self.clusters
 
-        if name not in cstat:
+        cluster_exists = False
+        cname = None
+        for cluster_name in cstat.keys():
```
LanceEa, line 112: @alexeikmun - Like I mentioned offline, I see this works. However, I'm hesitant to go forward with this approach mostly due to the question of `Does it scale for large clusters?` 

The `clusters_stats` function gets called for every cluster which means then this loop logic gets called for every cluster. So would this scale on a cluster with thousands of clusters and not hang? 

Also, it feels like we are losing the benefits of `cstats` being a dictionary. The purpose of that data structure is to allow us to do quick key lookups vs linear searches.

python/ambassador/diagnostics/diagnostics.py
```diff
@@ -118,8 +118,7 @@ def __init__(self, diag: "Diagnostics", estat: EnvoyStats, request) -> None:
 
         # Go ahead and grab Envoy cluster stats for all possible clusters.
         # XXX This might be a bit silly.
-        self.cluster_names = [cluster.envoy_name for cluster in self.diag.clusters.values()]
```
LanceEa, line 121: Since we are assigning `self.cluster_names` did you verify that this wasn't being referenced anywhere else? I just want to make sure nothing was referencing it since it is mutating the `DiagResults`

python/ambassador/diagnostics/diagnostics.py
```diff
@@ -118,8 +118,7 @@ def __init__(self, diag: "Diagnostics", estat: EnvoyStats, request) -> None:
...
-        self.cstats = {name: self.estat.cluster_stats(name) for name in self.cluster_names}
+        self.cstats = {cluster.envoy_name: self.estat.cluster_stats(cluster.stats_name) for cluster in self.diag.clusters.values()}
```
LanceEa, line 121: Can you add to the PR description a sample of the before and after for `cstats` here? I think I'm getting tripped up due to the fact that the cstats name is being reused in DiagResults and EnvoyStats so having the diff will help it make sense.

Just for information:

`envoy_name` is the cluster name but potentially mangled. The reason for this is that Envoy only allows 60 characters for the name of a cluster. So, we have logic that where there are really long names we will truncate them and handle name collisions. This mangled name is what is stored here and ultimately passed on to Envoy configuration.

Not sure if that matters but just wanted to call that out.

python/ambassador/diagnostics/diagnostics.py
alexeikmun: I've double-checked the `cluster_names`, and it seems that was created just like an aux variable, there's not any implementation elsewhere.

LukeShu (changes requested):
Your explanation is clear and helpful, and the code change itself looks good. But it needs a changelog entry and tests.

alexeikmun:
@LukeShu As we discussed on Friday with Lance, Diagnostics UI doesn't have any tests yet, I agree to create tests but not yet because that implies a lot of time.  This time the manual test should be enough until we create unit tests. Regarding changelog, I'm not sure in which version this little change would go.

LanceEa:
@alexeikmun - This will only go in 3.2 because the diagnostics feature wasn't introduced into the Agent until version 3.2 so clould can only support it as a going forward feature.

LukeShu (dismissed):
> As we discussed last Friday with Lance, Diagnostics UI doesn't have any tests yet, I agree to create tests but not yet because that implies a lot of time.

OK.

> This time the manual test should be enough until we create unit tests. Regarding changelog, I'm not sure in which version this little change would go.

Just stick it in the top-most section of `releaseNotes` and if it doesn't land for that release we'll handle it at merge.

LanceEa:
@alexeikmun - it looks like one test is failing. Normally, I would say it is one of the flaky ones but looking at this:

https://github.com/emissary-ingress/emissary/actions/runs/3084534857/jobs/5004150996#step:5:3727

It is getting a 500 when trying to call the diagd endpoint so it might be related. Let me know if you need help debugging. You can find the logs at the bottom of the summary page for download:

https://github.com/emissary-ingress/emissary/actions/runs/3084534857

LanceEa:
Looks like the last fix did it, if we could just run it and verify that does it then we can get this merged.

haq204 (approved):

LanceEa:
I verified it is working as well.

UI:

Screen Shot 2022-09-22 at 9 57 48 PM

Here is the cluster with the long service name causing the cluster name to be mangled but retaining the long name for the `alt_stat_name`
```json
{
        "alt_stat_name": "quote_with_reallly_loooooooooooooooooooooooong_naaaaaaamme_default",
        "connect_timeout": "3.000s",
        "dns_lookup_family": "V4_ONLY",
        "lb_policy": "ROUND_ROBIN",
        "load_assignment": {
          "cluster_name": "cluster_quote_with_reallly_loooooooooooo-1F2DC8766E451E56-0",
          "endpoints": [
            {
              "lb_endpoints": [
                {
                  "endpoint": {
                    "address": {
                      "socket_address": {
                        "address": "quote-with-reallly-loooooooooooooooooooooooong-naaaaaaamme.default",
                        "port_value": 80,
                        "protocol": "TCP"
                      }
                    }
                  }
                }
              ]
            }
          ]
        },
        "name": "cluster_quote_with_reallly_loooooooooooo-1F2DC8766E451E56-0",
        "type": "STRICT_DNS"
      }
```

One interesting thing to note is that they all show up grey until at least one call via the LB endpoint goes through. So, probably in most cases the `ambassador/v0/check_ready` and `ambassador/v0/check_alive` will be grey. 

A side note, we should investigate if those are routes are being actively used or just hanging around from the old python entrypoint.

LanceEa (approved):

Issue #4528, opened 2022-09-19
diagnostics: issue parsing vhosts stats


LanceEa:
**Describe the bug**
When using the diagnostics UI for emissary-ingress it calls the stats endpoint which is responsible for parsing out the Envoy provided metrics and matching them to the correct cluster for showing in the Diagnostics UI. Many of the areas such as cluster name or alt_stat_name are cleaned up and special characters are removed which allows us to do a simple string split on the `:` when parsing the key and value of the metric. 

example metric:
`cluster.quote_default.upstream_rq_2xx: 1`

However, the `name` field within the `VirtualHosts` of the http manager is not cleansed. This leads to us tryin

`vhost.ambassador-http-listener-example.com:443.vcluster.other.upstream_rq_retry: 0`

The current parsing logic will throw an error when trying to parse this line due to unpacking too many values. The `name` field in the `config.route.v3.VirtualHost` envoy configuration is what determines the stat name for these metrics. 

https://www.envoyproxy.io/docs/envoy/v1.23.1/api-v3/config/route/v3/route_components.proto#config-route-v3-virtualhost

**To Reproduce**
1. Setup a cluster
2. Create a Host which includes the Port number so `hostname: example.com:443`
3. Execute into the container or if you expose the diagnostics endpoints then call `http://localhost:8877/ambassador/v0/diag/?json=true` which is capable of setting the stats

This will cause an error and you can check the logs of the container to see the issue.

```log
022-06-21 11:41:59 diagd 2.3.1 [P20TThread-3] ERROR: could not update Envoy stats: too many values to unpack (expected 2)
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/ambassador-0.0.0.dev0-py3.9.egg/ambassador/diagnostics/envoy_stats.py", line 486, in update
    self.update_envoy_stats(last_attempt)
  File "/usr/lib/python3.9/site-packages/ambassador-0.0.0.dev0-py3.9.egg/ambassador/diagnostics/envoy_stats.py", line 328, in update_envoy_stats
    key, value = line.split(":")
ValueError: too many values to unpack (expected 2)
```

You can also take a look at envoy directly by looking at `http://localhost:8001/stats` to confirm there are non-parsable metrics

**Expected behavior**

We expect the parsing to not fail and to be able to correctly see the stats in the diagnostic ui.

Two approaches we can take:
1. Short-term - unblock the failure because these stats are currently ignored

[...]

PR #4529, opened 2022-09-19
diagnostics: properly parse metrics with colons in name

LanceEa:
## Description
This PR partially addresses #4528 by making sure that the parsing logic doesn't throw errors when parsing metric names with colons in them. A more robust solution will need to introduce behavior change and will take more time so this is to ensure users that are effected by this are unblocked.

See the issue for more details but TL;DR;

Parse from right-to-left and only look for a single colon. That way the metric name with the colon is still considered part of the key and will not throw an error.

## Related Issues
#4528 

## Testing
CI is green, pushed to cluster and tested to make sure that it wasn't throwing any errors.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,6 +107,11 @@ it will be removed; but as it won't be user-visible this isn't considered a brea
   the specified  non-negative window period in seconds before doing an Envoy reconfiguration.
   Default is "1" if not set.
 
+- Bugfix: If a `Host` or `TLSContext` contained a hostname with a `:` then when using the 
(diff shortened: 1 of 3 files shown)


haq204:
change wise this looks fine to me though I would update the label from `v2.4 -> v2.5`. 
For the long term change, I would say it should go only in 3.y and not be backported.

ddymko (approved):
looks good. Thanks for putting a comment about a follow up!

LanceEa:
@haq204 - This is green now and I have updated labels.

haq204 (approved):

PR #4532, opened 2022-09-20
[v1.14] releaseNotes.yml, CHANGELOG.tpl: Let an 'unplanned' date hide a release

LukeShu:
## Description

This way hypothetical "just in case" release notes for releases we don't actually intend to do don't show up in CHANGELOG.md.

I've also backported the dropping of `gotestsum`, because otherwise CI is broken on `release/v1.14`; see the commit message for more details.

Slack DMs from 2022-08-26:

> @kflynn:
>
> > So… is Emissary 3.0.1 a thing? `master` mentions it in
> > `docs/releaseNotes.yml` but it has a “TBD” date, it’s out of
> > order, it duplicates stuff previously listed for 3.1.0, and
> > there’s no `v3.0.1` tag.  :thinking_face:
>
> @AliceProxy:
>
> > It must be a typo for 3.1.0
>
> …
>
> @LukeShu:
>
> > 3.0.1 is probably not a thing, but isn't a typo either.  I'd added
> > a CI check that looks at the 'major.minor' of the first changelog
> > entry to make sure that you're targeting the correct branch.  And
> > so I needed to add a 3.0.1 section for the `release/v3.0` leg of
> > the v2.3→2.4→3.0→3.1→3.2 repatriation chain.
>
> @kflynn:
>
> > Ew. The 3.0.0 section wasn’t enough?
>
> @LukeShu:
>
> > Not if you wanna have them be in chronological order, since
> > various 2.3 patch releases happened after 3.0
>
> @kflynn:
>
> > Is there a gomplate special-case in there so that the CHANGELOG
> > doesn’t show an empty 3.0.1 release?
>
> @LukeShu:
>
> > Updating the `CHANGELOG.tpl` to skip a release if `len(.notes) == 0`
> > is a good idea
> >
> > Oh, but 3.0.1 isn't empty, since if we did one, it'd include
> > security patches that both 2.3 and 3.1 already have.
>
> @kflynn:
>
> > It’s empty until you actually do one, though.
>
> @LukeShu:
>
> > But its section in `releaseNotes.yml` isn't empty
> >
> > said patches are already on `release/v3.0`
>
> @kflynn:
>
> > Let me phrase that differently.  :slightly_smiling_face:
> >
> > It’s silly to have the CHANGELOG on `master` talk about a release
> > that does not exist.
>
> @LukeShu:
>
> > so maybe we should add a "don't include this release" flag in
> > `releaseNotes.yml`
>
> @kflynn:
>
> > That might be simplest, yeah.
>
> @LukeShu:
>
> > It already uses `date: "N/A"` to have it emit "not issued". So how
> > about `date: unplanned` to skip it entirely?
>
> @kflynn:
>
> > :+1:
>
> @LukeShu:
>
> > Will do in the near future. Probably not today.
>
> @kflynn:
>
> > Works for me.

## Related Issues
(none)

## Testing
(none)

## Checklist

 - [x] I made sure to update `CHANGELOG.md`. - that's all this is

    + Any new features
    + Any deprecations

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -204,7 +204,6 @@ commands:
     steps:
     # setup
     - amb-linux-install
-    - install-gotestsum
     - amb-checkout
     - amb-config-registry
     - run:
@@ -504,12 +503,6 @@ commands:
             echo "Chart version ${thisversion} doesn't match tag << pipeline.git.tag >>; aborting"
             exit 1
           fi
-  install-gotestsum:
-    steps:
-    - run:
-        name: "Install gotestsum"
-        command: |
-          go get gotest.tools/gotestsum
   install-redis:
     steps:
     - run:
--- a/.circleci/config.yml.d/amb_jobs.yml
+++ b/.circleci/config.yml.d/amb_jobs.yml
@@ -222,7 +222,6 @@ commands:
     steps:
       # setup
       - amb-linux-install
-      - install-gotestsum
       - amb-checkout
       - amb-config-registry
       - run:
--- a/.circleci/config.yml.d/generic_util.yml
+++ b/.circleci/config.yml.d/generic_util.yml
@@ -5,12 +5,6 @@
 version: 2.1
 
 commands:
-  "install-gotestsum":
-    steps:
-    - run:
-        name: "Install gotestsum"
-        command: |
-          go get gotest.tools/gotestsum
   "install-redis":
     steps:
     - run:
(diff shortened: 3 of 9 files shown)


kflynn (approved):
Hit it!

(Well. Once it's green...)

kflynn (approved):
Hit it! I'm glad to see `gotestsum` go.

AliceProxy (approved):
Nice, I had just noticed some people commenting on my PR for the gzip bugfix expecting a 1.14.5 release because of this note specifically.

PR #4543, opened 2022-09-22
Only watch tls secrets

raffis:
## Description
Emissary has quite some performance problems on our cluster. It's using 1 full core  and over 7G memory without any requests. Basically on standby.
Screenshot from 2022-09-22 09-48-43

Profiling:
Screenshot from 2022-09-22 09-39-11

The problem is that ambassador basically ends up reconciling non stop because it watches quite a big resource list including secrets, ingresses, ep and so on.
On a big cluster this can be quite cpu/mem heavy.
In our case we have ~10.000 secrets on the cluster and emissary globally (not single namespace) will read all  10.000 secrets.
On top of that some secrets change quite often which triggers basically an infinite reconcile loop.

The solution for this is:
* Remove ingress access for emissary from the clusterrole (we only use ambassador mappings)
* This pr, as ambassador should only read tls secrets. I don't see a reason why ambassador should read other secrets. (This reduces the number of secrets from 10.000 to like 30 in our case).

After those changes emissary uses like 120m/400Mi on average.

If non tls secrets are used for any reason there should be a possibility to declare a resource filter instead.

**Note**: I also updated the dev guide as the things there do not work as stated.

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/cmd/entrypoint/interesting_types.go
+++ b/cmd/entrypoint/interesting_types.go
@@ -3,6 +3,7 @@ package entrypoint
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/datawire/dlib/dlog"
 	"github.com/emissary-ingress/emissary/v3/pkg/kates"
@@ -41,6 +42,15 @@ func GetQueries(ctx context.Context, interestingTypes map[string]thingToWatch) [
 			FieldSelector: queryinfo.fieldselector,
 			LabelSelector: ls,
 		}
+
+		if selector := getWatchFieldSelector(queryinfo.typename); selector != "" {
+			query.FieldSelector = selector
+		}
+
+		if selector := getWatchLabelSelector(queryinfo.typename); selector != "" {
+			query.LabelSelector = selector
+		}
+
 		if query.FieldSelector == "" {
 			query.FieldSelector = fs
 		}
@@ -50,6 +60,42 @@ func GetQueries(ctx context.Context, interestingTypes map[string]thingToWatch) [
 	}
 
 	return queries
+}
+
+func getWatchFieldSelector(resourcegroup string) string {
+	fs := GetAmbassadorWatcherFieldSelector()
+	selectors := strings.Split(fs, ";")
+
+	for _, selector := range selectors {
+		byResourcegroup := strings.Split(selector, ":")
+		if len(byResourcegroup) > 1 {
+			if byResourcegroup[0] == resourcegroup {
+				return byResourcegroup[1]
+			}
+		} else {
+			return byResourcegroup[0]
+		}
+	}
+
+	return ""
+}
+
+func getWatchLabelSelector(resourcegroup string) string {
+	fs := GetAmbassadorWatcherLabelSelector()
+	selectors := strings.Split(fs, ";")
+
+	for _, selector := range selectors {
+		byResourcegroup := strings.Split(selector, ":")
+		if len(byResourcegroup) > 1 {
+			if byResourcegroup[0] == resourcegroup {
+				return byResourcegroup[1]
+			}
+		} else {
+			return byResourcegroup[0]
+		}
+	}
+
+	return ""
 }
 
 // GetInterestingTypes takes a list of available server types, and returns the types we think
--- a/cmd/entrypoint/interesting_types.go
+++ b/cmd/entrypoint/interesting_types.go
@@ -33,6 +33,9 @@ func GetQueries(ctx context.Context, interestingTypes map[string]thingToWatch) [
 	fs := GetAmbassadorFieldSelector()
 	ls := GetAmbassadorLabelSelector()
 
+	globalFieldSelector, groupFieldSelectors := selectorParser(GetAmbassadorWatcherFieldSelector())
+	globalLabelSelector, groupLabelSelectors := selectorParser(GetAmbassadorWatcherLabelSelector())
+
 	var queries []kates.Query
 	for snapshotname, queryinfo := range interestingTypes {
 		query := kates.Query{
@@ -43,13 +46,8 @@ func GetQueries(ctx context.Context, interestingTypes map[string]thingToWatch) [
 			LabelSelector: ls,
 		}
 
-		if selector := getWatchFieldSelector(queryinfo.typename); selector != "" {
-			query.FieldSelector = selector
-		}
-
-		if selector := getWatchLabelSelector(queryinfo.typename); selector != "" {
-			query.LabelSelector = selector
-		}
+		query.FieldSelector = electSelector(globalFieldSelector, groupFieldSelectors, queryinfo.typename)
+		query.LabelSelector = electSelector(globalLabelSelector, groupLabelSelectors, queryinfo.typename)
 
 		if query.FieldSelector == "" {
 			query.FieldSelector = fs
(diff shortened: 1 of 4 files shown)


cmd/entrypoint/interesting_types.go
```diff
@@ -80,9 +80,9 @@ func GetInterestingTypes(ctx context.Context, serverTypeList []kates.APIResource
 		//
 		// Note that we pull `secrets.v1.` in to "K8sSecrets".  ReconcileSecrets will pull
 		// over the ones we need into "Secrets" and "Endpoints" respectively.
-		"Services":   {{typename: "services.v1."}},                             // New in Kubernetes 0.16.0 (2015-04-28) (v1beta{1..3} before that)
-		"Endpoints":  {{typename: "endpoints.v1.", fieldselector: endpointFs}}, // New in Kubernetes 0.16.0 (2015-04-28) (v1beta{1..3} before that)
-		"K8sSecrets": {{typename: "secrets.v1."}},                              // New in Kubernetes 0.16.0 (2015-04-28) (v1beta{1..3} before that)
+		"Services":   {{typename: "services.v1."}},                                         // New in Kubernetes 0.16.0 (2015-04-28) (v1beta{1..3} before that)
+		"Endpoints":  {{typename: "endpoints.v1.", fieldselector: endpointFs}},             // New in Kubernetes 0.16.0 (2015-04-28) (v1beta{1..3} before that)
+		"K8sSecrets": {{typename: "secrets.v1.", fieldselector: "type=kubernetes.io/tls"}}, // New in Kubernetes 0.16.0 (2015-04-28) (v1beta{1..3} before that)
```
haq204, line 85: Not all folks put tls secrets as `kubernetes.io/tls`. you could create a normal secret and use the `tls.crt` and `tls.key` fields. So this could end up being a breaking change.

haq204 (changes requested):
My main concern is that this could end up breaking some customers if they don't store tls secrets as `kubernetes.io/tls`

Also we merged https://github.com/emissary-ingress/emissary/pull/4488 which also helps with constant re-configurations and will be available in v3.2 and v2.4

AliceProxy:
> Thanks for the contribution!
> 
> My main concern is that this could end up breaking some users if they don't store tls secrets as `kubernetes.io/tls`
> 
> Also we merged #4488 which also helps with constant re-configurations and will be available in v3.2 and v2.4

Also note that support for certificate revocation lists loads the CRL configuration for the feature from from non-tls secrets. I believe we also do watch for opaque secrets for validation contexts.

cmd/entrypoint/interesting_types.go
raffis: fair enough, how about the the 2nd proposal then to add the ability to configure a filter? It doesnt make sense to load and watch 10.000 resources in the case only a handful are required.
Where would that config go? Module crd?

raffis: or actually from an env only as the module res would need to be loaded first otherwise (if thats not the case, didnt check the code yet)

raffis:
> Also we merged https://github.com/emissary-ingress/emissary/pull/4488 which also helps with constant re-configurations and will be available in v3.2 and v2.4


v3.2 does not fix it. It also consumes like 10G of memory just because it reads the entire list of secrets from the cluster.

raffis:
> As for the idea to configure a filter, I think that works. In terms of where to place it, Module is a pretty old resource so we probably don't want to add new things in there. So I would probably prefer as an env var, to configure which secrets to watch based on label selectors? But that shouldn't be any different than `AMBASSADOR_LABEL_SELECTOR` https://www.getambassador.io/docs/emissary/latest/topics/running/environment/#ambassador_label_selector. I'll admit that we should make this option more visible in the performance and scaling section of the docs but have you tried setting this option and labeling your secrets to see if that resolves your issue?


I was actually not aware that env exists, this might solve it but it sounds unpractical since this involves labeling everything from mappings, services, secrets and so on right? 


What would be nice is to optionally specify the resourcegroup and support multiple selectors. Something like:
```
AMBASSADOR_LABEL_SELECTOR=secrets.v1:mylabel=myvalue,services.v1:mylabel=test
```

In my case I could then label my two secrets and define this env like:
```
AMBASSADOR_LABEL_SELECTOR=secrets.v1:emissary-secret=yes
```

Otherwise I would need to change like 50 repos and label services and so on as well.
This change might even be backwards compatible, didn't think that through yet though.

haq204:
in order to specify label selectors per resource, let's create a new env var idk call it `AMBASSADOR_WATCHER_LABEL_SELECTOR` and we can mark the `AMBASSADOR_LABEL_SELECTOR` as deprecated to be remove in a future version as the formats will be different.

We could have the same env var support the two different formats but I figure that would make the parsing logic more complicated so it might be more clear to have them as two separate env vars.

How does that sound?

raffis:
New functionality implemented, docs here: https://github.com/datawire/ambassador-docs/pull/700
I also added the same for the field selector (which was not even documented before).

Two comments, 
* When I run tests with `make gotest` it takes ages and fail, seeing some timeouts like `testutil_fake_test_test.go:86: Get timed out!` in the output and a lot of other output. This looks unrelated to my change. The tests added should work. They failed first time I implemented it and this was visible before all the follow up errors.
* I see that the env AMBASSADOR_LABEL_SELECTOR is also used in python/*, honestly no clue what the python stuff is for, and ARCHITECTURE.md is empty where I hoped to get informed about this. Not sure if something needs to be changed there.

However I tested the new feature on our cluster and it works so far.

haq204:
> New functionality implemented, docs here: [datawire/ambassador-docs#700](https://github.com/datawire/ambassador-docs/pull/700) I also added the same for the field selector (which was not even documented before).
> 
> Two comments,
> 
> * When I run tests with `make gotest` it takes ages and fails eventually, seeing some timeouts like `testutil_fake_test_test.go:86: Get timed out!` in the output and a lots of other output. This looks unrelated to my change. The tests added should work. They failed first time I implemented it and this was visible before all the follow up errors.
> * I see that the env AMBASSADOR_LABEL_SELECTOR is also used in python/*, honestly no clue what the python stuff is for, and ARCHITECTURE.md is empty where I hoped to get informed about this. Not sure if something needs to be changed there.
> 
> However I tested the new feature on our cluster and it works so far.

I pulled down your changes to check. `make gotest` also runs integration tests (and also runs in CI which unfortunately doesn't run correctly for community contributions due to insufficient permissions but we're actively working to fix that).

In order to properly run them you will need to set the following env vars:
```
DEV_REGISTRY # Docker registry to push/pull local builds to
DEV_KUBECONFIG #path to a valid kubeconfig file. Integrations tests will spawn a k3s cluster container
```

At this point, I can bring over your PR so that we can run CI on your changes.

for the python stuff, that is old legacy code. I don't think we need to update that but lets see what CI says.

haq204 (commented):
once the comment is addressed I can bring over your changes so that we can run CI on it.

raffis:
> I pulled down your changes to check. `make gotest` also runs integration tests (and also runs in CI which unfortunately doesn't run correctly for community contributions due to insufficient permissions but we're actively working to fix that).
> 
> In order to properly run them you will need to set the following env vars:
> 
> ```
> DEV_REGISTRY # Docker registry to push/pull local builds to
> DEV_KUBECONFIG #path to a valid kubeconfig file. Integrations tests will spawn a k3s cluster container
> ```
> 
> For the python stuff, that is old legacy code. I don't think we need to update that.

Thanks for the explanation 👍🏻

cmd/entrypoint/interesting_types.go
```diff
@@ -52,6 +62,42 @@ func GetQueries(ctx context.Context, interestingTypes map[string]thingToWatch) [
 	return queries
 }
 
+func getWatchFieldSelector(resourcegroup string) string {
```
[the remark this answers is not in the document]
raffis: Mhm I reckon thats micro optimization as this function is only called at bootstrapping as far as I can tell. But yeah lets change it 👍🏻

raffis:
I changed some bits and it supports now selectors without the apiversion for the resourcegroup.

raffis:
@haq204 can we run the ci job?

raffis:
Sorry to bother you but can I please 🙏🏻 have an update on this? @haq204 @AliceProxy

haq204:
Hi @raffis sorry for the delay got busy with other things. 

So change wise it looks good to me but I brought this up to the rest of the Maintainers team and we were discussing it internally. It turns out that this might actually be a regressions from all the way back to 1.14 Legacy Mode. As far as my understanding goes it's not supposed to watch for all secrets and only for this it cares about. 

Given this information, this seems to suggest that the proper approach now is to, well fix the regression rather than adding another env var toggle.

@LukeShu @kflynn I think your thoughts would be constructive to this.

raffis:
Whats the status with this? We still have to run my fork because of this bug.
This fix is generic and not only about secrets, you can filter all kind of resources this way.

iomarmochtar:
I face the same issue as what explained by @raffis and seems this changes can solve it, waiting forward for any progress in this MR @haq204

raffis:
This problem still exists, just tested again with the latest released

Issue #3959, opened 2021-12-02
Emissary can't find tlsSecret for Ingress resource


nuclear0wl:
**Describe the bug**
Hello, when I'm trying to create Ingress resource with TLS enables using the manifest below:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: kubia-ambassador-ingress
  annotations:
    kubernetes.io/ingress.class: ambassador
spec:
  tls:
  - hosts:
      - host.example.com
    secretName: example-tls-secret
  rules:
  - host: host.example.com
    http:
      paths:
      - path: /kubia
        pathType: Prefix
        backend:
          service:
            name: kubia-service
            port:
              number: 80
```
I'm getting the following error:
```
2021-12-02 16:50:31 diagd 2.0.5 [P21TAEW] ERROR: Secret example-tls-secret.default unknown
2021-12-02 16:50:31 diagd 2.0.5 [P21TAEW] ERROR: Host kubia-ambassador-ingress-0-0: invalid TLS secret example-tls-secret, marking inactive
```
At the same time if I will use native emissary resources then all is Ok with the same secret:
```
apiVersion: getambassador.io/v3alpha1
kind: Host
metadata:
  name: minimal-host
spec:
  tlsSecret:
    name: example-tls-secret
  hostname: host.example.com
  mappingSelector:
    matchLabels:
      examplehost: host
  requestPolicy:
    insecure:
      action: Route
---
apiVersion: getambassador.io/v3alpha1
kind: Mapping
metadata:
  name: quote-backend
spec:
  hostname: "*"
  prefix: /backend/
  service: kubia-service:80
```
Also I've tried to obtain TLS certificate automatically using **cert-manager** and in such case I'm getting the absolutely same error.

I have a bunch of my test resources in namespace _default_ and emissary-ingress in a separate one. Do I miss something or may this be a bug? Can you please help to understand.

**To Reproduce**
Deploy Ingress resource with TLS secret presented.

**Expected behavior**
Ingress resource works absolutely the same way as Host + Mapping resources.

**Versions (please complete the following information):**
 - Ambassador: 2.0.5
 - Azure Kubernetes Service
 - Version v1.19.11
 - Helm Chart version emissary-ingress-7.1.10

volatilemolotov:
[...]

PR #4607, opened 2022-10-08
Fix finding Ingress TLS secrets

olemarkus:
## Description

Emissary has not been able to save secrets from Ingress due to checking the wrong type.

## Related Issues

#3959 This one was closed due to there being a workaround

## Testing
Manually tested. No existing tests exists around any of this code.

## Checklist


AliceProxy:
Closing this one out since the commit was merged in https://github.com/emissary-ingress/emissary/pull/4610. Github just didn't close this out automatically due to my last rebase. Thanks for the writeup and fix!

PR #4618, opened 2022-10-12
fix: perform https redirect prior to ext_authz calls

LanceEa:
## Description

Fixes a regression with https_redirection when an AuthService is applied. This was introduced in v3+ due to upgrading Envoy where a behavior change was introduced for the ext_authz http filter. This restores the previous behavior by disabling ext_authz calls on a per route basis when an AuthService is applied.

Before change: 

```json
{
    "match": {
      "case_sensitive": true,
      "prefix": "/backend/",
      "runtime_fraction": {
        "default_value": {
          "denominator": "HUNDRED",
          "numerator": 100
        },
        "runtime_key": "routing.traffic_shift.cluster_quote_default_default"
      }
    },
    "redirect": {
      "https_redirect": true
    }
```

After change:

```json
{
    "match": {
      "case_sensitive": true,
      "prefix": "/backend/",
      "runtime_fraction": {
        "default_value": {
          "denominator": "HUNDRED",
          "numerator": 100
        },
        "runtime_key": "routing.traffic_shift.cluster_quote_default_default"
      }
    },
    "redirect": {
      "https_redirect": true
    },
    "typed_per_filter_config": {
      "envoy.filters.http.ext_authz": {
        "@type": "type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthzPerRoute",
        "disabled": true
      }
    }
  }
```

**IMPORTANT**: this is only applied to the https_redirect routes that are added by default for all Host which has `requestPolicy.insecure.action = Redirect` as the default action.

## Related Issues
#4620

## Testing

Manual Verification:
- reproduced in a cluster with 3.2.0
- tested dev build and verified https redirection occurs 
- visually inspected the envoy.json to ensure expected config was generated

Automtates Test:
- Added two new Unit Test to verify generated envoy config from diagd
- Added new KAT test to verify 301 redirect for non-tls request

## Checklist

### DIFF
--- a/.github/workflows/promote-ga.yml
+++ b/.github/workflows/promote-ga.yml
@@ -15,7 +15,7 @@ jobs:
       DEV_REGISTRY: ${{ secrets.DEV_REGISTRY }}
       RELEASE_REGISTRY: ${{ secrets.RELEASE_REGISTRY }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0
       - name: "Docker Login"
@@ -58,7 +58,7 @@ jobs:
     env:
       AMBASSADOR_RELENG_NO_GUI: "1"
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0
       - name: "Install Deps"
--- a/python/tests/kat/t_extauth.py
+++ b/python/tests/kat/t_extauth.py
@@ -1348,3 +1348,100 @@ service: {self.target.path.fqdn}
         assert self.results[3].backend.request.headers["kat-resp-extauth-protocol-version"] == [
             self.expected_protocol_version
         ]
+
+
+class AuthenticationHTTPSRedirectTest(AmbassadorTest):
+    """
+    AuthenticationHTTPSREdirect: We expect that the http-to-https will occur
+    without calling the AuthService (ext_authz).
+    """
+
+    target: ServiceType
+    auth: ServiceType
+
+    def init(self):
+        if EDGE_STACK:
+            self.xfail = "custom AuthServices not supported in Edge Stack"
+        self.target = HTTP()
+        self.auth = AHTTP(name="auth")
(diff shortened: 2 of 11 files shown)


LanceEa:
@LukeShu @ddymko  - I have the KAT testing passing now so this is good to go for a review.

As I was working on the KAT tests I added these two commits for CI improvements:
- d31ab7f408ba43b70661e89cfb872ecf4e651483
- 902bed428da6796fd50d34730f7a9500b4dd8b4f

I'm happy to PR them separately if they are sticking points, in favor of landing the fix commit.

.github/workflows/promote-ga.yml
```diff
@@ -58,7 +58,7 @@ jobs:
     env:
       AMBASSADOR_RELENG_NO_GUI: "1"
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/3
```
ddymko, line 61: `checkout@v` got accidentally removed

LanceEa: Good catch, I actually had fixed that locally but forgot to push it.

python/tests/kat/t_extauth.py
```diff
@@ -1348,3 +1348,100 @@ def check(self):
...
+        if EDGE_STACK:
+            self.xfail = "custom AuthServices not supported in Edge Stack"
+        self.target = HTTP()
+        self.auth = AHTTP(name="auth")
+        self.add_default_http_listener = False
+        self.add_default_https_listener = False
+
+    def manifests(self) -> str:
+        return (
+            self.format(
```
LukeShu, line 1372: When possible, I've been preferring f-strings to the old `self.format` magic.

LukeShu (approved):
There's the redirect-`Mapping` issue discussed today that still needs fixed.  But the changes here lgtm.

LanceEa:
> There's the redirect-`Mapping` issue discussed today that still needs fixed. But the changes here lgtm.

Cool, I will go ahead and land this so it can be in the next RC and will get to work on the fix for the mapping redirect issue as well.

PR #4619, opened 2022-10-13
Updated documentation to avoid unnecessary changes to ClientConfig

aosoriodw:
## Description
Documented that `NewClientFromConfigFlags` should be used when more control over the client settings is needed.

## Testing
Not required

## Checklist

    + Any new features
    + Any deprecations

### DIFF
--- a/pkg/kates/client.go
+++ b/pkg/kates/client.go
@@ -96,6 +96,12 @@ type Client struct {
 
 // The ClientConfig struct holds all the parameters and configuration
 // that can be passed upon construct of a new Client.
+// If you need support for additional parameters, consider creating a client with
+// NewClientFromConfigFlags like this:
+//
+// clientConfigFlags := NewConfigFlags(false)
+// clientConfigFlags.Impersonate = &username
+// client, err := NewClientFromConfigFlags(clientConfigFlags)
 type ClientConfig struct {
 	Kubeconfig string
 	Context    string


rhs:
FYI Kates already supports impersonating users and groups. If you use `k8s.NewKubeInfoFromFlags` to construct your client then you will automatically get all the common boilerplate options that kubectl supports, including the impersonation flags. I'd recommend using that pattern rather than wiring up each flag one by one just because there are a lot of those flags and the set keeps growing as client-go evolves. See here for an example of how to do that: https://github.com/emissary-ingress/emissary/blob/9e8c3ec0fee5c878cba7410a3a62f861862683ff/cmd/kubestatus/main.go#L26

rhs:
Just to be clear that's not a comment on this PR. The ClientConfig struct exists as it does because a) I didn't know about the generic client options portion of client-go when I wrote it and b) it's a reasonable convenience for a common case. I think it's reasonable to extend that convenience to include impersonation, the only other thing I'd suggest might be useful is to update the docs for ClientConfig to include a pointer to `k8s.NewKubeInfoFromFlags` along with a suggestion that if you're finding a need for some of the fancy options, then you probably just want to switch to that instead of using ClientConfig.

rhs (changes requested):
Looks good to me, the only change I'd suggest is updating the documentation as I mentioned in my earlier comment.

aosoriodw:
Agree that it's better to use `NewClientFromConfigFlags` when more control over the client configuration is needed. I've dropped the changes since they are not necessary and I have updated the documentation.

haq204:
@aosoriodw changes look fine to me. but looks like the DCO check is failing so you will need to add a `Signed-off-by: ` to your commit.

LanceEa (approved):

LanceEa:
Just needs the DCO, you will need to sign-off your commit. You can run `git commit --amend --signoff ` to amend the last commit and add sign-off signature.

Issue #4629, opened 2022-10-17
Temporarily misconfigured routes after mapping change


dynajoe:
I've noticed an issue where routes disappear for a very short period of time. We run very high frequency and notice some of our grpc clients are getting "Unimplemented" error code and a NR envoy code. I believe the issue is when a route is removed from route list in such a way that shifts all routes which seem to be named based on index in a list: 

https://github.com/emissary-ingress/emissary/blob/1d62b9587a990dc6ecc754be2f27be2e2f577922/pkg/ambex/transforms.go#L140 - I suspect that while this reconfiguration is happening (usually < 1 sec) the routes sorted AFTER the route being removed are briefly unavailable.

iNoahNothing:
I am running into this as well. In my case, the routes are not readded until another configuration change is applied. It seems like that configuration changes looks like it needs to be something other than a `Mapping` change or a restart of emissary in order for the envoy config to be generated correctly again.

LanceEa:
@dynajoe - I would love to get this one landed. As I mentioned in https://github.com/emissary-ingress/emissary/pull/4828, I would like to see some tests added for the behavior.

We have a Contribfest next week at Kubecon EU so I have marked with as a `good-first-issue` so that if you are ok with it, someone could finish up the tests to get this over the finish line. Let me know if you have any issues with that!

PR #4630, opened 2022-10-17
Fix routes shifting causing routing issues during a reconfiguration d…

dynajoe:
Addresses https://github.com/emissary-ingress/emissary/issues/4629

By hashing the filter match condition we can better ensure that routes have a stable name over a simple ordinal index.

### DIFF
--- a/pkg/ambex/transforms.go
+++ b/pkg/ambex/transforms.go
@@ -3,6 +3,9 @@ package ambex
 import (
 	// standard library
 	"context"
+	"crypto/md5"
+	"encoding/hex"
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


kflynn (approved):
This looks great. I'm curious to see how CI handles it (I don't _think_ anything in the test suite relies on this value, but... yeah), and I'll note that elsewhere we use SHA128 in similar situations just because I was feeling paranoid about MD5 collisions. Any other @emissary-ingress/emissary-maintainers have an opinion there?

I'm going to go ahead and approve with the understanding that CI needs to run on it.

PR #4642, opened 2022-10-22
release: prepare for v3.3.0

LanceEa:
## Description

Preparing change logs and readme for the next minor release version `v3.3`. Once this lands the branch will be tagged so that branch checks no longer fail.

## Related Issues
N/A

## Testing
Docs only

## Checklist
- [ ] **Does my change need to be backported to a previous release?**
- [x] **I made sure to update `CHANGELOG.md`.**
- [x] **This is unlikely to impact how Ambassador performs at scale.**
- [ ] **My change is adequately tested.**
- [ ] **I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently.**
- [ ] **The changes in this PR have been reviewed for security concerns and adherence to security best practices.**

PR #4644, opened 2022-10-23
cleanup: remove unused python imports

LanceEa:
## Description

After reviewing some of the dependabot PR's I noticed that some of them look as though they were no longer being used. So, this is to help get a better picture of what deps are actually being used and see which ones are no longer needed. This will help with dependabot and scanning tools so that we are not being flag for outdated deps that no longer get used.

The following was done:

- Executed `autoflake --remove-all-unused-imports  -r -i ./python` to
remove unused exports within the python.
- Reviewed changes to ensure re-exports were not removed.
- Executed `black ./python/` to fix lint issues
- Executed `pip-compile --allow-unsafe --no-build-isolation -q` to
update the requirements.txt to remove unused deps.
- Executed `isort ./python/` to ensure imports are sorted correctly
after updating them
- Executed `make generate` to ensure we capture updated dependenceis

## Related Issues
Example dependabot PR that is no longer used and we can remove the dependency. 

https://github.com/emissary-ingress/emissary/pull/4638

## Testing
CI Testing

## Checklist
 Yes, lets backport to `release/v2.5` to make it cleaner for cherry-picks. In general, may be cleaner to just run it there too instead because of the additional Envoy V2 files and some of these imports might still be used in the release/v2.5 branch.

### DIFF
--- a/python/ambassador/cache.py
+++ b/python/ambassador/cache.py
@@ -1,5 +1,5 @@
 import logging
-from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Set, Tuple
+from typing import Any, Callable, Dict, Optional, Set, Tuple
 
 
 class Cacheable(dict):
@@ -242,7 +242,6 @@ class NullCache(Cache):
         self.logger = logger
         self.logger.debug("NullCache: INIT")
         self.reset_stats()
-        pass
 
     def add(self, rsrc: Cacheable, on_delete: Optional[DeletionHandler] = None) -> None:
         pass
@@ -252,7 +251,6 @@ class NullCache(Cache):
 
     def invalidate(self, key: str) -> None:
         self.invalidate_calls += 1
-        pass
 
     def __getitem__(self, key: str) -> Any:
         self.misses += 1
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -145,7 +145,6 @@ libraries:
     PyYAML              5.4.1      MIT license
     Werkzeug            2.2.1      3-clause BSD license
     appdirs             1.4.4      MIT license
-    attrs               21.4.0     MIT license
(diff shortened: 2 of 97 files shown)


LanceEa:
Once PR https://github.com/emissary-ingress/emissary/pull/4642 is merged then the check branch will go away.

python/ambassador/cache.py
```diff
@@ -242,7 +242,6 @@ def __init__(self, logger: logging.Logger) -> None:
         self.logger = logger
         self.logger.debug("NullCache: INIT")
         self.reset_stats()
-        pass
```
haq204, line 245: was this manually removed or from a linter? seems slightly odd for an import checker to remove. Maybe autoflake?

haq204 (commented):
This makes me want to add pylint to our list of python linters which catches for things like unused imports, among other things.

I'm also fully relying on CI for testing

LanceEa:
> This makes me want to add pylint to our list of python linters which catches for things like unused imports, among other things.
> 
> I'm also fully relying on CI for testing

Yeah, I think we can add something going forward. I wanted to use this to help cleanup some of the dependabot stuff because if I have time this week I will address some of the other opens.

Since I'm not a python person I don't have a preference on pylint vs autoflake vs black so we can discuss that as some follow-up work to do.

python/ambassador/cache.py
haq204: yea `pass` is a no-op in python. it basically tells the python interpreter to do nothing

haq204 (dismissed):
lgtm

LanceEa:
Note: I will rebase and once we land the CI fix and that will also address the Branch version check too now that things have been tagged properly.

haq204 (approved):

AliceProxy (approved):
Love it, +1 to adding a linter that checks for this

PR #4649, opened 2022-10-26
charts: add startupProbe to emissary-ingress deployments

dethi:
## Description

emissary-ingress can be slow to start when the number of mappings grow. In that case, using a startupProbe is preferred over tweaking the initialDelaySeconds of the livenessProbe and readinessProbe.

startupProbe allow us to be generous with how long the pod takes to be ready, defined as `periodSeconds * failureThreshold`, but if it takes less time than allowed it will become ready faster (compared to having a high value for the initialDelaySeconds).

For example:
- startupProbe set with periodSeconds=10 failureThreshold=30 = 5mins
- livenessProbe/readinessProbe set with initialDelaySeconds=0

If the /check_ready endpoint returns OK after 2mins, the the startupProbe finish and the livenessProbe/readinessProbe immediatetly succeed, changing the pod state to live and ready.

If instead we didn't have the startupProbe but initialDelaySeconds was set to 300 (5min), the pod would always take 5min to be live and ready, even if /check_ready returns OK after 2mins.

## Related Issues

N/A

## Testing

Manual deploy

## Checklist

  - Any new features
  - Any deprecations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,13 @@ it will be removed; but as it won't be user-visible this isn't considered a brea
 
 ### Emissary-ingress and Ambassador Edge Stack
 
+- Change: emissary-ingress can be slow to start when there is a lot of mappings. Kubernetes
+  recommends using a startupProbe over tweaking the initialDelaySeconds of the livenessProbe and
+  readinessProbe. The Helm chart was updated to allow setting a startupProbe on the emissary-ingress
+  deployment. ([#4649])
+
+[#4649]: https://github.com/emissary-ingress/emissary/pull/4649
+
 ## [3.4.0] January 03, 2023
 [3.4.0]: https://github.com/emissary-ingress/emissary/compare/v3.3.0...v3.4.0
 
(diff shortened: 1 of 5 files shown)


dethi:
Two open questions:
- should this be added to the changelog?
- currently I have 1min as default, but if someone has customised their initialDelaySeconds to a larger value and upgrade the helm chart with this version, it will start failing. Maybe a better default would be to disable the startupProbe?

dethi:
Updated the change!

LanceEa:
@AliceProxy - I pulled this into AES and verified that there are no issues. https://github.com/datawire/apro/pull/3221

 I was able to deploy a cluster without setting enabling startup probe and with it enabled. No issues starting-up.

LanceEa (approved):
lgtm Thanks!

LanceEa:
@dethi - Thanks so much for the contribution! I landed this in https://github.com/emissary-ingress/emissary/pull/4779 and it will go out with the next minor release.

dethi:
Great, thanks!

PR #4655, opened 2022-10-28
Disable AMBASSADOR_GRPC_METRICS_SINK if agent is not enabled

Shadowssong:
## Description

When not using the ambassador-agent emissary-ingress will produce a high number of 5xx errors when attempting to send metrics to the `envoy_metrics_service`.  This in turn generates prometheus alerts based on these errors.  If you are not using ambassador-agent you should not set the environment variable `AMBASSADOR_GRPC_METRICS_SINK`.  Without this environment variable the static_resources.clusters config will not create the `envoy_metrics_service`.  The default config when this environment variable is set creates this cluster config:

```
       {
        "connect_timeout": "1s",
        "http2_protocol_options": {},
        "load_assignment": {
          "cluster_name": "envoy_metrics_service",
          "endpoints": [
            {
              "lb_endpoints": [
                {
                  "endpoint": {
                    "address": {
                      "socket_address": {
                        "address": "ambassador-external-agent",
                        "port_value": 80,
                        "protocol": "TCP"
                      }
                    }
                  }
                }
              ]
            }
          ]
        },
```

This is uses a semi-hardcoded value for the destination.  Without ambassador agent enabled this is an invalid host.  

Here are some prometheus graphs showing the high 5xx rate on my local development cluster (it is higher in production):

prometheus_5xx_steady_error_rate_ambassador_metrics_service

After applying this change the 5xx errors for this service stopped for both of our ambassador deployments:

prometheus_5xx_error_drop_ambassador_internal
prometheus_5xx_error_drop_ambassador_external

This is my first change to the ambassador/emissary-ingress ecosystem so please let me know any additional information that is needed.

## Related Issues

I did not find an existing issue and since I found the issue I decided to create a PR to address instead of an issue although I can still open an issue if that is preferred.

## Testing

To test this locally I did a helm pull and modified it locally while pointing our ArgoCD config at this change:

```
helm pull emissary-ingress --repo https://app.getambassador.io --version 8.0.0 --untar
```

I confirmed after this was applied that the ambassador config did no longer have the static_resources.clusters envoy_service_monitor config as well.

## Checklist

  - Any new features
  - Any deprecations

### DIFF
--- a/charts/emissary-ingress/templates/deployment.yaml
+++ b/charts/emissary-ingress/templates/deployment.yaml
@@ -172,8 +172,10 @@ spec:
             - name: admin
               containerPort: {{ .Values.adminService.port }}
           env:
+            {{- if .Values.agent.enabled }}
             - name: AMBASSADOR_GRPC_METRICS_SINK
               value: {{ include "ambassador.fullname" . }}-agent:80
+            {{- end }}
             - name: HOST_IP
               valueFrom:
                 fieldRef:


haq204:
Change wise this looks good to me. I don't think there are any other uses for `AMBASSADOR_GRPC_METRICS_SINK` besides creating the metrics cluster.

@LukeShu do you know of any other potential uses?

Shadowssong:
@haq204 or @LukeShu any suggestions on how to move this forward?  We have been silencing our alerts for this service for a while (and now adjusting the alerts to exclude this specific service).

PR #4666, opened 2022-11-07
build: upgrade openssl-dev in base-python image

haq204:
## Description
Alpine 3.15 updated openssl to 1.1.1s-r1 so the previous version 1.1.1q-r0 is not available anymore on pkgs.alpinelinux.org for the v3.15 channel. In order to avoid this in the future, we're just going to pin the minor version to 1.1.1 - openssl 1.X does not follow traditional semantic versioning; the scheme is `<major>.<minor>.<minor><patch-letter>`.

Also pin mypy. mypy released a new version [0.990](https://github.com/python/mypy/releases/tag/v0.990) today  which is making some breaking changes causing mypy runs to fail on our existing code. While we should address them, just pinning the mypy version for now so the rules don't change on us unexpectedly

## Related Issues

List related issues.

## Testing

CI

## Checklist

  - Any new features
  - Any deprecations

### DIFF
--- a/python/requirements-dev.txt
+++ b/python/requirements-dev.txt
@@ -1,7 +1,7 @@
 # Dev requirements
 gitpython
 httpretty
-mypy
+mypy<0.990
 packaging
 pexpect
 pyOpenSSL
--- a/docker/base-python/Dockerfile
+++ b/docker/base-python/Dockerfile
@@ -39,7 +39,7 @@ RUN apk --no-cache add \
   libcap-dev \
   libffi-dev \
   ncurses=6.3_p20211120-r1 \
-  openssl-dev=1.1.1q-r0 \
+  openssl-dev~=1.1.1 \
   py3-pip=~20.3.4 \
   python3=~3.9.13 \
   python3-dev \


LanceEa (dismissed):
lgtm...once green. 

Note: we should probably backport this to release/v2.5 branch as well because it will probably be bitten by it too

python/requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
 # Dev requirements
 gitpython
 httpretty
-mypy
+mypy<0.990
```
LanceEa, line 4: Is there a reason not to use `=`, if I'm reading the version specifier this says do not allow pre-releases, is that correct?

LanceEa (approved):
lgtm...let's get this landed to unblock the other PR's. I just had one question to make sure I understood the version specifier syntax.

python/requirements-dev.txt
haq204: it just means use the latest version that's below 0.990. so 0.990 is the upper bound. I just don't know how mypy releases things and it appears they don't do semantic versioning so this is just to cover scenarios if they do a future 0.8x release

AliceProxy (approved):
gucci gang

Issue #4274, opened 2022-06-10
IPv6 for healthcheck server


TimonOmsk:
## Problem

On January 10th, AWS launch IPv6 EKS cluster support. Any pod-to-pod communication within or across nodes always uses the pod’s IPv6 address. VPC CNI configures iptables to handle IPv6 while blocking any IPv4 connections. Announcement - https://aws.amazon.com/ru/blogs/containers/amazon-eks-launches-ipv6-support/

Emissary exposes healtcheck server for readiness\liveness checks only via IPv4. So, it makes impossible to check health of the pod by k8s on the such environemnts. We're at Kasten by Veeam using emissary so the IPv4-only healthcheck  blocks us from support of such IPv6 stacks. 

## Solution

I've found that listening only to IPv4 is hardcoded because some "Bad Things Happen" https://github.com/emissary-ingress/emissary/blob/e91276327cce20aa273e2d1be46c4f724de6953a/cmd/entrypoint/healthcheck_server.go#L113

Those "bad things" are described as:
> "if we don't explicitly say "tcp4" here, we seem to listen _only_ on v6". 

I don't know the initial motivation why the author made such decision(about IPv6 only), but nowadays it looks that this statement is not true. My proposal is to replace the "tcp4" with the "tcp" parameter in the net.Listen() func. In that case it will listen connections for both AF(AF_INET4 AND AF_INET6)

### Theory
There are 3 options(actually more, but other options are not relevant in this particular case) available for net.Listen(): tcp, tcp4, tcp6
"tcp4" and "tcp6" represents the options where the only specified type will be set as an allowed stack. Nothing to add. 

The interersting part is about "tcp" option. 

Here is how it described by the docs for socket creation function:
> A listen for a wildcard communication domain, "tcp" or
"udp", with a wildcard address: If the platform supports
both IPv6 and IPv4-mapped IPv6 communication capabilities,
or does not support IPv4, we use a dual stack, AF_INET6 and
IPV6_V6ONLY=0, wildcard address listen. The dual stack
wildcard address listen may fall back to an IPv6-only,
AF_INET6 and IPV6_V6ONLY=1, wildcard address listen.
Otherwise we prefer an IPv4-only, AF_INET, wildcard address listen.

[...]

PR #4301, opened 2022-06-22
Use environment variables to configure healthcheck server

TimonOmsk:
Signed-off-by: Dmitry Golushko <dmitry.golushko@veeam.com>

## Description

Full description and motivation can be found here:  https://github.com/emissary-ingress/emissary/issues/4274 

I decided to make it flexible and keep backward compatibility. 
I introduced 3 new environment variables:
|  Name | Description  | Default value  | Possible values | 
|---|---|---|---|
|  AMBASSADOR_HEALTHCHECK_BIND_ADDRESS |  allows to configure bind address for healthcheck server |  0.0.0.0 |  Any valid IPV4 or IPV6 address  |
|  AMBASSADOR_HEALTHCHECK_BIND_PORT |  allows to configure port for healthcheckserver | 8877  | Any valid port value  |
|  AMBASSADOR_HEALTHCHECK_IP_FAMILY |  allows to configure prefered IP family(IPV4, IPV6 or both) | ANY  | IPV4_ONLY, IPV6_ONLY, ANY  |

As I already described in the feature request in case of any wildcard IP(0.0.0.0, :: or just empty string) golang will try to bind to all available interface. Even if you specified IPV4 address (0.0.0.0) and host system is properly configured to use IPV4 mapped IPV6 address it will bind to IPV6 socket. To allow explicitly set preferred IP family the AMBASSADOR_HEALTHCHECK_IP_FAMILY was introduced. 

## Related Issues
https://github.com/emissary-ingress/emissary/issues/4274 

## Testing
A few sentences describing what testing you've done, e.g., manual tests, automated tests, deployed in production, etc.

## Checklist

 - [ ] I made sure to update `CHANGELOG.md`.

   Remember, the CHANGELOG needs to mention:
    + Any new features
    + Any changes to our included version of Envoy
    + Any non-backward-compatible changes
    + Any deprecations

 - [x] This is unlikely to impact how Ambassador performs at scale.

   Remember, things that might have an impact at scale include:
    + Any significant changes in memory use that might require adjusting the memory limits
    + Any significant changes in CPU use that might require adjusting the CPU limits
    + Anything that might change how many replicas users should use
    + Changes that impact data-plane latency/scalability

 - [ ] My change is adequately tested.

   Remember when considering testing:
    + Your change needs to be specifically covered by tests.
[...]

PR #4672, opened 2022-11-08
community: Use environment variables to configure healthcheck server

LanceEa:
## Description

This pulls in the work from the community contribution on making the healthcheck server configurable via Environment variables so that we can support IPV6 setups. This will include the PR run and just cleaning up the docs.

See existing community PR.
https://github.com/emissary-ingress/emissary/pull/4301

## Related Issues
Fixes https://github.com/emissary-ingress/emissary/issues/4274

## Testing
CI and this is being used in production for community members product.

## Checklist

### DIFF
--- a/cmd/entrypoint/env.go
+++ b/cmd/entrypoint/env.go
@@ -297,3 +297,36 @@ func GetSidecarUrl() string {
 func IsKnativeEnabled() bool {
 	return strings.ToLower(env("AMBASSADOR_KNATIVE_SUPPORT", "")) == "true"
 }
+
+// getHealthCheckHost will return address that the health check server will bind to.
+// If not provided it will default to all interfaces (`0.0.0.0`).
+func getHealthCheckHost() string {
+	return env("AMBASSADOR_HEALTHCHECK_BIND_ADDRESS", "0.0.0.0")
+}
+
+// getHealthCheckPort will return the port that the health check server will bind to.
+// If not provided it will default to port `8877`
+func getHealthCheckPort() string {
+	return env("AMBASSADOR_HEALTHCHECK_BIND_PORT", "8877")
+}
+
+// getHealthCheckIPNetworkFamily will return the network IP family that the health checker server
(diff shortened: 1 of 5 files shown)


LanceEa:
@TimonOmsk - FYI...due to the current state of our CI, it requires a maintainer to run it. Due to this is makes it a really slow back and forth process if I was to ask you to make changes. Therefore,  I have gone ahead and pulled in your commit, made some of my suggested changes (added unit tests, consolidated functions, doc comments) but overall the essence of your PR is still the same. 

Once this has been reviewed we will get it landed. Thanks again for the contribution and patience as we continue to evolve and improve our processes.

cmd/entrypoint/env.go
```diff
@@ -297,3 +297,36 @@ func GetSidecarUrl() string {
...
+func getHealthCheckIPNetworkFamily() string {
+	ipFamily := strings.ToUpper(env("AMBASSADOR_HEALTHCHECK_IP_FAMILY", "ANY"))
+
+	switch ipFamily {
+	case "IPV4_ONLY":
+		return "tcp4"
+	case "IPV6_ONLY":
+		return "tcp6"
+	}
+	return "tcp"
```
haq204, line 331: I feel like we should log a warning if a non supported value is set

haq204 (commented):
one minor comment but otherwise lgtm

haq204 (approved):
approving since I don't consider my comment serious enough to block

cmd/entrypoint/env.go
LanceEa: I don't want to muddy this simple function with introducing `ctx` so we can log. I think a trade-off would be to add a log line in the handler where it is passed to `net.Listen` so that it is explicit to the user.

LanceEa:
FYI... verifying in https://github.com/datawire/apro/pull/3161 in Edge Stack.

AliceProxy (approved):

PR #4674, opened 2022-11-08
backport: remove unused python deps

LanceEa:
## Description

**NOTE: this wasn't a direct cherry-pick due to envoy v2 configuration so I re-ran the same commands as I did for master to ensure I captured all the files. TL;DR; double check my work 😄.**

This is to help get a better picture of what deps are actually being used and see which ones are no longer needed. This will help with dependabot and scanning tools so that we are not being flag for deps that no longer get used.

The following was done:

- Executed `autoflake --remove-all-unused-imports  -r -i ./python` to remove unused exports within the python.
- Reviewed changes to ensure re-exports were not removed.
- Executed `black ./python/` to fix lint issues
- Executed `pip-compile --allow-unsafe --no-build-isolation -q` to update the requirements.txt to remove unused deps.
- Executed `isort ./python/` to ensure imports are sorted correctly after updating them
- Executed `make generate` to ensure we capture updated dependencies

## Related Issues
Normal steward practices

## Testing
CI 

## Checklist


python/tests/unit/test_bootstrap.py
```diff
@@ -23,7 +21,7 @@ def _test_bootstrap(yaml, v2, expectations={}):
         if expected is None:
             assert key not in bootstrap
         else:
-            import json
+            pass
```
haq204, line 24:
```python
pass

assert key in bootstrap
assert bootstrap[key] == expected
```

may not be what's intended here

LanceEa: yup...good catch!

LanceEa: Fixed, interestingly. I don't think anything in that file is called at all but I will leave looking into cleaning it up as a separate activity.

LanceEa:
I went ahead rebased because ewe had quite a few changes land in release/v2.5 so I wanted to make sure nothing was broken afterwards.

haq204 (approved):
pending CI passing

LanceEa:
Going to rebase to pull in the Kat fixes which should address flakes and address conflicts now which I kind of expected.

haq204 (approved):

AliceProxy (approved):

PR #4684, opened 2022-11-10
[v3.x] Bump the Python version

LukeShu:
## Description

> This is the 3.x equivalent of https://github.com/emissary-ingress/emissary/pull/4683

We're pinning Python 3.9.13, but Alpine now only has 3.9.15 available, so relax the pin in order to upgrade to that, in order to fix the build.  While I'm at it, clean up the pinned versions, and add a comment explaining a decent package pinning policy.

## Related Issues

(none)

## Testing

(CI should cover it)

## Checklist

- [x] **Does my change need to be backported to a previous release?** - yes, I've backported this to 2.5 as #4683

- [ ] **I made sure to update `CHANGELOG.md`.** - I don't think this is worthy of a changelog entry

  - Any new features
  - Any deprecations

- [x] **I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently.** - OK, not DEVELOPING.md, but I did add comments about this.

### DIFF
--- a/docker/base-python/Dockerfile
+++ b/docker/base-python/Dockerfile
@@ -40,8 +40,10 @@ ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local
 #   forced-bumps: 0
 #
 # The exception to this is that we pin 'python3' and 'py3-pip' because
-# we have other code that will need to change if those versions
-# change.
+# the 'pip-tools' version below will also need to change if those
+# versions change.  And even then, don't get more precise with the
+# version number than we need to ensure that the pip-tools version
+# agrees.
 RUN apk --no-cache add \
   bash \
   gcc \
@@ -58,8 +60,8 @@ RUN apk --no-cache add \
   libffi-dev \
   ncurses \
   openssl-dev \
-  py3-pip=~20.3.4 \
-  python3=~3.9.13 \
+  py3-pip=~20.3 \
+  python3=~3.9 \
   python3-dev \
   rust \
   cargo \
--- a/docker/base-python/Dockerfile
+++ b/docker/base-python/Dockerfile
@@ -24,12 +24,30 @@ WORKDIR /buildroot
 
 ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/go/bin:/buildroot/bin
 
+# For the most-part, you should not pin specific versions in this
+# `apk` command; it will always choose the latest version being
+# shipped by pinned version of Alpine.  This will automatically pull
+# in any patches, but not breaking changes.
+#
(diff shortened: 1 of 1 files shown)


LanceEa:
@LukeShu - Interesting, so the reason I pinned them before was that back 4 months ago when this same thing occurred that was the only way to force it to pick up the new versions without tripping the security scanner. As you mentioned in the comments the weekly cache of the base-python probably wasn't able to pickup these new versions. Then by happen chance the updating to a pin version was enough to cache-bust it.

LanceEa (approved):
lgtm...once green

docker/base-python/Dockerfile
```diff
@@ -24,12 +24,32 @@ WORKDIR /buildroot
...
 ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/go/bin:/buildroot/bin
 
+# For the most-part, you should not pin specific versions in this
+# `apk` command; it will always choose the latest version being
+# shipped by pinned version of Alpine.  This will automatically pull
+# in any patches, but not breaking changes.
+#
+# Because of how we cache the base image
+# (`docker/base-python.docker.gen`), it'll keep cache that for a week.
+# If there's a security patch patch that we want to pull in without
```
haq204, line 34: extra word patch.

Also `Because of how we cache the base image (docker/base-python.docker.gen), it'll keep cache that for a week` needs a grammar check

haq204 (commented):
just some grammar nits for the comment but otherwise lgtm

LukeShu:
v2: grammar fixes requested by Hamzah:
```patch
diff --git a/docker/base-python/Dockerfile b/docker/base-python/Dockerfile
index e21036f0c..29979fa2f 100644
--- a/docker/base-python/Dockerfile
+++ b/docker/base-python/Dockerfile
@@ -30,12 +30,12 @@ ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local
 # in any patches, but not breaking changes.
 #
 # Because of how we cache the base image
-# (`docker/base-python.docker.gen`), it'll keep cache that for a week.
-# If there's a security patch patch that we want to pull in without
-# waiting a week for the cache to roll over, instead of fussing with
-# pinning a specific package version, simply make an inconsequential
-# change that changes the file's checksum (such as bumping this
-# comment) :
+# (`docker/base-python.docker.gen`), it'll keep selected versions
+# around for a week.  If there's a security patch that we want to pull
+# in without waiting a week for the cache to roll over, instead of
+# fussing with pinning a specific package version, simply make an
+# inconsequential change that changes the file's checksum (such as
+# bumping this comment):
 #
 #   forced-bumps: 0
 #
```

haq204 (approved):

LanceEa (approved):

Issue #4691, opened 2022-11-13
Ingress controller does not work out of the box when installing the Helm chart and configuring AMBASSADOR_ID


eevdev:
**Describe the bug**
I encountered some problems while trying to get Emissary-ingress to act as an `Ingress` controller after installing Emissary-ingress using the Helm chart and configuring `AMBASSADOR_ID`:

1. Helm - generated `IngressClass` is missing the required `getambassador.io/ambassador-id` annotation
If `AMBASSADOR_ID` is set then the annotation `getambassador.io/ambassador-id` should be included in the `IngressClass` generated by the Helm chart, otherwise the `IngressClass` will be ignored ([source](https://github.com/emissary-ingress/emissary/blob/507bf91f95e489d7d5cb437ca7cd1d8f320d4a81/python/ambassador/fetch/ingress.py#L37-L41)).

2. Helm - misleading option to specify `IngressClass` controller value
The chart `IngressClass` template was added in https://github.com/emissary-ingress/emissary/issues/3975 ([PR](https://github.com/emissary-ingress/emissary/pull/4169/files)). Although a part of the suggested solution was to specify the controller class (default is `getambassador.io/ingress-controller`), the solution only touched the `IngressClass` and not other references; this means that Emissary-ingress still requires the controller value to be `getambassador.io/ingress-controller`, so it makes sense to remove the option to configure this value from the chart as changing it would not work.

3. Emissary-ingress - Ingress controller - it should be unnecessary to specify ambassador-id on both `Ingress` and `IngressClass`
If `AMBASSADOR_ID` is set, there's currently a requirement that the `getambassador.io/ambassador-id` annotation must be set on both the `Ingress` and the `IngressClass` (if used instead of the [deprecated `kubernetes.io/ingress.class` annotation](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation)). Setting the annotation on either `Ingress` or the assigned `IngressClass` should be sufficient.

**To Reproduce**
(I can't remember if it was necessary to restart Emissary-ingress pods after changing `Ingress` and/or `IngressClass` in the steps below)

1. Install Emissary-ingress using the Helm chart and specify `AMBASSADOR_ID` (in my case I have two installations in the same namespace, one for internal traffic and the other for public-facing services)
2. Set up an `Ingress` and assign the generated `IngressClass` to `spec.ingressClassName` (don't add the `getambassador.io/ambassador-id` annotation yet)
[...]

PR #4692, opened 2022-11-13
Ingressclass fixes

eevdev:
## Description

Make Emissary work as an Ingress controller out of the box when installed with Helm chart and AMBASSADOR_ID is set.

> fix(helm): don't allow changing controller value for generated IngressClass, as the expected controller name is hard-coded in class IngressClassProcessor
fix(helm): added missing annotation getambassador.io/ambassador-id to IngressClass when AMBASSADOR_ID is set
feat(ingress): allow setting the getambassador.io/ambassador-id annotation on either the Ingress or the referenced IngressClass instead of requiring it to be set on both

## Related Issues

https://github.com/emissary-ingress/emissary/issues/4691

## Testing

I haven't tested it yet, but managed to create the container images and will see if I can test those tomorrow/this week.

## Checklist

  - Any new features
  - Any deprecations


python/ambassador/fetch/ingress.py
```diff
@@ -127,8 +127,9 @@ def _process(self, obj: KubernetesObject) -> None:
             )
             return
 
-        # We don't want to deal with non-matching Ambassador IDs
-        if obj.ambassador_id != Config.ambassador_id:
+        # Ambassador ID, if specified, must match on either IngressClass or Ingress.
+        # The IngressClass, if we have one here, has already been checked for Ambassador ID in IngressClassProcessor._process
+        if not has_ingress_class and obj.ambassador_id != Config.ambassador_id:
```
eevdev, line 132: It seems this isn't enough to do what I intended, ie. avoid having to set the annotation on both ingress and ingressclass. I rarely have time to look at this though and last time I did I was a bit stuck, as in I don't know where to start.

I'll either need to figure out why this isn't enough or revert this part so that the remaining changes can move forward.

Bak3y:
Any updates/feedback on this MR?  If this fixes https://github.com/emissary-ingress/emissary/issues/4691 that'd be amazing

python/ambassador/fetch/ingress.py
eevdev: I've changed the PR to only deal with this issue now, so I'll just need to find some time to check this again to see if i can get it working.

PR #4695, opened 2022-11-14
Improve the dependabot configuration

LukeShu:
## Description

1. Generate `.github/dependabot.yml`, to reduce human strain
2. Teach GHA to have it run `make generate` on dependabot branches, so that humans don't need to edit them before merging.

We'll see if the GHA permissions are correct--I haven't tested that.

## Related Issues

(none)

## Testing

(none)

## Checklist

- [ ] **Does my change need to be backported to a previous release?** - dependabot doesn't run on old branches, so no

- [ ] **I made sure to update `CHANGELOG.md`.** - no applicable changes

  - Any new features
  - Any deprecations

- [ ] **This is unlikely to impact how Ambassador performs at scale.** - not a runtime change

- [ ] **My change is adequately tested.** - no

- [ ] **I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently.** - no new tricks


.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,41 @@
...
+# branches is pretty naive; it doesn't support us telling it that it
+# needs to run "make generate" after making an edit.
+#
+# But we can work around that by creating a workflow run "make
+# generate" on those branches!
+name: dependabot
+"on":
+  push:
+    branches:
+      - 'dependabot/**'
```
LanceEa, line 12: Should we limit this to only be on created so that it doesn't run if add additional commits on the PR?

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,41 @@
...
+
+jobs:
+  repatriate:
+    runs-on: ubuntu-latest
+    name: repatriate
```
LanceEa, line 17: bike shed: not sure this represents what it is doing and adds confusion since we removed repatriation from the workflow. Maybe just bluntly `generate` or `make-generate`

build-aux/generate.mk
```diff
@@ -35,6 +35,7 @@ generate/files      += $(OSS_HOME)/pkg/envoy-control-plane/  # recipe in _cxx/en
 # Individual files: Misc
 generate/files      += $(OSS_HOME)/DEPENDENCIES.md
 generate/files      += $(OSS_HOME)/DEPENDENCY_LICENSES.md
+generate-fast/files += .github/dependabot.yml
```
LanceEa, line 38: does this mean it is only re-generated when running `make generate-fast` or will it also be when we run `make generate`?

haq204: I believe both

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+      - uses: actions/checkout@v2
```
haq204, line 19: the current version is `actions/checkout@v3` now

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,41 @@
...
+          make generate
+
+          git add .
+          if test -z "$(git status -s)"; then
+            # already up-to-date
+            exit 0
+          fi
+
+          git commit --signoff --message='make go-mod-tidy && make generate'
+          git push origin
```
haq204, line 41: if the workflow is triggered on a push that doesn't cause `actions/checkout` to checkout in detached HEAD state using the commit that was just pushed?

build-aux/generate.mk
LukeShu: `make generate` does everything that `make generate-fast` does.

.github/workflows/dependabot.yml
LukeShu: Good question.

LukeShu:
v2:
```patch
diff --git a/.github/workflows/dependabot.yml b/.github/workflows/dependabot.yml
index 6e4d33142..248631ad1 100644
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -12,11 +12,10 @@ name: dependabot
       - 'dependabot/**'
 
 jobs:
-  repatriate:
+  generate:
     runs-on: ubuntu-latest
-    name: repatriate
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0
       - name: Install Deps
@@ -38,4 +37,4 @@ jobs:
           fi
 
           git commit --signoff --message='make go-mod-tidy && make generate'
-          git push origin
+          git push origin HEAD:${{ github.ref }}
```

.github/workflows/dependabot.yml
LukeShu: Updated.

.github/workflows/dependabot.yml
LukeShu: I've changed it to push `HEAD:${{ github.ref }}` so that it works either way.

.github/workflows/dependabot.yml
LukeShu: I've changed it to be named `generate`.

.github/workflows/dependabot.yml
haq204: > but dependabot updates its PRs I think, if a new version comes out before the original PR is merged

that is correct

.github/workflows/dependabot.yml
haq204: Looking at the CI run - yup looks it does checkout in detached HEAD state (good to know for the future)

the-wondersmith:
🍃 Housekeeping 🧹

PR #4700, opened 2022-11-16
deps: upgrade net and text pkg's in golang.org/x

LanceEa:
## Description

Upgrades packages to remove scan warnings for cve's:

golang.org/x/net - CVE-2022-27664
- only effected 1.19.0 and below,  we are currently on 1.19.3

golang.org/x/text - CVE-2022-32149
 - API not called in code base

## Related Issues
Keeping deps updated.

## Testing

CI is good enough

## Checklist

   - Yes, will backport to v2.5 branch too


AliceProxy (approved):

AliceProxy (commented):
Should add a changelog if it's in response to any CVE's (even if we're not impacted) otherwise I don't think it's necessary.

haq204 (approved):

LanceEa:
@AliceProxy - I left it off the change-log since it was me just being proactive to remove noise from scanners. If we do feel it is necessary later on then we can add it during release time and as part of doc updates.

PR #4706, opened 2022-11-19
deps: upgrade to envoy 1.24.1

LanceEa:
## Description

Updates Emissary-ingress to run on v1.24.1, the latest released version of Envoy. 

Two notable changes occurred:

1. LightStep and Envoy Maintainers removed support for the LightStep driver thus we can no longer support it. See below for recommended upgrade path
2. Envoy bug fix caused behavior change in Upstream Cluster Tracing naming which changes default naming for TracingService cluster

### LightStep Tracing Driver Upgrade Path

The LightStep team is a contributor to the Open Telemetry project and have decided to no longer support the native Envoy LightStep Tracing Driver. The Envoy Project maintainers do not want to support it either so the decision was made to remove support for it in favor of leveraging Open Telemetry Collector to forward other supported formats to LightStep. 

If using LightStep, users need to switch to using the Zipkin driver with the Open Telemetry Collector which can forward tracing data to LightStep. This needs to be done prior to upgrading to Emissary-ingress v3.4.0 to ensure no tracing data is dropped.

### Upstream Cluster Observability Bug Fix

In  commit https://github.com/envoyproxy/envoy/pull/23500/commits/230d0079899626700f71b5c09541a171285b1de5, Envoy 1.24 fixed a bug where the behavior for the tracing name created uses the `observabilityName` and not the `cluster` name.  The observabilityName is determined by using `cluster.alt_stat_name` if provided and then fallback to `cluster.name` as seen here 
https://github.dev/envoyproxy/envoy/blob/141f28cf628b488304015eae861b81ada293c833/source/common/upstream/upstream_impl.cc#L889

This means that spans produced for upstream clusters (child spans) will now use the correct logic rather than always being the cluster name. 

Now that this has been fixed and matches that behavior it caused an issue with the KAT tests for the `TracingService`. The `TracingService` will always have an `alt_stat_name` for its cluster because Emissary-ingress always populates one. We take the user provided `stats_name` from the `TracingService`  or else we fallback to using the `service`. The tests previously were asserting that the trace name matched cluster name but this is no longer true and instead should match service name. That has been fixed.

## Related Issues

N/A

## Testing

Pulled in custom commits on top of base envoy. Verified ability to build Envoy and that all Envoy tests pass and Emissary-ingress test passed.

I also pulled this into Edge Stack to make sure no issues there either.

Screen Shot 2022-11-19 at 12 19 57 AM

## Checklist


_cxx/envoy.mk
```diff
@@ -13,7 +13,8 @@ RSYNC_EXTRAS ?=
 
 # IF YOU MESS WITH ANY OF THESE VALUES, YOU MUST RUN `make update-base`.
```
haq204, line 14: out of curiosity what's the difference between `make update-base` and `make generate`

docs/releaseNotes.yml
```diff
@@ -86,6 +86,19 @@ items:
           This allows the healthcheck server to be configured to use IPv6-only k8s environments. 
           (Thanks to <a href="https://github.com/TimonOmsk">Dmitry Golushko</a>!).
 
+      - title: Upgrade to Envoy 1.24.1
+        type: feature
+        body: >-
```
haq204, line 91: perhaps you should link to Tenshin's new docs for OTel as well

haq204 (commented):
couple of minor comments but otherwise lgtm

docs/releaseNotes.yml
LanceEa: That isn't merged yet, so left it off. My plan was to do it to get it merged on getambassador.io.

_cxx/envoy.mk
LanceEa: I will do a session on this, but update-base will build envoy itself using the dev envoy build container.

tenshinhigashi (approved):

PR #4715, opened 2022-11-23
[CC] feat: add dedicated listener for ready check to be performed on envoy worker

LanceEa:
## Description

>Note: the only major changes from me was to rebase, fix some unit test due to rebase and to switch over to making logging of the ready endpoint to be turned off. This way users are bombarded with logs when they upgrade and will need to opt-in.

> I will squash the commits to clean-up the history.

As normal because our CI process needs some love, I have pulled in https://github.com/emissary-ingress/emissary/pull/4300. 
See that PR for details.

TL;DR;
This moves the `/ready` check to a dedicated Listener/Route rather than using Envoy's Admin `/ready` endpoint. This is recommend by the Envoy project and also done by other envoy based projects due to the fact that the Admin endpoint is blocking and can cause performance issues if xDS configuration and/or stats endpoint is being pinged at the same time. By moving it to a listener the ready check is done on an Envoy worker instead.

## Related Issues

See PR for details https://github.com/emissary-ingress/emissary/pull/4300

## Testing

KAT and unit tests have been added. I also pulled this into edge stack to verify all is good.

## Checklist

This will improve performance on large clusters because the ready check will no longer be blocking waiting on the `admin` server.


AliceProxy (approved):
seems fine

python/tests/kat/t_basics.py
```diff
@@ -30,6 +36,7 @@ def config(self) -> Generator[Union[str, Tuple[Node, str]], None, None]:
     def queries(self):
         yield Query(self.url("ambassador/v0/diag/?json=true&filter=errors"), phase=2)
         yield Query(self.url("_internal/v0/ping", scheme="http", port=8877), expected=403)
+        yield Query(self.url("ambassador/v0/check_ready", scheme="http", port=8877))
```
kflynn, line 39: Shouldn't this use port 8500?

kflynn (changes requested):
@LanceEa, I think there's one incorrect port number in a test? or am I mistaken?

python/tests/kat/t_basics.py
LanceEa: No, it can't because that Listener only runs on localhost (127.0.0.1) thus is not available to the `kat-client` pod where all the kat requests come from.

LanceEa: https://github.com/emissary-ingress/emissary/blob/master/cmd/entrypoint/healthcheck_server.go#L67

Copying from slack for context. It comes in on 8877 --> ambwatch.Ready check --> points at new Listener (127.0.0.1:8500) which is indirectly verifying two things:

1. that the new Listener is in fact working when configured with a different port
2. that diagd is able to process snapshots

kflynn: Ah right, I'm misremembering the flow through `entrypoint`. 🤦‍♂️

kflynn (approved):
Hit it! 🙂

PR #4719, opened 2022-11-28
deps: upgrade python dependencies

haq204:
## Description

Don't upgrade Kubernetes or orjson for now

Signed-off-by: Hamzah Qudsi <hqudsi@datawire.io>

## Related Issues

List related issues.

## Testing

## Checklist

  - Any new features
  - Any deprecations


LanceEa:
Can you pull this into AES and verify all is good?

Also, the one I'm probably most unsure about is the setup-tools (_just because I don't know what I don't know_). I thought we were pinned due to some major changes but CI is green so makes me think maybe it is not a big deal...does base python need bumped because of it?

haq204:
> Can you pull this into AES and verify all is good?
> 
> Also, the one I'm probably most unsure about is the setup-tools (_just because I don't know what I don't know_). I thought we were pinned due to some major changes but CI is green so makes me think maybe it is not a big deal...does base python need bumped because of it?

setuptools is used for packaging in python (i.e it's not used at runtime). no base python doesn't necessarily need to be bumped. You may be thinking of pip-tools.

LanceEa (approved):
lgtm

LanceEa:
Actually, this is what I was thinking of:

```shell
 DEPRECATION: expiringdict is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559
```

If you run `make python-dev-setup` it will install the venv and triggers this warning. I haven't really looked into this but the fact that I remember seeing this, I wasn't sure if upgrading setuptools would start causing some issues. Might be un-related but thought I would call it out.

tenshinhigashi (approved):

haq204:
> Actually, this is what I was thinking of:
> 
> ```shell
>  DEPRECATION: expiringdict is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559
> ```
> 
> If you run `make python-dev-setup` it will install the venv and triggers this warning. I haven't really looked into this but the fact that I remember seeing this, I wasn't sure if upgrading setuptools would start causing some issues. Might be un-related but thought I would call it out.

Ah ok. so yea upgrading setuptools won't cause this issue. What this deprecation notice is saying the setup.py installation method is going to go away in pip 23.1 (April 2023) and there is dependency `expiringdict` that doesn't provide pre-built wheels. 

So package distribution in python is really messy. setup.py install is a really really old way to distribute packages and libraries in python. More modern ways uses [wheels](https://pythonwheels.com/) from say a  `pyproject.toml`. setuptools provides some, well tools, to use in a  setup.py or as a build backend when using pyproject.toml. 

but for `expiringdict` specifically, we might not have to do much. they might start providing wheels in a future version. Otherwise we could add the `-use-pep517` flag so that it will build the wheels during installation. Either way we should track this an issue.

haq204:
https://github.com/emissary-ingress/emissary/issues/4723
https://github.com/emissary-ingress/emissary/issues/4724

Issue #4714, opened 2022-11-23
Mapping using hostname does not work


ericfuxealth:
**Describe the bug**
Using "hostname" in mapping does not work, but "host" works.

**To Reproduce**
1. Install emissary-ingress using helm chart: 
```
helm repo add datawire https://app.getambassador.io
helm repo update
kubectl apply -f https://app.getambassador.io/yaml/emissary/3.3.0/emissary-crds.yaml
kubectl wait --timeout=90s --for=condition=available deployment emissary-apiext -n emissary-system

helm install -n emissary --create-namespace \
     emissary-ingress datawire/emissary-ingress && \
 kubectl rollout status  -n emissary deployment/emissary-ingress -w

2. Notice deployment " emissary-ingress-agent" and "emissary-ingress" are using different version of the image: 
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: emissary-ingress-agent
  namespace: default
  labels:
    app.kubernetes.io/name: emissary-ingress-agent
    helm.sh/chart: emissary-ingress-8.3.0
    app.kubernetes.io/instance: emissary-ingress
    app.kubernetes.io/part-of: emissary-ingress
    app.kubernetes.io/managed-by: Helm
    product: aes
spec:
 ...
    spec:
      serviceAccountName: emissary-ingress-agent
      containers:
      - name: agent
        image: docker.io/emissaryingress/emissary:3.1.0
...
apiVersion: apps/v1
kind: Deployment
metadata:
  name: emissary-ingress
  namespace: default
  labels:
    app.kubernetes.io/name: emissary-ingress
    helm.sh/chart: emissary-ingress-8.3.0
    app.kubernetes.io/instance: emissary-ingress
    app.kubernetes.io/part-of: emissary-ingress
    app.kubernetes.io/managed-by: Helm
    product: aes
spec:
  replicas: 3
  selector:
    matchLabels:
      app.kubernetes.io/name: emissary-ingress
      app.kubernetes.io/instance: emissary-ingress
  strategy:
    type: RollingUpdate
  
  
  progressDeadlineSeconds: 600
  template:
    metadata:
      labels:
        app.kubernetes.io/name: emissary-ingress
        helm.sh/chart: emissary-ingress-8.3.0
        app.kubernetes.io/instance: emissary-ingress
        app.kubernetes.io/part-of: emissary-ingress
        app.kubernetes.io/managed-by: Helm
        product: aes
        profile: main
  ...
      containers:
[...]

PR #4761, opened 2022-12-14
Re-adding stand-alone agent

tenshinhigashi:
Signed-off-by: Tenshin Higashi <thigashi@datawire.io>

## Description

This PR re-adds the stand-alone agent that was removed due to regressions. Please note that this is NOT the best implementation of the stand-alone agent. 

A better solution would be to include the ambassador-agent as a dependency. 

## Related Issues

List related issues.

## Testing

## Checklist

  - Any new features
  - Any deprecations

### DIFF
--- a/charts/emissary-ingress/templates/_helpers.tpl
+++ b/charts/emissary-ingress/templates/_helpers.tpl
@@ -125,23 +125,3 @@ Define the http port of the Ambassador service
 {{- end -}}
 {{- end -}}
 {{- end -}}
-
-{{/*
-Set the image that should be used for the agent.
-Use agent.repository and agent.tag if present,
-Then use fullImageOverride if present,
-Then if the image repository is explicitly set, use "repository:image"
-*/}}
-{{- define "agent.image" -}}
-
-{{- if .Values.agent.image.repository -}}
-    {{- printf "%s:%s" .Values.agent.image.repository .Values.agent.image.tag -}}
-{{- else if .Values.image.fullImageOverride -}}
-    {{- .Values.image.fullImageOverride }}
-{{- else if hasKey .Values.image "repository" -}}
-    {{- printf "%s:%s" .Values.image.repository .Values.image.tag -}}
-{{- else -}}
-    {{- printf "%s:%s" "docker.io/emissaryingress/emissary" .Values.image.tag -}}
-{{- end -}}
-
-{{- end -}}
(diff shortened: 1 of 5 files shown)


LanceEa (commented):
Looks good. I think there was one other file that we can do clean-up on as well.

`charts/emissary-ingress/templates/_helpers.tpl` as seen here:

https://github.com/emissary-ingress/emissary/pull/4547/commits/aa70060c5977f647a9a30d2814470bfa6ac72470

charts/emissary-ingress/templates/_helpers.tpl
```diff
@@ -124,24 +124,4 @@ Define the http port of the Ambassador service
...
-{{- else if .Values.image.fullImageOverride -}}
-    {{- .Values.image.fullImageOverride }}
-{{- else if hasKey .Values.image "repository" -}}
-    {{- printf "%s:%s" .Values.image.repository .Values.image.tag -}}
-{{- else -}}
-    {{- printf "%s:%s" "docker.io/emissaryingress/emissary" .Values.image.tag -}}
-{{- end -}}
-
-{{- end -}}
+{{- end -}}
```
AliceProxy, line 127: nit: When you see this little Icon at the bottom of a file in the github changes view it means that there should be an empty new line at the end of the file. Try not to erase those on files you edit.

charts/emissary-ingress/templates/ambassador-agent.yaml
```diff
@@ -280,5 +259,4 @@ spec:
   selector:
     app.kubernetes.io/name: {{ include "ambassador.fullname" . }}-agent
     app.kubernetes.io/instance: {{ .Release.Name }}
-{{- end }}
-{{- end }}
+{{- end }}
```
AliceProxy, line 262: Same as my comment about the new line here

AliceProxy (changes requested):
Just my comment about the new lines but otherwise lgtm

charts/emissary-ingress/templates/_helpers.tpl
AliceProxy: If you use vscode then you can go look for the extension `EditorConfig.EditorConfig` and it will respect the contents of the `.editorconfig` file and do that automatically for you when you save. AFAIK most other editors can be made to respect `.editorconfig` as well, otherwise just follow the established patterns you see.

LanceEa (approved):
LGTM

LanceEa:
@alexgervais - I pinged you on this to make sure it looks good to you as well.

alexgervais (approved):

joebowbeer:
Fixes #4714 ?

PR #4779, opened 2023-01-04
[CC]charts: add startupProbe to emissary-ingress deployments

LanceEa:
## Description

I'm pulling in the a community-contribution for CI run. The original PR  can be found here https://github.com/emissary-ingress/emissary/pull/4649, with a rundown of what the PR does.

> emissary-ingress can be slow to start when the number of mappings grow. In that case, using a startupProbe is preferred over tweaking the initialDelaySeconds of the livenessProbe and readinessProbe.
> 
> startupProbe allow us to be generous with how long the pod takes to be ready, defined as `periodSeconds * failureThreshold`, but if it takes less time than allowed it will become ready faster (compared to having a high value for the initialDelaySeconds).
> 
> For example:
> - startupProbe set with periodSeconds=10 failureThreshold=30 = 5mins
> - livenessProbe/readinessProbe set with initialDelaySeconds=0
> 
> If the /check_ready endpoint returns OK after 2mins, the the startupProbe finish and the livenessProbe/readinessProbe immediatetly succeed, changing the pod state to live and ready.
> 
> If instead we didn't have the startupProbe but initialDelaySeconds was set to 300 (5min), the pod would always take 5min to be live and ready, even if /check_ready returns OK after 2mins.

## Related Issues

N/A

## Testing

CI and I will pull into AES to verify as well. General note to reviewers, we need to look over our helm chart tests and get a better handle on them, (in the future).

## Checklist

    The feature is opt-in so by default it will not get added unless a user actively attempts to add it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,13 @@ it will be removed; but as it won't be user-visible this isn't considered a brea
 
 ### Emissary-ingress and Ambassador Edge Stack
 
+- Change: emissary-ingress can be slow to start when there is a lot of mappings. Kubernetes
+  recommends using a startupProbe over tweaking the initialDelaySeconds of the livenessProbe and
+  readinessProbe. The Helm chart was updated to allow setting a startupProbe on the emissary-ingress
+  deployment. ([#4649])
+
+[#4649]: https://github.com/emissary-ingress/emissary/pull/4649
+
 ## [3.4.0] January 03, 2023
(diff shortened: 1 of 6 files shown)


haq204 (commented):
overall the changes lgtm however can we fix the change log and make sure its added in the chart change log?

LanceEa:
@haq204 - Can you do your PR review suggestions on the original, this PR is just being used to trigger CI.

haq204 (approved):

PR #4782, opened 2023-01-06
Image alpine-3.17_glibc-2.32 doesn't exist

bathina2:
## Description

Referring to a missing image in base-envoy.Dockerfile-stripped causes the build to `make update-base` to fail.

## Related Issues

List related issues.

## Testing

## Checklist

  - Any new features
  - Any deprecations

### DIFF
--- a/docker/base-envoy/Dockerfile.stripped
+++ b/docker/base-envoy/Dockerfile.stripped
@@ -16,7 +16,7 @@
 # If you have to change this file, you _must_ update BASE_ENVOY_RELVER
 # in the Makefile, then run "make update-base" to build and push the
 # new image.
-FROM frolvlad/alpine-glibc:alpine-3.17_glibc-2.32
+FROM frolvlad/alpine-glibc:alpine-3.17_glibc-2.34
 
 # We use rsync to move things in and out of the container.
 RUN apk add rsync


docker/base-envoy/Dockerfile.stripped
```diff
@@ -16,7 +16,7 @@
 # If you have to change this file, you _must_ update BASE_ENVOY_RELVER
 # in the Makefile, then run "make update-base" to build and push the
 # new image.
-FROM frolvlad/alpine-glibc:alpine-3.17_glibc-2.32
+FROM frolvlad/alpine-glibc:alpine-3.17_glibc-2.34
```
haq204, line 19: Looks like `BASE_ENVOY_RELVER` in `_cxx/Makefile` wasn't updated when this was last updated. Instead of changing the base image here, bump `BASE_ENVOY_RELVER` and try running `make update-base` again which should push a new version of the image.

haq204: my bad, you're right it should be updated to frolvlad/alpine-glibc:alpine-3.17_glibc-2.34, just also bump BASE_ENVOY_RELVER as well

LanceEa: @haq204 - yes this is needed to build envoy.

LanceEa (approved):

PR #4792, opened 2023-01-13
charts: added list verb for secrets to ambassador agent

jacoblbeck:
## Description

Added List verb to Secrets for the ambassador agent service account. The Ambassador Agent now needs the ability to list secrets not only `get` them.

## Related Issues

List related issues.

## Testing

## Checklist

  - Any new features
  - Any deprecations


charts/emissary-ingress/templates/ambassador-agent.yaml
```diff
@@ -47,7 +47,7 @@ rules:
   verbs: [ "get", "list", "watch" ]
 - apiGroups: [""]
   resources: [ "secrets"]
-  verbs: [ "get", "create", "delete", "patch", "watch"]
+  verbs: [ "get", "create", "delete", "patch", "watch", "list"]
```
haq204, line 50: just curious as to why the agent needs to list secrets and what secrets its listing?

Is there a new agent version that requires this?

jacoblbeck: Yeah theres a new agent that requires this for 2 reasons, there is a change in the k8s library we are using that lists secrets instead of getting a specific one, and also now the agent secret can come from multiple sources, either emissary/ES install or telepresence, which is prefixed differently, so the agent lists the secrets and filters on the specific suffix of the secret since the exact naming isn't the same

haq204: in that case you will also want to update the agent image version for the new version that requires this new action + update the chart changelog.

charts/emissary-ingress/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ numbering uses [semantic versioning](http://semver.org).
 
 - Upgrade Emissary to v3.5.0 [CHANGELOG](https://github.com/emissary-ingress/emissary/blob/master/CHANGELOG.md)
 - Adds opt-in settings to configure a startupProbe for the Emissary-ingress pod
+- Updates the ambassador agent to v1.0.8
```
haq204, line 10: you should also add that we update the rbac setttings for the agent

knlambert:
@jacoblbeck I recommand to NOT update the agent to the last version for now since it now also require lease permissions + a bug in the metric reporting.

knlambert:
@jacoblbeck ?

knlambert:
@jacoblbeck what is the status for this one ?

PR #4806, opened 2023-01-19
Pinning build dependency version

tenshinhigashi:
Signed-off-by: Tenshin Higashi <thigashi@datawire.io>

## Description

The purpose of this PR is to pin the `build` dependency to version 0.9.0. 

`build` version 0.10.0 omitted its license info from `pip show` which `py-mkopensource` relies on to generate `DEPENDENCIES.md`. Pinning `build` to an older version allows us to properly generate `DEPENDENCIES.md` on python 3.10. 

## Related Issues

List related issues.

## Testing

## Checklist

  - Any new features
  - Any deprecations

- [ x ] **This is unlikely to impact how Ambassador performs at scale.**

### DIFF
--- a/docker/base-python/Dockerfile
+++ b/docker/base-python/Dockerfile
@@ -75,7 +75,8 @@ RUN apk --no-cache add \
 # https://github.com/jazzband/pip-tools/#versions-and-compatibility to
 # select a pip-tools version that corresponds to the 'py3-pip' and
 # 'python3' versions above.
-RUN pip3 install pip-tools==6.12.1
+# Pinning build version due to missing license info from pip show in newer versions
+RUN pip3 install pip-tools==6.12.1 build==0.9.0
 
 RUN curl --fail -L https://dl.google.com/go/go1.19.4.linux-amd64.tar.gz | tar -C /usr/local -xzf -
 


docker/base-python/Dockerfile
```diff
@@ -75,7 +75,8 @@ RUN apk --no-cache add \
 # https://github.com/jazzband/pip-tools/#versions-and-compatibility to
 # select a pip-tools version that corresponds to the 'py3-pip' and
 # 'python3' versions above.
-RUN pip3 install pip-tools==6.12.1
+# Pinning build version due to missing license info from pip show in newer versions
+RUN pip3 install pip-tools==6.12.1 build==0.9.0
```
haq204, line 79: To give more context behind this specifically. `pip-tools` does not pin `build` so we're pinning the `build` package to ensure consistency. If you update `pip-tools` you should also update `build` as well. We could have also just updated `py-mkopensource` for the `build` license info but we could run into the same problem again when a new version of `build` is released. So now we deal with any license omissions when we are explicitly updating the version and not randomly.

haq204 (approved):

LanceEa (approved):

PR #4809, opened 2023-01-20
Add a feature to IngressProcessor to resolve named ports

### LINKED ISSUE #4322: upstream request timeout with ingress v1 when specifying backend.service.port.name
**Describe the bug**

When specifiying ` backend.service.port.name` instead of ` backend.service.port.number`, emissary ingress will not function:

```
HTTP/1.1 504 Gateway Timeout
content-length: 24
content-type: text/plain
date: Fri, 08 Jul 2022 01:45:49 GMT
server: envoy

upstream request timeout
```

**To Reproduce**

1. create a simple pod/service/ingress.

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: test
spec:
  ingressClassName: ambassador
  rules:
  - host: test.example.com
    http:
      paths:
      - backend:
          service:
            name: my-service
            port:
              number: 8080 <<<<<< here
        path: /
        pathType: Prefix
```

This will work fine. Now change the `port.number` and use `port.name` in the service instead. emissary ingress will no longer work for that ingress.

**Expected behavior**

Service backends should work with either `port.number` and `port.name`.

 - Ambassador: 3.0.0
 - Kubernetes environment: openshift v4.10
 - Version k8s version 1.23

**Additional context**

I found a TODO here(https://github.com/emissary-ingress/emissary/blob/master/pkg/kates/k8s_resource_types/ingress.go#L20), not sure if it is causing this problem.

As k8s v1.23+ has deprecated and dropped support of ingress v1beta, supporting ingress v1 is essential.

antonu17:
## Problem statement

The Ingress processor ignores the named service port and replaces it with `80`. 

## Related Issues

closes #4322

## Testing

Added a test case to `test_fetcher.py`

## Checklist

  - would be nice to backport this to v2.4 as well (#4810)

### DIFF
--- a/python/ambassador/fetch/dependency.py
+++ b/python/ambassador/fetch/dependency.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 from typing import (
     Any,
     Collection,
+    Dict,
     Iterator,
     Mapping,
     MutableSet,
@@ -13,7 +14,7 @@ from typing import (
     TypeVar,
 )
 
-from .k8sobject import KubernetesObject
+from .k8sobject import KubernetesObjectKey, KubernetesObject
 
 
 class Dependency(Protocol):
@@ -35,9 +36,11 @@ class ServiceDependency(Dependency):
     """
 
     ambassador_service: Optional[KubernetesObject]
+    discovered_services: Dict[KubernetesObjectKey, KubernetesObject]
 
     def __init__(self) -> None:
         self.ambassador_service = None
+        self.discovered_services = {}
 
     def watt_key(self) -> str:
         return "service"
--- a/python/ambassador/fetch/dependency.py
+++ b/python/ambassador/fetch/dependency.py
@@ -14,7 +14,7 @@ from typing import (
     TypeVar,
 )
 
-from .k8sobject import KubernetesObjectKey, KubernetesObject
+from .k8sobject import KubernetesObject, KubernetesObjectKey
 
 
 class Dependency(Protocol):
--- a/python/ambassador/fetch/ingress.py
+++ b/python/ambassador/fetch/ingress.py
@@ -104,19 +104,19 @@ class IngressProcessor(ManagedKubernetesProcessor):
                 f"Not reconciling Ingress {obj.name}: observed and current statuses are in sync"
             )
 
-    def _resolve_service_port_number(self, namespaceName, serviceName, servicePort):
+    def _resolve_service_port_number(self, namespace, service_name, service_port):
         for k8s_svc in self.service_dep.discovered_services.values():
             key = f'{k8s_svc.name}.{k8s_svc.namespace}'
 
-            if key != f'{serviceName}.{namespaceName}':
+            if key != f'{service_name}.{namespace}':
                 continue
 
             for port in k8s_svc.spec.get('ports', []):
-                if servicePort == port.get('name', None):
-                    return port.get('port', servicePort)
-            self.logger.debug(f"Could not resolve port '{servicePort}' for service '{key}'")
+                if service_port == port.get('name', None):
+                    return port.get('port', service_port)
+            self.logger.debug(f"Could not resolve port '{service_port}' for service '{key}'")
 
-        return servicePort
+        return service_port
 
     def _process(self, obj: KubernetesObject) -> None:
         ingress_class_name = obj.spec.get("ingressClassName", "")
--- a/python/ambassador/fetch/ingress.py
+++ b/python/ambassador/fetch/ingress.py
@@ -104,7 +104,7 @@ class IngressProcessor(ManagedKubernetesProcessor):
                 f"Not reconciling Ingress {obj.name}: observed and current statuses are in sync"
             )
 
-    def _resolve_service_port_number(self, namespace, service_name, service_port):
+    def _try_resolve_service_port_number(self, namespace, service_name, service_port):
         self.logger.debug(f"Resolving named port '{service_port}' in service '{service_name}'")
 
         key = KubernetesObjectKey(KubernetesGVK("v1", "Service"), namespace, service_name)
@@ -241,7 +241,7 @@ class IngressProcessor(ManagedKubernetesProcessor):
                 try:
                     service_port = int(service_port)
                 except:
-                    service_port = self._resolve_service_port_number(
+                    service_port = self._try_resolve_service_port_number(
                         obj.namespace, service_name, service_port
                     )
 
--- a/python/ambassador/fetch/ingress.py
+++ b/python/ambassador/fetch/ingress.py
@@ -104,6 +104,20 @@ class IngressProcessor(ManagedKubernetesProcessor):
                 f"Not reconciling Ingress {obj.name}: observed and current statuses are in sync"
             )
 
+    def _resolve_service_port_number(self, namespaceName, serviceName, servicePort):
+        for k8s_svc in self.service_dep.discovered_services.values():
+            key = f'{k8s_svc.name}.{k8s_svc.namespace}'
+
+            if key != f'{serviceName}.{namespaceName}':
+                continue
+
+            for port in k8s_svc.spec.get('ports', []):
+                if servicePort == port.get('name', None):
+                    return port.get('port', servicePort)
+            self.logger.debug(f"Could not resolve port '{servicePort}' for service '{key}'")
+
+        return servicePort
+
     def _process(self, obj: KubernetesObject) -> None:
         ingress_class_name = obj.spec.get("ingressClassName", "")
 
@@ -220,6 +234,11 @@ class IngressProcessor(ManagedKubernetesProcessor):
                 service_name = path_backend.get("serviceName", None)
                 service_port = path_backend.get("servicePort", None)
                 path_location = path.get("path", "/")
+
+                try:
+                    service_port = int(service_port)
+                except:
+                    service_port = self._resolve_service_port_number(obj.namespace, service_name, service_port)
 
                 if not service_name or not service_port or not path_location:
                     continue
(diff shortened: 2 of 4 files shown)


python/ambassador/fetch/ingress.py
```diff
@@ -221,6 +235,11 @@ def _process(self, obj: KubernetesObject) -> None:
                 service_port = path_backend.get("servicePort", None)
                 path_location = path.get("path", "/")
 
+                try:
+                    service_port = int(service_port)
+                except:
+                    service_port = self._resolve_service_port_number(obj.namespace, service_name, service_port)
```
haq204, line 241: if `_resolve_service_port_number` fails to resolve to a port number then it would return the port name if I'm reading the above method correctly right?

So later on it would default to port 80?

haq204 (commented):
thanks for the addressing this!

Just have one question.

python/ambassador/fetch/ingress.py
antonu17: that's right, in case the port can't be resolved `_resolve_service_port_number` returns the same `service_port` value. This is possible in a situation if `Service` name specified in the `Ingress` spec doesn't exist, or if the `Service` doesn't have a port with the name provided in the `Ingress`. Both cases are not liable for any ingress controller.

haq204:
@antonu17 I've opened https://github.com/emissary-ingress/emissary/pull/4813 so that CI can run. Our CI is a bit funky in that only maintainers have the proper permissions to run it. 

It appears the only issue is linting. The easiest way to fix that would be to run `make format`

antonu17:
> @antonu17 I've opened #4813 so that CI can run. Our CI is a bit funky in that only maintainers have the proper permissions to run it.
> 
> It appears the only issue is linting. The easiest way to fix that would be to run `make format`

Got it, pushed `make format` changes

haq204:
> > @antonu17 I've opened #4813 so that CI can run. Our CI is a bit funky in that only maintainers have the proper permissions to run it.
> > It appears the only issue is linting. The easiest way to fix that would be to run `make format`
> 
> Got it, pushed `make format` changes

Looks like mypy is still not happy...
```
python/tests/unit/test_fetch.py:331: error: Argument 1 to "len" has incompatible type "Optional[Dict[str, ACResource]]"; expected "Sized"
python/tests/unit/test_fetch.py:333: error: Item "None" of "Optional[Dict[str, ACResource]]" has no attribute "values"
```

antonu17:
@haq204 Thanks for re-trying CI, i see it still fails with:

```
python/tests/unit/test_fetch.py:331: error: Argument 1 to "len" has incompatible type "Optional[Dict[str, ACResource]]"; expected "Sized"
python/tests/unit/test_fetch.py:333: error: Item "None" of "Optional[Dict[str, ACResource]]" has no attribute "values"
Found 2 errors in 1 file (checked 189 source files)
```

I think I understand why. Pushed another commit that fixes it.

haq204:
> @haq204 Thanks for re-trying CI, i see it still fails with:
> 
> ```
> python/tests/unit/test_fetch.py:331: error: Argument 1 to "len" has incompatible type "Optional[Dict[str, ACResource]]"; expected "Sized"
> python/tests/unit/test_fetch.py:333: error: Item "None" of "Optional[Dict[str, ACResource]]" has no attribute "values"
> Found 2 errors in 1 file (checked 189 source files)
> ```
> 
> I think I understand why. Pushed another commit that fixes it.

@antonu17 thanks! I updated the other PR for changelog updates and rebased it onto master. CI is green there so once it gets approvals from the other maintainers it should be good to merge.

> @haq204 Thanks for re-trying CI, i see it still fails with:
> 
> ```
> python/tests/unit/test_fetch.py:331: error: Argument 1 to "len" has incompatible type "Optional[Dict[str, ACResource]]"; expected "Sized"
> python/tests/unit/test_fetch.py:333: error: Item "None" of "Optional[Dict[str, ACResource]]" has no attribute "values"
> Found 2 errors in 1 file (checked 189 source files)
> ```
> 
> I think I understand why. Pushed another commit that fixes it.

antonu17:
> @antonu17 thanks! I updated the other PR for changelog updates and rebased it onto master. CI is green there so once it gets approvals from the other maintainers it should be good to merge.

@haq204 thanks for your review. should I close this PR?

python/ambassador/fetch/ingress.py
```diff
@@ -104,6 +104,20 @@ def _update_status(self, obj: KubernetesObject) -> None:
                 f"Not reconciling Ingress {obj.name}: observed and current statuses are in sync"
             )
 
+    def _resolve_service_port_number(self, namespace, service_name, service_port):
+        for k8s_svc in self.service_dep.discovered_services.values():
```
LanceEa, line 108: discovered_services is a dict, can't we just use `self.service_dep.discovered_services.get(f"{service_name}.{namespace}", None)`? 

If the cluster has a larger number of services we potentially would have to do a lot of looping here.

LanceEa (changes requested):
@antonu17 @haq204 - Thanks for this and this will be super useful for folks using Ingress resources. 

I left a few notes on the code and test. I need familiarize myself a little more with this part of the code base so some of my comments/questions might be obvious once I dive deeper into it.

python/ambassador/fetch/dependency.py
```diff
@@ -35,9 +36,11 @@ class ServiceDependency(Dependency):
     """
 
     ambassador_service: Optional[KubernetesObject]
+    discovered_services: Dict[KubernetesObjectKey, KubernetesObject]
```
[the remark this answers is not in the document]
antonu17: I saw this code for the first time in my live this week xD, sorry in advance if I misunderstood something.

So `ambassador_service` seems to be ambassador's loadbalancer service which is used for referencing in Ingress status info.
And `discovered_services` dictionary used to be a field of the `InternalServiceProcessor`, but I moved it to `ServiceDependency` class to make it available for IngressProcessor for port lookups.

python/ambassador/fetch/ingress.py
antonu17: My bad 🤦  will update soon

python/tests/unit/test_fetch.py
```diff
@@ -278,6 +279,64 @@ def test_mapping_v1(self):
         assert mapping.prefix == valid_mapping_v1.spec["prefix"]
         assert mapping.service == valid_mapping_v1.spec["service"]
 
+    def test_ingress_with_named_port(self):
```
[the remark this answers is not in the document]
antonu17: Alright, I'll add more test cases to address these additional cases

antonu17:
@LanceEa, @haq204 I addressed all PR comments and updated the code and tests. have a look once more please.

python/ambassador/fetch/ingress.py
```diff
@@ -105,17 +105,20 @@ def _update_status(self, obj: KubernetesObject) -> None:
...
 
-            for port in k8s_svc.spec.get("ports", []):
-                if service_port == port.get("name", None):
-                    return port.get("port", service_port)
-            self.logger.debug(f"Could not resolve port '{service_port}' for service '{key}'")
+        for port in k8s_svc.spec.get("ports", []):
+            if service_port == port.get("name", None):
+                return port.get("port", service_port)
 
+        self.logger.debug(f"Could not find port '{service_port}' in service '{service_name}'")
```
haq204, line 121: since line 119 also could also return the service port, this debug log wouldn't always get outputted if `service_port` is returned

haq204 (commented):
one nit but otherwise lgtm once @LanceEa takes a look

python/ambassador/fetch/ingress.py
antonu17: Correct. If the port is found we don't need to display "not found" message

python/tests/unit/test_fetch.py
```diff
@@ -309,33 +309,74 @@ def test_ingress_with_named_port(self):
...
+
+        mapping_health = mappings.get("quote-0-2")
+        assert mapping_health
+        assert mapping_health.prefix == "/health"
+        assert mapping_health.service == "quote.default:9000"
+
+        mapping_missed_name = mappings.get("quote-0-3")
+        assert mapping_missed_name
+        assert mapping_missed_name.prefix == "/missed-name"
+        assert mapping_missed_name.service == "missed.default:missed"
```
LanceEa, line 374: maybe I'm missing something but aren't we trying to resolve the names to port numbers here? 

I would expect this one to be rejected and dropped or fallback to something, same with the metrics above too.

antonu17: Hello, @LanceEa! 
The fallback behavior is to retain the named port in the `aconf` mapping structure (this is how the ingress fetcher works at the moment). The bad port will be replaced to 80/443 later here: https://github.com/emissary-ingress/emissary/blob/aee9d485917365a2efa53ce90f121483da5c95a0/python/ambassador/ir/ircluster.py#L212-L215

antonu17: The point of these additional ingress fetcher tests is to demonstrate that port number resolution works whenever it is possible: referenced service and ports exist. In case of ingress' paths reference non-existing services or ports fetcher behaves as before.
Unfortunately, there were no tests for ingress fetcher before, so there's nothing to compare these tests with.
One thing I can do to better emphasize that the port number resolution is not guaranteed is to add the word `try` to the name of the method.

LanceEa: ahh got it. Like I mentioned, I wasn't as familiar with this part of the codebase 😄. My initial gut instinct was to assume that the resolution would have happened at this stage rather than delaying until later stages. So, LGTM

LanceEa (approved):

haq204:
@antonu17 can you rebase onto `main`? Once that is done we can run CI and get it merged

haq204:
@antonu17 CI in https://github.com/emissary-ingress/emissary/pull/4813 is green and has been approved and merged.

I'm going to close this one out. 

Thanks again for your contribution!

PR #4810, opened 2023-01-20
backport: IngressProcessor named ports support to v2.4

antonu17:
## Description

Backport #4809 to v2.4 release branch


LanceEa:
@antonu17 - I'm going to close this PR because we landed this in v2.5 and will only be landing fixes like this on the latest minor version. 

Thanks again for helping with this and getting this landed!

PR #4828, opened 2023-02-03
CI for #4630

kflynn:
CI for @dynajoe's PR #4630

### DIFF
--- a/pkg/ambex/transforms.go
+++ b/pkg/ambex/transforms.go
@@ -3,6 +3,9 @@ package ambex
 import (
 	// standard library
 	"context"
+	"crypto/md5"
+	"encoding/hex"
+	"encoding/json"
 	"fmt"
 
 	// third-party libraries
@@ -89,7 +92,7 @@ import (
 //                 "config_source": {
 //                   "ads": {}
 //                 },
-//                 "route_config_name": "ambassador-listener-8443-routeconfig-0"
+//                 "route_config_name": "ambassador-listener-8443-routeconfig-376bf87fb310abb282f452533940481d-0"
 //               }
 //             }
 //           }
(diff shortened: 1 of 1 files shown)


pkg/ambex/transforms.go
```diff
@@ -135,10 +142,20 @@ func V3ListenerToRdsListener(lnr *v3listener.Listener) (*v3listener.Listener, []
 					if rc.Name == "" {
 						// Generate a unique name for the RouteConfiguration that we can use to
 						// correlate the listener to the RDS record. We use the listener name plus
-						// an index because there can be more than one route configuration
+						// filter_chain_match hash because there can be more than one route configuration
 						// associated with a given listener.
-						rc.Name = fmt.Sprintf("%s-routeconfig-%d", l.Name, len(routes))
+						filterChainMatch, _ := json.Marshal(fc.GetFilterChainMatch())
```
LanceEa, line 147: @kflynn - Is there any downside to using the `json` package vs `protojson`. Based on the docs it says it can produce a different output https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson. Which probably in this scenario isn't important since it just being used for a hash. Assuming between go-versions it doesn't change.

again, I don't think there is an issue here but just asking for curiosity sake.

kflynn: Ugh, I missed this. I don't believe that this is a problem: the `protojson` package is mostly about supporting protobuf-typed-stuff, and this isn't.

LanceEa: No worries :)

LanceEa:
@kflynn - I would like to see a test added to this. I think its an easy one to add a test since it is a pure function without a lot of deps.

What do you think if we flag as a good candidate for the ContribFest at Kubecon EU?

PR #4830, opened 2023-02-05
deps: update k3d and k3s

LanceEa:
## Description

This updates k3d to the latest version and fixes args that were renamed in v5 of k3d. Also, updated to the latest minor release of k3s 1.22.Y.

As part of this, I looked into the KAT flakiness that we see. I have landed a few additional fixes to address this:

- introduced startupProbes to `emissary-apiext` to ensure it has time to generate self-signed certificate before running readiness check so that it doesn't needlessly restart pods
- ensure Kat test wait for `emisary-apiext` to be ready before trying to apply yaml
- capture kat-client logs in github actions to help with debugging
- add some additional logs lines
- add `--egress-selector-mode=disabled` to k3d to get around bug in k3s 1.22 (note: 1.23 this is fixed see commit messages for details_)

## Related Issues

General keeping up with the deps updated and improve our QOL of CI and kat test flakiness.

## Testing

Locally tested creating and tearing down cluster. Also, CI is green.

## Checklist

### DIFF
--- a/docs/releaseNotes.yml
+++ b/docs/releaseNotes.yml
@@ -66,6 +66,16 @@ items:
         - title: "#4809"
           link: https://github.com/emissary-ingress/emissary/pull/4809
 
+      - title: Add starupProbe to emissary-apiext server
+        type: change
+        body: >-
+          The <code>emissary-apiext</code> server is a Kubernetes Conversion Webhook that converts between the 
+          Emissary-ingress CRD versions. On startup, it ensures that a self-signed cert is available
+          so that K8s API Server can talk to the conversion webhook (*TLS is required by K8s*). We 
+          have introduced a startupProbe to ensure that emissary-apiext server has enough time to
+          configure the webhooks before running liveness and readiness probes. This is to ensure 
+          slow startup doesn't cause K8s to needlessly restart the pod.
+
   - version: 3.4.0
     prevVersion: 3.3.0
     date: '2023-01-03'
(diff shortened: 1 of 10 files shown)


AliceProxy (approved):

docs/releaseNotes.yml
```diff
@@ -66,6 +66,16 @@ items:
         - title: "#4809"
           link: https://github.com/emissary-ingress/emissary/pull/4809
 
+      - title: Add starupProbe to emissary-apiext server
```
haq204, line 69: should we add this in the chart changelog too?

python/kat/harness.py
```diff
@@ -1616,6 +1616,15 @@ def _setup_k8s(self, selected):
             ):
                 raise RuntimeError("Failed applying CRDs")
 
+            print("waiting for emissary-apiext server to become available")
+            if os.system(
```
haq204, line 1620: there seems to be a utility function `ShellCommand.run_with_retry` already. Why not use that to keep consistency?

docs/releaseNotes.yml
LanceEa: No because its not part of Helm, it is part of the crds and apiext yaml that is generated. This is why I put it here and not in the chart changelog.

python/kat/harness.py
LanceEa: It is consistent with what is there in that file. If you look at that file we call `os.system` many times. Even a few lines below that. The `ShellCommand.run_with_retry` is currently only used in scenarios where we want a retry mechanism. 

In this scenario, I don't think we need a retry mechanism because I kept the existing applying mapping retry. So, between the startup probe and the mapping apply retry if it still errors at that point then more than likely something is actually wrong.

haq204 (approved):

PR #4837, opened 2023-02-07
[v3.4.1] release pr

LanceEa:
## Description

This pulls in latest z patch for Envoy 1.24.2 which incudes a patch for the BoringSSL and also includes an update to the c-ares dependency to address an issue with upstream wildcard dns resolution.

## Related Issues

N/A

## Testing

CI and smoke testing.

## Checklist

Not a direct backport but we need to make sure this lands in master and also need to research if v2.5 needs patching too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,15 @@ it will be removed; but as it won't be user-visible this isn't considered a brea
 
 ## RELEASE NOTES
 
+## [3.4.1] February 07, 2023
+[3.4.1]: https://github.com/emissary-ingress/emissary/compare/v3.4.0...v3.4.1
+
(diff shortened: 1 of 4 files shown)


haq204 (approved):
what happened to updating the base envoy image? Did that end up not being required?

AliceProxy (approved):

LanceEa:
> what happened to updating the base envoy image? Did that end up not being required?

Are you meaning the Envoy commit ID or the `frolvlad/alpine-glibc:alpine-3.12_glibc-2.32`? The former is the new build of envoy with the patch and the latter is still available since `release/v3.4` was based on alpine-3.12. I will have to do another build tomorrow bumping the `relver` since we need a unique tag to generate an alpine 3.17 base envoy container.

PR #4838, opened 2023-02-07
Fix build failures when dependabot updates a dependency

aosoriodw:
## Description

This PR adds a new job to the `job-promote-to-passed` workflow that will check if a build is running because of a dependabot change. This job checks if `make generate` made changes to DEPENDENCIES.md/DEPENDENCY_LICENSES.md, and will commit them and start a new build.

This will simplify merging PRs created by dependabot, that would otherwise fail because dependency information is out of date.

If the job detects changes to DEPENDENCIES.md/DEPENDENCY_LICENSES.md it will commit these files, as well as go.mod and go.sum, if applicable.

go.mo and go.sum can change because `go-mkgenerate` will update them if necessary. For example, when a module is updated, a package may be removed from the new version, and if other dependency uses the removed package, `go build` will fail.  `go-mkgenerate` fixes this by updating any dependencies that use removed packages.

Note: The workflow `job-promote-to-passed` was updated to run on `workflow_dispatch` because this is required for builds to be triggered after the action updates the dependency information.

## Related Issues

N/A

## Testing

1. [Test](https://github.com/emissary-ingress/emissary/actions/runs/4126864912/jobs/7129344822) that workflow `job-promote-to-passed` fails if dependencies files are out of date and committer is not dependabot.
2. Test that when committer is dependabot, dependency information is updated and committed to the repo. This [workflow](https://github.com/emissary-ingress/emissary/actions/runs/4126874706/jobs/7129368869) is when dependencies are updated, and [this](https://github.com/emissary-ingress/emissary/actions/runs/4126972240) one is when it runs again with the latest changes from the repo.

## Checklist

  - Any new features
  - Any deprecations


go.sum
```diff
@@ -221,8 +221,12 @@ github.com/datawire/dlib v1.3.0 h1:KkmyXU1kwm3oPBk1ypR70YbcOlEXWzEbx5RE0iRXTGk=
...
 github.com/datawire/dtest v0.0.0-20210928162311-722b199c4c2f h1:Yz+xVYWkSARvfJ0e3LWwF1N5F+zJI9pCZ9R41OtkrKE=
 github.com/datawire/dtest v0.0.0-20210928162311-722b199c4c2f/go.mod h1:D/jb71PHlfUjfNSUVpXQAscXsXy3x+CY+c52lPnW2qU=
-github.com/datawire/go-mkopensource v0.0.0-20220218163159-cc298d9fabc4 h1:brHMnjdzg0cFM8HGwXyYhpEK1esLHPmrFu1N/cRFRPA=
-github.com/datawire/go-mkopensource v0.0.0-20220218163159-cc298d9fabc4/go.mod h1:u8J6vyaKZZSKl/sFtiQLV8I6n8xqDdPOXdY9mAQzmNg=
+github.com/datawire/go-mkopensource v0.0.2 h1:r7b/jtIQPF0pvwrE+LiIiLBg6upuNsKRSxXSkygviiA=
+github.com/datawire/go-mkopensource v0.0.2/go.mod h1:6WIetr8QL8XpzFUjJAb6kA/yQALBOGxUCd10UXEQwyY=
+github.com/datawire/go-mkopensource v0.0.3 h1:YZUG3HXMRewrULt8c+it9dbJSTj5jFfeKZ6AsX0SQqk=
+github.com/datawire/go-mkopensource v0.0.3/go.mod h1:6WIetr8QL8XpzFUjJAb6kA/yQALBOGxUCd10UXEQwyY=
+github.com/datawire/go-mkopensource v0.0.4 h1:l7FHSMjY4PjI/GTIHuf7otS3jYkBFVtj6+oh7dKJsX8=
+github.com/datawire/go-mkopensource v0.0.4/go.mod h1:6WIetr8QL8XpzFUjJAb6kA/yQALBOGxUCd10UXEQwyY=
```
AidanWenzel, line 229: Why does this include so many versions?

AidanWenzel (changes requested):

go.sum
aosoriodw: I've removed old references.

LanceEa: The `go.sum` shouldn't be manually manipulated and it is to ensure that what runs locally also runs the same in CI. It is maintained by `go mod tidy` or in the case of emissary-ingress we use `make go-mod-tidy` because it takes care of running it recursively in multiple directories.

Here is an explanation on the reason for multiple entries in the file: https://stackoverflow.com/questions/69825533/why-does-go-sum-include-so-many-older-packages

.github/workflows/execute-tests-and-promote.yml
```diff
@@ -55,6 +56,16 @@ jobs:
         shell: bash
         run: |
           make generate
+      - name: "Update dependency information after dependabot change"
+        uses: datawire/go-mkopensource/actions/save-dependabot-changes@v0.0.4
+        id: changed-by-dependabot
+        with:
+          branches_to_skip: master
```
LanceEa, line 63: Branch protection should prevent this but good to be explicit that we don't want any automated commits landing on master with our current setup.

LanceEa (approved):
Thanks! lgtm

haq204 (approved):

aosoriodw:
Updated the action and `go-mkopensource` tool to the latest version to fix an issue that could make the license info and go.mod go out of sync.

haq204 (approved):

LanceEa (approved):

PR #4869, opened 2023-02-23
py-mkopensource: Support license classifiers

LukeShu:
## Description
    
The "License" field (which `py-mkopensource` has been using) has been deprecated for a long time; packages are supposed to use Trove "Classifiers".  It's been working "good enough" because packages have been slow to switch, and when they do they usually set both.  But it's becoming more and more of a problem.

So yeah, implement parsing of license classifiers.

This depends on cleanup in https://github.com/datawire/go-mkopensource/pull/52.  At the time of this writing, this PR gets around that with `replace github.com/datawire/go-mkopensource => github.com/LukeShu/go-mkopensource v0.0.0-20230222225849-63083f3d8800`, but it shouldn't be merged that way. Once https://github.com/datawire/go-mkopensource/pull/52 lands, I'll rebase this PR not have the `replace`. But it can still be reviewed/approved before then.

cc: @tenshinhigashi for the change in the `Dockerfile` in the last commit.

## Related Issues

I woulda sworn that about a year ago I wrote up an issue about license classifiers, but I can't find it. Maybe it was in apro.git?

## Testing

I checked that many of the special-cases in `py-mkopensource` are no longer necessary.

## Checklist

- [x] **I made sure to update `CHANGELOG.md`.** - no applicable changes

  - Any new features
  - Any deprecations

- [x] **This is unlikely to impact how Ambassador performs at scale.** - not a runtime change

- [x] **My change is adequately tested.** - I'd say so

- [x] **I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently.** - no tricks


LanceEa:
@aosoriodw  and team have been working on go-mkopensource and I don't have much opinion in this area so I would lean on working with him. 

Once that lands, I can give this one a better look over but general speaking it seems fine.

the-wondersmith:
🍃 Housekeeping 🧹

PR #4912, opened 2023-03-07
helm: Added support for HPA behavior

styler:
## Description

This pull request adds support for Horizontal Pod Autoscaler (HPA) behavior.

https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#configurable-scaling-behavior

## Related Issues

List related issues.

## Testing

Manual testing was conducted to ensure that the new HPA behavior was working as expected.

## Checklist

  - Any new features
  - Any deprecations

- [ X] **The changes in this PR have been reviewed for security concerns and adherence to security best practices.**


LanceEa:
@styler - thanks for the PR, would you mind writing up an issue first outlining the reasoning for wanting to add this.

Also, I took a quick glance at the link and it looks like this feature wasn't stable until 1.23 and we support back to `1.22` still so I just want to make sure we are handling this in backwards compatible way.

styler:
@LanceEa For sure!

NikitaCloudRuntime:
Hello 
We're also interested in this feature :)
@styler did you have any chance you work on it and adding the issue mentioned in the comment above?

LanceEa:
@styler - were you able to write up an issue on this topic? The key thing for me is outlining the plan for handling backwards compatibility since this API didn't become stable until 1.23 and we still maintain support back to 1.21.

PR #4929, opened 2023-03-21
Remove kubectl dependency from images

rick-a-lane-ii:
Removes all use of kubectl from images.

## Description

Including the `kubectl` binary in the Emissary image is no longer necessary, and it would be better to remove it completely.

## Testing

* Built and deployed the full environment locally. Confirmed the local environment was fully functional.
* Manually confirmed that the `kubectl` binary was no longer present in the Emissary image. 
* Ran the full suite of tests.

## Checklist

- [X] **Does my change need to be backported to a previous release?**

  No.

  Unnecessary for this change.

  This change will not impact performance in any way.

  Unnecessary for this change.

### DIFF
--- a/build-aux/Dockerfile
+++ b/build-aux/Dockerfile
@@ -128,7 +128,6 @@ RUN adduser ambassador -u 8888 -G root -D -H -s /bin/false
 # External stuff that should change infrequently
 RUN apk --no-cache add bash curl python3=${py_version} libcap htop
 RUN apk upgrade --no-cache
-COPY --from=artifacts /usr/bin/kubectl /usr/bin/kubectl
 COPY --from=artifacts /usr/lib/libyaml* /usr/lib/
 
 # Other installers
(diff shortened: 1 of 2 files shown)


rick-a-lane-ii:
> @rick-a-lane-ii you should have permissions to upload directly to emissary-ingress.git and not have to open a PR from your fork (our CI doesn't work on community forks due to permission issues though @LanceEa has been looking into fixing that).

Yep, I'll close this out and open a PR normally against the main repo. I wanted to specifically adhere to the community-based process for one of these PRs.

> Also can you update https://github.com/emissary-ingress/emissary/blob/master/docs/releaseNotes.yml for the change? when you update `docs/releaseNotes.yml`, you will have to run `make generate` which will generate the new `CHANGELOG.md`

I went back and forth on whether this should have an updated CHANGELOG or not, because it doesn't feel like this should have been an advertised/inherent feature in the first place, and it feels a bit odd to announce that it no longer supports it. But no matter, I will update accordingly.

PR #4938, opened 2023-03-29
Hqudsi/envoy 1.25 upgrade

haq204:
## Description

## Related Issues

List related issues.

## Testing

## Checklist

  - Any new features
  - Any deprecations


LanceEa:
@haq204 - I'm guessing you fixed the conflicts during the cherry-pick. Do you recall what changes were made to the ResponseMap filter...was there a refactor due to renamed functions or what? Might be good to capture that context for future us.

Are you doing a separate PR for the go-control-plane update? 


Just to be clear on the CVE's listed in the commit message and PR description. We are not affected by these because these were only relevant if using the Kafka filter and wasm filters which we do not use, correct?

haq204:
> @haq204 - I'm guessing you fixed the conflicts during the cherry-pick. Do you recall what changes were made to the ResponseMap filter...was there a refactor due to renamed functions or what? Might be good to capture that context for future us.

mostly changing function signatures for new types

> Are you doing a separate PR for the go-control-plane update?
> 
Looks like the updated protos didn't require any changes to be made in emissary. which is what I was really referring to 

> Just to be clear on the CVE's listed in the commit message and PR description. We are not affected by these because these were only relevant if using the Kafka filter and wasm filters which we do not use, correct?

Correct we're not affected by any of them just adding them in for context

LanceEa:
Ok gotcha, I have been also bumping the go-control-plane with envoy version bumps as well. Generally, speaking I think the risk is low, with not doing that due to the fact Envoy protos are fairly stable and they are good doing things in backwards compatible way.

The `go-control-plane` repo syncs with upstream when changes are made so we look for the latest commit that matches the same protos used so that are aligned. You can see directions here: https://github.com/emissary-ingress/emissary/blob/f39bda4e0b7550df721ef840dee7b06cba223f41/_cxx/envoy.mk#L40.

The PR LGTM but just let me know if you think its worth updating go-control-plane as well

LanceEa (dismissed):

haq204:
@LanceEa ah got it.

I went ahead and updated `ENVOY_GO_CONTROL_PLANE_COMMIT` while I'm at it

LanceEa (approved):
LGTM

Issue #4968, opened 2023-04-15
emissary-ingress TLS not working in AWS ELB NLB


robd003:
If you're using ```protocol: HTTPSPROXY``` it will always fail.

Solution: ```protocolStack: [ "PROXY", "TLS", "HTTP", "TCP" ]```

robd003:
@cindymullins-dw It seems kind of silly to not support the largest cloud provider in the world.

Can we at least document how to use emissary-ingress when running on AWS and using a Network Loadbalancer with proxy v2 support for native HTTP and TLS?

blacksd:
@cindymullins-dw for the records, it breaks the default VPCgen2 Load Balancer in IBM Cloud too, I just experienced it first-hand.
At the very least, this should be explicitly stated in the documentation - from that PoV, as of today `protocol` is preferred against `protocolStack`.

PR #4969, opened 2023-04-16
Fix broken HTTPSPROXY

robd003:
## Description

Fixes problem where Emissary tries to parse TLS before parsing HAproxy protocol

## Related Issues

https://github.com/emissary-ingress/emissary/issues/4968
https://github.com/emissary-ingress/emissary/issues/4153

## Testing

Tested on AWS cluster

## Checklist

- [] **Does my change need to be backported to a previous release?**

### DIFF
--- a/docs/releaseNotes.yml
+++ b/docs/releaseNotes.yml
@@ -42,6 +42,8 @@ items:
           This upgrades $productName$ to be built on Envoy v1.25.3. This provides $productName$ with the latest
           security patches, performances enhancments, and features offered by the envoy proxy.
 
+          Fixed longstanding issue with HTTPSPROXY not working.
+
   - version: 3.5.0
     prevVersion: 3.4.0
     date: '2023-02-15'
--- a/python/ambassador/ir/irlistener.py
+++ b/python/ambassador/ir/irlistener.py
@@ -47,7 +47,7 @@ class IRListener(IRResource):
         # HTTPPROXY: accepts cleartext HTTP/1.1 sessions using the HAProxy PROXY protocol over TCP.
         "HTTPPROXY": ["PROXY", "HTTP", "TCP"],
         # HTTPSPROXY: accepts encrypted HTTP/1.1 or HTTP/2 sessions using the HAProxy PROXY protocol over TLS over TCP.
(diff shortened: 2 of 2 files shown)


robd003:
@AliceProxy could you approve the CI workflow for this pull request?

AliceProxy:
> @AliceProxy could you approve the CI workflow for this pull request?

Currently our CI system requires access to the secrets for docker pushing/pulling. We're in the process of reworking that but for now I'll have to open a separate PR with your commit to get CI running. If I just approve CI on this PR then it will fail since the PR author doesn't have write access and the secrets will show up as empty strings. You can refer to the linked PR for the CI run and you will need to sign your commit with `git rebase HEAD~1 --signoff` to get the DCO check to pass (it's a requirement for CNCF projects that contributors sign their commits :slightly_smiling_face:)

I seem to recall it being the case that this field was working for Azure users but not AWS users. Per [this comment in the issue you linked](https://github.com/emissary-ingress/emissary/issues/4153#issuecomment-1059374558) it would indeed seem that if that is the case then Azure is just doing things wrong and we should merge this assuming no issues with CI.

AliceProxy:
Looks like there was just a small problem with `make generate` due to the changelog and the DCO bot needing the commit to be signed. I can update the changelog if we merge this. Can you take a look @LanceEa? I'm a +1 for just updating this and calling it out in the release notes but it might be a good idea to test it manually with AKS/GKE just to know what the expected behaviour for each cloud provider is.

robd003:
Any update on this? @AliceProxy

LanceEa:
@robd003 - I apologize for the slow response on this one. The maintainers discussed it a little while back and I was supposed to respond but forgot to before I went on vacation.

As far as we know, the current ordering works for some cloud vendors (we believe Azure) therefore this would be a breaking change to someone. Since the `protocol` field is just syntactic sugar over the `protocolStack` we would not want to introduce a breaking change but would rather recommend using `protocolStack` so that the user can be explicit about the ordering.


> Note: if someone was to show that it doesn't work in any cloud vendor then we could reconsider and treat it as a bug but at this time the maintainers do not have the bandwidth to do that exercise.

Again sorry for the slow response.

Issue #4987, opened 2023-04-21
Internal cache files prevent `make images` from functioning consistently


rick-a-lane-ii:
**Describe the bug**

`make images` needs to always make the images if they do not exist. Currently that is not the case under some circumstances. Internal cache files can prevent `make images` from functioning. And while `make clean` should prevent this from happening, requiring this is not sufficient.

**To Reproduce**

* `make images`
* `docker image prune --all` or `docker system prune --all`
* `make images`

```console
#9 [internal] load metadata for emissary.local/base:latest
#9 sha256:31db512ce9d91e1bf165bdfff09664949b2ea72f2a1396b50dbbc3a59137ced5
#9 ERROR: failed to do request: Head "https://emissary.local/v2/base/manifests/latest": dialing emissary.local:443 with direct connection: resolving host emissary.local: lookup emissary.local: no such host

#8 [internal] load metadata for emissary.local/base-pip:latest
#8 sha256:19dadcdde8ee75b24f73095d8e8cfebe9171db4be5e178a7b6c8997261b89d3c
#8 ERROR: failed to do request: Head "https://emissary.local/v2/base-pip/manifests/latest": dialing emissary.local:443 with direct connection: resolving host emissary.local: lookup emissary.local: no such host
------
 > [internal] load metadata for emissary.local/base-pip:latest:
------
------
 > [internal] load metadata for emissary.local/base:latest:
------
failed to solve with frontend dockerfile.v0: failed to solve with frontend gateway.v0: rpc error: code = Unknown desc = failed to do request: Head "https://emissary.local/v2/base/manifests/latest": dialing emissary.local:443 with direct connection: resolving host emissary.local: lookup emissary.local: no such host
```

**Expected behavior**

`make images` should always make images as a baseline.

**Versions (please complete the following information):**

N/A

**Additional context**

PR #4995, opened 2023-04-25
Clean up outdated cache files during make images

rick-a-lane-ii:
Adds automated support for inspecting images based on the internal cache files, and cleans up any outdated files that would cause `make images` to subsequently fail.

Previously, the existing caching mechanism would prevent `make images` from functioning if the referenced images no longer exist. While both `make clean` and `make clobber` would allow `make images` to function again under those circumstances, this should not be required, and makes the build system cumbersome. With these changes in place, `make images` should function more smoothly under more circumstances, and should largely default to recreating images if necessary.

## Description

`make images`, `make push`, and `make deploy` (indirectly) will now inspect the docker images referenced in the internal cache files to determine if the images exist. If not, it will now remove the outdated cache files, and rebuild images as normal instead of outright failing.

## Related Issues

#4987

## Testing

Local environment testing focused primarily on the build system.

## Checklist

### DIFF
--- a/build-aux/builder.mk
+++ b/build-aux/builder.mk
@@ -151,16 +151,16 @@ clean: docker/$(LCNAME).docker.clean
 
 REPO=$(BUILDER_NAME)
 
-_inspect-images = base base-envoy base-pip base-python
-images-inspect: $(foreach i,$(_inspect-images), docker/$i.docker.inspect)
-.PHONY: images-inspect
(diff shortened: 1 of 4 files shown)


haq204 (approved):

LanceEa (approved):
lgtm

haq204 (approved):

PR #5006, opened 2023-04-28
Refactor the NAME environment variable

rick-a-lane-ii:
NAME is overly simplistic, not descriptive, and conflicts in some cases with existing environment variables on some systems (i.e. WSL 2), where it represents the local hostname. We should move away from using names like NAME, LCNAME, and BUILDER_NAME. Several hardcoded references to `emissary` exist as well, and some of these cause issues when using a different name (or using an environment where NAME already exists).

## Description

* Renames `NAME` to `EMISSARY_NAME` (although there may be better options here)
* Sets `LCNAME` to reference `EMISSARY_NAME` (it's probably worth renaming `LCNAME` in the near future)
* Arguably it may be worth removing this complexity altogether and just using `emissary` instead.
* Fixes a few places where `emissary` is hardcoded.
* Removes the complexity from the `WSL` documentation. WSL 2 smoothly integrates these days, and does not need too much documentation around it. Additionally, the whole related NAME issue with it just needs to be removed instead of adding to the complexity.
* The CHANGELOG may need updated with this change, just in case (although it's unlikely NAME is being changed often...)

## Related Issues

N/A

## Testing

Copious amounts of local testing. Spot testing in CI.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,15 +56,16 @@ endif
 
 # Everything else...
 
-NAME ?= emissary
+EMISSARY_NAME ?= emissary
+
 _git_remote_urls := $(shell git remote | xargs -n1 git remote get-url --all)
 IS_PRIVATE ?= $(findstring private,$(_git_remote_urls))
 
 include $(OSS_HOME)/build-aux/ci.mk
 include $(OSS_HOME)/build-aux/deps.mk
 include $(OSS_HOME)/build-aux/main.mk
-include $(OSS_HOME)/build-aux/check.mk
 include $(OSS_HOME)/build-aux/builder.mk
+include $(OSS_HOME)/build-aux/check.mk
 include $(OSS_HOME)/_cxx/envoy.mk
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -56,7 +56,9 @@ endif
 
 # Everything else...
 
-NAME ?= emissary
+EMISSARY_NAME ?= emissary
+LCNAME := $(shell echo $(EMISSARY_NAME) | tr '[:upper:]' '[:lower:]')
```
haq204, line 60: you might be able to remove LCNAME here because it will be imported from the below include statement. It appears its first usage in this file is after the imports.

haq204: you might also need to swap
```
include $(OSS_HOME)/build-aux/check.mk
include $(OSS_HOME)/build-aux/builder.mk
```
to make that work

haq204:
For makefile changes, we generally don't update the CHANGELOG

LanceEa (approved):

PR #5009, opened 2023-04-30
deps: update crd generation tools and k8s/* deps 

LanceEa:
## Description

Updates dependencies for generating a our CRD's and conversion between our CRD types which in turn bumps our k8s dependencies as well.

Each of the dependencies has been updated to be in-sync for our lowest supported k8s version of 1.22. This matches our current k3d version that the each pull-request has e2e test ran against.

> I didn't want to tackle it in this PR but we should shift these over to go:generate style scripts.

Below is an outline of each of these components with some information on each.

### Code-generator
We use `conversion-gen` to automate creating conversion code between our crd versions. Some of our typings do not play nice with the upstream tools so we have a fork with custom commits on top: https://github.com/emissary-ingress/code-generator/tree/emissary/v1.22

Ideally, we wouldn't maintain our own fork but for now I have left it. If we wanted to go back to upstream we would need to either get these commits upstreamed into Kubernetes or convert the functions that don't play nice over to hand-written ones and maintain them ourselves.

For the time being, I opted to pulling our commits on to the latest v1.22 branch due to time/effort to switch. 
If we were to The code-generator repo is part of k8s and would require us to 

### Controller-tools

We use `controller-gen` to generate the CRD YAML and the DeepCopy functions for our types. We had a fork with a single commit on top of it to fix an issue with DeepCopy and alias types.  I have opted to convert the function to a hand-written function for the `v2.MappingSpec` so that we can reference upstream instead. 

The outputted CRD's no longer contain descriptions unless a go doc comment is on the field. Previously, if a field in a struct had a non-primitive type (`Foo`) it would take the doc comment of that type and use it for the description in openapi spec for the field. However, this has changed thus the reason for the diff in descriptions.

> Note: I think this is a good candidate for upstreaming and should be fairly straight-forward. We would just need to write proper tests and submit it to upstream. Then once landed we could remove the handwritten version.

### Controller-runtime

Bumped this up to `v0.10` which is based on the k8s 1.22 libraries and also bumps the webhook conversion review type from `v1beta1` to `v1` for the apiext. 

### Fix-Crds

This is a lib in our tools directory that mutates and consolidates our CRD's into a single `crd.yaml` after being generated with `controller-gen`. It is also what injects the conversion-webhook component into the yaml. It has been updated to use `v1` as the `conversionReviewVersions` in the outputed `crd.yaml`

### kube-openapi

Bumping the k8s deps required also bumping `kube-openapi` due to its usage in mutliple pkgs (kates, go-client). Updating to the latest didn't work and keeping it the same didn't either due to build errors here and in apro.git.

After investigation, I found that I needed to take the version of the library prior to a switch in `gnostic` from `googlapies/gnostic` --> `github.com/google/gnostic`. Once we bump the k8s libraries to 1.24+ we should be able to pull in the latest version of this library.

## Related Issues

general dep upgrads

## Testing

CI here and pulled into apro.git to verify all is good: https://github.com/datawire/apro/pull/3296

## Checklist


rick-a-lane-ii (approved):

go.mod
```diff
@@ -67,6 +67,7 @@ exclude (
 	k8s.io/legacy-cloud-providers v0.0.0
 	k8s.io/metrics v0.0.0
 	k8s.io/mount-utils v0.0.0
+	k8s.io/pod-security-admission v0.0.0
```
haq204, line 70: Just wondering where this comes from since I don't see it being directly used and isn't part of the indirect dep list.

haq204 (approved):

go.mod
LanceEa: It was introduced in v1.22 as a dependency for K8s. https://pkg.go.dev/k8s.io/pod-security-admission/admission?tab=versions

So, with out having this exclude in there it would throw an error run `go mod tidy`, similar to what occurs if we don't exclude the other libraries.

haq204: Ah that makes. If I recall correctly Pod Security Policies were removed in 1.22 and replaced with Pod Security Admissions

PR #5049, opened 2023-05-15
feat: add go_filter

haq204:
## Description

Inject a go filter for EdgeStack. When running pure Emissary, the go filter is not injected. 
The go library will reach out to a filter plugin service to handle things like WAF. The Go filter runs after CORS (so that preflights can continue to work) but before the ext_authz filter so that WAF can run before OAuth.

Note that we will need to update our Envoy build to include the go-filter

## Related Issues

List related issues.

## Testing

Added unit tests. The unit tests test that the go filter is being injected into the xDS. Because the go filter is meant to be injected only when the `EDGE_STACK` env var is set to true, some test cases are marked as `edgestack` and would run in apro CI which will be handled separately

I've confirmed that the unit tests do pass with both settings for the `EDGE_STACK` flag:
image

Full Integration and E2E tests will be added in a separate PR.

## Checklist

  - Any new features
  - Any deprecations

Note that integration and E2E tests will be added in a separate PR

### DIFF
--- a/python/ambassador/ir/ir.py
+++ b/python/ambassador/ir/ir.py
@@ -31,6 +31,7 @@ from .irbasemappinggroup import IRBaseMappingGroup
 from .ircluster import IRCluster
 from .irerrorresponse import IRErrorResponse
 from .irfilter import IRFilter
+from .irgofilter import IRGOFilter
 from .irhost import HostFactory, IRHost
 from .irhttpmapping import IRHTTPMapping
 from .irlistener import IRListener, ListenerFactory
@@ -408,11 +409,14 @@ class IR:
 
         # After the Ambassador and TLS modules are done, we need to set up the
         # filter chains. Note that order of the filters matters. Start with CORS,
-        # so that preflights will work even for things behind auth.
+        # so that preflights will work for all filters.
 
         self.save_filter(
             IRFilter(ir=self, aconf=aconf, rkey="ir.cors", kind="ir.cors", name="cors", config={})
         )
+
+        # We always add the Go filter calling to our filter plugin service if applicable before auth
+        self.save_filter(IRGOFilter(ir=self, aconf=aconf))
 
         # Next is auth...
         self.save_filter(IRAuth(self, aconf))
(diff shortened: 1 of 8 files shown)


python/ambassador/ir/irgofilter.py
```diff
@@ -0,0 +1,72 @@
...
+            name=name,
+            namespace=namespace,
+            type="decoder",
+        )
+
+    # We want to enable this filter only in Edge Stack
+    def setup(self, ir: "IR", _: Config) -> bool:
+        if ir.edge_stack_allowed:
+            if not go_library_exists(GO_FILTER_OBJECT_FILE):
+                self.post_error(f"{GO_FILTER_OBJECT_FILE} not found, disabling Go filter...")
```
AliceProxy, line 65: Do you think we should change the message or something so that emissary users don't read this and think it's a runtime/config problem?

AliceProxy (dismissed):
this seems fine

python/ambassador/ir/irgofilter.py
haq204: Emissary users wouldn't see this message. It would only post in Edge Stack mode if the library was missing

python/ambassador/ir/ir.py
```diff
@@ -408,12 +409,16 @@ def __init__(
 
         # After the Ambassador and TLS modules are done, we need to set up the
         # filter chains. Note that order of the filters matters. Start with CORS,
-        # so that preflights will work even for things behind auth.
+        # so that preflights will work even for things behind auth and WAF.
```
LanceEa, line 412: I would leave this off or make it generic.

python/ambassador/ir/ir.py
```diff
@@ -408,12 +409,16 @@ def __init__(
...
 
         self.save_filter(
             IRFilter(ir=self, aconf=aconf, rkey="ir.cors", kind="ir.cors", name="cors", config={})
         )
 
+        # We always add the Go filter calling to our filter plugin service if applicable before auth
+        # so that things like WAF can run before auth.
```
LanceEa, line 419: I would leave specifics off and keep use case generic.

python/conftest.py
LanceEa, line 1: Is there a reason for having this at the root and not nested in the tests folder?

LanceEa (commented):
Left a few small comments but generally speaking it looks good to me.

AliceProxy (approved):

LanceEa (approved):

PR #5057, opened 2023-05-17
envoy: add contrib extensions

haq204:
## Description
Companion to https://github.com/emissary-ingress/emissary/pull/5049

Add protobuf for go filter under api/contrib. Because the go filter is not part of the core envoy protobufs (`api/envoy`), the Makefiles have been updated to also build select `api/contrib` protos based on need. You can specify which additional contrib protos to build in `_cxx/envoy.mk` and it will build only the contrib protos listed in addition to the core envoy protobufs to minimize the total build time for `make generate`.

Update envoy build to include contrib extensions which includes extensions such as go-filter. Ideally, we would have enabled only the extensions that we needed but Envoy's bazel system doesn't include any contrib extensions whether you enable it or not in the normal source build. The contrib build enables all contrib extensions by default and modifying the bazel source files to exclude certain extensions would have added to maintanance complexity so all contrib extensions are built which adds only a few MB to the final build.

## Related Issues

List related issues.

## Testing

Verified that the new envoy build can run the go-filter using the example filter from the envoy sandbox:

```
docker compose images
Container                     Repository                  Tag                 Image Id            Size
golang-helloworld_service-1   golang-helloworld_service   latest              a3adf50359af        154MB
golang-proxy-1                golang-proxy                latest              03f4a3f53c32        130MB

curl -v localhost:10000 2>&1 | grep rsp-header-from-go
< rsp-header-from-go: bar-test

```

## Checklist

  - Any new features
  - Any deprecations


_cxx/envoy.mk
```diff
@@ -254,36 +258,39 @@ envoy-shell: $(ENVOY_BASH.deps)
 #
 # These targets are depended on by `make generate` in `build-aux/generate.mk`.
 
-# Raw protobufs
-$(OSS_HOME)/api/envoy: $(OSS_HOME)/api/%: $(OSS_HOME)/_cxx/envoy
-	rsync --recursive --delete --delete-excluded --prune-empty-dirs --include='*/' --include='*.proto' --exclude='*' $</api/$*/ $@
+# Raw Envoy protobufs
+$(OSS_HOME)/api/envoy $(addprefix $(OSS_HOME)/api/,$(envoy-api-contrib)): $(OSS_HOME)/api/%: $(OSS_HOME)/_cxx/envoy
```
LanceEa, line 262: Is it possible to just write these to `api/envoy/extensions/filters/http` so that they are consistent or do you think that could get us in trouble?

The benefit is that a lot of this wouldn't have to change and we would have the same import pattern in ambex.

LanceEa (commented):
Does this mean we wont need a custom build of envoy? For the most part looks good, just one question?

LanceEa (commented):
Does this mean we wont need a custom build of envoy? For the most part looks good, just one question?

haq204:
> Does this mean we wont need a custom build of envoy? For the most part looks good, just one question?

No we'll still need to update the build. Just tested trying to use the example go-filter with our current envoy build and it crashed because the type isn't compiled so will have to update that (separate PR).

_cxx/envoy.mk
haq204: I tried doing that but it makes the make recipes more complicated and probaly requires some custom recipes to be created for `contrib/` and I wanted to reuse what we have as much as possible. 

I didn't want to go down that rabbit hole for the sake of package purity right now

haq204: personally I think a better approach is migrate to buf

LanceEa (approved):

_cxx/envoy.mk
```diff
@@ -18,7 +18,7 @@ RSYNC_EXTRAS ?=
   ENVOY_COMPILATION_MODE ?= opt
   # Increment BASE_ENVOY_RELVER on changes to `docker/base-envoy/Dockerfile`, or Envoy recipes.
   # You may reset BASE_ENVOY_RELVER when adjusting ENVOY_COMMIT.
-  BASE_ENVOY_RELVER ?= 0
+  BASE_ENVOY_RELVER ?= 2
```
haq204, line 21: I tried another bazel build flag before but that didn't work hence why it's bump to 2.

This can be reset on the next Envoy version upgrade

LanceEa (approved):

AliceProxy (approved):

PR #5058, opened 2023-05-18
Dont specify a default cpu limits

scottillogical:
Cpu limits might be necessary or helpful in certain multi tenant environments, but I think its far more common for emissary to be running on k8s nodes where the extra cpu cycles should be used instead of limiting cpu by default.   As such this pr removes the default cpu limit

### DIFF
--- a/charts/emissary-ingress/values.yaml.in
+++ b/charts/emissary-ingress/values.yaml.in
@@ -300,7 +300,6 @@ deploymentAnnotations: {}
 resources: # +doc-gen:break
   # Recommended resource requests and limits for Ambassador
   limits:
-    cpu: 1000m
     memory: 600Mi
   requests:
     cpu: 200m


cindymullins-dw:
Hi @scottillogical , FYI you're welcome to join our [Monthly Contributors meeting](https://github.com/emissary-ingress/emissary/blob/master/Community/MEETING_SCHEDULE.md#monthly-contributors-meeting) tomorrow, Wed June 7 at 3:30pm ET with our engineers.) to discuss this PR if that would be helpful to you.  Thanks for contributing!

NikitaCloudRuntime:
Hello,

we're also interested in not setting the default for CPU limit
@scottillogical could you share what is the current status? Did you have a chance to join [Monthly Contributors meeting](https://github.com/emissary-ingress/emissary/blob/master/Community/MEETING_SCHEDULE.md#monthly-contributors-meeting) mentioned in the comment above?
thank you

LanceEa:
@scottillogical - Thanks for the PR. Although you are right it may be more common to want higher or no limits on an install where the Nods for ingress are isolated, this would be a breaking change for some one.

Although, I understand the workaround isn't the ideal solution, it does work to set the values to really high numbers so we wouldn't accept the proposed change at this time.


>Hey folks, adding weight to this issue is the fact that if you are an edge-stack user, you cannot remove the default cpu limit via helm, only change it. Setting it to null [helm docs] (https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key) doesn't work because of https://github.com/helm/helm/issues/9027 with that feature and subcharts, as you can see here

We are aware of the limitation of sub-charts and are tracking the issue you posted. It does appear there is some traction on this effort via: https://github.com/helm/helm/pull/12162

scottillogical:
Wasn't adding the limit itself backwards incompatible change?  If anyone cares about cpu limiting ambassador, they should set this value themselves - this is an env specific setting that should not be a default and that is not common.   I would imagine most users have the limits set by default and don't even realize and wouldn't want it.

PR #5103, opened 2023-06-13
diagd: atomically write the ADS config

### LINKED ISSUE #5093: ambex: failing to load `envoy.json` can result in "empty" snapshot being pushed to Envoy
**Describe the bug**
If `ambex` fails to load any Envoy config files, the file in question is completely ignored BUT the new snapshot update continue. This cause `ambex` to save and push a new snapshot without the correct configuration. See https://github.com/emissary-ingress/emissary/blob/v3.5.2/pkg/ambex/main.go#L405-L409

For example, when `envoy/envoy.json` is skipped due to a loading error, this cause Envoy to be reconfigured with none of our Listeners/Mappings/etc. Listener then go in drain mode, requests are failing with various response flags (NC, UAEX, UH).

**To Reproduce**

Haven't reproduced locally (yet), but I would expect the following to work:
1. Run ambex locally
2. Add a valid config file in the directory being watched 
3. Validate that Ambex has pushed a snapshot and Envoy has loaded the config
4. Mutate the file to make it invalid
5. Validate that ambex failed to load the file, but still pushed a snapshot
6. Validate that Envoy loaded config is wrong (missing the configuration from files)

**Expected behavior**
`ambex` should handle the error properly and not continue the snapshot generating if it failed to load any of the given input files.

 - Emissary-Ingress: v3.5.2
 - GKE 1.25.8-gke.1000	

**Additional context**

Logs showing the issue:
```
2023-06-05 22:42:29 diagd 3.5.2 [P31TAEW] INFO: EnvoyConfig: Generating V3
2023-06-05 22:42:34 diagd 3.5.2 [P31TAEW] INFO: V3Listener <V3Listener HTTP https-listener on 0.0.0.0:443 [XFP]>: generated ===========================
2023-06-05 22:42:35 diagd 3.5.2 [P31TAEW] INFO: V3Ready: ==== listen on 127.0.0.1:8006
time="2023-06-05 22:42:50.8810" level=warning msg="/ambassador/envoy/envoy.json: proto: unexpected EOF" func=github.com/emissary-ingress/emissary/v3/pkg/ambex.update file="/go/pkg/ambex/main.go:407" CMD=entrypoint PID=1 THREAD=/ambex/main-loop
time="2023-06-05 22:42:50.8814" level=info msg="Saved snapshot v1165" func=github.com/emissary-ingress/emissary/v3/pkg/ambex.csDump file="/go/pkg/ambex/main.go:350" CMD=entrypoint PID=1 THREAD=/ambex/main-loop
time="2023-06-05 22:42:51.0230" level=info msg="Pushing snapshot v1165" func=github.com/emissary-ingress/emissary/v3/pkg/ambex.updaterWithTicker file="/go/pkg/ambex/ratelimit.go:159" CMD=entrypoint PID=1 THREAD=/ambex/updater
```

Note: we run with `envoy_validation_timeout=0` (disabled) because with the size of our config (328.0MB), the validation uses too much memory and is too slow. In that particular error case `proto: unexpected EOF`, it is possible that `envoy/envoy.json` was corrupted and that having the validation enabled would have catch this before it reach ambex. But the `Decode()` function has a lot of other case where it could fail, like failing to read the file, which would end up in the same situation.

You can also see that since ambex didn't return an error, diagd think we successfully reconfigured Envoy with all the expected config (S13379 L1 G9199 C8503)
```
2023-06-05 22:42:59 diagd 3.5.2 [P31TAEW] INFO: configuration updated (incremental) from snapshot snapshot (S13379 L1 G9199 C8503)
```

dethi:
## Description

Write the generated ADS config to a temporary file first, then rename the file. This ensure that the update is atomic and that ambex won't load a partially written file.

## Related Issues

Fix #5093

## Testing

- Manual testing
- Deployed to all our prod clusters

## Checklist

  - Any new features
  - Any deprecations

### DIFF
--- a/python/ambassador_diag/diagd.py
+++ b/python/ambassador_diag/diagd.py
@@ -1771,8 +1771,19 @@ class AmbassadorEventWatcher(threading.Thread):
         with open(app.bootstrap_path, "w") as output:
             output.write(dump_json(bootstrap_config, pretty=True))
 
-        with open(app.ads_path, "w") as output:
-            output.write(dump_json(ads_config, pretty=True))
+        # Write the ADS config to a temporary file first and then rename to
+        # ensure atomic update. Otherwise, ambex could be reading the config
+        # while we are still writing it.
+        #
+        # Rename are only atomic if in the same FS. To ensure that this work
+        # with any setup, we store the temporary file at the same path but with
+        # an extension that is ignored by ambex.
+        ads_tmp_path = app.ads_path + ".tmp"
+        with open(ads_tmp_path, "w") as output:
+            # Don't use pretty here, to reduce the config size, and improve the
+            # dump and load speed of the config.
+            output.write(dump_json(ads_config, pretty=False))
+        os.rename(ads_tmp_path, app.ads_path)
 
         with open(app.clustermap_path, "w") as output:
             output.write(dump_json(clustermap, pretty=True))


dethi:
Does this need to be included in the CHANGELOG?

LanceEa:
@dethi - Thanks for this contribution! I definitely think there is some work we can do here to improve this. I will comment on your issue as well but here is a few things:

1. I will pull this in and kick off CI
2. On the changlog

> Does this need to be included in the CHANGELOG?

I wouldn't worry about it in this case, this is more internal implementation details so we typically don't put them in the change log explicitly unless its user facing or we feel important for the users to know. Ideally we would have a more automated change-log to avoid manual effort and merge conflicts but its something we haven't had time to put effort into.

3. This code isn't written in the most test friendly fashion. Any ideas on a good way to test it?

In the meantime, for my own curiosity I put together a quick fsnotify test example to see how often the file watch events are triggered using your provided solution and roughly mimicking the behavior of ambex. 

Here was my output:

```shell
2023/06/21 08:08:50 event: CREATE        "data/envoy-output.tmp" has .tmp in the prefix, ignoring....
2023/06/21 08:08:50 event: WRITE         "data/envoy-output.tmp" has .tmp in the prefix, ignoring....
2023/06/21 08:08:50 event: CREATE        "data/envoy-output.json"
2023/06/21 08:08:50 event: RENAME        "data/envoy-output.tmp" has .tmp in the prefix, ignoring....
```
Events 1 & 2 should get ignored by ambex because of the calls to `.isDecodable` will ignore the `.tmp` since its not `json/pb`. https://github.com/emissary-ingress/emissary/blob/c28d9fa4fc51774ae7e268af8b72e95be8fe484b/pkg/ambex/main.go#L399

Event 3 would trigger the loading and parsing of the file. So, its not clear to me if that event indicates that no additional writing will be done or not.

Event 4 would also trigger the loading and parsing of the file because ambex only uses the event as a trigger to call update and instead uses the configured `dirs` (only `/ambassador/envoy`) by default to check for files with the correct extension.

So, my two open questions are:
1. Does event 3 still run the risk of not being atomic and still have the potential to get large configs such as yours into a broken state?
2. I think there might be a secondary issue/opportunity to improve ambex so that it doesn't unnecessarily run update on both CREATE and RENAME.

LanceEa:
@dethi - Actually thinking about it more, the behavior I just described is on my Mac but I believe Linux would have the similar behavior but I need to confirm that.

dethi:
From what I read of the code, fsnotify is not used at all on that config folder. It's only used to watch the itsio certs folder. Correct me if I'm wrong.

LanceEa:
> From what I read of the code, fsnotify is not used at all on that config folder. It's only used to watch the itsio certs folder. Correct me if I'm wrong.

Yes, we do use fsnotify for watching istio but that comes in via watchallthings and I believe it triggers the diagd snapshot process so it goes through the full configuration path. Somewhere here: https://github.com/emissary-ingress/emissary/blob/c28d9fa4fc51774ae7e268af8b72e95be8fe484b/cmd/entrypoint/watcher.go#L292
But, it has been a bit sense I looked at this code so might be missing something.

Diagd communicating with Ambex is done as well via fsnotify events. When Ambex is started it is passed the envoy folder (`/ambassador/envoy`) folder which is where the diagd generated config is generated. 
https://github.com/emissary-ingress/emissary/blob/c28d9fa4fc51774ae7e268af8b72e95be8fe484b/cmd/entrypoint/entrypoint.go#L172

Then we register the watcher here:
https://github.com/emissary-ingress/emissary/blob/c28d9fa4fc51774ae7e268af8b72e95be8fe484b/pkg/ambex/main.go#L650


Does that make sense or maybe I'm missing something?

dethi:
> Diagd communicating with Ambex is done as well via fsnotify events

Not by default. Diagd notify ambex via signal (`-kick`). The fsnotify version is behind a flag that default to false (`-watch`) and that isn't present in the emissary-ingress helm chart. Also, someone using `-watch=true` would be even more prone to the race condition that I explained in the linked issue. I would be surprise that anyone use it in production.

Do you have more context around why this was added / when it's expected to be used? Meanwhile I can make an additional change to ignore fsnotify events that aren't useful.

dethi:
> > Diagd communicating with Ambex is done as well via fsnotify events
> 
> Not by default. Diagd notify ambex via signal (`-kick`). The fsnotify version is behind a flag that default to false (`-watch`) and that isn't present in the emissary-ingress helm chart. Also, someone using `-watch=true` would be even more prone to the race condition that I explained in the linked issue. I would be surprise that anyone use it in production.
> 
> Do you have more context around why this was added / when it's expected to be used?

@LanceEa would you have time answering the question above?

LanceEa:
@dethi - this has been on my TODO list and finally got some free time to dive in. I wanted to run some experiments and make sure my understanding was correct as well. 

>Do you have more context around why this was added / when it's expected to be used?

I don't have a ton of context on this since I joined the project after a lot of this design and development was already done so mostly just trying to read tea leaves 😄 . But, I believe the idea behind the file watch was to be able to break-up the `envoy.json` into smaller files to reduce the loading of one larger envoy.json. But, as you mentioned it isn't currently being used.

>Not by default. Diagd notify ambex via signal (-kick). The fsnotify version is behind a flag that default to false (-watch)

Yes, you are right. I had missed this line: https://github.com/emissary-ingress/emissary/blob/60867c43299c9940d243a3791870918b2eafc644/pkg/ambex/main.go#L656

Current default args are:

```golang
&ambex.Args{
    watch:false, 
    adsNetwork:"tcp",
    adsAddress: "127.0.0.1:8003", 
    dirs:[]string{"/ambassador/envoy"}, 
    snapdirPath: "/ambassador/snapshots", 
    numsnaps: 30, 
    edsBypass:false,
}
```
I ran a test build and verified that the os HUP is what is triggering ambex to update and load the `envoy.json` file for passing it along to Envoy via xDS and that no where in the codebase are we providing a mechanism to actually set `--watch=true`.

I will go ahead and get your PR running in CI to verify all is good and we can get this merged in. Thanks for the patience on it!

LanceEa:
@haq204 - CI is green as seen here: https://github.com/emissary-ingress/emissary/pull/5258

dethi:
Thanks @LanceEa! Since you have access to internal knowledge, if you can find if inotify is still being planned or not in the future, this way I can work on another PR for it.

AliceProxy (approved):

LanceEa:
Thanks again for the PR! 

> Since you have access to internal knowledge, if you can find if inotify is still being planned or not in the future, this way I can work on another PR for it.

I will talk with the other Maintainers but I personally lean towards cleaning it up to simplify the code base and remove the confusion.

PR #5202, opened 2023-08-03
deps: upgrade to envoy 1.27

LanceEa:
## Description

This upgrades to Emissary-ingress  to the latest release of Envoy 1.27 and the corresponding update to go-control-plane as well. The upgrade from Envoy 1.26 -> 1.27 included a few refactoring's from upstream that needed to be addressed:

1. Multiple functions were refactored requiring us to update our custom ResponseMap http filter to conform to the new function signatures.
2. CI scripts for generating protobufs were refactored requiring this project to update how it builds the Envoy api protobufs.

Other than that, there were no behavioral changes observed.

> Reviewers please take a look over the new commits in https://github.com/datawire/envoy/commits/rebase/release/v1.27.0. If there are no concerns then I will finish uploading to the mirrors and tagging it so that the `check-envoy-version` job will turn green as well.

## Related Issues

N/A - normal policy of keeping up with Envoy releases.

## Testing

CI is green, no additional tests have been added. Verified that their are no issues with Edge Stack as seen here: https://github.com/datawire/apro/pull/3391

## Checklist


haq204:
just a quick question: 
in the Envoy 1.27 rebase branch, will the updates made to response made be squashed in the original commit before tagging and pushing?

Otherwise lgtm

LanceEa:
> just a quick question: in the Envoy 1.27 rebase branch, will the updates made to response made be squashed in the original commit before tagging and pushing?
> 
> Otherwise lgtm

I think we should leave them for now so that we can be clear about what has changed. I cleaned up the others previously because it was like 6-7 but I think going forward a better approach is to keep a commit sitting on top to be clear that it is fix refactors. 

 Ideally we can take some time to assess whether ResponseMap can be replaced by out-of-box CustomResponse filter. Fingers crossed a simple func could translate between the two with an opt-out flag. Then in some future release cut over to it if we see no issues with it and then eventually drop the ResponseMap commits.

LanceEa:
Cleaned up commit messages. This should be good to review once it is green.

rick-a-lane-ii (approved):

haq204 (approved):

PR #5215, opened 2023-08-10
bug: make canary groups take into account host mappingSelector labels

### LINKED ISSUE #4170: Canary grouping must take labels into account
The following configuration attempts to use labels to associate the `quote-backend-foo`
`Mapping` only with `foo.example.com`, and `quote-backend-bar` only with `bar.example.com`:

```yaml
---
kind: Host
metadata:
  name: foo-host
spec:
  hostname: foo.example.com
  mappingSelector:
    matchLabels:
      host: foo
  ...
---
kind: Host
metadata:
  name: bar-host
spec:
  hostname: bar.example.com
  mappingSelector:
    matchLabels:
      host: bar
  ...
---
kind: Mapping
metadata:
  name: quote-backend-foo
  labels:
    host: foo
spec:
  prefix: /test/
  service: quote
---
kind: Mapping
metadata:
  name: quote-backend-bar
  labels:
    host: bar
spec:
  prefix: /test/
  service: quote
```

Since the two `Mapping`s have different `host` labels, they will associate with different `Host`s
and should not be placed in the same canary group. However, as of Emissary 2.2.2, the canary-group
logic doesn't pay attention to labels, so the two `Mapping`s will (wrongly) be placed in the same
canary group. Since Envoy-config generation operates on groups rather than on individual `Mapping`s,
the effect here is that `https://foo.example.com/test/` will be correctly routed, but
`https://bar.example.com/test/` will 404.

As an additional constraint, consider the same `Host`s above, but these two `Mapping`s:

```yaml
---
kind: Mapping
metadata:
  name: quote-labels-1
  labels:
    host: foo
    irrelevant-label: "1"
spec:
  prefix: /labels/
  service: quote-1
---
kind: Mapping
metadata:
  name: quote-labels-2
  labels:
    host: foo
    irrelevant-label: "2"
spec:
  prefix: /labels/
  service: quote-2
```

These `Mapping`s _should_ be placed in the same canary group, since the `Host`s don't differentiate
based on the `irrelevant-label` label. So we mustn't look at _every_ label -- only the ones that
the `Host`s actually care about matter.

To work around this, you can either:

1. Give the two `Mapping`s a `precedence` with _different values_ so that they can't be canaried together, or
2. Use the `hostname` rather than labels.

haq204:
## Description

If two mappings have different host labels but share the same prefix, then they should not be placed in the same canary group. Previously, labels were not taken into account when grouping mappings into canary groups, which caused mappings sharing the same prefix to be part of the same canary group when using host labels (hostname or host field was not specified in the mapping).

In order to solve this, we update the mapping group_id logic to incorporate host selector labels. This means that the mappings are grouped based on the following criteria:
- HTTP method
- Prefix
- Headers
- Query parameters
- Host labels
- Precedence

Because this affects only how mappings are grouped, the following behavior is retained:
- hostname, host, and :authority header fields take precendence over host labels for the purposes of grouping.
- If hostname, host, and/or :authority header fields are specified along with host labels, then they must match before they can be associated.
- A canary group can be associated with multiple Hosts based on selector labels. 

## Related Issues

#4170.

## Testing

Added additional unit tests for canary grouping.

## Checklist

  - Any new features
  - Any deprecations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,15 @@ it will be removed; but as it won't be user-visible this isn't considered a brea
 
 ### Emissary-ingress and Ambassador Edge Stack
 
+- Bugfix: As of v2.2.2, if two mappings were associated with different Hosts through host
+  mappingSelector labels but share the same prefix, the labels were not taken into account which
+  would cause one Mapping to be correctly routed but the other not.
+  This change fixes this issue so
+  that Mappings sharing the same prefix but associated with different Hosts will be correctly
+  routed. ([Canary grouping must take labels into account])
+
+[Canary grouping must take labels into account]: https://github.com/emissary-ingress/emissary/issues/4170
+
 ## [3.7.2] July 25, 2023
 [3.7.2]: https://github.com/emissary-ingress/emissary/compare/v3.7.1...v3.7.2
 
--- a/docs/releaseNotes.yml
+++ b/docs/releaseNotes.yml
@@ -35,7 +35,19 @@ items:
   - version: 3.8.0
     prevVersion: 3.7.2
     date: TBD
-    notes: []
+    notes:
+      - title: Account for matchLabels when associating mappings with the same prefix to different Hosts
+        type: bugfix
+        body: >-
+          As of v2.2.2, if two mappings were associated with different Hosts through host
+          mappingSelector labels but share the same prefix, the labels were not taken into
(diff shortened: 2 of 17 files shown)


AliceProxy (approved):
lgtm

LanceEa (approved):
lgtm...lets get it pulled into Edge Stack to make sure there are no gotcha's

PR #5227, opened 2023-08-15
Ambassador agent labels

jnieuwen:
## Description

Add deploymentLabels and podLabels to the ambassador-agent deployment. 

## Related Issues

N/A

## Testing

Manual test.

## Checklist

- [N/A] **Does my change need to be backported to a previous release?**

- [N/A ] **I made sure to update `CHANGELOG.md`.**

  - Any new features
  - Any deprecations

Only adding some extra labels on the deployment, should not impact performance of the deployment itself.

- [N/A] **I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently.**

- [N/A] **The changes in this PR have been reviewed for security concerns and adherence to security best practices.**

### DIFF
--- a/charts/emissary-ingress/templates/ambassador-agent.yaml
+++ b/charts/emissary-ingress/templates/ambassador-agent.yaml
@@ -190,6 +190,9 @@ metadata:
     app.kubernetes.io/name: {{ include "ambassador.fullname" . }}-agent
     {{- include "ambassador.labels" . | nindent 4 }}
     product: aes
+    {{- if .Values.deploymentLabels }}
+      {{- toYaml .Values.deploymentLabels | nindent 4 }}
+    {{- end }}
 spec:
   replicas: 1
   selector:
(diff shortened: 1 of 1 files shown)


LanceEa:
@jnieuwen - Thanks for the PR. The Agent code was extracted to its own component and now has its own chart so can you take a look here to see if that meets your needs and if not submit a PR there? 

https://github.com/datawire/ambassador-agent/tree/master/helm/ambassador-agent 

We would like to move away from Agent specific work in this repository and maintaining duplicate templates. So, our idea would be either to remove the templating in the emissary-ingress chart. We can then either sub-chart it the Agent or not allowing it to evolve at a faster cadence independent of emissary-ingress releases.

jnieuwen:
Hi @LanceEa - that also might work for our use case. Thanks for the reply. From my point the PR can be closed.

LanceEa:
@jnieuwen - Sounds good, going to go ahead and close this PR.

PR #5258, opened 2023-08-22
[CI Run] diagd: atomically write the ADS config

### LINKED ISSUE #5093: ambex: failing to load `envoy.json` can result in "empty" snapshot being pushed to Envoy

LanceEa:
**DO NOT MERGE** - CI run for #5103 

>note: I have rebased on recent master but no other changes.

Write the generated ADS config to a temporary file first, then rename the file. This ensure that the update is atomic and that ambex won't load a partially written file.

Fix #5093

### DIFF
--- a/python/ambassador_diag/diagd.py
+++ b/python/ambassador_diag/diagd.py
@@ -1771,8 +1771,19 @@ class AmbassadorEventWatcher(threading.Thread):
         with open(app.bootstrap_path, "w") as output:
             output.write(dump_json(bootstrap_config, pretty=True))
 
-        with open(app.ads_path, "w") as output:
-            output.write(dump_json(ads_config, pretty=True))
+        # Write the ADS config to a temporary file first and then rename to
+        # ensure atomic update. Otherwise, ambex could be reading the config
+        # while we are still writing it.
+        #
+        # Rename are only atomic if in the same FS. To ensure that this work
+        # with any setup, we store the temporary file at the same path but with
+        # an extension that is ignored by ambex.
+        ads_tmp_path = app.ads_path + ".tmp"
+        with open(ads_tmp_path, "w") as output:
+            # Don't use pretty here, to reduce the config size, and improve the
+            # dump and load speed of the config.
+            output.write(dump_json(ads_config, pretty=False))
+        os.rename(ads_tmp_path, app.ads_path)
 
         with open(app.clustermap_path, "w") as output:
             output.write(dump_json(clustermap, pretty=True))


LanceEa:
Merged original PR already and closing CI run PR.

PR #5261, opened 2023-08-24
kill docker demo mode

AliceProxy:
## Description

Removes the docker demo mode that is pretty much never used and has minimal functionality.

## Related Issues

List related issues.

## Testing

## Checklist

  - Any new features
  - Any deprecations


DevDocumentation/DEVELOPING.md
```diff
@@ -538,7 +538,7 @@ If you do not have envoy in your path already, the entrypoint will use
 docker to run it. At the moment this is untested for macs which probably
 means it is broken since localhost communication does not work by
 default on macs. This can be made to work as soon an intrepid volunteer
-with a mac reaches out to me (rhs@datawire.io).
+with a mac reaches out to me (<rhs@datawire.io>).
```
haq204, line 541: probably need to assign someone also to replace Rafael :smile:

haq204 (approved):

cmd/entrypoint/env.go
```diff
@@ -206,7 +188,7 @@ func GetDiagdFlags(ctx context.Context, demoMode bool) []string {
 	// XXX: this was not in entrypoint.sh
 	result = append(result, "--port", GetDiagdBindPort())
 
-	cdir := GetConfigDir(demoMode)
+	cdir := env("AMBASSADOR_CONFIG_DIR", path.Join(GetAmbassadorConfigBaseDir(), "ambassador-config"))
```
LanceEa, line 191: Do we need this at all anymore? What still uses the filesystem flag for diagd?

LanceEa (commented):
generally looks good just a couple of comments

cmd/entrypoint/env.go
AliceProxy: I don't know of anything that does, but we still have one endpoint that listens for a request and enables it

LanceEa: What do you mean by `endpoint`? 

Looking around the directory for `ambassador-confg` isn't used at all throughout the repository so I think it is essentially legacy code. It apparently is a `Values.ambassadorConfig` in the Helm chart but other than that there is no reference to it. Do you have any context on it? It is probably a separate PR from this one but I'm thinking we should remove it as well because we don't actively support/document it or know if it even works. 

https://github.com/emissary-ingress/emissary/blob/b9657518353b87dc05fef18a4c06b58e252da4dc/charts/emissary-ingress/templates/deployment.yaml#L233

build-aux/Dockerfile
```diff
@@ -110,7 +110,6 @@ RUN mkdir -p /ambassador/sidecars && \
     ln -s /buildroot/ambassador/python/kubewatch.py /ambassador/kubewatch.py
 
 # These will be extracted into the optimized image later
-ADD demo demo
 ADD manifests/emissary/emissary-crds.yaml.in manifests/emissary/emissary-crds.yaml
```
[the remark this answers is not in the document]
LanceEa: I think we have to keep this because of this comment below: https://github.com/emissary-ingress/emissary/pull/5261#discussion_r1305655291

cmd/entrypoint/env.go
LanceEa: As we discussed offline, lets resolve that one independently of this PR.

LanceEa (approved):

PR #5276, opened 2023-08-31
cleanup: remove older telepresence make targets and setup

LanceEa:
## Description

WIP: it looks like some of our tests rely on these so more investigation needs to be done.

These make targets and usages of telepresence have gone stale and are not part of the normal dev loop. To reduce confusion and maintenance they are being cleaned up from the repository. We can re-introduce on a as needed basis.

## Related Issues

N/A - general cleanup

## Testing

CI is green, pulled into Edge Stack to verify no assumptions about tooling are being made.

## Checklist

- [x] **I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently.**


LanceEa:
Closed because we still need these for the time being until we re-work some test infrastructure.

PR #5346, opened 2023-10-10
Apply envoy security patches

AliceProxy:
## Description

My original goal was to update envoy to 1.26.5, but Emissary's build system is completely broken by commits that have landed there. So instead I am just applying the security patches from the 1.26.5 release on top of our 1.26.4 branch. Next we can just jumpo to 1.27 since @LanceEa is working on CI and builds for that.


LanceEa (approved):

AliceProxy:
The Envoy check is only failing because I do not have permission to push to the `gcr` mirror and neither does the `d6edevautomaton` bot.

tenshinhigashi (approved):

PR #5350, opened 2023-10-11
update: envoy version in comment to v1.26.5

mrnonz:
Due to HTTP/2 (CVE-2023-44487). I've already update comment to match with `ENVOY_COMMIT ?= 9962dc50f4dc8c5067bea567b1b2a4028a265bc3`.

Thanks


LanceEa:
@mrnonz - Thanks for the PR but the comment is already correct as you can see the commit corresponds to the following branch: https://github.com/datawire/envoy/commits/rebase/release/v1.26.4

Upstream Envoy has backported a lot of CI changes recently that was causing issues with our custom Envoy build process. Rather than delay it while we figure those out, we chose to apply the CVE' commits to the 1.26.4 branch which is why the comment was left.

Issue #5436, opened 2023-11-16
Api ext pod errors causing restart


BChancusi:
**Describe the bug**

Applying Emissary as yaml and waiting on the deployment fails.

> deployment.apps/emissary-apiext condition met
> Error from server: error when creating "STDIN": conversion webhook for getambassador.io/v2, Kind=Module failed: Post "https://emissary-apiext.emissary-system.svc:443/webhooks/crd-convert?timeout=30s": proxy error from 127.0.0.1:6443 while dialing 10.42.0.46:8443, code 502: 502 Bad Gateway

Caused by the api-ext pod restarting post being flagged as ready/passing the wait check.

> time="2023-11-16 13:12:46.4818" level=info msg="Emissary Ingress apiext (version \"3.9.0\")" func=github.com/emissary-ingress/emissary/v3/cmd/apiext.Main file="/go/cmd/apiext/main.go:16" CMD=apiext PID=1
time="2023-11-16 13:12:46.4827" level=info msg="APIEXT_LOGLEVEL=info" func="github.com/emissary-ingress/emissary/v3/pkg/apiext.(*WebhookServer).Run" file="/go/pkg/apiext/server.go:89" CMD=apiext PID=1
time="2023-11-16 13:12:47.1958" level=info msg="Serving HTTPS on port 8443" func=github.com/emissary-ingress/emissary/v3/pkg/apiext/internal.ServeHTTPS file="/go/pkg/apiext/internal/serve.go:88" CMD=apiext PID=1 THREAD=/serve-https
time="2023-11-16 13:12:47.2323" level=info msg="Configuring conversion for \"authservices.getambassador.io\"" func=github.com/emissary-ingress/emissary/v3/pkg/apiext/internal.updateCRD file="/go/pkg/apiext/internal/inject.go:137" CMD=apiext PID=1 THREAD=/configure-crds
time="2023-11-16 13:12:47.2439" level=info msg="Configuring conversion for \"consulresolvers.getambassador.io\"" func=github.com/emissary-ingress/emissary/v3/pkg/apiext/internal.updateCRD file="/go/pkg/apiext/internal/inject.go:137" CMD=apiext PID=1 THREAD=/configure-crds
[...]

PR #5449, opened 2023-11-20
Add /status permission to ApiExt ClusterRole

BChancusi:
Fixes #5436

## Description

Fix permission errors by adding the permission needed to update the sub resource "status" required by https://github.com/emissary-ingress/emissary/commit/28ff28173998aad252dc12fa95d717fde5cee30e.

## Related Issues

#5436

## Testing

Manually tested in personal cluster and tested using companies CICD pipeline which includes emissary deployment.

There is also:
https://github.com/emissary-ingress/emissary/blob/f76e5d46427cc02b88966f4d7ad223fbdcce756d/python/tests/integration/manifests/knative_serving_0.18.0.yaml#L157-L159 which already includes the change and likely why issue not caught as part of test cases.

## Checklist

- [ ] **Does my change need to be backported to a previous release?**
  - What backport versions were discussed with the Maintainers in the Issue?

- [ ] **I made sure to update `CHANGELOG.md`.**

   Remember, the CHANGELOG needs to mention:
  - Any new features
  - Any changes to our included version of Envoy
  - Any non-backward-compatible changes
  - Any deprecations

- [x] **This is unlikely to impact how Ambassador performs at scale.**

   Remember, things that might have an impact at scale include:
  - Any significant changes in memory use that might require adjusting the memory limits
  - Any significant changes in CPU use that might require adjusting the CPU limits
  - Anything that might change how many replicas users should use
  - Changes that impact data-plane latency/scalability

- [ ] **My change is adequately tested.**

   Remember when considering testing:
  - Your change needs to be specifically covered by tests.
    - Tests need to cover all the states where your change is relevant: for example, if you add a behavior that can be enabled or disabled, you'll need tests that cover the enabled case and tests that cover the disabled case. It's not sufficient just to test with the behavior enabled.
  - You also need to make sure that the _entire area being changed_ has adequate test coverage.
    - If existing tests don't actually cover the entire area being changed, add tests.
    - This applies even for aspects of the area that you're not changing – check the test coverage, and improve it if needed!
  - We should lean on the bulk of code being covered by unit tests, but...
  - ... an end-to-end test should cover the integration points

- [x] **I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently.**

- [ ] **The changes in this PR have been reviewed for security concerns and adherence to security best practices.**

PR #5468, opened 2023-12-06
apiext: add rbac for updating crd statuses

LanceEa:
## Description

On some locked down clusters, the status update for CRD's will fail due to missing RBAC. This updates the fix-crds tool so that it includes `customresourcedefinitions/status` in the RBAC permissions when updating CRDs. AFAIK, it seems like it will report an error but still update status in our CI tests which is why we didn't catch this.

This supersedes https://github.com/emissary-ingress/emissary/pull/5449 because the file changed in that PR is generated and would get overridden as soon as make generate was called again. The file is generated via the `tools/src/fix-crds` using the template defined here: `tools/src/fix-crds/apiext.yaml` so that has been updated so that `make generate` is clean now.

## Related Issues

Fixes https://github.com/emissary-ingress/emissary/issues/5436

## Testing

CI Tests are still passing. Manually testing from user as noted here: https://github.com/emissary-ingress/emissary/pull/5449#issue-2002048514

## Checklist

### DIFF
--- a/manifests/emissary/emissary-crds.yaml.in
+++ b/manifests/emissary/emissary-crds.yaml.in
@@ -5238,7 +5238,7 @@ rules:
     resources: [ "customresourcedefinitions" ]
     verbs: [ "list", "watch" ]
   - apiGroups: [ "apiextensions.k8s.io" ]
-    resources: [ "customresourcedefinitions" ]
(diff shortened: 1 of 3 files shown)


joebowbeer:
@AliceProxy Why hasn't this been released?

PR #5583, opened 2024-02-26
module: add listener_max_connection_lifetime_ms option

### LINKED ISSUE #2900: Upper bound downstream connection lifetime in ambassador
**Summary**
In grpc server there is max_connection_age property which you can set . This enables the server to close connections after certain time .  Similar to this max_connection_duration property is there in envoy proxy .  Please refer to below links for more details . 

https://github.com/envoyproxy/envoy/pull/8591
https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/protocol.proto

Is there similar configuration that i can do in ambassador ? 

**Why**
Grpc doesn't honor DNS TTL . Which leads to improper load balancing during upscale / downscale . Another issue it leads to is you can't switch traffic using dns . In order to solve this issue we tried using max_connection_age in our grpc servers . But as ambassador sits in front of our grpc services and our clients connect to ambassador directly , we are not able to solve this problem . If ambassador provides this support of setting maximum connection age , our problem will get resolved . Please suggest how to do this in ambassador

dethi:
## Description

Add listener_max_connection_lifetime_ms, similar to how listener_idle_timeout_ms is implemented. This limit the maximum age of a downstream connections by adding max_connection_duration to the listener's common_http_protocol_options.

This option could already be set on the clusters (upstream connections), but setting it on the downstream connections could be useful in multiple scenarios:

- to periodically rebalance the load between multiple emissary-ingress, especially when most of the clients are using HTTP2/gRPC and connection stick for a really long time.

- when using mTLS (client certificate), to ensure that clients reconnect with their renewed certificate, as Envoy doesn't close the connection when the certificate expires.

## Related Issues

Fix #2900

## Testing

- Added automated tests
- Not yet deployed in a real cluster

## Checklist

  - No, this is a new feature.

  - Any new features
  - Any deprecations

### DIFF
--- a/python/ambassador/envoy/v3/v3listener.py
+++ b/python/ambassador/envoy/v3/v3listener.py
@@ -523,6 +523,15 @@ class V3Listener:
                     "idle_timeout": "%0.3fs" % (float(listener_idle_timeout_ms) / 1000.0)
                 }
 
+        listener_max_connection_lifetime_ms = self.config.ir.ambassador_module.get(
+            "listener_max_connection_lifetime_ms", None
+        )
+        if listener_max_connection_lifetime_ms:
+            common_http_options = base_http_config.setdefault("common_http_protocol_options", {})
+            common_http_options["max_connection_duration"] = "%0.3fs" % (
+                float(listener_max_connection_lifetime_ms) / 1000.0
+            )
+
         if "headers_with_underscores_action" in self.config.ir.ambassador_module:
             if "common_http_protocol_options" in base_http_config:
                 base_http_config["common_http_protocol_options"][
--- a/python/ambassador/ir/ir.py
+++ b/python/ambassador/ir/ir.py
@@ -1160,6 +1160,9 @@ class IR:
         od["listener_idle_timeout_ms"] = self.ambassador_module.get(
             "listener_idle_timeout_ms", None
         )
+        od["listener_max_connection_lifetime_ms"] = self.ambassador_module.get(
+            "listener_max_connection_lifetime_ms", None
+        )
         od["headers_with_underscores_action"] = self.ambassador_module.get(
             "headers_with_underscores_action", None
         )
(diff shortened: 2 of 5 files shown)


dethi:
Two questions:
- any wording recommendation for the CHANGELOG
- how can I update the doc?

rgs1:
@dethi any chance you could rebase this? It might be nice to support this on upstream connections too, since the same type of issues apply there as well (e.g.: stickiness, ensuring mTLS still works, etc). Thanks!

rgs1:
@kflynn any chance we could get this one reviewed/merged? It solves a couple of issues we've seen in prod...

dethi:
Upstream option already exist today.

I can try to rebase, but I think I was missing something in this change as well for the CRD side of things. My testing kind of stopped when I saw that there was no maintainer activity on the repo.

rgs1:
> Upstream option already exist today.

Oh nice, missed that (https://github.com/emissary-ingress/emissary/blob/master/python/ambassador/envoy/v3/v3cluster.py#L122 for others wondering where it is). 


> I can try to rebase, but I think I was missing something in this change as well for the CRD side of things. My testing kind of stopped when there wasn’t much maintainers activity on PR.

Yeah maybe the commit that added support for upstream is a good reference wrt what is needed for CRDs:

https://github.com/emissary-ingress/emissary/commit/69bdabf2f3d26a1e40c5f64b27cc6cf4950eeba8


> But I can try to get back to it. Would be nice to have the CI test running to see what’s missing.

cc: @kflynn or @cindymullins-dw to kick CI 🙏

dethi:
Rebased the change on master. Thanks for the example for the implementation. I took a bit of time to dig into my change again and now I remember where I was left confused last time: `listener_idle_timeout_ms` which I was taking as inspiration for my change (since it's in the same Envoy config section and applies to listener too) doesn't show up at all in the CRD yaml. It's only present in the Python code.

Now with the commit you shared, I took a look at the Go section and from what I can see, the [Module CRD accept a `config` key that is completely untyped](https://github.com/emissary-ingress/emissary/blob/master/pkg/api/getambassador.io/v2/crd_module.go#L29). Thus why I couldn't find it last time. So based on this, I don't think I need to change anything else.

I will get it to run on one of our dev clusters to validate the config, and if we can get the CI tests to run as well.

PR #5684, opened 2024-05-16
deps: upgrade envoy to 1.30.1 (patched)

AliceProxy:
## Description

Upgrades Envoy to 1.30.1 with our custom commits. Lots of changed files since we're going from 1.27.2 -> 1.30.1.

Screenshot from 2024-05-16 09-20-05

(two tests from that screenshot since they were known flakes that had to be re-ran, the rest already passed and were just cached)


AliceProxy:
I initially cherry-picked Lance's old 1.28 commits to help since I was running into issues. They aren't needed in this branch anymore, but since they also aren't hurting anything, I was just going to leave them in unless anyone cares since the rebase to drop them would be annoying.

rick-a-lane-ii (approved):

kflynn (approved):
Wow, what a mess! Yeah, let's get this landed. Thanks!!

PR #5691, opened 2024-05-28
Switch to Envoy 1.30.1

kflynn:
Switch to Envoy 1.30.1 (and pull in some other dependency bumps from `main` as well)

- **Merge master for dependency bumps**
Update Envoy to 1.30.1.

Update Go to 1.21.10 in the base-python image (which, uh, we don't really use in this branch!).

Update Go packages:
  golang.org/x/sys v0.15.0 => v0.18.0
  google.golang.org/protobuf v1.32.0 => v1.33.0

Update Python packages:
  blinker 1.7.0 => 1.8.2
  durationpy 0.5 => 0.6
  flask 3.0.0 => 3.0.3
  gunicorn 21.2.0 => 22.0.0
  idna 3.4 => 3.7
  packaging 23.2 => 24.0
  typing-extensions 4.8.0 => 4.11.0
  urllib3 2.1.0 => 2.2.1

PR #5692, opened 2024-05-28
[BREAKING CHANGE] 4.x chart mod

kflynn:
https://github.com/emissary-ingress/emissary/pull/5691 needs to land before this.

Four significant changes here:

1. Make the chart version match the image version.

This will appear to be a conflict with existing Emissary charts, but the 4.x charts and the older charts don't live in the same Helm repository, so it doesn't matter.

**NOTE WELL**: this _may_ be something we need to change later, since right now a Helm-chart-only change will require making new binaries. The goal here is that releases get to be sufficiently lightweight that we don't care.

2. Also, change the default naming so that we don't end with insane things like a Service named "emissary-ingress-emissary-chart-admin".

3. _Also_, change the `app.kubernetes.io/name` annotation _not_ to default to `.Chart.Name` -- https://helm.sh/docs/chart_best_practices/labels/ differentiates this from the `helm.sh/chart` annotation, so use Emissary's name for the `app.kubernetes.io/name` annotation.

4. Ditch the `a8r.io/*` annotations that were left over from older Ambassador Cloud stuff.

Also, I cleaned up the chart README some. Holy _crap_ that was out of date.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -14,7 +14,7 @@ before:
     - rm -rf python/ambassador.version
     - make VERSION=v{{ .Version }} python/ambassador.version
     # Make the chart as part of the build process.
-    - make VERSION={{ .Version }} CHART_VERSION={{ .Version }} chart
+    - make VERSION=v{{ .Version }} chart
 
 builds:
   - id: busyambassador
--- a/Makefile
+++ b/Makefile
@@ -27,16 +27,21 @@ ifneq ($(MAKECMDGOALS),$(OSS_HOME)/build-aux/go-version.txt)
   endif
 
   VERSION := $(or $(VERSION),$(shell go run ./tools/src/goversion))
-  $(if $(or $(filter v3.%,$(VERSION)),$(filter v0.0.0-%,$(VERSION))),\
-    ,$(error VERSION variable is invalid: It must be v3.* or v0.0.0-$$tag, but is '$(VERSION)'))
+  $(if $(or $(filter v4.%,$(VERSION)),$(filter v0.0.0-%,$(VERSION))),\
+    ,$(error VERSION variable is invalid: It must be v4.* or v0.0.0-$$tag, but is '$(VERSION)'))
   $(if $(findstring +,$(VERSION)),\
     $(error VERSION variable is invalid: It must not contain + characters, but is '$(VERSION)'),)
   export VERSION
 
-  CHART_VERSION := $(or $(CHART_VERSION),$(shell go run ./tools/src/goversion --dir-prefix=chart))
-  $(if $(or $(filter v8.%,$(CHART_VERSION)),$(filter v0.0.0-%,$(CHART_VERSION))),\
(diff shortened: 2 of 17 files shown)


ppeble (approved):

python/tests/integration/manifests/rbac_namespace_scope.yaml
```diff
@@ -35,7 +35,7 @@ metadata:
   labels:
     app.kubernetes.io/instance: kat-rbac-singlenamespace
     app.kubernetes.io/managed-by: kat
-    app.kubernetes.io/name: emissary-ingress
+    app.kubernetes.io/name: emissary-chart
```
AliceProxy, line 38: using `emissary-chart` for all these seems weird, but otherwise seems ok

AliceProxy (approved):

chen-anders (approved):

python/tests/integration/manifests/rbac_namespace_scope.yaml
kflynn: Yeah, you're not wrong. Here was my thinking:

First, we're publishing a few different artifacts that are all in `oci://ghcr.io/emissary-ingress`. Having them all be in this registry makes sense. Calling the main Docker image within that registry `emissary` also makes sense.

So... what do we call the charts? We're going to have two: the main chart and the CRD chart. In the non-OCI world, the chart was named "emissary-ingress" and so it set this label to "emissary-ingress". In the OCI world, it really doesn't make sense to have a Docker image named `emissary` and a chart named `emissary-ingress`, though. I ended up settling on naming the main chart `emissary-chart` and the CRD chart `emissary-crd-chart` -- and that implies that this label should change.

We could also borrow from Linkerd, and have charts named `emissary-crds` and `emissary-control-plane`. I'm not sure I really like that, either, since the main Emissary chart doesn't just install the control plane...

So, yeah, how strongly do you feel about that? 🙂 It is definitely _possible_ to force these labels back to `emissary-ingress`, I just don't think it makes sense if the chart isn't named that...

AliceProxy: Yeah no, we've chatted about that before and that all makes sense with the chart naming. I mean specifically the

```yaml
 app.kubernetes.io/name: emissary-chart
 app.kubernetes.io/part-of: emissary-ingress
```

It feels a little weird to use these labels for helm stuff since  the deployment or other k8s resoruces aren't called `emissary-chart`. I just meant more like, should we stick to label conventions as prescribed by https://helm.sh/docs/chart_best_practices/labels/?

kflynn: D'oh! Thank you, I _completely_ misunderstood what you were getting at because I wasn't paying enough attention to the name of the label. 🤦‍♂️ Yes, we should indeed follow the Helm best practices. Lemme fix that...

chen-anders (approved):

AliceProxy (approved):

ppeble (approved):

PR #5694, opened 2024-05-30
Chainsaw, part 1

kflynn:
This is part 1 of taking a chainsaw to a bunch of stuff we don't need. There's more to come.

**NOTE**: you'll need to have a `v4.0.0-test.0` tag to do much with these changes, but please don't push such a tag yet.

There are three commits in this PR. I **strongly** recommend reading the review notes below.

- **[WIP] Chainsaw: ditch the _cxx directory and start ditching machinery around it**

This is a monster commit. **DON'T** try to review it in the GitHub UI; there are some 1200 deleted files, since this commit deletes the `_cxx`, `api/envoy`, and `pkg/envoy-control-plane` directories completely: we simply don't need these if we're no longer building our own Envoy. Of course, that means that we need to change a bunch of Go imports to use the stock Envoy packages rather than the ones we compile.

To see just the _changes_ in this commit, the easiest way is probably

```bash
git show 6fb669364bd7c70fcffb92afd2da7f6155575bea -- \
$(git show --name-only 6fb669364bd7c70fcffb92afd2da7f6155575bea | fgrep -v '_cxx
api/envoy
pkg/envoy-control-plane')
```

- **[WIP] Chainsaw: Ditch base-python, base-pip, and base-envoy.**

We also don't need to screw around with the `base-python`, `base-pip`, and `base-envoy` Docker images at all. `base-envoy` goes away because we're not building our own Envoy; `base-python` and `base-pip` go away because using distroless rather than Alpine simplifies the world enough that the complexity of these two images isn't worth it: I can do a full rebuild in about five minutes on my laptop.

This commit is OK to review in GitHub, it's only a dozen files or so.

- **[WIP] Chainsaw: use goreleaser to build all the Docker images**

Finally, this commit switches to using goreleaser to build other Docker images. I _think_ this gets the ones we need to actually run tests, but maybe I'm wrong.

Once this commit lands, then `goreleaser release` should build multiarch images for everything.


AliceProxy (approved):
love goreleaser :rocket:

chen-anders (approved):

the-wondersmith:
Been a long time coming, _awesome_ to see this happening



<br/><br/>

> ... taking a chainsaw to a bunch of stuff we don't need. ...

PR #5697, opened 2024-06-10
add ghcr badge

ipitio:
ghcr.io's api doesn't let you see the pull count so I created ghcr-pulls, a daily scraper that updates the count and makes this badge possible. I noticed that you use ghcr and have a Docker Hub badge and thought you might find this useful.


the-wondersmith:
@ipitio I'm not 100% sure when it was added, but the [`README`](https://github.com/emissary-ingress/emissary/blob/main/README.md?plain=1#L15) does now have a GHCR pull count badge. Closing this as "uncertain of when, but done nevertheless" 😅

PR #5708, opened 2024-06-27
Chainsaw, part 2

kflynn:
Chainsaw, part 2!

This one is probably best reviewed commit by commit.

- 069f35a9: [WIP] Chainsaw: Restore IRErrorResponse imports since the code that needs it wasn't removed yet. Sigh.

    I was a little hasty in Chainsaw, part 1, when I ripped out a bunch of imports related to the error-rewriting code that's not supported in Emissary 4 _without_ actually ripping out the code using those imports -- d'oh. This puts the imports back so that unit tests have a hope of running. (Ripping out the feature code will come in part 3 of the chainsaw commits).

- 092d98b7: [WIP] Chainsaw: Rip out tons of docker.mk stuff.

    This takes the chainsaw to the rest of the Docker build system and uses Goreleaser to build our ancillary Docker images instead. I'm not delighted by this, since it makes Goreleaser take five minutes to run on my Mac, but it's a starting point.

- f0d598ad: [WIP] Chainsaw: rip out vestiges of Edge Stack

    This takes the chainsaw to all the "if edgestack" stuff in Emissary, including (at least most of the) stuff related to the intercept agent, Edge Stack gofilters, Edge Stack's built-in internal auth & ratelimit services, and the concept that Emissary called a sidecar and was really arbitrary binaries that could be included in the image and  would get started at boot time.

    This commit does _not_ entirely rip out the ACME holepuncher, because we might want that later.

- 6f30e3b7: [build] Make the Envoy image a build arg defined in .goreleaser.yaml

    Defining build variables in .goreleaser.yaml to me makes more sense than having them hardcoded in the Dockerfiles.

- c7a37908: Use the correct Docker image for the Gateway API transform test -- but disable it on Mac because Docker networking is annoying.

    Putting the Envoy base image in .goreleaser.yaml makes it pretty simple for the Gateway API transform test, which wants to run Envoy in Docker, to find the right image. Unfortunately the test doesn't work on my Mac because of Docker networking madness, so skip it on Mac for the moment.
 
- 5c3f40c2: Switch away from setuptools over to importlibs for finding schemas and templates, so Python unit tests can run.

    Finally, we used to be using a deprecated setuptools API to let the Python code find out where to look for schemas and templates at runtime. Switch to the supported importlib alternative.

After all that, Python and Go unit tests pass. To verify that for yourself:

```
make pytest-unit
make gotest
```

My basic functional test also passes, so this PR probably leaves us in mostly-tolerable shape to go after E2E tests locally and in CI.

(Note that there's at least one more chainsaw PR coming, to finish ripping out all the stuff related to error rewrites and header casing, and to do a closer review of things I commented out rather than deleting in this PR.)


python/ambassador/envoy/v3/v3listener.py
```diff
@@ -906,15 +906,15 @@ def compute_http_routes(self) -> None:
...
-                        elif (
-                            self.config.ir.edge_stack_allowed
-                            and (route_precedence == -1000000)
-                            and (
-                                rv.route["match"].get("safe_regex", {}).get("regex", None) == "^/$"
-                            )
-                        ):
-                            extra_info = " (force Route for fallback Mapping)"
-                            action = "Route"
+                        # elif (
```
AliceProxy, line 909: Leaving this here to come back to?

AliceProxy (approved):
LGTM, much cleaner, especially without the edge stack logic in the tests.

.goreleaser.yaml
```diff
@@ -222,6 +227,66 @@ dockers:
     extra_files:
       - docker/test-stats
 
+  ## kat-client
+  - use: buildx
+    goos: linux
+    goarch: arm64
```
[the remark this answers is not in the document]
kflynn: 🤦‍♂️ Somehow I thought that was only a paid Goreleaser feature, augh. Thanks! I'll use that in a later PR.

python/ambassador/envoy/v3/v3listener.py
kflynn: More accurately leaving it to take a closer look at to see if there's anything there that would be smart to keep. Next chainsaw PR for that stuff. 🙂

chen-anders (approved):

the-wondersmith:
LGTM as well. For historical reasons I'll voice a concern that the CI is just completely wrekt here, but I ran that past @kflynn and it's (currently) a known and expected state of affairs that is slated to be rectified in the next pass of chainsawing.

tl;dr 👍✅

kflynn:
To be clear, I think CI is going to take more than just a chainsaw. 😉 But I think that this round of chainsawing gets us to a place where it's sane to rebuild CI.

Issue #5714, opened 2024-07-03
Incorrect Cache Key for Mapping


sekar-saravanan:
**Describe the bug**
_cache_key is getting generated incorrectly for mappings in ir.json, when using header with regex.

**Steps to reproduce the behavior:**
Create mapping with regex header (attached mapping.yaml)
Fetch ir.json generated by emissary (attached sample ir.json)
Check the _cache_key of the mappings inside the groups block.
Always _cache_key should be in the format of [{kind}-{version}-{name}-{namespace}](https://github.com/emissary-ingress/emissary/blob/master/python/ambassador/ir/irbasemapping.py#L166-L181). But header_name is getting updated in the place of mapping name 

**Imapct**
If we plan to use the _cache_key in any feature ticket, it could get impacted.

**Expected behavior**
Always _cache_key should be in defined format. {kind}-{version}-{name}-{namespace}

**Versions (please complete the following information):**
 - Ambassador: 3.9.1
 - Kubernetes environment: EKS
 - Version: 1.28.9

**Additional context**
Attached the mapping.yaml and ir.json files.
[incorrect-cache-key.zip](https://github.com/user-attachments/files/16076855/incorrect-cache-key.zip)

sekar-saravanan:
Fix added for this issue: https://github.com/emissary-ingress/emissary/pull/5715/files

Issue #5702, opened 2024-06-20
Emissary creating duplicate routes in ir.json and econf.json


sekar-saravanan:
**Describe the bug**
Emissary creating duplicate routes in ir.json and econf.json while adding weight for canary release

**To Reproduce**
1. Create mapping for original and canary release with weight. It will recreate the ir.json and econf.json. (No duplicates found)
2. Do perform update on any other existing mapping. It will recreate the ir.json and econf.json. (Now we can see the duplicate routes for above created mapping)

**Expected behavior**
Emissary should not create duplicate routes in ir.json and econf.json for canary release

**Versions (please complete the following information):**
 - Ambassador: [3.9.2]
 - Kubernetes environment [EKS]
 - Version [1.28.9]

**Additional context**
Have attached the deployment.yaml, service.yaml, mappings.yaml, host.yaml and ir.json and econf.json with duplicate routes.

Duplicate routes presents under,
ir.json: -> groups -> mappings
econf.json: -> listeners -> filter_chains -> filters -> route_config -> virtual_hosts -> routes
[docs.zip](https://github.com/user-attachments/files/15910230/docs.zip)

**What is the Impact ?**
No impacts on traffic. 

**Other Imapcts:**
- Getting IR MISMATCH & ENVOY CONFIG MISMATCH error frequently in the logs, whenever updating any of the mapping object. [[https://github.com/emissary-ingress/emissary/blob/master/python/ambassador_diag/diagd.py#L545](https://github.com/emissary-ingress/emissary/blob/master/python/ambassador_diag/diagd.py#L545), [https://github.com/emissary-ingress/emissary/blob/master/python/ambassador_diag/diagd.py#L551](https://github.com/emissary-ingress/emissary/blob/master/python/ambassador_diag/diagd.py#L551)]
- Taking cpu consumption while getting json_diff. Sometime getting 100% cpu spike when running with bunch of mappings and which leads to envoy failure [[https://github.com/emissary-ingress/emissary/blob/master/python/ambassador_diag/diagd.py#L547](https://github.com/emissary-ingress/emissary/blob/master/python/ambassador_diag/diagd.py#L547), [https://github.com/emissary-ingress/emissary/blob/master/python/ambassador_diag/diagd.py#L553](https://github.com/emissary-ingress/emissary/blob/master/python/ambassador_diag/diagd.py#L553)]
[...]

PR #5726, opened 2024-07-20
v4 fixes for 5702 and 5714

kflynn:
This is @sekar-saravanan's fixes for #5702 and #5714 combined with my test cases.

Reviewers, this is **definitely** best reviewed commit by commit.

Note also that **irtype.go is known to be incomplete** -- but, as demonstrated, it's already good enough to do useful testing work!

Finally, note also that I have _not_ tried to backport this work to the v3 branch. It's definitely doable, but since this involves a couple of chainsaw commits, there'll be some weirdness around the Go import paths, I'm sure.


kflynn:
(I initially submitted this as a PR against `master`, which is why the labels were so disturbing at first. 🙂)

kflynn:
@sekar-saravanan and @the-wondersmith, it'd be good for y'all to look at this too.

cmd/entrypoint/irtype.go
```diff
@@ -0,0 +1,177 @@
...
+type IRClusterTarget struct {
+	IP         string `json:"ip"`
+	Port       int    `json:"port"`
+	TargetKind string `json:"target_kind"`
+}
+
+type IRCluster struct {
+	IRResource
+	BarHostname      string               `json:"_hostname"`  // Why this _and_ hostname?
+	BarNamespace     string               `json:"_namespace"` // Why this _and_ namespace?
```
AliceProxy, line 33: I don't like this but I get why it's here for now

AliceProxy (approved):
seems fine to me. I'm not going to bother nit picking until CI is in a place where tests are passing and reliable indicators again and when the PR for merging this branch into main comes.

cmd/entrypoint/irtype.go
kflynn: You and me both. I'll figure out what up with this soon.

PR #5731, opened 2024-08-01
Add Helm Chart Support for Opt-In Support for Exposing Port 8878 in the AdminService

w-h37:
…in for access to the Go Plugin Filter Metrics

## Description

Adding support in the Helm chart to allow an Opt-In feature for users to configure a new port to open in the adminService. This new port will allow users to set `Values.adminService.goPluginMetricsPort=8878` to expose port `8878` for users to get access to the Go Filter Plugin Metrics, which we added in Edge Stack release `3.11.0`. This allows users to expose an additional port so that they may use metric scraping tools to scrape the new Go Plugin metrics using Helm.

## Related Issues

List related issues.

## Testing

Running `helm install --dry-run --debug <LOCAL-CHART>` has the expected default output, where `goPluginMetricsPort` inside the adminService file is empty, the adminService remains unchanged, and emissary-ingress installs with no changes. 

Running `helm install --dry-run --debug <LOCAL-CHART> --set adminService.goPluginMetricsPort=8878` has the expected output where the `goPluginMetricsPort` field inside the adminService has the value `8878`, the environment variable `ENABLE_PLUGIN_FILTER_METRICS` is set to `"true"`, port `8878` is opened on the emissary-ingress deployment, and the adminService has the port `8878` exposed on it. 

## Checklist

  - Any new features
  - Any deprecations

### DIFF
--- a/charts/emissary-ingress/templates/admin-service.yaml
+++ b/charts/emissary-ingress/templates/admin-service.yaml
@@ -32,6 +32,12 @@ spec:
       {{- if (and (eq .Values.adminService.type "NodePort") (not (empty .Values.adminService.nodePort))) }}
       nodePort: {{ int .Values.adminService.nodePort }}
       {{- end }}
+      {{- if .Values.adminService.goPluginMetricsPort}}
+    - port: {{ .Values.adminService.goPluginMetricsPort }}
+      targetPort: 8878
+      protocol: TCP
+      name: go-plugin-filter-metrics
+      {{- end }}
     - port: {{ .Values.adminService.snapshotPort }}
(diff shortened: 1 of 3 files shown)


kflynn:
In the short term, I think lint should be made to pass and then we should take this...

In the longer term, adding things to the Emissary chart for Edge Stack features feels weird to me at this point. 😐 Or am I mistaken and this is a thing an Emissary user can get value from?

kflynn (approved):
Taking this one, but let's not add any more Edge-Stack-only things into the charts (or the code, really).

PR #5733, opened 2024-08-06
Re-adding HostIP to helm charts and deployments

tenshinhigashi:
## Description

There was some friction in the upgrade process when users noticed that hostIP was missing. This reduces that friction.

### DIFF
--- a/charts/emissary-ingress/templates/deployment.yaml
+++ b/charts/emissary-ingress/templates/deployment.yaml
@@ -216,6 +216,10 @@ spec:
             - name: admin
               containerPort: {{ .Values.adminService.port }}
           env:
+            - name: HOST_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
             {{- if .Values.prometheusExporter.enabled }}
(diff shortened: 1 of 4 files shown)


kflynn:
Hey @tenshinhigashi! What was the friction?

kflynn:
(It's kinda tough to see this being a bad idea, but I'm very curious where you're seeing it get used.)

tenshinhigashi:
@kflynn There was friction with users who use Datadog (and I think maybe other tracing services). Our docs use `HOST_IP` in the tracing services, and thus users who followed our docs had invalid deployments when upgrading. Im updating the getambassador docs to also include a message around this as well.

https://www.getambassador.io/docs/emissary/latest/howtos/tracing-datadog#3-configure-the-tracingservice

kflynn:
Ugh, got it. Thanks!

kflynn:
...how long has `make lint` been failing? That should really be fixed before we start landing stuff. 🙂

kflynn (approved):
Hit it!

PR #5761, opened 2024-08-09
dev(chainsaw): refactor ci

the-wondersmith:
PR deprecates the existing CI workflow(s) in favor of a dramatically simplified version primarily relying on [goreleaser](https://goreleaser.com/intro/).

Notes:
  - PR is not actually feature complete, and is being made at @kflynn's request
    in order to push progress forward
  - the `build` stage of the new workflow still needs work
    - it *does* build emissary binaries, but nothing else is currently done with them
    - a post-build hook needs to be added to the `.goreleaser.yml` so that wheels
      for all the packages in the `python/` directory are built
  - the `release` stage of the new workflow still needs work
    - the stage *may* end up being omitted entirely, depending on further fiddling
      with the [goreleaser configuration](https://goreleaser.com/customization/release/#github)
  - all existing workflows have been moved to `.github/workflows/deprecated`
    - all `on` values except for `workflow_dispatch` have been commented out
      on the deprecated workflows so that they simply no longer run automatically
      but can still be triggered manually if need be while the new CI process is
      being worked out

There's *probably* more I'm forgetting here, but I trust that anything not mentioned will be worked out in the comments here 😅


kflynn (approved):
Looks good for building on, many thanks!! 🙂

PR #5762, opened 2024-08-09
Allow CI to run on dev/v4 only for now, and make the "we would release here" step succeed so we can start enjoying green checkmarks.

kflynn:
Those green checkmarks won't be too meaningful, but they _will_ unblock things and allow us to start paying attention to red things again.


kflynn:
(Going to recreate this to see if Circle is gone.)

PR #5763, opened 2024-08-09
Activate CI for dev/v4

kflynn:
Three changes here:

- Allow CI to run on dev/v4 only for now
- Make the "we would release here" step succeed so that hopefully CI will pass
- Lint fixes

Having CI pass is mostly about starting to pay attention when it doesn't pass.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -68,7 +68,7 @@ jobs:
       - name: Create Python Virtual Env
         shell: bash
         run: >-
-          make python-dev-setup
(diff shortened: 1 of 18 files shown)


the-wondersmith (approved):
Can't *believe* I missed a couple of these things 🥲

LGTM though

kflynn:
The weird thing to me is that I had to mess with the unit tests execution, because it worked fine the old way on my machine?? But having a place to stand and iterate on is good enough for now.

ppeble (approved):

PR #5764, opened 2024-08-10
ruff prep

kflynn:
I've been messing with uv and ruff (thanks, @the-wondersmith!) and wow, they are fast enough to legit be game-changers. This PR cleans up the Python code in line with what ruff wants.

Reviewers: going commit by commit will probably be simpler but it's not that big a deal.

- **Clean up imports and __init__.py's**
- **Clear up typing for some booleans**
- **Only have one copy of is_ip_address**
- **Shut up regex warning (even though irerrorresponse.py needs to go away)**
- **Clean up some things ruff doesn't like**
- **Ditch dead code**
- **Clean up is, is not, not in**
- **Drop unneeded f-strings**
- **Clean up bare excepts**

### DIFF
--- a/python/ambassador/compile.py
+++ b/python/ambassador/compile.py
@@ -31,6 +31,10 @@ class _CompileResult(TypedDict):
     xds: NotRequired[EnvoyConfig]
 
 
+def null_file_checker(path: str) -> bool:
+    return True
+
+
 def Compile(
     logger: logging.Logger,
     input_text: str,
@@ -56,7 +60,7 @@ def Compile(
     """
 
     if not file_checker:
-        file_checker = lambda path: True
+        file_checker = null_file_checker
 
     if not secret_handler:
         secret_handler = NullSecretHandler(logger, None, None, "fake")
--- a/python/ambassador/config/config.py
+++ b/python/ambassador/config/config.py
@@ -241,7 +241,7 @@ class Config:
             # Make sure it's a list. Yes, this is Draconian,
             # but the jsonschema will allow only a string or a list,
             # and guess what? Strings are Iterables.
-            if type(allowed_ids) != list:
+            if type(allowed_ids) is not list:
                 allowed_ids = typecast(StringOrList, [allowed_ids])
 
             if Config.ambassador_id in allowed_ids:
--- a/python/ambassador/config/config.py
+++ b/python/ambassador/config/config.py
@@ -281,7 +281,7 @@ class Config:
         the set of ACResources to be sorted in some way that makes sense.
         """
 
-        self.logger.debug(f"Loading config")
+        self.logger.debug("Loading config")
 
         rcount = 0
 
--- a/python/ambassador/config/config.py
+++ b/python/ambassador/config/config.py
@@ -430,7 +430,7 @@ class Config:
 
             try:
                 handler(resource)
-            except Exception as e:
+            except Exception:
                 # Bzzzt.
                 raise
                 # return RichStatus.fromError("%s: could not process %s object: %s" % (resource, resource.kind, e))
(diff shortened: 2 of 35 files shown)


python/ambassador/config/config.py
```diff
@@ -241,7 +241,7 @@ def good_ambassador_id(self, resource: dict):
             # Make sure it's a list. Yes, this is Draconian,
             # but the jsonschema will allow only a string or a list,
             # and guess what? Strings are Iterables.
-            if type(allowed_ids) != list:
+            if type(allowed_ids) is not list:
```
the-wondersmith, line 244:
```suggestion
            if not isinstance(allowed_ids, list):
```

python/ambassador/compile.py
```diff
@@ -31,6 +31,10 @@ class _CompileResult(TypedDict):
     xds: NotRequired[EnvoyConfig]
 
 
+def null_file_checker(path: str) -> bool:
```
the-wondersmith, line 34:
```suggestion
def null_file_checker(_: str) -> bool:
```

Tiny nit - unused variables shouldn't be bound to names

python/ambassador/config/config.py
```diff
@@ -430,7 +430,7 @@ def process(self, resource: ACResource) -> RichStatus:
 
             try:
                 handler(resource)
-            except Exception as e:
+            except Exception:
```
the-wondersmith, line 433: If nothing is being done here, why is the exception being caught? Usually, you'd expect something like:

```python
try:
    handler(resource)
except Exception as error:
    raise SomeCustomError from error
```

So that the traceback and context are preserved.

Haven't had time to dive into the full context here so it's a genuine question, not a criticism or anything 😅

python/ambassador/envoy/v3/v3_static_resources.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, config: "V3Config") -> None:
 
         self.update(
             {
-                "listeners": [l.as_dict() for l in config.listeners],
+                "listeners": [x.as_dict() for x in config.listeners],
```
the-wondersmith, line 27:
```suggestion
                "listeners": [listener.as_dict() for listener in config.listeners],
```

Absolutely a tiny nit, please feel 100% free to ignore it, just a personal preference for short-but-descriptive variable names over generic or shortest-possible ones.

python/ambassador/envoy/v3/v3route.py
```diff
@@ -406,13 +406,13 @@ def __init__(
 
         request_headers_to_remove = group.get("remove_request_headers", None)
         if request_headers_to_remove:
-            if type(request_headers_to_remove) != list:
+            if type(request_headers_to_remove) is not list:
```
the-wondersmith, line 409:
```suggestion
            if not isinstance(request_headers_to_remove, list):
```

python/ambassador/envoy/v3/v3route.py
```diff
@@ -406,13 +406,13 @@ def __init__(
...
                 request_headers_to_remove = [request_headers_to_remove]
             self["request_headers_to_remove"] = request_headers_to_remove
 
         response_headers_to_remove = group.get("remove_response_headers", None)
         if response_headers_to_remove:
-            if type(response_headers_to_remove) != list:
+            if type(response_headers_to_remove) is not list:
```
the-wondersmith, line 415:
```suggestion
            if not isinstance(response_headers_to_remove, list):
```

python/ambassador/ir/ircors.py
```diff
@@ -68,7 +68,7 @@ def _cors_normalize(value: Any) -> Any:
         are returned unaltered.
         """
 
-        if type(value) == list:
+        if type(value) is list:
```
the-wondersmith, line 71:
```suggestion
        if isinstance(value, list):
```

the-wondersmith (changes requested):

PR #5777, opened 2024-08-16
chore(build/ci): switch to uv and ruff

the-wondersmith:
PR completes the "migration" to fully using [`uv`](https://docs.astral.sh/uv/) and [`ruff`](https://docs.astral.sh/ruff/) for python environment management and linting/formatting (respectively).

There are _numerous_ changes to the `.py` files, (and they _should_ be eyeballed at the very least), but they're all just linting fixes to get the python side of things "up to snuff" with ruff's configured lints.

The important parts here that actually require significant scrutiny are the changes to all the non-`.py` files. If possible, I think it would be ideal for reviewers to pull down the branch locally into a "fresh" directory and attempt to run through "common" dev workflows to ensure everything still works as intended.

Also, it's worth noting that the CI workflow _does_ install `uv` and `ruff` automatically, but I don't think anything in any of the makefiles does the same for non-ci environments. Presumably that should be rectified prior to merging the branch 🤔?  Reviewers, please chime in on it.


python/ambassador-diag/src/ambassador_diag/diagd.py
```diff
@@ -64,9 +79,7 @@
     parse_bool,
     parse_json,
 )
-
-if TYPE_CHECKING:
-    from ambassador.ir.irtlscontext import IRTLSContext  # pragma: no cover
+from ambassador_diag.templates import TEMPLATE_PATH
```
kflynn, line 82: When I run `gmake lint/mypy` in this branch _before_ you make this edit, `mypy` doesn't flag 

```
if TYPE_CHECKING:
    from ambassador.ir.irtlscontext import IRTLSContext  # pragma: no cover
```

as needing attention, but it does say 

```
pyproject.toml: [mypy]: Unrecognized option: show_none_errors = False
```

So... what are you running? 🙂

kflynn (changes requested):
This is pretty awesome and I love how fast uv & ruff are. Couple of small things, but also we need to have ruff wrap at 100 columns correctly, and mypy is making me nervous here because I can't get it to tell me about an error that you fixed.... 🙂

build-aux/lint.mk
```diff
@@ -33,24 +33,14 @@ clean: .dmypy.json.rm .mypy_cache.rm-r
...
 lint-goals += lint/black
 lint/black: $(OSS_HOME)/.venv
-	uv run black --check ./python/
+	uv run ruff format --check ./python/
 .PHONY: lint/black
 
 format-goals += format/black
 format/black: $(OSS_HOME)/.venv
-	uv run black ./python/
+	uv run format ./python/
```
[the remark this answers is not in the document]
the-wondersmith: Fixed

.github/actions/setup-deps/action.yml
```diff
@@ -67,4 +67,4 @@ runs:
       env:
         DEBIAN_FRONTEND: noninteractive
       run: |
-        uv --no-config pip install -U awscli packaging
+        uv --no-config pip install -U ruff awscli packaging
```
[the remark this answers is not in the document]
the-wondersmith: Fixed

build-aux/tools.mk
```diff
@@ -19,7 +19,8 @@ clobber: clobber-tools
 
 .PHONY: clobber-tools
 clobber-tools:
-	rm -rf $(tools.bindir) $(tools.dir)/include $(tools.dir)/downloads
+	rm -rf uv.lock $(tools.bindir) $(tools.dir)/include $(tools.dir)/downloads
+	find . -type d -name '.*cache' -exec rm -rf {} \; || true
```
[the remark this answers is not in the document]
the-wondersmith: Mmmm... yeah, probably 😅. Fixed.

python/ambassador-diag/src/ambassador_diag/diagd.py
kflynn: Ew. On `dev/v4` before this change, `mypy` _also_ doesn't flag this line. 😛

python/ambassador/src/ambassador/VERSION.py
```diff
@@ -23,7 +23,9 @@
 Version = "MISSING(FILE)"
 Commit = "MISSING(FILE)"
 try:
-    with open(os.path.join(os.path.dirname(__file__), "..", "ambassador.version")) as version:
+    with open(
+        os.path.join(os.path.dirname(__file__), "..", "ambassador.version")
+    ) as version:
```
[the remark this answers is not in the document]
the-wondersmith: Reset the line length to 100, re-ran `uv run -- ruff check --config='pyproject.toml' python/`. Should now be fixed 🤦

python/ambassador-diag/src/ambassador_diag/diagd.py
the-wondersmith: I fixed that `show_none_errors` warning (copy/paste error). Uhhh... not sure about the lack of flagging though?

python/tests/src/tests/unit/test_mapping.py
```diff
@@ -29,7 +29,7 @@ def test_mapping_host_star_error():
     errors = ir.aconf.errors["bad-mapping.default.1"]
     assert len(errors) == 1, f"Expected 1 error but got {len(errors)}"
 
-    assert errors[0]["ok"] == False
+    assert errors[0]["ok"] is False
```
kflynn, line 32: Ew. Again, not gonna scream if we don't have a way to tweak `ruff` for this, but while I can vaguely get behind `is None`, `is False` makes me cringe. 😂

python/tests/src/tests/unit/test_error_response.py
```diff
@@ -143,7 +149,7 @@ def _test_errorresponse(yaml, expectations, expect_fail=False):
             % (expected_body_format_override, expected_filter)
         )
         print("checking m: ", m)
-        actual_filter = m["filter"]
+        # actual_filter = m["filter"]
```
kflynn, line 152: Another one to just delete, I think.

python/tests/src/tests/kat/test_ambassador.py
```diff
@@ -0,0 +1,15 @@
+# Import all the real tests from other files, to make it easier to pick and choose during development.
+
```
kflynn, line 2: The comment should go if all the imports have gone.

IIRC the point of all this stuff had to do with the way `pytest` found tests -- I assume that's different now...

python/tests/src/tests/kat/t_circuitbreaker.py
```diff
@@ -121,7 +124,7 @@ def check(self):
                 elif "X-Envoy-Overloaded" in result.headers:
                     pending_overloaded += 1
 
-            failed = False
+            # failed = False
```
kflynn, line 127: Thinking this can just be deleted.

python/tests/src/tests/kat/t_active_health_check.py
```diff
@@ -122,7 +128,7 @@ def check(self):
                 valid += 1
             elif self.results[i].status == 500:
                 errors += 1
-        msg = "Errors: {}, Valid: {}".format(errors, valid)
+        # msg = "Errors: {}, Valid: {}".format(errors, valid)
```
kflynn, line 131: Should just delete this. Originally I think it was used as the assertion string (cf the `ActiveHealthCheckTest` function earlier in this file) but it's a terrible string for that anyway.

.github/actions/setup-deps/action.yml
```diff
@@ -51,30 +63,24 @@ runs:
...
+        curl -LsSf https://astral.sh/uv/install.sh | bash
     - name: Install Python Requirements
       shell: bash
+      if: ${{ inputs.install-python == 'true' }}
       env:
         DEBIAN_FRONTEND: noninteractive
       run: |
-        python -m pip install -U awscli packaging
+        uv venv
+        uv --no-config pip install -U awscli
```
kflynn, line 86: I think `awscli` is in the dev dependencies now, no?

(I also expect we'll likely get rid of it – I _think_ it's only ALabs-specific stuff.)

kflynn (approved):
Man, this is a huge PR. 😂 Approving with some comments that can be addressed in a future PR. The ones I feel most strongly about are that we shouldn't really have any `#noqa: E402` comments (we should just be able to move the `import`s up above the executable code) and that I'd really prefer `==` when comparing specifically with `True` and `False`...

Thanks much!! This must've been a huge, huge, massive pain.

the-wondersmith:
> ...some comments that can be addressed in a future PR. 

I'll get to work on them posthaste 😅

> The ones I feel most strongly about are that we shouldn't really have any `#noqa: E402` comments (we should just ...

_Agreed_. For historical purposes though - I left the imports "out of order" and added the `E402` ignores because I didn't have the bandwidth to go through and ensure that the patching routine that's happening in between the import blocks isn't necessary any more w/ the new package structure and switch to `uv`.

> ... and that I'd really prefer `==` when comparing specifically with `True` and `False`...

Ummm.... that _might_ be a sticking point? There actually _is_ a meaningful difference between `is` and `==` in Python [[ref](https://wiingy.com/learn/python/difference-between-python-and-is-operators-in-python/)]. Plus it's a linting violation in ruff specifically [[ref](https://docs.astral.sh/ruff/rules/true-false-comparison/)], but also in several other linters / style guides.

> Thanks much!! This must've been a huge, huge, massive pain.

My pleasure! It wasn't that big a pain - labors of ~~love~~ hyperfixation rarely are 😂

python/ambassador-diag/src/ambassador_diag/diagd.py
```diff
@@ -1840,12 +1847,21 @@ def _load_ir(
...
             "configuration updated (%s) from snapshot %s (S%d L%d G%d C%d)"
-            % (config_type, snapshot, service_count, listener_count, group_count, cluster_count)
+            % (
+                config_type,
+                snapshot,
+                service_count,
+                listener_count,
+                group_count,
+                cluster_count,
+            )
```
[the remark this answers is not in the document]
the-wondersmith: Because `ruff` preserves behavior from `black`, and therefore because of the trailing comma rule in black [[ref](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#c-skip-magic-trailing-comma)], at least I assume?

python/ambassador/src/ambassador/config/config.py
```diff
@@ -430,7 +445,7 @@ def process(self, resource: ACResource) -> RichStatus:
 
             try:
                 handler(resource)
-            except Exception as e:
+            except Exception:
                 # Bzzzt.
                 raise
                 # return RichStatus.fromError("%s: could not process %s object: %s" % (resource, resource.kind, e))
```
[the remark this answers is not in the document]
the-wondersmith: Yes. It _absolutely should_.

python/ambassador/src/ambassador/envoy/v3/v3route.py
```diff
@@ -273,7 +274,7 @@ def __init__(
 
         # Passing a list to set is _very important_ here, lest you get a set of
         # the individual characters in group.host!
-        self["_host_constraints"] = set([group.get("host") or "*"])
+        self["_host_constraints"] = {group.get("host") or "*"}
```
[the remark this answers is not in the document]
the-wondersmith: It _is_ indeed a `ruff` thing, "prefer literals over function calls" is the convention (at least AFAIK)

python/tests/src/tests/unit/test_mapping.py
the-wondersmith: In all technicality that _should_ be `assert not errors[0]["ok"]` which (arguably) isn't much better 🙃

.github/actions/setup-deps/action.yml
the-wondersmith: I did move everything else that _used_ to be installed in that step into the dev deps, but nothing in the Python packages explicitly used anything from `awscli`, so I figured maybe it was a CI-specific thing and just left it.

We can absolutely sunset it / remove it entirely though, if it's an ALabs-exclusive

PR #5790, opened 2024-10-02
[WIP] v4alpha1 CRDs

kflynn:
This has been... ugly.

We're starting from the point of view that the v4alpha1 CRDs should be camelCase internally, but that we'll need to still use snake_case when talking to Python (purely as a practical matter of paranoia). Managing that... wow. The Go ecosystem, you see, does _not_ give you control over the tag used when serializing or deserializing JSON or YAML.

The Right Way to manage this may be a PR into `encoding/json` and `sigs.k8s.io/yaml`. The short-term way? I've copied those into Emissary's codebase and hacked them. 😛 

The end result is:

- `encoding/json` uses the `json` tag (for camelCase v4alpha1 nativeness)
- `github.com/emissary-ingress/emissary/v3/pkg/json` uses the `v3` tag but falls back to the `json` tag (for snake_case v3alpha1 madness)
- `sigs.k8s.io/yaml` uses the `yaml` and `json` tags (for camelCase v4alpha1 nativeness)
- `github.com/emissary-ingress/emissary/v3/pkg/yaml` uses the `v3` tag but falls back to the `yaml` and `json` tags (for snake_case v3alpha1 madness)

and `cmd/migrator` has the beginnings of a conversion tool. I say "beginnings" because it currently only does snake_case YAML to camelCase YAML but it's a start... and I used it to convert the testdata manifests for the v4alpha1 CRDs.


the-wondersmith:
🍃 Housekeeping 🧹

PR #5794, opened 2024-10-18
ambex: Remove usage of md5

wlynch:
## Description

When trying to run emissary in crypto-restricted environments (e.g. FIPS), usage of md5 can be problematic:

```
time="2024-10-17 19:20:29.7063" level=error msg="shut down with error error: PANIC: openssl: MD5 failed" func=github.com/emissary-ingress/emissary/v3/pkg/busy.Main file="github.com/emissary-ingress/emissary/v3/pkg/busy/busy.go:87" CMD=entrypoint PID=1
```

This replaces the usage with [xxhash](https://github.com/cespare/xxhash), which avoids the usage of md5 but keeps the same deterministic hash behavior. This was chosen over other stdlib hash functions that would be conformant (e.g. sha256) to minimize hash size since long hash length may be problematic for [k8s name length restrictions (253 characters)](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names).

## Related Issues

n/a

## Testing

Updated unit tests.

## Checklist

  Shouldn't need backport.

  - Any new features
  - Any deprecations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,6 +107,8 @@ it will be removed; but as it won't be user-visible this isn't considered a brea
   instead of the Mapping name, which could reduce the cache's effectiveness. This has been fixed so
   that the correct key is used. ([Incorrect Cache Key for Mapping])
 
+- Change: Changed Ambex suffix hashing algorithm to use xxhash64 instead of md5.
+
 [Incorrect Cache Key for Mapping]: https://github.com/emissary-ingress/emissary/issues/5714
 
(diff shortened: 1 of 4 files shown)


Alice-Lilith:
Sounds good to me. @kflynn, any concerns or shall I kick-off CI for this?

kflynn:
@Alice-Lilith i think running CI sounds great, thanks!

kflynn:
Merged by #5795 with our thanks!

Issue #5167, opened 2023-07-11
endpointslices support for service discovery


ajaychoudhary-hotstar:
**Please describe your use case / problem.**
k8s endpoints object support max 1000 endpoint. This is a bottleneck to scale pods beyond 1000. K8s support endpointslices from 1.21 to resolve this issue but it seems the ambassador doesn't support service discovery from endpointslices. 
https://kubernetes.io/docs/concepts/services-networking/service/#over-capacity-endpoints 

**Describe the solution you'd like**
Add support for endpointslices for service discovery. 

**Describe alternatives you've considered**

**Additional context**
Add any other context about the feature request here.

saravind-hotstar:
Is there a mitigation available here apart from vertical scaling @cindymullins-dw

PR #5798, opened 2024-10-28
Added support for endpointslices

ajaychoudhary-hotstar:
## Description

Currently Ingress is doing service discovery from endpoint objects, there is is hard limit of 1K IP address that k8s endpoint objects can hold which breaks load balancing if some backends scale more than 1K. 

## Related Issues

(https://github.com/emissary-ingress/emissary/issues/5167)

## Testing

Tested in lower env, scaled the backend system to 1.2K pods, and noticed the same 1.2K in backend metrics envoy_cluster_membership_total. 

## Checklist

  - Any new features
  - Any deprecations

- [x] **This is unlikely to impact how the Ambassador performs at scale.**

- [ x] **My change is adequately tested.**


cmd/entrypoint/endpoint_routing.go
```diff
@@ -20,13 +20,21 @@ func makeEndpoints(ctx context.Context, ksnap *snapshot.KubernetesSnapshot, cons
...
-		if !ok {
-			continue
-		}
-		for _, ep := range k8sEndpointsToAmbex(k8sEp, svc) {
-			result[ep.ClusterName] = append(result[ep.ClusterName], ep)
+	svcEndpointSlices := map[string][]*kates.EndpointSlice{}
+
+	// Collect all the EndpointSlices for each service
+	for _, k8sEndpointSlice := range ksnap.EndpointSlices {
+		svcKey := fmt.Sprintf("%s:%s", k8sEndpointSlice.Namespace, k8sEndpointSlice.Labels["kubernetes.io/service-name"])
```
kflynn, line 27: Is it possible for the `kubernetes.io/service-name` label to be missing? (I don't think so, but if it somehow is, maybe we should just skip this object?)

kflynn (changes requested):
@ajaychoudhary-hotstar, many thanks for diving into this! The mechanics of how you're dealing with EndpointSlices look reasonable to me, but I need to make your life a little harder, I'm afraid -- I've run into at least one case where there were no EndpointSlices but there were Endpoints, so I think we should probably look at EndpointSlices if present and Endpoints if not. Does that sound sane?

cmd/entrypoint/endpoint_routing.go
ajaychoudhary-hotstar: make sense, added check for label.

ajaychoudhary-hotstar:
> @ajaychoudhary-hotstar, many thanks for diving into this! The mechanics of how you're dealing with EndpointSlices look reasonable to me, but I need to make your life a little harder, I'm afraid -- I've run into at least one case where there were no EndpointSlices but there were Endpoints, so I think we should probably look at EndpointSlices if present and Endpoints if not. Does that sound sane?

Thanks, @kflynn Have added endpoints as fallback.

kflynn (commented):
@ajaychoudhary-hotstar, I think this looks good, many thanks! Let me sort out CI for it...

Alice-Lilith (commented):
Nice, and thanks for the PR! 🙂 🚀

kflynn:
Landed as PR #5808! 🙂

Issue #4744, opened 2022-12-08
Emissary Ingress - does not reread Istio certificates after a few days


szdonat:
**Describe the bug**
After integrating Emissary Ingress with Istio mTLS, 2 or 3 days later Emissary Ingress does not reread the TLS certificates coming from Istio any more and reports TLS error.
I found 2 quick fixes for the problem:
- restart Emissary Ingress with a cron job every day
- turn of diagnostics with the Helm parameter agent.reportDiagnostics: false, and adding a module with diagnostics.enabled: false and diagnostics.allow_non_local: false
 
**To Reproduce**
1. Install Emissary Ingress with Helm Chart and integrate mTLS with Istio, based on the documentation https://www.getambassador.io/docs/emissary/latest/howtos/istio
2. Wait 3 days
3. Cannot access Emissary Ingress, it reports TLS error
4. See error: upstream connect error or disconnect/reset before headers. reset reason: connection failure, transport failure reason: TLS error: 268436501:SSL routines:OPENSSL_internal:SSLV3_ALERT_CERTIFICATE_EXPIRED

**Expected behavior**
The Emissary Ingress - Istio integration should work at least for a week without turning diagnostics off.

**Versions (please complete the following information):**
 - Emissary Ingress: [3.3.0]
 - Emissary Ingress Chart: [8.3.0]
 - Kubernetes environment [bare metal, 1.21.8]
 - Istio [1.15.3]

**Additional context**
The Emissary Ingress - Istio integration always survives 36 hrs, the TLS problem comes after 2-3 day.
It is interesting that after turning off the diagnostics, I continue to see the Ambassador Diagnostic Overview page, but the integration survives for at least 8 days. I attached the values file used with the Helm Chart.
[values.yaml.txt](https://github.com/emissary-ingress/emissary/files/10185511/values.yaml.txt)

LiVe:
Unfortunately this issue has been unresolved for quite a while :/

See for example:
https://github.com/emissary-ingress/emissary/issues/4484

cindymullins-dw:
I believe we're aware of this issue and working on it.  Are you seeing any 503 errors?  Can you reproduce it consistently or able you able to output Ambassador pod logs /snapshot when it occurs?  Does it happen at the same time or at random intervals?  Any specs you can share with us to assist with replication would be helpful.

LiVe:
[...]

PR #5820, opened 2025-02-11
Fix istio cert rotation bug for issue #4744

jonathanelbailey:
## Description

This PR fixes a known issue when Emissary is configured to connect to an Istio mTLS network.  When Emissary's Istio sidecar initiates a cert rotation the `istio-certs` secret generates a cache entry, but does not generate a delta since the `istio-certs` cache entry does not map to an actual resource on the Kubernetes cluster.  This can eventually resolve itself if a delta is generated that can initiate a change, but other times it results in an unrecoverable error that can cause traffic disruption without custom health checks.

## Related Issues

https://github.com/emissary-ingress/emissary/issues/4744

## Testing

* Replicated the issue using recommended steps from https://github.com/emissary-ingress/emissary/issues/4744#issuecomment-1406798652
* Using this [test image](https://hub.docker.com/r/jebaile7964/emissary) that includes the fix, the test case could no longer be replicated
* Tested on Istio 1.23.2
* Added parametrized unit tests for improved coverage of this test case

## Checklist

  - Any new features
  - Any deprecations

### DIFF
--- a/python/ambassador/ir/ir.py
+++ b/python/ambassador/ir/ir.py
@@ -135,56 +135,54 @@ class IR:
         to_invalidate: List[str] = []
         invalidate_groups_for: List[str] = []
 
-        # OK. If we don't have a cache, just skip all this crap.
-        if cache is not None:
-            # We have a cache. Start by assuming that we'll need to reset it,
+        # OK. If we don't have a cache and there are no deltas, just skip all this crap.
+        if (cache and fetcher.deltas) is not None:
+            # We have a cache and deltas is non null. Start by assuming that we'll need to reset it,
             # unless there are no deltas at all.
             reset_cache = len(fetcher.deltas) > 0
 
-            # Next up: are there any deltas?
-            if fetcher.deltas:
-                # Yes. We're going to walk over them all and assemble a list
-                # of things to delete and a count of errors while processing our
-                # list.
+            # Yes. We're going to walk over them all and assemble a list
+            # of things to delete and a count of errors while processing our
+            # list.
 
-                delta_errors = 0
+            delta_errors = 0
 
-                for delta in fetcher.deltas:
-                    logger.debug(f"Delta: {delta}")
+            for delta in fetcher.deltas:
+                logger.debug(f"Delta: {delta}")
 
-                    # The "kind" of a Delta must be a string; assert that to make
-                    # mypy happy.
+                # The "kind" of a Delta must be a string; assert that to make
+                # mypy happy.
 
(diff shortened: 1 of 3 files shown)


kflynn:
Hey @jonathanelbailey, are you on the CNCF Slack by any chance? I'd like to chat about this one... 🙂

python/tests/unit/test_ir.py
```diff
@@ -58,3 +67,50 @@ class TestCase:
...
+        (
+            "null-cache-without-deltas",
+            None,
+            None,
+            {"config_type": "complete", "reset_cache": True, "invalidate_groups_for": []}
+        )
+        
+        ]
+    )
+    def test_check_deltas(self, name, cache_entry, deltas, expected, caplog):
```
kflynn, line 100: Silly question: does this test fail without your change? 🙂

kflynn (commented):
I'm leaning toward taking this but there are some corner cases I'm thinking about -- nice work in any case, many thanks! 🙂 I'm most curious, I think, about the test case, and about what testing you've put this through in general?

Thanks!! and happy to talk on Slack if you'd rather.

python/ambassador/ir/ir.py
```diff
@@ -135,56 +135,54 @@ def check_deltas(
         to_invalidate: List[str] = []
         invalidate_groups_for: List[str] = []
 
-        # OK. If we don't have a cache, just skip all this crap.
-        if cache is not None:
-            # We have a cache. Start by assuming that we'll need to reset it,
+        # OK. If we don't have a cache and there are no deltas, just skip all this crap.
+        if (cache and fetcher.deltas) is not None:
```
kflynn, line 139: (Restoring the comment that @the-wondersmith accidentally deleted 😂):

This logic feels a little weird – overall I think you're going for 'if we have a cache and there are deltas", and I think that'd be better expressed as

```
if (cache is not None) and fetcher.deltas:
```

My reasoning is that `fetcher.deltas` actually cannot ever be `None` – it's either an empty list or a list with some things in it.

ppeble:
FYI, I have attempted to address the comments here since we've been so delayed in responding ourselves. I have a new PR open here that contains those updates: https://github.com/emissary-ingress/emissary/pull/5935

I will likely close this in favor of the other one unless there are objections.

ppeble:
I am closing this in favor of https://github.com/emissary-ingress/emissary/pull/5935. The other PR will contain these changes plus the updates asked for in the comments.

PR #5824, opened 2025-02-12
Rework charts

kflynn:
This is mostly to have the CRDs chart and the main Emissary chart committed in this branch. It's definitely a departure from before -- less in the content of the charts, more in the way they're built.


the-wondersmith:
> `...` It's definitely a departure from before -- less in the content of the charts, more in the way they're built.

Can you give a short rundown of the difference (for posterity 😅)?

the-wondersmith:
🍃 Housekeeping 🧹

PR #5831, opened 2025-03-26
Update the TRY-3.10 document for 3.10.0-rc.3.

kflynn:


Alice-Lilith (approved):

TRY-3.10.md
```diff
@@ -32,7 +34,7 @@ First, install the CRDs and the conversion webhook:
 ```bash
 helm install emissary-crds \
  --namespace emissary-system --create-namespace \
```
joebowbeer, line 36: Is there no way to push RC charts and images to this repo?

I'd be more willing to try this, and would be more confident in the resulting release, if this RC were published to where it will eventually be released from.

PR #5835, opened 2025-04-28
feat(service): provide options to pass `loadBalancerClass`

sunnykrGupta:
In order to use dedicated load balancer controller (in my case, its   [AWS Load Balancer Controller](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.12/deploy/configurations/)  which has feature to create network LB and Application LB both. We need to provide `loadBalancerClass` to allow events to be listened by the controller deployed by the operator. Without this flag, default cloud operator picks up the changes and process it to provision cloud resources. 

Submitting this feature request to enable it as part of the emissary charts. 

ref: https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class


sunnykrGupta:
Done. Thanks for taking care of it. 

> Hello! Thank you for submitting this, I think it's a good addition. Would you mind pointing your PR at the branch `dev/v4`? We are gearing up to release the next major version and we are trying to get changes into the new branch so we don't have to port it later.
> 
> Thank you! Once that is done I'll approve and merge.

ppeble:
Closing this in favor of https://github.com/emissary-ingress/emissary/pull/5867

PR #5867, opened 2025-08-29
feat(service): provide options to pass `loadBalancerClass`

ppeble:
## Description

This is a new PR with commits targeting `dev/v4` from https://github.com/emissary-ingress/emissary/pull/5835.

Original description:

In order to use dedicated load balancer controller (in my case, its   [AWS Load Balancer Controller](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.12/deploy/configurations/)  which has feature to create network LB and Application LB both. We need to provide `loadBalancerClass` to allow events to be listened by the controller deployed by the operator. Without this flag, default cloud operator picks up the changes and process it to provision cloud resources. 

Submitting this feature request to enable it as part of the emissary charts. 

ref: https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class

### DIFF
--- a/charts/emissary-ingress/templates/service.yaml
+++ b/charts/emissary-ingress/templates/service.yaml
@@ -23,6 +23,9 @@ spec:
   {{- if .Values.service.loadBalancerIP }}
   loadBalancerIP: "{{ .Values.service.loadBalancerIP }}"
   {{- end }}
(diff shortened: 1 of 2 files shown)


kflynn (approved):
Ship it! 🙂

PR #5878, opened 2025-11-14
dev/v4: release ci artifacts

the-wondersmith:
PR represents the initial pass at adding a `release` step to the CI workflow that publishes built images and chart artifacts to `ghcr.io` and cuts a tagged GitHub release.


kflynn (approved):
OK, gonna take this one -- we've taken it as far as we can without letting GitHub Actions run it for real. Many many thanks @the-wondersmith! 🙂

PR #4953, opened 2023-04-07
Add loadBalancerSourceRanges to service template in values.yaml

bothra90:
## Description
The service.loadBalancerSourceRanges configuration flag is supported but not documented:
 https://github.com/emissary-ingress/emissary/blob/232a480530bf68d83f4fdabd0f6988da68850843/charts/emissary-ingress/templates/service.yaml#L71

## Testing

A few sentences describing what testing you've done, e.g., manual tests, automated tests, deployed in production, etc.

## Checklist

- [x] **Does my change need to be backported to a previous release?**
 Doesn't need to be backported

- [x] **I made sure to update `CHANGELOG.md`.**
This is just a documentation fix

- [x] **This is unlikely to impact how Ambassador performs at scale.**

- [x] **My change is adequately tested.**

- [x] **I updated `DEVELOPING.md` with any any special dev tricks I had to use to work on this code efficiently.**

- [x] **The changes in this PR have been reviewed for security concerns and adherence to security best practices.**

PR #5888, opened 2025-11-22
chart: add loadBalancerSourceRanges to service template in values.yaml

ppeble:
Copy of https://github.com/emissary-ingress/emissary/pull/4953 that points to our target branch.


ppeble:
Closing, there is no DCO on the original commit so I have to do this myself.

PR #5889, opened 2025-11-22
chart: add loadBalancerSourceRanges to service template in values.yaml

ppeble:
Change matching https://github.com/emissary-ingress/emissary/pull/4953 to target dev/v4 branch.

### DIFF
--- a/charts/emissary-ingress/values.yaml
+++ b/charts/emissary-ingress/values.yaml
@@ -204,6 +204,9 @@ service:
 
   annotations: {}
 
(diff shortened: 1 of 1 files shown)


kflynn (changes requested):
One change to request -- thanks! (I was very surprised to find that the chart honors these already. 🙂)

the-wondersmith (dismissed):

kflynn (approved):
Hit it!  🙂 Thanks @ppeble!

PR #5921, opened 2026-04-01
Set traced ingress listeners to INBOUND

### LINKED ISSUE #5907: Ingress traffic creates CLIENT spans instead of SERVER spans with OpenTelemetry
**Describe the bug**

While reviewing OpenTelemetry tracing behavior across multiple ingress controllers, I noticed that Emissary differs from other Envoy-based ingress controllers in how it models ingress entry spans.

When OpenTelemetry tracing is enabled, Emissary creates CLIENT spans for incoming HTTP requests, whereas other ingress controllers create SERVER root spans for the same traffic.

**To Reproduce**

1. Deploy Emissary `v4.0.0-rc.1` with OpenTelemetry tracing enabled:

```
kind: TracingService
metadata:
  name: otel-tracing
  namespace: emissary
spec:
  service: "otel-collector.opentelemetry.svc.cluster.local:4317"
  driver: opentelemetry
  config:
    service_name: emissary-ingress
```

2. Configure a backend service and Mapping
3. Send HTTP traffic through Emissary
4. Inspect exported OTLP traces

Observed: ingress entry span has span.kind = 3 (CLIENT)
Expected: ingress entry span has span.kind = 2 (SERVER)

**Expected behavior**

Ingress entry-point spans should be modeled as SERVER root spans, consistent with Envoy semantics and with other Envoy-based ingress controllers.

**Versions**

* Emissary: v4.0.0-rc.1
* Kubernetes: kind 

**Additional context**
Comparison with other Envoy-based ingress controllers
* Contour: creates span.kind = 2 (SERVER) 
* kgateway: creates span.kind = 2 (SERVER) 
* Emissary: creates span.kind = 3 (CLIENT)

**Possible cause**
This appears to be related to the traffic direction configured for Envoy listeners when tracing is enabled:

`python/ambassador/envoy/v3/v3listener.py#L572`

```
self.traffic_direction = "OUTBOUND"
```

For ingress listeners, this likely should be `"INBOUND"` so that Envoy emits SERVER spans for incoming requests.

aviralgarg05:
## Description

This change fixes the listener metadata emitted when tracing is enabled on Emissary-managed ingress listeners.

Today, traced HTTP listeners are generated with `traffic_direction: OUTBOUND`. In Envoy, that causes spans created from those listeners to be treated as client-side spans, which is incorrect for ingress traffic. This patch changes traced ingress listeners to use `traffic_direction: INBOUND` instead, so the generated spans reflect the actual role of the listener.

The change is intentionally small and scoped to listener generation. A targeted unit test was added to cover traced listeners explicitly and to guard against this regression going forward.

## Related Issues

- Fixes #5907

## Testing

I verified the change with focused and broader automated test coverage.

Focused validation:
- `uv run --group dev pytest --tb=short python/tests/src/tests/unit/test_listener.py -k traced_listeners_are_inbound`
- Re-ran the targeted tracing listener test after the final cleanup pass

Area validation:
- `uv run --group dev pytest --tb=short python/tests/src/tests/unit/test_listener.py`

Broader regression validation:
- `uv run --group dev pytest --tb=short python/tests/src/tests/unit`

Linting:
- `uv run --group dev ruff check python/ambassador/src/ambassador/envoy/v3/v3listener.py python/tests/src/tests/unit/test_listener.py`
- `make lint`

Note:
- On this workstation, `make lint` needed to be run with a local Python 3.13 shim for the chart-lint step because the pinned `yamale` tool crashes under host Python 3.14. This was only a local verification detail; no repository files were changed to work around it.

## Checklist

  - No backport versions were discussed in the issue thread while preparing this fix.

   This change does not add a feature, change the bundled Envoy version, introduce a deprecation, or create a backward-incompatible behavior change. I did not update `CHANGELOG.md`.

   This only changes the emitted listener `traffic_direction` value for traced ingress listeners. It does not change routing, allocation patterns, concurrency, replica guidance, or request processing cost in the data plane.

   The fix is covered by a new targeted unit test for traced listeners, and the full Python unit test suite was also run to verify the surrounding listener-generation behavior did not regress.

   No reusable project-specific development trick was introduced by this change.

   This change only adjusts listener tracing metadata and adds test coverage. It does not expand privileges, alter authn/authz behavior, or introduce new network-facing configuration surfaces.

### DIFF
--- a/python/ambassador/src/ambassador/envoy/v3/v3listener.py
+++ b/python/ambassador/src/ambassador/envoy/v3/v3listener.py
@@ -588,7 +588,7 @@ class V3Listener:
             base_http_config["generate_request_id"] = True
 
             base_http_config["tracing"] = {}
-            self.traffic_direction = "OUTBOUND"
+            self.traffic_direction = "INBOUND"
 
             custom_tags = self.config.ir.tracing.get("custom_tags", [])
             if custom_tags:
--- a/python/tests/src/tests/unit/test_listener.py
+++ b/python/tests/src/tests/unit/test_listener.py
@@ -10,7 +10,13 @@ from ambassador.config import Config
 from ambassador.envoy import EnvoyConfig
 from ambassador.fetch import ResourceFetcher
 from ambassador.utils import EmptySecretHandler
-from tests.utils import Compile, default_http3_listener_manifest, econf_compile, logger
+from tests.utils import (
+    Compile,
+    default_http3_listener_manifest,
+    default_listener_manifests,
+    econf_compile,
+    logger,
+)
 
 
 def _ensure_alt_svc_header_injected(listener, expectedAltSvc):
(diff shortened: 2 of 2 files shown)


kflynn:
@aviralgarg05 Hey, sorry for the delay here! Can you rebase this back to `main` so we can turn CI loose on it again? Thanks!

PR #5924, opened 2026-04-09
fix: suppress noisy diagnostics UI warnings

### LINKED ISSUE #5923: Gateway API and TLS warnings in diagnostics UI should be configurable or removed
As described in #2138 and #2139 from back when, it doesn't always make sense for these errors to be unconditional. At this point, we should simply remove the Gateway API CRD warning, and we should allow disabling the TLS warning.

I'm opening this new issue because it will be easier to track the two together, and while we're at it, because I should reexamine #4143 and figure out why I closed it back when. 🤔

officialasishkumar:
## Description

Suppress the diagnostics UI warnings that were making otherwise-healthy installs look broken:

- filter the noisy missing-CRD messages out of the diagnostics UI while leaving the underlying logged errors intact
- add `diagnostics.missing_tls_ok: true` on the Ambassador Module to suppress the missing-`TLSContext` warning for intentionally cleartext deployments

## Related Issues

Fixes #5923.

## Testing

- `ruff check python/ambassador-diag/src/ambassador_diag/diagd.py python/tests/src/tests/unit/test_diagd.py`
- `/tmp/emissary/.venv/bin/python -m pytest python/tests/src/tests/unit/test_envoy_stats.py python/tests/src/tests/unit/test_diagd.py`

## Checklist

    None were discussed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,10 @@ from the Helm charts.
 - Fix: The diagnostics UI refers to Emissary (instead of Ambassador) and
   links to emissary-ingress.dev's docs.
 
+- Fix: The diagnostics UI no longer treats missing CRD definitions as a
+  warning, and setting `diagnostics.missing_tls_ok: true` on the
+  Ambassador Module suppresses the missing-`TLSContext` warning.
+
 - Fix: Emissary now uses the current `string_match` Envoy HeaderMatcher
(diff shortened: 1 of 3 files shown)


kflynn:
Hey @officialasishkumar! Can you wrangle DCO on this? I'd like to actually take this, with apologies for the delay while we sorted out some other stuff. 🤦‍♂️

PR #5930, opened 2026-04-14
docs: correct example prestop hook in helm values.yaml

RichardoC:
## Description

The main container of emissary doesn't include sleep, curl, or almost anything. For this reason, I've changed these example commands to Python which works with the python3 version baked into the container (though called `python` not `python3`)

## Related Issues

List related issues.

## Testing

Run these commands manually with 

```
kubectl --namespace emissary-ingress exec -it emissary-ingress-59f48c4657-5gk8k -- python -c 'import urllib.request; urllib.request.urlopen(urllib.request.Request("http://localhost:8001/healthcheck/fail", method="POST")).read(); import time ; time.sleep(780)'
```

## Checklist

No

Not needed

### DIFF
--- a/charts/emissary-ingress/values.yaml
+++ b/charts/emissary-ingress/values.yaml
@@ -434,10 +434,43 @@ createDefaultListeners: false
 lifecycle: # +doc-gen:break
   # Example block that works well if the main container is running behind a load balancer that
   # is slow to deregister pods - e.g. with IP-type registration in an AWS TargetGroup:
-  # lifecycle:
-  #   preStop:
-  #     exec:
-  #       command:
-  #         - "python" # The Python in the container image.
-  #         - "-c"
-  #         - import urllib.request; urllib.request.urlopen(urllib.request.Request("http://localhost:8001/healthcheck/fail", method="POST")).read(); import time ; time.sleep(780)
+  # Also, if you're giving more than 30 seconds, set the following, to higher valyes to avoid k8s terminating the pod too early
+  # terminationGracePeriodSeconds: 660
+  # progressDeadlines:
+  #   ambassador: 900
+
+  # preStop:
+  #   exec:
+  #     command:
+  #       # The python is actually the python3 included in the image, it's just called python despite convention
+  #       - "python"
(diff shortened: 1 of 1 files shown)


RichardoC:
One for the reviewer, this really doesn't work as I expected. It looks like (regardless of this prestop hook) the pod will terminate far faster.

I'm using the following through the helm chart, and it's not waiting anywhere near 780 seconds

```yaml
lifecycle: # slow down shutdown to allow litellm to finish processing requests
  # based on https://github.com/emissary-ingress/emissary/blob/6460e5181f527317e8bba7eaf7f5e1fc6fa17f77/charts/emissary-ingress/values.yaml#L436
  preStop:
    exec:
      command:
        # These Python's are actually the python3 included in the image, it's just called python despite convention
        - "python"
        - "-c"
        - import urllib.request; urllib.request.urlopen(urllib.request.Request("http://localhost:8001/healthcheck/fail", method="POST")).read(); import time ; time.sleep(780)
```

RichardoC:
That took a while, but this should be the combination of things which actually allow long lived connections to drain.

Where should I (also) add this to the docs?

kflynn:
Hey @RichardoC! I think we're inclined to shift this drain code into Emissary itself -- rather than rely on a `preStop` hook, just listen for `SIGTERM` and do the draining. That'd make it a bit simpler, I think, to give the user better control of the termination grace period, etc. 

Is that something you'd be interested in looking into?

RichardoC:
> Hey @RichardoC! I think we're inclined to shift this drain code into Emissary itself -- rather than rely on a `preStop` hook, just listen for `SIGTERM` and do the draining. That'd make it a bit simpler, I think, to give the user better control of the termination grace period, etc.
> 
> Is that something you'd be interested in looking into?

That's far better than this work around!

I'd like to, but it's likely to be > weeks until I'd be able to even look at it due to work pressures

kflynn (approved):
Taking this one, with apologies for the delay! (I still want to move this into Emissary itself, but upon consideration I'd rather bring it in while that happens.)

PR #5935, opened 2026-04-21
Fix istio cert rotation bug for issue #4744

ppeble:
## Description

This is building off of the existing PR https://github.com/emissary-ingress/emissary/pull/5820, which attempts to address https://github.com/emissary-ingress/emissary/issues/4744. This was originally submitted by @jonathanelbailey. I will include his original post below:

> This PR fixes a known issue when Emissary is configured to connect to an Istio mTLS network. When Emissary's Istio sidecar initiates a cert rotation the istio-certs secret generates a cache entry, but does not generate a delta since the istio-certs cache entry does not map to an actual resource on the Kubernetes cluster. This can eventually resolve itself if a delta is generated that can initiate a change, but other times it results in an unrecoverable error that can cause traffic disruption without custom health checks.

I have included a commit that address the comments from the previous PR so that we can still credit the original author while addressing the questions.

## Related Issues

https://github.com/emissary-ingress/emissary/issues/4744

## Testing

I tested locally with k3d with both `main` and the `peble:fix-istio-cert-rotation-4744` branches:

* Reproduced the bug on a `main` image: emissary crashed with `TypeError: object of type 'NoneType' has no len()` at ir.py:158 on every empty-delta snapshot, never reaching configuration updated
* Verified the fix on a `ppeble:fix-istio-cert-rotation-4744` image built from this branch on the same cluster: 0 `TypeError `/ could not reconfigure events across ~60 min of runtime with SECRET_TTL=1m (cert rotation every ~30 s).

## Checklist

No.

  - Any new features
  - Any deprecations

See testing section above.

N/A

### DIFF
--- a/python/ambassador/src/ambassador/ir/ir.py
+++ b/python/ambassador/src/ambassador/ir/ir.py
@@ -151,16 +151,14 @@ class IR:
         to_invalidate: List[str] = []
         invalidate_groups_for: List[str] = []
 
-        # OK. If we don't have a cache and there are no deltas, just skip all this crap.
-        if (cache and fetcher.deltas) is not None:
-            # We have a cache and deltas is non null. Start by assuming that we'll need to reset it,
-            # unless there are no deltas at all.
-            reset_cache = len(fetcher.deltas) > 0
-
-            # Yes. We're going to walk over them all and assemble a list
-            # of things to delete and a count of errors while processing our
-            # list.
-
+        # If we don't have a cache, or there are no deltas, there's nothing to do here --
+        # the defaults above (complete reconfigure, reset the cache) are what we want.
+        # Note that fetcher.deltas is always a list: it starts as [] and is reassigned
+        # from watt_dict.get("Deltas", []), so we check truthiness rather than `is not None`.
+        if (cache is not None) and fetcher.deltas:
+            # We have a cache and at least one delta. Walk over all the deltas and
+            # assemble a list of things to delete and a count of errors while processing
+            # our list.
             delta_errors = 0
 
             for delta in fetcher.deltas:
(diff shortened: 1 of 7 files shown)


python/ambassador/src/ambassador/ir/ir.py
```diff
@@ -151,66 +151,53 @@ def check_deltas(
...
-                        cache.invalidate(key)
-
-            # When all is said and done, it's an incremental if we don't need to reset
-            # the cache.
-            if not reset_cache:
+        # If we don't have a cache, or there are no deltas, there's nothing to do here --
+        # the defaults above (complete reconfigure, reset the cache) are what we want.
+        # Note that fetcher.deltas is always a list: it starts as [] and is reassigned
+        # from watt_dict.get("Deltas", []), so we check truthiness rather than `is not None`.
+        if (cache is not None) and fetcher.deltas:
```
ppeble, line 158: FYI @kflynn you said in the other PR that `fetcher.deltas` is always `Not None` but it turns out that this is not true. Here is your original comment: https://github.com/emissary-ingress/emissary/pull/5820#discussion_r2008503731

The actual json produced by the go process and given to this python process ends up looking like this:

```
{
    "AmbassadorMeta": {
      "cluster_id": "90fa5e76-73b4-5ad0-a6b5-452f45580845",
      "ambassador_id": "default",
      "ambassador_version": "4.0.1-SNAPSHOT-205e64914",
      "kube_version": "v1.30.5+k3s1",
      "sidecar": null
    },
    "Deltas": null,
    "Invalid": null
 }
```

The `"Deltas": null` is what will trigger this.

The good news is that the current implementation will catch this as we intend! 

On a larger note, would you object to addressing this when we process the fetched json here? Referring to this: https://github.com/emissary-ingress/emissary/blob/main/python/ambassador/src/ambassador/fetch/fetcher.py#L254

We can do what we already do below with other things, like this: https://github.com/emissary-ingress/emissary/blob/main/python/ambassador/src/ambassador/fetch/fetcher.py#L268

Thoughts? I don't see a danger to doing this as it would make explicit what you already thought was true.

ppeble: Actually, I just checked and the only thing that receives the `self.deltas` is this function! So unless you object I am going to add that into this PR.

kflynn: I do not object, to be clear. 🙂

kflynn (approved):
Looks good from here, thanks! 🙂

PR #5936, opened 2026-04-24
Initial CI end to end tests

ppeble:
## Description

Adds a preliminary set of e2e tests to the CI pipeline. The goal of these tests is to be incredibly basic 'end to end' tests of functionality to ensure that our changes work outside of only unit tests.

This is an intentionally very basic test setup. It starts k3d, builds images, applies some fixtures, and then uses a bash script to verify that it all worked. It is not designed for big or complex scenarios.

I investigated two other possible implementations:

* [Chainsaw](https://github.com/kyverno/chainsaw)
* [KUTTL](https://github.com/kudobuilder/kuttl)

Both felt very heavy for what we wanted to do but I'm not opposed to investigating if we think that this may end up being a much more complicated suite.

## Testing

All of this runs in the CI and can be run locally via `make` commands added to `build-aux/e2e.mk`.

## Checklist

- [X] **I updated `CONTRIBUTING.md` with any special dev tricks I had to use to work on this code efficiently.**


the-wondersmith (dismissed):

the-wondersmith (approved):

kflynn (approved):
Let's get this landed. 🙂 Thanks @ppeble! There's (obviously) so much more we can do here which we should talk through, but I love having the basics alive again.

PR #5939, opened 2026-04-30
Add RELEASE.md and remove old pro CHANGELOG that is no no longer necessary

ppeble:
Add in explicit instructions for maintainers on how to perform releases.

Also removes an old 'pro' CHANGELOG that is no longer used.


the-wondersmith (dismissed):

kflynn (approved):
Ship it! 🙂 I'm sure we'll need to tweak as we go, but let's land it.

PR #5940, opened 2026-05-05
Move e2e test framework from custom built to chainsaw

ppeble:
## Description

Replaces the custom e2e harness with [Chainsaw](https://kyverno.github.io/chainsaw/), plus the two pieces the KAT suite needs before it can be ported.

**`probe.sh`** is the assertion mechanism: it runs `kat-client` against a `queries.json` and retries until a `jq` expression passes. That expression is the direct equivalent of KAT's Python `check()`:

| KAT | Chainsaw |
|---|---|
| `manifests()` / `config()` | `manifests.yaml` |
| `queries()` | `queries.json` |
| `Query(url, expected=404)` | `.[0].result.status == 404` |
| `check()` / `assert` | the `jq` expression |
| `r.backend.name == "x"` | `.[0].result.json.backend == "x"` |
| one Emissary per test | a slot |

Retries live in the helper because Chainsaw does not retry `script` steps and Envoy config propagation is async; every fixture needs 2-3 attempts.

**Slots** solve isolation. Most KAT tests cannot share one Emissary, since a `Module`, `AuthService`, `RateLimitService`, `TracingService` or `LogService` configures the whole instance. A slot is an isolated Emissary keyed by `AMBASSADOR_ID` rather than by namespace: an Emissary only acts on CRDs whose `ambassador_id` matches its own, and a CRD omitting the field defaults to `"default"`. So a slot is just another helm release of the shipped chart with `env.AMBASSADOR_ID` set. No chart changes were needed.

Slots are **allocated at run time, never named by a fixture.** A fixture declares only whether it needs isolation:

```yaml
metadata:
  labels:
    slot: exclusive     # or: shared
```

and every Emissary CRD uses one expression, identical in every fixture:

```yaml
spec:
  ambassador_id: [(env('SLOT'))]
```

`slots.sh` exports `SLOT` per shard and decides placement: `default` for the shared shard, and the `exclusive` fixtures dealt round-robin across the available slots. Because an Emissary with `AMBASSADOR_ID` unset self-identifies as `"default"`, the shared case needs no special-casing, so switching a fixture between kinds is a one-word change. Exclusive shards target their fixtures with `--include-test-regex`, which chainsaw feeds to Go's `-run`.

This matters beyond tidiness: with hardcoded slots, two fixtures picking the same number would quietly share an Emissary and clobber each other's global config, for a wrong-but-green result with nothing pointing at the cause. A fixture can no longer name a slot, so it cannot name the wrong one.

Two chart collisions are worked around per slot (`module.enabled=false` and a distinct `ingressClassResource.name`); both are documented at length in `slots.sh`.

## What is included

- `test/e2e/probe.sh` and `test/e2e/slots.sh` (slot list, port math, concurrent install, dynamic dealing, sharded runner, plus a `check` that fails the suite on a missing or unknown `slot` label or a name that would not survive being put in a regex -- any of which would mean a fixture silently never runs)
- 8 slots on a 4-port stride out of the `8100-8159` block, installed concurrently
- `build-aux/e2e.mk` and `.github/workflows/test-images.yaml` both drive `slots.sh`, so the slot list has one definition
- Shared diagnostics in `.chainsaw.yaml` via `spec.error.catch`; a fixture-level `catch` appends to it rather than replacing it, so fixtures carry none
- Fixtures need no `bindings` and no `template: true`: `script` steps inherit the process environment, manifests call `(env('KAT_SERVER_IMAGE'))` and `(env('SLOT'))` directly, and templating is on by default
- 5 fixtures ported from KAT: `header-routing` (`t_headerrouting.py`), `gzip-minimum` / `gzip-content-type` / `gzip-unsupported-content-type` (`t_gzip.py`), `server-name` (`t_basics.py::ServerNameTest`)
- An unrelated fix for `hatchling` 1.32.0, without which every CI job fails at dependency install (see those commits)

A fixture is ~19 lines, holding only what is specific to it: a name, a kind, and an assertion.

## Testing

Green on amd64 and arm64. 8 slots install in ~40s; the runner deals 4 exclusive fixtures across them and all 8 fixtures pass, 22.6-31.8s each. Spare slots get no process at all.

Each probe's retry progression and final response were checked in the CI logs, not just the exit status. `gzip-unsupported-content-type` asserts an *absence*, so its `jq` also asserts status and backend; otherwise a `404` from an Emissary that had not yet loaded the route would pass for the wrong reason.

- 175 of 183 KAT classes still to port.
- ~47 queries across 17 KAT files hit admin port 8877, which is not published.
- A few tests need container env rather than CRDs (`AMBASSADOR_SINGLE_NAMESPACE`, `LUA_SCRIPTS_ENABLED`). Those are genuine presets, and unlike the anonymous slots they warrant real names.
- `manifests.yaml` is now the bulk of a fixture (~40 lines of kat-server Deployment/Service boilerplate). A shared `backends/` library would cut that, and the port needs one anyway for KAT's 11 `ServiceType` backends.
- The global `catch` has not yet executed, since nothing has failed since it landed. Worth one deliberate broken assertion against a live cluster.
- `pkg/envoytest` has a pre-existing data race on `logCtx` that intermittently fails `TestGatewayMatches` under `-race`. Untouched here.


the-wondersmith (approved):

kflynn (approved):
Yeah, let's land this and start iterating. 🙂 Thanks @ppeble!

PR #5975, opened 2026-08-08
CHAINSAW: finish removing the Ambassador error response override filter

kflynn:
I thought we'd finished ripping this out some time ago, but nope. It just got more important with the need to always set `@type` in the Envoy configuration objects -- if we leave this in, we're wrestling a type that's no longer part of the Envoy core.


Alice-Lilith (dismissed):

ppeble (approved):
🚀

PR #5979, opened 2026-08-12
docs: update PR template and contributing guidelines with reproducibi…

the-wondersmith:
PR documents `emissary`'s updated requirements for bug-fix PRs


kflynn (approved):
Yeah, let's do it. Probably worth checking over the rest of the text, too, since we haven't in a while. (That should be a separate thing though.)

PR #5980, opened 2026-08-14
test(e2e): port the extauth suite, with labelled probe assertions

ppeble:
## Description

Ports the most assertion-dense test in the KAT suite, plus the framework change it needed and two bugs that the first local run of the e2e suite turned up.

### `probe.sh` takes labelled assertions

A fixture could only pass one jq expression, so a test with many assertions had to AND them together and got one bit back on failure. `AuthenticationTestV1` is 69 asserts; a single boolean there is undebuggable.

```sh
probe.sh queries.json <label> <jq> [<label> <jq> ...]
```

Each assertion is evaluated individually every attempt. Progress lines report how many pass and name the first failure; on final failure only the failing ones are listed. The bare single-expression form is unchanged, so the 8 existing fixtures are untouched.

### `extauth-http`: `t_extauth.py::AuthenticationTestV1`

28 queries (KAT's 9 literal `Query()` calls plus a 20-iteration loop) and 69 asserts, as 37 labelled assertions. Three kat-server backends: a plain target plus two auth services with `INCLUDE_EXTAUTH_HEADER=yes`, which makes kat-server echo back a JSON document describing what it received in an `extauth` header. Two AuthServices share a `path_prefix` so Envoy load-balances across them.

Covers request headers forwarded to auth, non-whitelisted headers dropped in *both* directions, `Location` passthrough, default `Authorization` whitelisting, `X-Forwarded-Proto` forced to http, `bypass_auth` skipping auth entirely, `status_on_error`, and load balancing.

Two assertions could not be derived by reading, so they were measured against a live cluster rather than guessed:

- `l5d-dst-override` is `["extauth:80"]` in **10/10** samples. I expected ~50/50 since only `auth2` sets `add_linkerd_headers`; it is 100% because the setting applies to the merged ext_authz config, not per-AuthService.
- The 20-request loop hits both auth services in **5/5** runs at a 9-11 split, so asserting both got traffic is not the coin flip KAT's "make it more likely" comment implies.

## Testing

Full 9-fixture suite green locally in **17s**; `extauth-http` passes 37/37 on the first attempt. Fixtures run 8.5-11.3s locally against 25-32s on CI runners, where 8 Emissary pods contend for 4 vCPUs.

- 174 of 183 KAT classes still to port. The next tranche is the 10 files that port in full (25 classes): `dns_type`, `requesheader`, `envoy_logs`, `queryparameter_routing`, `headerswithunderscoresaction`, `max_req_header_kb`, `grpc_bridge`, `grpc_web`, `listeneridletimeout`.
- ~45 classes touch admin paths, but only 4 queries use `port=8877` explicitly; the rest hit `ambassador/v0/diag`, `check_ready`, `check_alive` and `/metrics` on the normal edge port. Worth testing before building port-forward infrastructure.
- ~33 classes need container env rather than CRDs, which no anonymous slot provides. About 5 named presets would cover 30 of them.
- The 4-port slot stride allows one TCP port per slot; 4 of the 18 `extra_ports` tests need more, which means recreating the k3d cluster.
- `manifests.yaml` is the bulk of a fixture now. A shared `backends/` library would cut it, and KAT's 11 ServiceTypes need one anyway.


kflynn (approved):
Yeah, I mean, it's a little hairy, but this _is_ a hairy test. 😂 

I think that with a little more experience we can think about how to improve some of those JQ things to make life a little easier, but this is still (to me) more understandable than KAT. Let's land this!

PR #5988, opened 2026-09-09
test(e2e): port the TLS and invalid-protocol rate limit tests, and fix kat-server's cert

ppeble:
## Description

Ports two more KAT classes from `t_ratelimit.py` to Chainsaw fixtures, and fixes a defect in the kat-server image that the first of them uncovered.

### `fix(kat-server)`: the baked-in cert was a CA cert, not a server leaf

kat-server terminates TLS on 8443 with the certificate baked into its image. That certificate was created with step's `root-ca` profile:

```
subject=CN=kat.server.local
X509v3 Key Usage: critical
    Certificate Sign, CRL Sign
X509v3 Basic Constraints: critical
    CA:TRUE, pathlen:1
```

No Digital Signature bit and no `serverAuth` EKU, which a TLS server leaf needs. Envoy's BoringSSL enforces key usage during the handshake regardless of whether the peer verifies the chain, so it aborts with `KEY_USAGE_BIT_INCORRECT` and Emissary cannot reach kat-server over TLS at all.

For a `RateLimitService` the failure is silent rather than loud: `failure_mode_deny` defaults to false, so the filter fails open and requests that should have been rate limited return 200.

Switching to the `self-signed` profile produces a correct leaf:

```
X509v3 Key Usage: critical
    Digital Signature
X509v3 Extended Key Usage:
    TLS Web Server Authentication, TLS Web Client Authentication
X509v3 Subject Alternative Name:
    DNS:kat.server.local
```

That profile needs `--subtle` before step will issue a self-signed leaf, and `--not-after` because it defaults to a 24-hour lifetime where `root-ca` defaulted to ten years.

This blocked more than the fixture that found it. Any test exercising kat-server's own HTTPS listener would have hit the same wall, so it needed fixing before the downstream-TLS ports from `t_tls.py` and `t_hosts.py`.

### `ratelimit-tls`, from `t_ratelimit.py::RateLimitV1WithTLSTest`

The existing `ratelimit-grpc` fixture plus a `Secret` and a `TLSContext` on the `RateLimitService`'s `tls` field. The TLS is on Emissary's upstream connection to the rate-limit service rather than the downstream listener, so the queries and assertions are unchanged and nothing in `slots.sh` or `build-aux/e2e.mk` needed to move. The cert is the `ratelimit.datawire.io` one KAT used, inlined as `stringData` so it stays readable in review.

### `ratelimit-protocol-invalid`, from `t_ratelimit.py::RateLimitVerTest`

That class has `v2`, `v3` and `default` variants but only two distinct outcomes. `v3` is already covered by `ratelimit-grpc`, and `default` behaves identically to `v2` because `IRRateLimit.setup()` defaults an unset `protocol_version` to `v2` and then rejects it, so only the one invalid case is ported.

Worth noting where the rejection lives: the CRD enum accepts both `v2` and `v3` and the field is optional, so this is not schema validation. `IRRateLimit.setup()` posts an error and returns `False`, which drops the rate-limit filter entirely. The fixture asserts that: all three queries reach the target with a 200 and none of the rls response headers appear.

`RateLimitV0Test` is deliberately not ported. Its config is identical to the already-ported `RateLimitV1Test` apart from a `/labels/` Mapping that no query exercises.

## Testing

The cert fix was verified by falsification, rebuilding the kat-server image both ways and running `ratelimit-tls` against each:

- `root-ca` cert: **fails 6 of 10 assertions**, all on the rate-limit rejection path, confirming the fail-open behaviour.
- `self-signed` leaf: **10 of 10 pass**.

Full suite after the fix: **12 of 12 fixtures pass** across 9 shards, 0 failed, 0 skipped.

## Notes for reviewers

`ratelimit-protocol-invalid` proves the filter is absent but cannot prove it is absent *because* the protocol version was invalid, since that would need the diagnostics on the admin port. `ratelimit-grpc` covers the positive path with the same topology, so the two together are meaningful, but nothing enforces the pairing.


test/e2e/fixtures/ratelimit-tls/manifests.yaml
```diff
@@ -0,0 +1,197 @@
...
+    L9TC4gq0rwKBgA9lvQ1nR+yrqzQoB/uOCduCHySUG00cMje5XQYBFac+bMBvj7Ex
+    K6LmCP9Ltz2iVzIbNRX6LnnAW0/r2og4F4pgzSA59yFNAO9WFgPj2zm99M6lgldS
+    C5vmwW0xmBRVdMvTrUAp62xAkye7ZHyOaovgz4fJLCT1HK0mrapg09SpAoGANMNt
+    i16E1AnVA3gTkLsPCttaXnibhZu33ViKmwPpi6IqgshWeS42wDq3NKyErT8evFmE
+    HRvwoX1R7mEt9qKxABHBFXSRKeZOAtIXQht+37nouoVADU6+vzh/5ZwL7s8YS+Kd
+    l7HR4u5VxN+vYj1WLzNMTKZSURoYr2Cg2Hb1PcsCgYAfEAy1r1cf+v3Y3fql1ka5
+    Hcj1XI1pD9msFGV+ipBqUMRSjZwe3owiY065ZKxf57naG+KwZ+bjkPTsW/2NAQ8S
+    PhuKUWfI8FYJfImzKlXV6NghAxZUKvzWnyHV10w/qEE+3NvdSXyj/9eTNpkAbmXW
+    IT0KrQbTnETnPL4jomup+w==
+    -----END PRIVATE KEY-----
```
the-wondersmith: I know it's literally just a testing fixture, but do we **_have_** to commit certificates to the repo? It's not a hill I'm precisely willing to die on, but I *will* put it at least a token effort in the name of general repo hygiene 😅.

ppeble: No, this is reasonable. I changed it so it generates the cert at run time using the test framework steps.

the-wondersmith (approved):

PR #5992, opened 2026-09-14
test(e2e): port the query-parameter and regex-rewrite routing tests

ppeble:
## Description

Ports five KAT classes from the dormant Python suite to Chainsaw fixtures, continuing the `test/e2e` migration. Routing-focused: selecting a backend by query parameter, and rewriting a request path by regex.

| Fixture | KAT class |
|---|---|
| `queryparameter-routing` | `t_queryparameter_routing.py::QueryParameterRoutingTest` |
| `queryparameter-routing-regex` | `t_queryparameter_routing.py::QueryParameterRoutingWithRegexTest` |
| `queryparameter-present` | `t_queryparameter_routing.py::QueryParameterPresentRoutingTest` |
| `regexrewrite-forwarding` | `t_regexrewrite_forwarding.py::RegexRewriteForwardingTest` |
| `regexrewrite-forwarding-extract` | `t_regexrewrite_forwarding.py::RegexRewriteForwardingWithExtractAndSubstituteTest` |

One fixture per class, matching how `t_gzip.py`'s three classes became three fixtures. All are `slot: shared`, since they configure only Mappings.

Two notes on the translation:

**Prefixes are namespaced to the fixture.** `shared` fixtures run concurrently against the same Emissary, so reusing KAT's literal `/target/` and `/foo/` would have these fixtures stealing each other's traffic. Each prefix is named after its fixture, as `header-routing` and `http-basic` already do. The regex `pattern` and `substitution` literals were adjusted to match, preserving the same rewrite shapes: one literal match and one capture-group substitution.

**Negative expectations are paired with positive ones.** A bare 404 assertion also passes when nothing was ever configured, so every fixture with a non-matching query also asserts that its matching query reached the expected backend in the same probe run.

`x-envoy-original-path` was not previously exercised by any fixture. It behaves as the KAT tests expect: present on every 200, carrying the pre-rewrite path.

## Related Issues

None.

## Testing

- Each of the five fixtures individually via `make e2e/run/<fixture>`: all pass.
- `make e2e/run` (full suite, sharded): **18 of 18 fixtures pass**, 0 failed, 0 skipped. That covers the 13 pre-existing fixtures alongside the 5 new ones.

## Checklist

  - Nothing here ships in the image or touches the data plane.

  - The change is test coverage, exercised in CI by the `Test Images` jobs which run the whole `fixtures` directory. Each fixture covers both the matching and non-matching cases of the behaviour it tests, and the full suite was run to confirm nothing pre-existing regressed.

  - No, this is test coverage.

  - No new tricks needed.

  - No credentials, certificates, or key material are introduced. The fixtures create only Mappings and kat-server Deployments in each test's ephemeral namespace.


kflynn (approved):
Hit it! 🙂

PR #5993, opened 2026-09-14
test(e2e): port fifteen single-backend tests from seven KAT files

ppeble:
## Description

Ports fifteen KAT classes from seven files to Chainsaw fixtures, continuing the `test/e2e` migration. All are single-backend tests asserting on status codes or headers, so they are mechanically similar and grouped here as one batch. **There is one commit per source file**, which is the intended way to review this.

| Commit | Fixtures | KAT source |
|---|---|---|
| `test(e2e): port t_dns_type.py's five live classes` | `dns-type-logical`, `dns-type-logical-endpoint`, `dns-type-ttl`, `dns-type-ttl-endpoint`, `dns-type-ttl-logical` | `t_dns_type.py` |
| `test(e2e): port three of t_request_header.py's four classes` | `request-header-preserve-id`, `request-header-envoy-headers`, `request-header-suppress-envoy-headers` | `t_request_header.py` |
| `test(e2e): port t_headerswithunderscoresaction.py's two classes` | `headers-underscores-allow`, `headers-underscores-reject` | `t_headerswithunderscoresaction.py` |
| `test(e2e): port t_max_req_header_kb.py's two classes` | `max-request-header-kb`, `max-request-header-kb-max` | `t_max_req_header_kb.py` |
| `test(e2e): port t_cors.py::GlobalCORSTest` | `cors-global` | `t_cors.py` |
| `test(e2e): port t_grpc_web.py::AcceptanceGrpcWebTest` | `grpc-web` | `t_grpc_web.py` |
| `test(e2e): port t_grpc_bridge.py::AcceptanceGrpcBridgeTest` | `grpc-bridge` | `t_grpc_bridge.py` |

`t_request_header.py::XRequestIdHeaderDefaultTest` is not ported; KAT marks it `self.xfail` ("Need to figure out passing header through external connections from KAT").

### Slot assignment is per class, not per file

A `Module` is instance-global, so any fixture declaring one must be `slot: exclusive` while the rest can be `shared` and run concurrently. That split does not follow file boundaries: `t_request_header.py` declares a Module in two of its three ported classes, and `t_headerswithunderscoresaction.py` in only one of two. Each fixture's slot is set from what that class actually configures.

Worth flagging for anyone grepping this suite later: much of KAT writes `kind:  Module` with two spaces, so `grep 'kind: Module'` misses roughly half of them. `grep -E '^ *kind: +Module'` is the pattern that finds all of them. `t_cors.py::GlobalCORSTest` is one of the ones the naive pattern misses, and it does set an instance-wide default CORS policy that its Mapping then overrides per route, hence `exclusive`.

### Two other translation notes

**Prefixes and resolver names are namespaced per fixture.** `shared` fixtures share an Emissary, so a reused `prefix:` would have two fixtures stealing each other's traffic. The two DNS fixtures needing a `KubernetesEndpointResolver` name it after the fixture rather than reusing KAT's literal `endpoint`, which keeps both safely `shared`.

**The two `max_request_headers_kb` fixtures have no `queries.json`.** Their oversized headers are 30KB and 97KB of filler, so the probe step builds the query set with `jq` into a temp file at run time rather than committing that much padding.

## Related Issues

None.

## Testing

- Each of the fifteen fixtures individually via `make e2e/run/<fixture>`: all pass.
- `make e2e/run` (full suite, sharded): **28 of 28 fixtures pass**, 0 failed, 0 skipped, across all 9 shards. That covers the 13 pre-existing fixtures alongside the 15 new ones.
- Separately audited all 28 fixtures for the Module/slot hazard described above: every Module-declaring fixture is `exclusive` and every `shared` one declares none, so there is no fixture that would race intermittently.

None of these expectations turned out to be stale. Every ported class reproduced KAT's original assertions unchanged, including both the 30KB and 96KB header caps and all six CORS origin-matching cases.

## Checklist

  - Nothing here ships in the image or touches the data plane. It does add eleven `exclusive` fixtures, so slots now run more than one fixture serially and the e2e job takes longer, but no slots were added.

  - The change is test coverage, exercised in CI by the `Test Images` jobs which run the whole `fixtures` directory. Where a behaviour has an enabled and a disabled form, both are covered by separate fixtures: headers-with-underscores allow and reject, Envoy headers present and suppressed, and the header-size cap at two different limits.

  - No, this is test coverage.

  - No new tricks needed.

  - No credentials, certificates, or key material are introduced. Fixtures create only Emissary CRDs and kat-server Deployments in each test's ephemeral namespace, and the generated oversized-header payloads are written to a temp file outside the repository.


the-wondersmith (approved):

the-wondersmith:
lgtm

PR #5994, opened 2026-09-16
test(e2e): port eleven TCPMapping tests from t_tcpmapping.py

ppeble:
## Description

Ports eleven `t_tcpmapping.py` classes to Chainsaw fixtures, continuing the `test/e2e` migration. These are the classes declaring `extra_ports = [6789]`, i.e. needing exactly one TCP port, which is what a slot already publishes at `<base>+2`. No changes to the slot scheme or the cluster were needed.

**One commit per tier**, growing in complexity, which is the intended review order:

| Commit | Fixtures | KAT classes |
|---|---|---|
| no-cert | `tcpmapping-plain`, `tcpmapping-cross-namespace`, `tcpmapping-tls-origination-scheme`, `tcpmapping-weight` | `TCPMappingBasicTest`, `TCPMappingCrossNamespaceTest`, `TCPMappingTLSOriginationV3SchemeTest`, `TCPMappingWeightTest` |
| TLS origination | `tcpmapping-tls-origination-context`, `-context-dot`, `-context-cross-namespace` | the three `TCPMappingTLSOriginationContext*Test` classes |
| TLS termination | `tcpmapping-tls-termination-basic`, `-cross-namespace` | `TCPMappingTLSTerminationBasicTest`, `TCPMappingTLSTerminationCrossNamespaceTest` |
| SNI | `tcpmapping-sni-shared-context`, `tcpmapping-sni-separate-contexts` | `TCPMappingSNISharedContextTest`, `TCPMappingSNISeparateContextsTest` |

All eleven are `slot: exclusive`, because a TCPMapping binds the instance's TCP listener port. The existing `tcpmapping-basic` fixture stays `shared` and untouched; it binds the base release's 6789, so it does not contend with these.

Certificates are generated per run by an `openssl` script step into each test's own namespace, as the existing TLS fixtures do. Nothing certificate-shaped is committed.

### Deliberate gaps, so a reviewer does not have to find them

**`TCPMappingTLSOriginationBoolTest` is not ported.** The CRD the charts install types `TCPMapping.spec.tls` as a plain string, so `tls: true` is rejected by the API server and there is no boolean form to port. The test's configuration reduces to the same `https://` scheme prefix on `service` that `TCPMappingTLSOriginationV3SchemeTest` already covers, so a fixture for it would have been byte-identical to that one.

Worth flagging separately: the generated `pkg/api/getambassador.io/crds.yaml` declares this field as `oneOf: [string, boolean]`, while `charts/emissary-crds/templates/tcpmappings.yaml` declares `type: string`. Since installs use the chart, the string form is what actually applies. That divergence looks like a pre-existing inconsistency rather than anything to do with this change, so it is left alone here.

**Only the `tlscontext` half of the parameterised classes is ported.** The four TLS-termination and SNI classes inherit a `variants()` that yields both a `tlscontext` and a `host` spelling of where the certificate comes from. These fixtures cover the `tlscontext` variant; the `host` variant is not yet ported.

**`tcpmapping-weight` does not assert the 70/30 ratio** its KAT counterpart does. It sends 40 requests and asserts that all succeed and that both weighted backends appear at least once. A single finite probe batch cannot bound a distribution without either a very large sample or accepting flakiness, so it asserts what is deterministic instead.

## Related Issues

None.

## Testing

- Each of the eleven fixtures individually via `make e2e/run/<fixture>`: all pass.
- `make e2e/run` (full suite, sharded) on this branch rebased onto current main: **44 of 44 fixtures pass**, 0 failed, 0 skipped, in about 50 seconds wall clock. That is 16 shared and 28 exclusive, with the exclusive ones dealt three or four to a slot and run serially.
- Checked after the run that no namespaces leak. The cross-namespace fixtures create a second Namespace through an `apply` step rather than a raw `kubectl` call inside a script, so Chainsaw owns it and deletes it with the rest.

## Checklist

  - Nothing here ships in the image or touches the data plane. It does add eleven `exclusive` fixtures, so the e2e job does more serial work per slot; the full suite still completes in about 50 seconds locally.

  - The change is test coverage, exercised in CI by the `Test Images` jobs which run the whole `fixtures` directory. TCP behaviour is now covered in its passthrough, TLS-originating, TLS-terminating, and SNI forms, across same-namespace and cross-namespace placements.

  - No, this is test coverage.

  - No new tricks needed.

  - No credentials or key material are committed. Every certificate is generated at run time as an explicit leaf (`CA:FALSE`, `digitalSignature`, `serverAuth`) with SANs matching only the hostnames the test queries, loaded into the test's ephemeral namespace and discarded with it.


kflynn (approved):
Yeah, let's get this in -- thanks @ppeble!! 🙂

